Issue #52, opened 2018-01-26
Specifying driver and executor pod properties


foxish:
Starting a couple of discussions around specification of properties.

```
spec:
  deps: {}
  driver:
    cores: "0.1"
    image: liyinan926/spark-driver:v2.3.0
  executor:
    image: liyinan926/spark-executor:v2.3.0
    instances: 1
    memory: 512m
``` 

Should this mimic the driver/executor properties more closely?
For example: `spark.kubernetes.driver.limit.cores`. 

/xref: https://dist.apache.org/repos/dist/dev/spark/v2.3.0-rc2-docs/_site/running-on-kubernetes.html

liyinan926:
Yes, I think at least each Kubernetes specific property + some most common Spark general properties should have a corresponding field in the spec.

PR #56, opened 2018-01-26
Added support for core limits, labels, and annotations, node selector, and pull policy

liyinan926:
Ref #52.

@foxish


pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -200,6 +200,8 @@ type DriverSpec struct {
 	// Cores is used to set spark.driver.cores.
 	// Optional.
 	Cores *string `json:"cores,omitempty"`
+	// CoreLimit specifies a hard limit on CPU cores for the driver pod.
+	CoreLimit *string `json:"coreLimit,omitempty"`
```
foxish: Wondering if it's just easier to take the podspec field from types.go and allow only a subset? Since that has all the possible parameterization possible.

liyinan926: Yes, I was actually thinking about that. It seems a bit overkill as we really only allow setting a few fields. We can extend that through the initializer/mutation webhook, and I think we can revisit this later when we have better ideas on what fields should be customizable.

liyinan926:
Just realized that `DriverSpec` and `ExecutorSpec` share most of the fields. Combined them into `SparkPodSpec` and embed `SparkPodSpec` into both `DriverSpec` and  `ExecutorSpec`. Also added a TODO for investigation use of `v1.PodSpec` instead.

PR #150, opened 2018-05-10
Fix the typos in quick start guide and adding the namespace explicitly

Jiri-Kremser:
Without the `-n sparkoperator` I wasn't able to see the deployment.

wrt. `kubectl describe sparkapplication spark-pi` it just looks better, although the previous form was also working


liyinan926:
Thanks for fixing!

PR #168, opened 2018-05-24
Explicitly delete the driver pod and UI service when a SparkApp gets deleted

### LINKED ISSUE #167: resources are not cleaned up when a sparkapp is deleted
right now the onDelete handler in the sparkapp controller doesn't really do anything. this is the last opportunity for it to clean up the pods / services created. seems like `deleteDriverAndUIService()` should be called in `onDelete()`

liyinan926:
Fixes #167.

@scotthew1


scotthew1:
hey thanks for the quick fix!

liyinan926:
No problem. The image at `gcr.io/spark-operator/spark-operator:v2.3.0-v1alpha1-latest` has also been updated in case you use it.

Issue #172, opened 2018-05-30
support for operating on a specific namespace


scotthew1:
the current spark operator setup assumes that it runs in its own namespace and manages the sparkapp crds for the whole cluster. we were hoping that we could deploy this alongside our application which currently has the requirement that it has namespace isolation. with our current setup, if we have multiple applications on our cluster, each spark-operator sees all the crds on the cluster regardless of namespace. i was hoping i could restrict this with different role bindings but alas...
```
E0529 21:44:38.579020       1 reflector.go:205] k8s.io/spark-on-k8s-operator/pkg/client/informers/externalversions/factory.go:117: Failed to list *v1alpha1.SparkApplication: sparkapplications.sparkoperator.k8s.io is forbidden: User "system:serviceaccount:sreisor:ideal-tuatara-sparkoperator" cannot list sparkapplications.sparkoperator.k8s.io at the cluster scope
E0529 21:44:38.579370       1 reflector.go:205] k8s.io/spark-on-k8s-operator/pkg/client/informers/externalversions/factory.go:117: Failed to list *v1alpha1.ScheduledSparkApplication: scheduledsparkapplications.sparkoperator.k8s.io is forbidden: User "system:serviceaccount:sreisor:ideal-tuatara-sparkoperator" cannot list scheduledsparkapplications.sparkoperator.k8s.io at the cluster scope
```
is there anything we can do to restrict spark operator listing only within its namespace or a specific namespace? it's our first dependency that would break our ability to `helm install` our application into an isolated namespace.

liyinan926:
Thanks for creating this. We also have a use case that requires limiting the operator to a single namespace. It should be straightforward to add an option to do that. Note that if the operator manipulates SparkApplications in a namespace that is different than the one of the operator, it still needs a ClusterRoleBinding.

scotthew1:
cool! glad to hear we're not the only ones looking for this functionality. if you point me in the right direction i'd be happy to help add the option.

liyinan926:
[...]

PR #173, opened 2018-05-31
add -namespace flag to manage only CRDs in that namespace

scotthew1:
change for #172

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/golang/glog"
 
+	apiv1 "k8s.io/api/core/v1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


scotthew1:
signed the CLA

main.go
```diff
@@ -60,6 +60,8 @@ var (
 	submissionRunnerThreads = flag.Int("submission-threads", 3, "Number of worker threads "+
 		"used by the SparkApplication submission runner.")
 	resyncInterval = flag.Int("resync-interval", 30, "Informer resync interval in seconds")
+	namespace = flag.String("namespace", "", "The Kubernetes namespace to manage. "+
+		"Will manage CRDs for the whole cluster if unset.")
```
liyinan926: Changing `CRDs` to `custom resource objects of the managed CRD types` makes it clearer.

main.go
```diff
@@ -105,12 +107,17 @@ func main() {
 		}
 	}
 
-	factory := crdinformers.NewSharedInformerFactory(
+	var factoryOpts []crdinformers.SharedInformerOption
+	if *namespace != "" {
```
liyinan926: I think we can be more defensive by doing `if namespace != nil && *namespace != ""`.

scotthew1: ah good call

main.go
```diff
@@ -60,6 +60,8 @@ var (
 	submissionRunnerThreads = flag.Int("submission-threads", 3, "Number of worker threads "+
 		"used by the SparkApplication submission runner.")
 	resyncInterval = flag.Int("resync-interval", 30, "Informer resync interval in seconds")
+	namespace = flag.String("namespace", "", "The Kubernetes namespace to manage. "+
```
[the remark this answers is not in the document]
scotthew1: i like this idea 👍

main.go
```diff
@@ -105,12 +108,20 @@ func main() {
 		}
 	}
 
-	factory := crdinformers.NewSharedInformerFactory(
+	var factoryOpts []crdinformers.SharedInformerOption
+	if namespace == nil {
```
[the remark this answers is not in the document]
scotthew1: ok, i thought this was the case before where the default value was an empty string, but wanted to keep my bases covered after your previous comment. i will clean this up.

PR #174, opened 2018-05-31
use multi-stage Dockerfile for reliable builds

scotthew1:
as-is this project requires a lot of manual installation to build, and nothing is ensuring versions for go, dep, and the code generator. this PR provides a multi-stage dockerfile that can build in a reliable environment as well as some changes to the codegen. also, these changes make the current makefile unnecessary. 

i understand some of these changes are a bit opinionated, so i'd be happy to make changes according to feedback. i didn't bother making changes to the documentation until i had some feedback, but if you guys like this approach, i'm happy to make those changes too.


hack/custom-boilerplate.go.txt
```diff
@@ -0,0 +1,17 @@
+// Code generated by k8s code-generator DO NOT EDIT.
+
+/*
+Copyright 2017 Google LLC
```
liyinan926: s/`2017`/`2018`/.

scotthew1:
@liyinan926 do you think it's necessary that the docs still explain how to install all the dependencies locally for a local build? or do you think being able to build this docker image is enough?

scotthew1:
i updated the docs a bit. my editor cleans up end of line whitespace so there's a bit of changes from that, but i put those in a separate commit for easier review.

liyinan926:
LGTM.

PR #180, opened 2018-06-08
Made DNS check optional (on by default)

liyinan926:


mrow4a:
@liyinan926 what is the use-case of disabling the DNS check? Meaning the driver now will hang for some time waiting, instead of failing fast, am I correct?

liyinan926:
@mrow4a Some clusters don't have kube-dns, for example, OpenShift clusters. This check will fail on those clusters even if they may have other dns systems installed.

PR #189, opened 2018-06-14
Add an entrypoint to spark-operator for handling non-root users

tmckayus:
In order to use the spark-on-k8s-operator on a platform where a user
is assigned based on a security context (OpenShift), the /etc/passwd
file needs to be modified in an entrypoint like it is in the standard
Apache Spark 2.3 image to support an assigned user (note that it
won't be changed for a root user). Change the spark-operator.yaml to
call the entrypoint and pass /usr/bin/spark-operator as the first argument.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,5 @@ RUN go generate && go build -o /usr/bin/spark-operator
 
 FROM gcr.io/ynli-k8s/spark:v2.3.0
 COPY --from=builder /usr/bin/spark-operator /usr/bin/
-ENTRYPOINT ["/usr/bin/spark-operator"]
+COPY ./entrypoint /entrypoint
+ENTRYPOINT ["/entrypoint"]
--- a/entrypoint
+++ b/entrypoint
@@ -34,4 +34,4 @@ if [ -z "$uidentry" ] ; then
         echo "Container ENTRYPOINT failed to add passwd entry for anonymous UID"
     fi
 fi
-exec /usr/bin/spark-operator "$@"
+exec "$@"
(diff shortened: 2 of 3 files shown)


tmckayus:
@liyinan926 please take a look. I've been playing with the spark-on-k8s-operator OpenShift and I think this is a transparent change to make it work there

erikerlandson:
I'm wondering if there is any clever way we can use the spark upstream's `entrypoint`, since this is layered on top of those images

tmckayus:
fyi, see https://issues.apache.org/jira/browse/SPARK-24534

manifest/spark-operator.yaml
```diff
@@ -43,7 +43,6 @@ spec:
       - name: sparkoperator
         image: gcr.io/spark-operator/spark-operator:v2.3.0-v1alpha1-latest
         imagePullPolicy: Always
-        command: ["/usr/bin/spark-operator"]
```
tmckayus: fyi, going to tweak this and use /entrpoint here and move the command into the arg list

entrypoint
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+#
```
liyinan926: nit: missing `Copyright 2018 Google LLC` in the license header.

liyinan926: We can ignore this for now.

Dockerfile
```diff
@@ -29,4 +29,5 @@ RUN go generate && go build -o /usr/bin/spark-operator
 
 FROM gcr.io/ynli-k8s/spark:v2.3.0
```
liyinan926: This will also require an image built using the Dockerfile in the latest spark/master, right?

skonto:
We have similar code running fine on Openshift will contribute it here and push forward this PR is useful. @yuchaoran2011 ^^

liyinan926:
This is covered by #343.

PR #190, opened 2018-06-15
Adjust to kubernetes api v1.10 and client-go v7.0

mrow4a:
As in the title. Spark v1.10 brings some more features, e.g. binaryData in config maps and updates/removes some dependencies. 

@liyinan926, CC @prasanthkothuri


sparkctl/cmd/create.go
```diff
@@ -329,15 +331,21 @@ func buildHadoopConfigMap(appName string, hadoopConfDir string) (*apiv1.ConfigMa
 		if err != nil {
 			return nil, err
 		}
-		hadoopConfigFiles[file.Name()] = string(content)
+
+		if utf8.Valid(content) {
```
liyinan926: What if the file encoding is not utf-8?

mrow4a: It gets stored as new type: binaryData. For legacy configMap, the files with non-utf8 were anyways corrupted, since they were converted to string. So in the cluster you could see the data, but when you try using it, you get errors about bad format or corruption. I guess, with binaryData flag for legacy configMap, it will be just ignored (instead of being corrupted) - probably worth testing

liyinan926: Copied from 1.10 change logs: `ConfigMap objects now support binary data via a new binaryData field. When using kubectl create configmap --from-file, files containing non-UTF8 data will be placed in this new field in order to preserve the non-UTF8 data.`. So this looks like the right approach. I think it worths testing this in a 1.9 cluster.

liyinan926:
@mrow4a have you got a chance to test this on a 1.9 cluster?

mrow4a:
@liyinan926 Will test it later this week.

mrow4a:
@liyinan926 I have tested it on cluster v1.9.3, and binaryConfig gets just ignored - it does not show in the final definition of ConfigMap

```
$ kubectl get configmap/cluster-init-job-hadoop-config -o yaml
apiVersion: v1
kind: ConfigMap
metadata:
  creationTimestamp: 2018-07-06T14:41:42Z
  name: cluster-init-job-hadoop-config
  namespace: default
  resourceVersion: "3682"
  selfLink: /api/v1/namespaces/default/configmaps/cluster-init-job-hadoop-config
  uid: b326abbf-812a-11e8-bccf-fa163eea31cb

$ kubectl get configmap/cluster-init-job-hadoop-config -o yaml
apiVersion: v1
data:
  test.txt: |
    utf8 test
kind: ConfigMap
metadata:
  creationTimestamp: 2018-07-06T14:42:43Z
  name: cluster-init-job-hadoop-config
  namespace: default
  resourceVersion: "3855"
  selfLink: /api/v1/namespaces/default/configmaps/cluster-init-job-hadoop-config
  uid: d7069d7f-812a-11e8-bccf-fa163eea31cb

$ kubectl describe configmap/cluster-init-job-hadoop-config
Name:         cluster-init-job-hadoop-config
Namespace:    default
Labels:       <none>
Annotations:  <none>

Data
====
test.txt:
----
utf8 test

Events:  <none>
```

liyinan926:
Awesome!

PR #207, opened 2018-07-11
Fast forwarding 2.2 branch to latest master branch

kumare3:
There are few fixes in master that should be pulled into 2.2
Also, this will set the stage to make the larger change on top of this branch.


kumare3:
I signed it!

kumare3:
@liyinan926 can you please approve the merge? This is just fast forwarding to the latest master. We have tested this, and barring the race condition bug, which we will fix through the refactor it works just as expected

Dockerfile
```diff
@@ -14,6 +14,24 @@
...
+RUN chmod +x /usr/bin/dep
+
+WORKDIR ${GOPATH}/src/k8s.io/spark-on-k8s-operator
+COPY Gopkg.toml Gopkg.lock ./
+RUN dep ensure -vendor-only
+COPY . ./
+RUN go generate && go build -o /usr/bin/spark-operator
+
+
+FROM gcr.io/ynli-k8s/spark:v2.3.0
```
liyinan926: This should be `FROM kubespark/spark-base:v2.2.0-kubernetes-0.5.0` for 2.2.

Dockerfile
```diff
@@ -14,6 +14,24 @@
...
+COPY --from=builder /usr/bin/spark-operator /usr/bin/
+>>>>>>> master
```
liyinan926: Remove this.

Dockerfile
```diff
@@ -14,6 +14,24 @@
 # limitations under the License.
 #
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
kumare3: hey good catch, will do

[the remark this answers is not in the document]
kumare3: Also I will follow up with a commit to use scratch to build the go binary. That makes the image really tiny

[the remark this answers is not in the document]
kumare3: updated

Dockerfile
kumare3: done

Dockerfile
kumare3: done

liyinan926:
@kumare3 Looks like this is still missing a few commits made to the master branch.

kumare3:
@liyinan926 thank you for bringing that to my attention. I have pulled the recent tip. Can you check and approve?

PR #220, opened 2018-07-18
Spark Operator: Prometheus Metrics Integration

akhurana001:
Reference: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/67


akhurana001:
PTAL @liyinan926 and let me know if you have any questions.

mrow4a:
Check your commits - it looks very strange - commit does not seem to represent functionality, rather show the history of your changes (maybe squash?)

manifest/spark-operator.yaml
```diff
@@ -44,5 +47,9 @@ spec:
         image: gcr.io/spark-operator/spark-operator:v2.3.0-v1alpha1-latest
         imagePullPolicy: Always
         command: ["/usr/bin/spark-operator"]
+        ports:
+          - containerPort: 10254
         args:
         - -logtostderr
+        - -enable-metrics=true
```
mrow4a: What happens when I don't use prometheus?

liyinan926:
@mrow4a This PR contains some merge PRs that bring the `spark-2.2-support` up to date with the master. The actual diff shows only the relevant changes.

manifest/spark-operator.yaml
akhurana001: Right now this only supports Prometheus. Let me know what else you believe we should support ? 
I can generalize this a little more by adding wrapper interfaces for each Prometheus constructs like Counter/Gauge so that we can add other implementations going further.  Some of the operator arguments like metrics-endpoint etc will still be specific to Prometheus and we can rename them appropriately. 

@mrow4a/@liyinan926  let me know what you guys think about this.

akhurana001:
@liyinan926 can we turn on squash support for the repo ?  That way we won't have to push all the commits. Otherwise I can create another branch and do a rebase -i to generate a PR with just one commit if that's preferred.

liyinan926:
@akhurana001 Yes, squash and merge is an option. It might be better to create a separate PR against the master branch also for https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/commit/6e0e422b55966e8f4813da7293e6711f106cce3b.

akhurana001:
Closing this PR in favor of the following 2 separate PRs: 
Operator Spark-submit: Never Wait for Completion : https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/226
SparkOperator: Prometheus Metrics Integration: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/227

PR #221, opened 2018-07-18
Shorten the suffix appended to the name of the driver UI service

### LINKED ISSUE #201: UI service has a random hash in its name
The Operator appends a 32 bit hash to the appName specified in the spec (https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L729 ) . Is there a particular reason to append the 32 bit hash ? I understand this is used to create the UI service but won't the appName/resourceName already be unique in kubernetes and can be used directly for the same ? 
 
Also kubernetes enforces a 63 character limit on the name and the hash (with the "ui-svc" string) ends up consuming a good part of it.  We ended up breaching this threshold.

liyinan926:
Fixes #201. By shortening the suffix appended to the name of the driver UI service, we give user more room to use longer application name. Particularly, the service name is changed from `appId-ui-svc` to `appName-ui-svc` instead as `appIs` is derived as `appName-hash`.

### DIFF
--- a/pkg/controller/sparkapplication/sparkui.go
+++ b/pkg/controller/sparkapplication/sparkui.go
@@ -45,9 +45,15 @@ func createSparkUIService(
 		return "", -1, fmt.Errorf("invalid Spark UI port: %s", portStr)
 	}
 
+	serviceName := app.Name + "-ui-svc"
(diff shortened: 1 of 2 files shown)


liyinan926:
@akhurana001 @kumare3

PR #223, opened 2018-07-19
Shorten the name of the driver UI service

liyinan926:
@akhurana001 @kumare3


akhurana001:
👍 , thanks for picking this up @liyinan926

Issue #199, opened 2018-07-02
Potential bug in Operator State management


kumare3:
The operator seems to monitor events (informer) for the CRD and the executor/driver pods independently. When a pod is updated the state is preserved in etcd using optimistic concurrency. The problem is, we do not do a CAS style operation, which ensures that a previously successful/completion job does not go back to running.
Let us take an example,
 1. Job is submitted, state is submitted
 2. driver goes to running, but fails write and waits before a retry
 3. driver completes and status update to complete is successful
 4. now the previously pending write is applied, which causes the state to be reverted to submitted/running
Basically the problem is out of order updates being applied to a consistent data store
The manifestation is that the app state get stuck in a non terminal state and affects the observer.

We have a simple hack to resolve this issue for now
https://github.com/lyft/spark-on-k8s-operator/blob/spark-2.2-support/pkg/controller/sparkapplication/controller.go#L521-L524

A problem with the hack is we can lose the executor states (as we just ignore non terminal states)
We will work on cleaning up this and upstreaming the fix if you think this is a problem

kumare3:
On a side note, we could simplify the state machine management, by making sure all the updates to the state happen in the same thread (goroutine). For example, any pod update (for spark) just triggers the top level CRD's onUpdate and then we use the listers to determine and take the right action. @liyinan926 we would love to discuss this and help you refactor the code. For now we will implement the bug fix and share that with you.

liyinan926:
@kumare3 It's a good point that state updates should be handled by a single thread in a defined order, e.g., FIFO. We could trigger the top-level `OnUpdate` whenever there's an status update of the driver or executor pods. Specifically, whenever the controller receives a driver or executor pod status update through the `appStateReportingChan` channel, the key of the corresponding `SparkApplication` object gets enqueued. By doing this, however, instead of updating the status of a `SparkApplication` based on the received status updates of the driver or executor pods, within `OnUpdate` the operator needs to list the driver and executor pods and update the application status based on the result of the listing. This is much closer to what other custom controllers/operators do to achieve to so-called "level-based triggering".

kumare3:
@liyinan926 yes that is what I was suggesting. We have this implemented in a few Operators internally (which we plan to open source) We would love to contribute this if you think this could be a great addition. I feel this would simplify the operator tremendously and will help us add new features much faster? Lets discuss this during our call tomorrow.

liyinan926:
[...]

PR #227, opened 2018-07-23
SparkOperator: Prometheus Metrics Integration

akhurana001:
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/67


docs/quick-start-guide.md
```diff
@@ -50,6 +50,41 @@ $ kubectl describe deployment sparkoperator -n sparkoperator
 
 ```
 
+### Metrics
+
+The operator exposes a set of metrics via the metric endpoint to be scraped by `Prometheus`. This is enabled by default and can be disabled by setting  the flag `-enable-metrics=false` and running
```
liyinan926: I'm thinking maybe the this should default to `false` since not everyone is running Prometheus.

main.go
```diff
@@ -20,29 +20,28 @@ package main
...
 	crdclientset "k8s.io/spark-on-k8s-operator/pkg/client/clientset/versioned"
 	crdinformers "k8s.io/spark-on-k8s-operator/pkg/client/informers/externalversions"
 	"k8s.io/spark-on-k8s-operator/pkg/controller/scheduledsparkapplication"
 	"k8s.io/spark-on-k8s-operator/pkg/controller/sparkapplication"
 	"k8s.io/spark-on-k8s-operator/pkg/crd"
 	ssacrd "k8s.io/spark-on-k8s-operator/pkg/crd/scheduledsparkapplication"
 	sacrd "k8s.io/spark-on-k8s-operator/pkg/crd/sparkapplication"
+	"k8s.io/spark-on-k8s-operator/pkg/util"
 	"k8s.io/spark-on-k8s-operator/pkg/webhook"
+	"os"
```
liyinan926: Built-ion packages should come at the top.

main.go
```diff
@@ -73,6 +80,12 @@ func main() {
 		glog.Fatal(err)
 	}
 
+	var metricsBundle *util.PrometheusMetrics
+	if *enableMetrics {
+		glog.Info("Enabling metrics")
```
liyinan926: `enabling metrics collecting and exporting to Prometheus`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +567,149 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
+		if newStatus == validStatus {
+			return true
+		}
+	}
+	return false
+}
+
+// isValidExecutorStatusTransition determines if the executor state transition is valid
+func isValidExecutorStatusTransition(oldStatus, newStatus v1alpha1.ExecutorState) bool {
+
```
liyinan926: Please remove the extra empty line here and in several places below.

pkg/controller/sparkapplication/controller.go
```diff
@@ -55,6 +53,38 @@ const (
 
 var (
 	keyFunc = cache.DeletionHandlingMetaNamespaceKeyFunc
+
+	// Valid Driver State Transitions: map with valid pairs of oldState-> {newStates}
```
liyinan926: I think it's worth putting code related to state transition validation into a separate go file, e.g., state_validation.go?

pkg/controller/sparkapplication/controller.go
```diff
@@ -500,21 +532,33 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
 
 func (c *Controller) updateSparkApplicationStatusWithRetries(
 	original *v1alpha1.SparkApplication,
-	updateFunc func(*v1alpha1.SparkApplicationStatus)) *v1alpha1.SparkApplication {
-	toUpdate := original.DeepCopy()
+	updateFunc func(*v1alpha1.SparkApplicationStatus),
+) *v1alpha1.SparkApplication {
```
liyinan926: You can revert this line. It doesn't need to be in a new line.

pkg/controller/sparkapplication/controller.go
```diff
@@ -500,21 +532,33 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
...
+		if !isValidDriverStatusTransition(original.Status.AppState.State, toUpdate.Status.AppState.State) {
+			toUpdate.Status.AppState = original.Status.AppState
+		}
+		updated, err := c.tryUpdateStatus(original, toUpdate)
 		if err == nil {
+			c.postMetrics(original, updated)
 			return updated
 		}
 		lastUpdateErr = err
+		glog.Errorf("[Attempt: %v] failed to update SparkApplication [%v] %v", i, toUpdate, err)
```
liyinan926: The first `%v` can be `%d`. Also do you really want to log the entire `SparkApplication` object instead of just the app name?

pkg/controller/sparkapplication/controller.go
```diff
@@ -500,21 +532,33 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
...
 	original *v1alpha1.SparkApplication,
-	updateFunc func(*v1alpha1.SparkApplicationStatus)) *v1alpha1.SparkApplication {
-	toUpdate := original.DeepCopy()
+	updateFunc func(*v1alpha1.SparkApplicationStatus),
+) *v1alpha1.SparkApplication {
 
 	var lastUpdateErr error
 	for i := 0; i < maximumUpdateRetries; i++ {
-		updated, err := c.tryUpdateStatus(original, toUpdate, updateFunc)
+		toUpdate := original.DeepCopy()
```
liyinan926: You only need to make a deep copy once outside the loop. You can assign `toUpdate` to is fetched from the API server below as the one returned is a copy anyway.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +567,149 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
+	for _, validStatus := range validExecutorStateTransitions[oldStatus] {
+		if newStatus == validStatus {
+			return true
+		}
+	}
+
+	return false
+}
+
+func fetchMetricLabels(specLabels map[string]string, labels []string) map[string]string {
```
liyinan926: Let's put metrics related code into its own file, e.g., metrics.go.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -32,9 +31,13 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 
+	"github.com/prometheus/client_golang/prometheus"
+	prometheus_model "github.com/prometheus/client_model/go"
 	"k8s.io/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1alpha1"
 	crdclientfake "k8s.io/spark-on-k8s-operator/pkg/client/clientset/versioned/fake"
 	crdinformers "k8s.io/spark-on-k8s-operator/pkg/client/informers/externalversions"
+	"k8s.io/spark-on-k8s-operator/pkg/util"
+	"net/http"
```
liyinan926: Built-in package should be at the top.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -17,13 +17,12 @@ limitations under the License.
 package sparkapplication
 
 import (
+	"github.com/stretchr/testify/assert"
```
liyinan926: Please revert this position change.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +567,149 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
+		if metricLabels[label] == "" {
+			metricLabels[label] = "Unknown"
+		}
+	}
+	return metricLabels
+}
+
+func (c *Controller) postMetrics(oldState, newState *v1alpha1.SparkApplication) {
+	if c.metrics == nil {
+		glog.V(2).Infof("Metrics not enabled. Skipping posting metrics")
```
liyinan926: `Metrics not enabled so skipping exporting metrics`. The same below.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +567,149 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func fetchMetricLabels(specLabels map[string]string, labels []string) map[string]string {
+
+	metricLabels := map[string]string{}
```
liyinan926: Please use `make(map[string]string)` to be consistent with the rest of the code base.

pkg/util/metrics.go
```diff
@@ -0,0 +1,250 @@
+package util
```
liyinan926: Missing license header.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +567,149 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
+	for _, label := range labels {
+		metricLabels[label] = specLabels[label]
+		if metricLabels[label] == "" {
+			metricLabels[label] = "Unknown"
+		}
+	}
+	return metricLabels
+}
+
+func (c *Controller) postMetrics(oldState, newState *v1alpha1.SparkApplication) {
```
liyinan926: `exportMetrics`, and let's rename `oldState` and `newState` to `oldApp` and `newApp`.

akhurana001:
Hi, I have updated based on the PR comments.  I have refactored metrics implementation by moving the SparkApplication specific logic under  pkg/sparkapplication/sparkapp_metrics.go and only keeping the common logic in util/metrics.go.  Also updated the state-transition map to let terminal states move to Submitted as we support RetryPolicy of Always.

A couple of other pointers:
Based on my testing, the executor running count is off because the operator doesn't seem to be getting all executor events. I am guessing this is due to the driver pod/spark-distribution itself killing the executors on completion ? We can potentially mark executors as done if we get a terminal state for a driver. Thoughts ? 

Also I was trying to run a FailedSubmission retry example and looks like we might not be enforcing retryAttempts count on submission failure potentially due to race-conditions in state-management . I don't think that is related to this PR and I will work on a fix for that in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/199 . Let me know if you feel this PR can cause this behavior and I can take a deeper look.

docs/quick-start-guide.md
akhurana001: done

main.go
akhurana001: ran goimports on this and other files. I was using gofmt earlier which ended up reworking the imports in alphabetical order.

main.go
akhurana001: done

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go
akhurana001: sure, makes sense.

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go
akhurana001: I wanted to log more details since this is an error state. For now, I changed this to just log the name. Will probably add more details once we have the state management change out and this should be pretty rare

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go
akhurana001: sure, refactored as below and moved this out

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go

pkg/util/metrics.go
akhurana001: done, here and other files as well

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+		metricLabels[label] = validSpecLabels[label]
+		if metricLabels[label] == "" {
+			metricLabels[label] = "Unknown"
+		}
+	}
+	return metricLabels
+}
+
+func (sm *SparkAppMetrics) exportMetrics(oldApp, newApp *v1alpha1.SparkApplication) {
+
```
liyinan926: This line can be removed.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+	metricLabels := fetchMetricLabels(newApp.Labels, sm.Labels)
+	glog.V(2).Infof("Posting Metrics for %s. OldStatus: %v NewStatus: %v", newApp.GetName(), oldApp.Status, newApp.Status)
+
+	oldStatus := oldApp.Status.AppState.State
```
liyinan926: Can you rename them to `oldState` and `newState`, respectively?

pkg/util/metrics.go
```diff
@@ -0,0 +1,184 @@
...
+
+type PositiveGauge struct {
+	mux         sync.RWMutex
+	name        string
+	gaugeMetric *prometheus.GaugeVec
+}
+
+// Gauge with conditional decrement ensuring its value is never negative.
+func NewPositiveGauge(name string, description string, labels []string) *PositiveGauge {
+
```
liyinan926: This line can be removed.

docs/quick-start-guide.md
```diff
@@ -50,6 +50,41 @@ $ kubectl describe deployment sparkoperator -n sparkoperator
 
 ```
 
+### Metrics
+
+The operator exposes a set of metrics via the metric endpoint to be scraped by `Prometheus`. This is disabled by default and can be enabled by setting  the flag `-enable-metrics=true` and running 
+ `kubectl apply -f manifest/spark-operator.yaml`. You will also need to add the prometheus annotations `prometheus.io/scrape`
+  and `prometheus.io/port` which are used by Prometheus to scrape the metric endpoint if you are enabling metrics.
+  
+If enabled, the operator generates the following metrics:
```
liyinan926: I suggest we use a table that explains what each metric is for.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+		labels,
+	)
+	sparkAppExecutorSuccessCount := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: util.CreateValidMetricNameLabel(prefix, "spark_app_executor_success_count"),
+			Help: "Spark App Successful Executor Count via the Operator",
+		},
+		labels,
+	)
+	sparkAppRunningCount := util.NewPositiveGauge(util.CreateValidMetricNameLabel(prefix, "spark_app_running_count"),
```
liyinan926: Why are these two gauges, whereas the other app-related metrics are `CounterVec`?

pkg/controller/sparkapplication/controller.go
```diff
@@ -501,15 +508,25 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
...
+		// Apply update
+		updateFunc(&toUpdate.Status)
+
+		// Let's keep the old App status if this is not a valid transition.
+		if !isValidDriverStatusTransition(original.Status.AppState.State, toUpdate.Status.AppState.State) {
+			toUpdate.Status.AppState = original.Status.AppState
+		}
+		updated, err := c.tryUpdateStatus(original, toUpdate)
 		if err == nil {
+			c.postMetrics(original, updated)
```
liyinan926: We can add a condition around `postMetrics`, i.e., if `updated != nil` we will do `postMetrics`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -523,17 +540,30 @@ func (c *Controller) updateSparkApplicationStatusWithRetries(
...
-		glog.Errorf("failed to update SparkApplication %s: %v", toUpdate.Name, lastUpdateErr)
+		glog.Errorf("failed to update SparkApplication %s: %v", original.Name, lastUpdateErr)
 	}
 
 	return nil
 }
 
+func (c *Controller) postMetrics(oldApp, newApp *v1alpha1.SparkApplication) {
+
+	if c.metrics == nil {
```
liyinan926: Like mentioned above, these two conditions can be moved out of this function.

liyinan926:
> Based on my testing, the executor running count is off because the operator doesn't seem to be getting all executor events. I am guessing this is due to the driver pod/spark-distribution itself killing the executors on completion ? We can potentially mark executors as done if we get a terminal state for a driver. Thoughts ?

What if you collect executor metrics in `SparkPodMonitor` instead of when updating application status?

liyinan926:
> Also I was trying to run a FailedSubmission retry example and looks like we might not be enforcing retryAttempts count on submission failure potentially due to race-conditions in state-management . I don't think that is related to this PR and I will work on a fix for that in #199 . Let me know if you feel this PR can cause this behavior and I can take a deeper look.

Let's not worry about that in this PR. I will also take a look at that issue.

kumare3:
@liyinan926 I agree that we should maintain readability and quality of code. Could we achieve this through something like a linter? https://github.com/alecthomas/gometalinter

We could set that up as the CI process

liyinan926:
@kumare3 making style and correctness checks part of the CI process using tools like `go vet` or `go lint` sounds like a great idea.

kumare3:
@liyinan926 since you are the owner of the repo, can you setup something, and we can all follow

liyinan926:
@kumare3 Sure.

liyinan926:
@kumare3 @akhurana001 I will try setting something up for future PRs. For this PR, thanks for fixing the styles and formatting manually. The code style should look good to me after the fixes. Thanks!

pkg/controller/sparkapplication/controller.go
akhurana001: removed this function

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: Counters can only be incremented while Gauges can be both incremented/decremented hence for running count, we need Gauges

docs/quick-start-guide.md
akhurana001: makes sense, done

pkg/util/metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+
+	util.RegisterMetric(sparkAppSubmitCount)
+	util.RegisterMetric(sparkAppSuccessCount)
+	util.RegisterMetric(sparkAppFailureCount)
+	util.RegisterMetric(sparkAppSuccessExecutionTime)
+	util.RegisterMetric(sparkAppFailureExecutionTime)
+	util.RegisterMetric(sparkAppExecutorFailureCount)
+	util.RegisterMetric(sparkAppExecutorSuccessCount)
+
+	metricBundle := &SparkAppMetrics{
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+		sparkAppFailureExecutionTime,
+		sparkAppExecutorRunningCount,
+		sparkAppExecutorFailureCount,
+		sparkAppExecutorSuccessCount,
+	}
+	return metricBundle
+}
+
+func fetchMetricLabels(specLabels map[string]string, labels []string) map[string]string {
+
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+import (
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1alpha1"
+	"k8s.io/spark-on-k8s-operator/pkg/util"
+)
+
+type SparkAppMetrics struct {
```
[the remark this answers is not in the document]
akhurana001: makes sense, done

pkg/util/metrics_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPositiveGauge_EmptyLabels(t *testing.T) {
+
```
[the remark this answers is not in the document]
akhurana001: removed

pkg/util/metrics_test.go
```diff
@@ -0,0 +1,70 @@
...
+	assert.Equal(t, fetchGaugeValue(gauge.gaugeMetric, emptyMap), float64(0))
+
+	gauge.Inc(emptyMap)
+	assert.Equal(t, fetchGaugeValue(gauge.gaugeMetric, emptyMap), float64(1))
+	gauge.Dec(map[string]string{})
+	assert.Equal(t, fetchGaugeValue(gauge.gaugeMetric, emptyMap), float64(0))
+}
+
+func TestPositiveGauge_WithLabels(t *testing.T) {
+
```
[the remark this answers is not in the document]
akhurana001: removed

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,233 @@
...
+	SparkAppSuccessExecutionTime *prometheus.SummaryVec
+	SparkAppFailureExecutionTime *prometheus.SummaryVec
+
+	SparkAppExecutorRunningCount *util.PositiveGauge
+	SparkAppExecutorFailureCount *prometheus.CounterVec
+	SparkAppExecutorSuccessCount *prometheus.CounterVec
+}
+
+func NewSparkAppMetrics(prefix string, labels []string) *SparkAppMetrics {
+
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/controller.go
```diff
@@ -148,7 +157,6 @@ func (c *Controller) Start(workers int, stopCh <-chan struct{}) error {
 		// the worker after one second.
 		go wait.Until(c.runWorker, time.Second, stopCh)
 	}
-
```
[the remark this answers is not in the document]
akhurana001: added back

pkg/controller/sparkapplication/controller.go
```diff
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-
```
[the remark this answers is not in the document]

main.go
```diff
@@ -20,28 +20,28 @@ package main
...
 import (
 	"flag"
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
 	"time"
 
 	"github.com/golang/glog"
-
```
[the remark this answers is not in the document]
akhurana001: done

akhurana001:
> Based on my testing, the executor running count is off because the operator doesn't seem to be getting all executor events. I am guessing this is due to the driver pod/spark-distribution itself killing the executors on completion ? We can potentially mark executors as done if we get a terminal state for a driver. Thoughts ?
> 
> What if you collect executor metrics in SparkPodMonitor instead of when updating application status?

Based on my debugging,  we are getting Executor Pod Deleted updates with podStatus still as Pending/Running which is why in  both SparkApplication Status and metrics, we see some of the executors stuck as pending/running. I am guessing this happens because on completion, the driver deletes the executors which might still be pending/running ? 
This can potentially be handled similar to the way we default to `PodFailed` state on Driver Deletion in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/spark_pod_monitor.go#L164

If this sounds ok, I can test this a little more and create a PR with fix for both Executor Status in SparkApplication as well as metrics.

akhurana001:
Also FYI, I created a separate `spark-operator-with-metrics.yaml` with all the additional configurations required to set-up the metrics. 

Since we don't want to turn these on by default, I reverted the old `spark-operator.yaml` to not have any metric related configs.

mrow4a:
Good idea, this will also play well with Helm

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+	// Transform spec labels since our labels names might be not same as specLabels if we removed invalid characters.
+	validSpecLabels := make(map[string]string)
+	for labelKey, v := range specLabels {
+		newKey := util.CreateValidMetricNameLabel("", labelKey)
+		validSpecLabels[newKey] = v
+	}
+
+	metricLabels := make(map[string]string)
+	for _, label := range labels {
+		metricLabels[label] = validSpecLabels[label]
```
liyinan926: It's better to do the following instead:
```
if value, ok := validSpecLabels[label]; ok {
  metricLabels[label] = value
} else {
  metricLabels[label] = "Unknown"
}
```

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+		if metricLabels[label] == "" {
+			metricLabels[label] = "Unknown"
+		}
+	}
+	return metricLabels
+}
+
+func (sm *sparkAppMetrics) exportMetrics(oldApp, newApp *v1alpha1.SparkApplication) {
+	metricLabels := fetchMetricLabels(newApp.Labels, sm.Labels)
+	glog.V(2).Infof("Posting Metrics for %s. OldStatus: %v NewStatus: %v", newApp.GetName(), oldApp.Status, newApp.Status)
```
liyinan926: Can we change all occurrences of `posting` to `exporting`?

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+				glog.Errorf("Error while posting metrics: %v", err)
+			} else {
+				m.Inc()
+			}
+		}
+	}
+
+	// Potential Executor status updates
+	for executor, newExecState := range newApp.Status.ExecutorState {
+		if newExecState == v1alpha1.ExecutorRunningState && (isValidExecutorStateTransition(oldApp.Status.ExecutorState[executor], newExecState)) {
```
liyinan926: Can we use a `switch` block that checks `newExecState` instead of embedding the check into the if blocks?

pkg/controller/sparkapplication/sparkapp_metrics_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	wg.Wait()
+	assert.Equal(t, float64(10), fetchCounterValue(metrics.SparkAppSubmitCount, app1))
+	assert.Equal(t, float64(10), fetchCounterValue(metrics.SparkAppFailureCount, app1))
+	assert.Equal(t, float64(10), fetchCounterValue(metrics.SparkAppSuccessCount, app1))
+	assert.Equal(t, float64(10), fetchCounterValue(metrics.SparkAppExecutorFailureCount, app1))
+	assert.Equal(t, float64(10), fetchCounterValue(metrics.SparkAppExecutorSuccessCount, app1))
+	assert.Equal(t, float64(5), metrics.SparkAppExecutorRunningCount.Value(app1))
+	assert.Equal(t, float64(5), metrics.SparkAppRunningCount.Value(app1))
+
```
liyinan926: This empty line can be removed.

pkg/controller/sparkapplication/state_validation.go
```diff
@@ -0,0 +1,78 @@
...
+			v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedState},
+
+		v1alpha1.SubmittedState: {v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedState},
+		v1alpha1.RunningState:   {v1alpha1.CompletedState, v1alpha1.FailedState},
+		// The application can be potentially restarted based on the RestartPolicy
+		v1alpha1.FailedState:    {v1alpha1.SubmittedState},
+		v1alpha1.CompletedState: {v1alpha1.SubmittedState},
+	}
+
+	// Valid Executor State Transitions: map with valid paris of oldState-> {newStates}
```
liyinan926: Typo: `paris` -> `pairs`.

pkg/controller/sparkapplication/state_validation.go
```diff
@@ -0,0 +1,78 @@
...
+		v1alpha1.NewState: {v1alpha1.UnknownState, v1alpha1.SubmittedState,
+			v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedSubmissionState, v1alpha1.FailedState},
+
+		v1alpha1.UnknownState: {v1alpha1.NewState, v1alpha1.SubmittedState,
+			v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedSubmissionState, v1alpha1.FailedState},
+
+		v1alpha1.FailedSubmissionState: {v1alpha1.UnknownState, v1alpha1.SubmittedState,
+			v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedState},
+
+		v1alpha1.SubmittedState: {v1alpha1.RunningState, v1alpha1.CompletedState, v1alpha1.FailedState},
```
liyinan926: Can we also use empty lines to separate the following entries out for consistency?

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+func CreateValidMetricNameLabel(prefix, name string) string {
+	// "-" aren't valid characters for prometheus metric names or labels
+	return strings.Replace(prefix+name, "-", "_", -1)
+}
+
+// Best effort metric registration with Prometheus.
+func RegisterMetric(metric prometheus.Collector) {
+	if err := prometheus.Register(metric); err != nil {
+		glog.Warningf("Error while registering Prometheus metric: [%v]", err)
```
liyinan926: You might want to use `Errorf` instead.

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+}
+
+type WorkQueueMetrics struct {
+	prefix string
+}
+
+func InitializeMetrics(endpoint string, port string, prefix string) {
+
```
liyinan926: This empty line can be removed.

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	workQueueMetrics := WorkQueueMetrics{prefix: prefix}
+	workqueue.SetProvider(&workQueueMetrics)
+}
+
+// Depth Metric for kubernetes workqueue
+func (p *WorkQueueMetrics) NewDepthMetric(name string) workqueue.GaugeMetric {
+	depthMetric := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: CreateValidMetricNameLabel(p.prefix, name+"_depth"),
+		Help: "Current depth of workqueue: " + name,
```
liyinan926: Can we use `fmt.Sprintf` for formatting the `Help` strings?

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+
+	} else {
+		glog.V(2).Infof("Incrementing %s with labels %s", c.name, labelMap)
+		m.Inc()
+	}
+}
+
+// Decrement the metric only if its positive for the labels specified
+func (c *PositiveGauge) Dec(labelMap map[string]string) {
+
```
liyinan926: This empty line can be removed.

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+func (c *PositiveGauge) Value(labelMap map[string]string) float64 {
+	c.mux.RLock()
+	defer c.mux.RUnlock()
+	return fetchGaugeValue(c.gaugeMetric, labelMap)
+}
+
+// Increment the Metric for the labels specified
+func (c *PositiveGauge) Inc(labelMap map[string]string) {
+
```
liyinan926: This empty line can be removed.

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	if m, err := c.gaugeMetric.GetMetricWith(labelMap); err != nil {
+		glog.Errorf("Error while posting metrics: %v", err)
+
```
liyinan926: This empty line can be removed.

pkg/util/metrics.go
```diff
@@ -0,0 +1,182 @@
...
+type WorkQueueMetrics struct {
+	prefix string
+}
+
+func InitializeMetrics(endpoint string, port string, prefix string) {
+
+	// Start the metrics endpoint for Prometheus to scrape
+	http.Handle(endpoint, promhttp.Handler())
+	go http.ListenAndServe(port, nil)
+	glog.Infof("Started Metrics server at localhost%s%s", port, endpoint)
```
liyinan926: Should it be `"Started metrics server at %s:%s", endpoint, port)`?

liyinan926:
> Based on my debugging, we are getting Executor Pod Deleted updates with podStatus still as Pending/Running which is why in both SparkApplication Status and metrics, we see some of the executors stuck as pending/running. I am guessing this happens because on completion, the driver deletes the executors which might still be pending/running? This can potentially be handled similar to the way we default to PodFailed state on Driver Deletion in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/spark_pod_monitor.go#L164

Let me think about a bit more on this.

liyinan926:
@akhurana001 Please rebase onto the latest master to resolve the conflict.

main.go
```diff
@@ -108,7 +122,7 @@ func main() {
 		time.Duration(*resyncInterval)*time.Second,
 		factoryOpts...)
 	applicationController := sparkapplication.NewController(
-		crdClient, kubeClient, apiExtensionsClient, factory, *submissionRunnerThreads, *namespace)
+		crdClient, kubeClient, apiExtensionsClient, factory, *submissionRunnerThreads, *enableMetrics, *metricsPrefix, metricsLabels, *namespace)
```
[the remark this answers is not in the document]
akhurana001: makes sense, done

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -287,10 +296,18 @@ func TestOnDelete(t *testing.T) {
 }
 
 func TestProcessSingleDriverStateUpdate(t *testing.T) {
+
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+		prometheus.CounterOpts{
+			Name: util.CreateValidMetricNameLabel(prefix, "spark_app_failure_count"),
+			Help: "Spark App Failure Count via the Operator",
+		},
+		labels,
+	)
+	sparkAppSuccessExecutionTime := prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name: util.CreateValidMetricNameLabel(prefix, "spark_app_success_execution_time_microseconds"),
+			Help: "Spark App Runtime via the Operator",
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+		},
+		labels,
+	)
+	sparkAppFailureExecutionTime := prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name: util.CreateValidMetricNameLabel(prefix, "spark_app_failure_execution_time_microseconds"),
+			Help: "Spark App Runtime via the Operator",
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1alpha1"
+	"k8s.io/spark-on-k8s-operator/pkg/util"
+)
+
+type sparkAppMetrics struct {
+	Labels               []string
+	prefix               string
+	SparkAppSubmitCount  *prometheus.CounterVec
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1alpha1"
+	"k8s.io/spark-on-k8s-operator/pkg/util"
+)
+
+type sparkAppMetrics struct {
+	Labels               []string
```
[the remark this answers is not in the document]
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,229 @@
...
+	util.RegisterMetric(sparkAppSubmitCount)
+	util.RegisterMetric(sparkAppSuccessCount)
+	util.RegisterMetric(sparkAppFailureCount)
+	util.RegisterMetric(sparkAppSuccessExecutionTime)
+	util.RegisterMetric(sparkAppFailureExecutionTime)
+	util.RegisterMetric(sparkAppExecutorFailureCount)
+	util.RegisterMetric(sparkAppExecutorSuccessCount)
+
+	return &sparkAppMetrics{
+		labels,
```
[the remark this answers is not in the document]
akhurana001: These are just the label names passed in while registering the metrics. The values are passed in to the exportMetrics function

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics_test.go
akhurana001: done

pkg/controller/sparkapplication/state_validation.go
akhurana001: done

pkg/controller/sparkapplication/state_validation.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: So the endpoint is actually just the path. Right now this log line looks like `Started Metrics server at localhost:10254/metrics` which looks easy to understand  to me. 

Also got rid of `:` from user-passed port and explicitly appending it to the user-specified port.

pkg/controller/sparkapplication/controller.go
akhurana001: We actually need this and I have reverted it back to make a deep copy inside the loop (https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/227/commits/f600336b353287b0cd983148071c2bf9640b1135)

This is because for metrics,  we need the oldState of the CRD which we actually updated and since we run `updateFunc(&toUpdate.Status)` , `toUpdate` is no longer the oldState which was read from the API server nor is `original`. Let me know if you have any questions.

docs/quick-start-guide.md
```diff
@@ -50,6 +50,43 @@ $ kubectl describe deployment sparkoperator -n sparkoperator
...
+| `spark_app_failure_execution_time_microseconds` |Execution time for applications which failed. |
+| `spark_app_executor_success_count` | Total number of Spark Executors which completed successfully. |
+| `spark_app_executor_failure_count` | Total number of Spark Executors which failed. |
+| `spark_app_executor_running_count` | Total number of Spark Executors which are currently running. |
+
+The following is a list of all the configurations the operators supports for metrics: 
+
+```bash
+-enable-metrics=true
+-metrics-port=10254
```
liyinan926: It's worth mentioning that if users use a different port, they also need to update the annotation `prometheus.io/port` and `containerPort` in the yaml.

pkg/controller/sparkapplication/controller.go
```diff
@@ -501,20 +507,33 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
...
-		updated, err := c.tryUpdateStatus(original, toUpdate, updateFunc)
+		toUpdate := original.DeepCopy()
+
+		glog.V(2).Infof("Trying to update SparkApplication %s", original.Name)
+		// Apply update
+		updateFunc(&toUpdate.Status)
+
+		// Let's keep the old App status if this is not a valid transition.
+		if !isValidDriverStateTransition(original.Status.AppState.State, toUpdate.Status.AppState.State) {
+			toUpdate.Status.AppState = original.Status.AppState
```
liyinan926: We might need to log a `warning` here with the old and new state.

pkg/controller/sparkapplication/controller.go
```diff
@@ -501,20 +507,33 @@ func (c *Controller) processSingleExecutorStateUpdate(update *executorStateUpdat
 func (c *Controller) updateSparkApplicationStatusWithRetries(
 	original *v1alpha1.SparkApplication,
 	updateFunc func(*v1alpha1.SparkApplicationStatus)) *v1alpha1.SparkApplication {
-	toUpdate := original.DeepCopy()
-
 	var lastUpdateErr error
 	for i := 0; i < maximumUpdateRetries; i++ {
-		updated, err := c.tryUpdateStatus(original, toUpdate, updateFunc)
+		toUpdate := original.DeepCopy()
```
liyinan926: Since the client always return a copy, the original approach that makes a copy once outside the loop and assign what the client returns to `toUpdate` is more desirable.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -287,10 +297,17 @@ func TestOnDelete(t *testing.T) {
 }
 
 func TestProcessSingleDriverStateUpdate(t *testing.T) {
+	type metrics struct {
```
liyinan926: Looks like the same type is used and defined in `TestProcessSingleAppStateUpdate` below, and a similar type is defined in `TestProcessSingleExecutorStateUpdate ` for executors. I think we can define a single type at the package level for this and use it for all three tests.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,235 @@
...
+	sparkAppSuccessExecutionTime *prometheus.SummaryVec
+	sparkAppFailureExecutionTime *prometheus.SummaryVec
+
+	sparkAppExecutorRunningCount *util.PositiveGauge
+	sparkAppExecutorFailureCount *prometheus.CounterVec
+	sparkAppExecutorSuccessCount *prometheus.CounterVec
+}
+
+func newSparkAppMetrics(prefix string, labels []string) *sparkAppMetrics {
+	for i, label := range labels {
```
liyinan926: To be safe, let's make and use a copy of `labels` here. The in-place changes to `labels` here are visible outside the caller of this function.

pkg/util/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+
+type PositiveGauge struct {
+	mux         sync.RWMutex
+	name        string
+	gaugeMetric *prometheus.GaugeVec
+}
+
+// Gauge with conditional decrement ensuring its value is never negative.
+func NewPositiveGauge(name string, description string, labels []string) *PositiveGauge {
+	for i, label := range labels {
```
liyinan926: Ditto. Let's make and use a copy of `labels`.

pkg/controller/sparkapplication/sparkapp_metrics.go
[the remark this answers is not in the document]
liyinan926: I meant you should construct this `sparkAppMetrics` in the following form:
```
return &sparkAppMetrics{
  labels: labels,
  prefix: prefix,
  sparkAppSubmitCount: sparkAppSubmitCount,
  ...
```

This is much more robust than rely in the right order of the fields when initializing such struct.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,235 @@
...
+	case v1alpha1.FailedSubmissionState:
+		fallthrough
+	case v1alpha1.FailedState:
+		if isValidDriverStateTransition(oldState, newState) {
+			if !newApp.Status.SubmissionTime.Time.IsZero() && !newApp.Status.CompletionTime.Time.IsZero() {
+				d := newApp.Status.CompletionTime.Time.Sub(newApp.Status.SubmissionTime.Time)
+				if m, err := sm.sparkAppFailureExecutionTime.GetMetricWith(metricLabels); err != nil {
+					glog.Errorf("Error while exporting metrics: %v", err)
+				} else {
+					m.Observe(float64(d / time.Microsecond))
```
liyinan926: Can we do `float64(d) / time. Microsecond` instead here?

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	case v1alpha1.CompletedState:
+		if isValidDriverStateTransition(oldState, newState) {
+			if !newApp.Status.SubmissionTime.Time.IsZero() && !newApp.Status.CompletionTime.Time.IsZero() {
+				d := newApp.Status.CompletionTime.Time.Sub(newApp.Status.SubmissionTime.Time)
+
+				if m, err := sm.sparkAppSuccessExecutionTime.GetMetricWith(metricLabels); err != nil {
+					glog.Errorf("Error while exporting metrics: %v", err)
+				} else {
+					m.Observe(float64(d / time.Microsecond))
```
liyinan926: Can we do `float64(d) / time. Microsecond` instead here?

pkg/controller/sparkapplication/sparkapp_metrics_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"net/http"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSparkAppMetrics(t *testing.T) {
+	http.DefaultServeMux = new(http.ServeMux)
+	// Invalid label.
```
liyinan926: What does this comment mean?

pkg/util/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+*/
+
+package util
+
+import (
+	"net/http"
+	"strings"
+	"sync"
+
+	"fmt"
```
liyinan926: This import should be at the top.

pkg/util/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type MetricConfig struct {
+	MetricsEndpoint string
+	MetricsPort     string
+	MetricsPrefix   string
+	MetricsLabels   []string
+}
+
+type PositiveGauge struct {
```
liyinan926: Can we add a doc for this like `PositiveGauge` is a variant of Prometheus Gauge that only holds non-negative values. BTW: this can be zero, right?

pkg/util/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+
+// Decrement the metric only if its positive for the labels specified
+func (c *PositiveGauge) Dec(labelMap map[string]string) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	// Decrement only if positive
+	val := fetchGaugeValue(c.gaugeMetric, labelMap)
+	if val > 0 {
+		glog.V(2).Infof("Decrementing %s with labels %s metricVal to %v", c.name, labelMap, val-1)
```
liyinan926: Looks like indention is not consistent: 2 in some places, whereas 4 in other places. Please run `go fmt ./...` to format the code.

docs/quick-start-guide.md
akhurana001: sure, done

pkg/controller/sparkapplication/controller.go
akhurana001: done, just FYI, this will be pretty common since same state transitions are also invalid. e.g. we will log going from RUNNING -> RUNNING.

pkg/controller/sparkapplication/controller_test.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: done

pkg/util/metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: incompatible types: time.Microsecond is duration

pkg/controller/sparkapplication/sparkapp_metrics.go
akhurana001: same as above

pkg/util/metrics.go
akhurana001: sure, i have been running `goimports` on all files.

pkg/util/metrics.go
akhurana001: yes, this can be zero. Added comment

pkg/util/metrics.go
akhurana001: done

pkg/controller/sparkapplication/sparkapp_metrics_test.go
akhurana001: So prometheus doesn't support "-" in metric or label names. Since we were specifying "-" in our workQueue names, I added a helper method to convert "-" to "_" and not error out (https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/227/files#diff-48f67873a59be740236d811d032ce6f6R33) .. This test just validates that we are correctly handling that case as well.

pkg/controller/sparkapplication/controller.go
akhurana001: As I mentioned, this works fine if we aren't publishing metrics.

For metrics as well as to verify if the stateTransition is valid, we need to maintain the last returned by crdClient. 
We can set `toUpdate` as the last-state returned by the crdClient but then in the next attempt, we also do `	updateFunc(&toUpdate.Status)`  at https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/227/files/f600336b353287b0cd983148071c2bf9640b1135#diff-b7eea120d7ac0e8587781b863b7db8e4R516 . At this point, we lose the last state returned by the `crdClient`. 

For us to  validate `isValidDriverStateTransition` and post metrics, we need to maintain that last state returned by the `crdClient` which is what `original` will be at this point ? 


I might be missing something but this is my understanding of the logic here. I be happy to jump on a quick call if need be.

liyinan926: Makes sense.

pkg/controller/sparkapplication/controller.go
liyinan926: Oh, right, I forgot that transition. Let's add another condition before the check for validity like the following:
```
if toUpdate.Status.AppState.State != original.Status.AppState.State && !isValidDriverStateTransition(original.Status.AppState.State, toUpdate.Status.AppState.State)
```

akhurana001:
Thanks @liyinan926 . Please also update the image in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/manifest/spark-operator-with-metrics.yaml#L48 once you build the operator with metrics.

liyinan926:
Will do.

PR #243, opened 2018-08-03
 Extend sparkctl to stream SparkApplication events

mrow4a:
SparkOperator will additionally 
trigger events on driver/executor `pending`, `running` and `unknown`.

- [x] Add documentation when logic approved

Now `$ sparkctl events <name>` will trigger `watch` and display:
```
Events: Type	Reason	Age	Message
Normal	SparkApplicationAdded	1h	SparkApplication spark-pi was added, enqueued it for submission
Normal	SparkApplicationTerminated	1h	SparkApplication spark-pi terminated with state: COMPLETED
Normal	SparkDriverCompleted	1h	Driver spark-pi-1533304888525-driver completed


Normal	SparkExecutorCompleted	1h	Executor spark-pi-1533304888525-exec-1 completed
```

It is also good in early catching early mistakes
```
Normal	SparkApplicationDeleted	6s	SparkApplication spark-pi was deleted
Normal	SparkApplicationAdded	6s	SparkApplication spark-pi was added, enqueued it for submission
Warning	SparkApplicationSubmissionFailed	6s	SparkApplication spark-pi failed submission: Error: Cannot load main class from JAR https://storage.googleapis.com/spark-on-k8s-cluster/spark-app-dependencies/default/spark-pi/spark-service-examples_2.11-0.3.0.jar with URI https. Please specify a class through --class.
```

Additionaly, `$ sparkctl status spark-pi` will pre-format to a bit different status:
```
application state:
+-----------+---------------+---------------+-------------------------------+-----------+---------+
|   STATE   | SUBMITTED AGE | COMPLETED AGE |          DRIVER POD           | DRIVER UI | RETRIES |
+-----------+---------------+---------------+-------------------------------+-----------+---------+
| COMPLETED | 10m           | 10m           | spark-pi-1533309141745-driver | N.A.      |       0 |
+-----------+---------------+---------------+-------------------------------+-----------+---------+
executor state:
+-------------------------------+-----------+
|         EXECUTOR POD          |   STATE   |
+-------------------------------+-----------+
| spark-pi-1533309141745-exec-1 | COMPLETED |
+-------------------------------+-----------+
```

@liyinan926

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -675,19 +675,35 @@ func (c *Controller) getNodeExternalIP(nodeName string) string {
 
 func (c *Controller) recordDriverEvent(
 	app *v1alpha1.SparkApplication, phase apiv1.PodPhase, name string) {
-	if phase == apiv1.PodSucceeded {
+
+	switch phase {
+	case apiv1.PodSucceeded:
 		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkDriverCompleted", "Driver %s completed", name)
-	} else if phase == apiv1.PodFailed {
+	case apiv1.PodPending:
+		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkDriverPending", "Driver %s is pending", name)
+	case apiv1.PodRunning:
+		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkDriverRunning", "Driver %s is running", name)
+	case apiv1.PodFailed:
 		c.recorder.Eventf(app, apiv1.EventTypeWarning, "SparkDriverFailed", "Driver %s failed", name)
+	case apiv1.PodUnknown:
+		c.recorder.Eventf(app, apiv1.EventTypeWarning, "SparkDriverUnknownState", "Driver %s in unknown state", name)
 	}
 }
 
 func (c *Controller) recordExecutorEvent(
 	app *v1alpha1.SparkApplication, state v1alpha1.ExecutorState, name string) {
-	if state == v1alpha1.ExecutorCompletedState {
+		
+	switch state {
+	case v1alpha1.ExecutorCompletedState:
 		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorCompleted", "Executor %s completed", name)
-	} else if state == v1alpha1.ExecutorFailedState {
+	case v1alpha1.ExecutorPendingState:
+		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorPending", "Executor %s is pending", name)
+	case v1alpha1.ExecutorRunningState:
+		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorRunning", "Executor %s is running", name)
+	case v1alpha1.ExecutorFailedState:
 		c.recorder.Eventf(app, apiv1.EventTypeWarning, "SparkExecutorFailed", "Executor %s failed", name)
+	case v1alpha1.ExecutorUnknownState:
+		c.recorder.Eventf(app, apiv1.EventTypeWarning, "SparkExecutorUnknownState", "Executor %s in unknown state", name)
 	}
 }
 
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -692,7 +692,7 @@ func (c *Controller) recordDriverEvent(
 
 func (c *Controller) recordExecutorEvent(
 	app *v1alpha1.SparkApplication, state v1alpha1.ExecutorState, name string) {
-		
+
 	switch state {
 	case v1alpha1.ExecutorCompletedState:
 		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorCompleted", "Executor %s completed", name)
--- a/sparkctl/README.md
+++ b/sparkctl/README.md
@@ -101,6 +101,20 @@ Usage:
 $ sparkctl status <SparkApplication name>
 ```
 
+### Events
+
+`events` is a sub command of `sparkctl` for listing `SparkApplication` events in the namespace 
+specified by `--namespace`. 
+
+The `events` command also supports streaming the events with the `--follow` or `-f` flag. 
+The command will display events since last creation of the `SparkApplication`
+for the specific `name`, and continues to stream events even if `ResourceVersion` changes.
+
+Usage:
+```bash
+$ sparkctl events <SparkApplication name> [-f]
+```
+
 ### Log
 
 `log` is a sub command of `sparkctl` for fetching the logs of a pod of `SparkApplication` with the given name in the 
--- a/sparkctl/cmd/events.go
+++ b/sparkctl/cmd/events.go
@@ -1,0 +1,175 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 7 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -677,17 +677,29 @@ func (c *Controller) recordDriverEvent(
 	app *v1alpha1.SparkApplication, phase apiv1.PodPhase, name string) {
 	if phase == apiv1.PodSucceeded {
 		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkDriverCompleted", "Driver %s completed", name)
+	} else if phase == apiv1.PodPending {
```
liyinan926: Let's switch to a `case` block.

pkg/controller/sparkapplication/controller.go
```diff
@@ -677,17 +677,29 @@ func (c *Controller) recordDriverEvent(
...
+	} else if phase == apiv1.PodUnknown {
+		c.recorder.Eventf(app, apiv1.EventTypeWarning, "SparkDriverUnknownState", "Driver %s in unknown state", name)
 	}
 }
 
 func (c *Controller) recordExecutorEvent(
 	app *v1alpha1.SparkApplication, state v1alpha1.ExecutorState, name string) {
 	if state == v1alpha1.ExecutorCompletedState {
 		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorCompleted", "Executor %s completed", name)
+	} else if state == v1alpha1.ExecutorRunningState {
```
liyinan926: Ditto. Let's switch to a `case` block..

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2017 Google LLC
```
liyinan926: `2017` -> `2018`

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/kubernetes/pkg/util/interrupt"
+	crdclientset "k8s.io/spark-on-k8s-operator/pkg/client/clientset/versioned"
+	"strings"
```
liyinan926: These two imports should go with the other built-in ones at the top.

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,105 @@
...
+	},
+}
+
+func doShowEvents(name string, crdClientset crdclientset.Interface) error {
+	app, err := getSparkApplication(name, crdClientset)
+	if err != nil {
+		return fmt.Errorf("failed to get SparkApplication %s: %v", name, err)
+	}
+
+	kubeClientset, err := getKubeClient()
```
liyinan926: Let's move this to the `Run` function body and pass the `kubeClient` in as in `create.go`.

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return err
+	}
+
+	stringAppKind := string(app.Kind)
```
liyinan926: I don't think you need to do an explicit type conversion here and below. `GetFieldSelector ` and `ResourceVersion ` should be able to take `&app.Kind` and `app.ResourceVersion`.

sparkctl/cmd/status.go
```diff
@@ -54,32 +54,37 @@ func doStatus(name string, crdClientset crdclientset.Interface) error {
...
 		table.SetHeader([]string{"Executor Pod", "State"})
-		table.SetColumnColor(tablewriter.Colors{tablewriter.FgBlueColor}, tablewriter.Colors{tablewriter.FgGreenColor})
 		for executorPod, state := range app.Status.ExecutorState {
 			table.Append([]string{executorPod, string(state)})
 		}
 		table.Render()
 	}
+
+	if app.Status.AppState.ErrorMessage != "" {
+		fmt.Printf("application error message: %s\n", app.Status.AppState.ErrorMessage)
```
liyinan926: Let's add an empty line before this, e.g., `\napplication error message`.

sparkctl/cmd/utils.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2017 Google LLC
```
liyinan926: `2017` -> `2018`.

mrow4a:
@liyinan926 addressed the review

sparkctl/README.md
```diff
@@ -101,6 +101,18 @@ Usage:
 $ sparkctl status <SparkApplication name>
 ```
 
+### Events
+
+`events` is a sub command of `sparkctl` for streaming `SparkApplication` events in the namespace 
+specified by `--namespace`. The command will display events since last creation of the `SparkApplication`
+for the specific `name`, and continues to stream events even 
+if `ResourceVersion` or `UID` changes due to completion or deletion of `SparkApplication`.
```
liyinan926: I don't think `UID` will change for the lifetime of a `SparkApplication` object. And does it make sense to continue streaming events even if the `SparkApplication` gets deleted?

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,106 @@
...
+	},
+}
+
+func doShowEvents(name string, crdClientset crdclientset.Interface, kubeClientset kubernetes.Interface) error {
+	app, err := getSparkApplication(name, crdClientset)
+	if err != nil {
+		return fmt.Errorf("failed to get SparkApplication %s: %v", name, err)
+	}
+
+	// Since ResourceVersion of SparkApplication changes on CREATE/SUCCESS, we take all events for
```
liyinan926: `ResourceVersion` changes whenever the object gets updated. Why this matters?

sparkctl/cmd/events.go
```diff
@@ -0,0 +1,106 @@
...
+	eventsInterface := kubeClientset.CoreV1().Events(Namespace)
+	selector := eventsInterface.GetFieldSelector(&app.Name, &app.Namespace, &appKind, nil)
+	options := metav1.ListOptions{FieldSelector: selector.String(), Watch: true}
+
+	events, err := eventsInterface.Watch(options)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Events: Type\tReason\tAge\tMessage\n")
```
liyinan926: I'm thinking it's better to use a `tablewriter` to print the events so the header and event lines are aligned column wise.

sparkctl/README.md
mrow4a: The problem I faced is that the stream after deletion event still continues. So I would need to check for "SparkApplicationDeleted" event so streaming terminates, or it will wait forever (till timeout)

sparkctl/cmd/events.go
mrow4a: The problem is that I cannot filter by "at least as early as resource version" since after application completes I will not see history for this application.

sparkctl/cmd/events.go
mrow4a: I tried, but this would mean I will need to render the "small event raw" each time as a table. I can do it, but just thought it would be too big overhead.

mrow4a: if you do `kubectl get pods --watch` it watches for everything. Events in `kubectl describe` are per application uid indeed. I will play with `kubectl watch` to get some good practice. Thanks!

mrow4a: Ohh sorry, I think I misunderstood the question. We are watching for events, so I would need to render table with each new event comming. This is why I just print. What I can do, is to not stream and just print all events for specific UID.

liyinan926: I think we can also add an option `-watch` that if specified, enables watching of events. But by default, you just print all events available for a `SparkApplication`.

PR #247, opened 2018-08-07
Modified gencerts.sh script to be used in Helm chart automatic secret creation. 

yuchaoran2011:
Modified gencert.sh so that it can be run inside a pod container when running as a k8s Job. It can still be manually run the same way as before. This PR is to be reviewed together with this [commit](https://github.com/helm/charts/pull/6239/commits/de8d491d000edebc3bff1fc2c9c0b974d958b439).

### DIFF
--- a/hack/gencerts.sh
+++ b/hack/gencerts.sh
@@ -43,7 +43,7 @@ function parse_arguments {
       shift 2
       continue
       ;;
-      -p|--pod-container) # Whether the script is running inside a pod container or invoked manually
+      -p|--in-pod)
       export IN_POD=true
       shift 1
       continue
--- a/hack/gencerts.sh
+++ b/hack/gencerts.sh
@@ -25,7 +25,7 @@ function usage {
   Options:
   -h | --help               	   Display help information.
   -n | --namespace <namespace>   The namespace where the Spark operator is installed.			
-  -p | --pod-container           whether the script is running inside a pod container or manually invoked.
+  -p | --in-pod                  Whether the script is running inside a pod or not.
 EOF
 }
 
--- a/hack/gencerts.sh
+++ b/hack/gencerts.sh
@@ -17,11 +17,61 @@
 # Generates a CA certificate, a server key, and a server certificate signed by the CA.
 
 set -e
+SCRIPT=`basename ${BASH_SOURCE[0]}`
 
-NAMESPACE=$1
-if [ -z $NAMESPACE ]; then
-	NAMESPACE="sparkoperator"
-fi
+function usage {
+  cat<< EOF
+  Usage: $SCRIPT
+  Options:
+  -h | --help               	 This help info.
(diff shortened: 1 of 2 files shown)


hack/gencerts.sh
```diff
@@ -17,6 +17,61 @@
 # Generates a CA certificate, a server key, and a server certificate signed by the CA.
 
 set -e
+SCRIPT=`basename ${BASH_SOURCE[0]}`
+
+function usage {
+  cat<< EOF
+  Usage: $SCRIPT
+  Options:
+  -h | --help               	 This help info.
```
liyinan926: `Display help information.`. And please make the line indented consistently with other lines.

yuchaoran2011: Now fixed. Thanks!

liyinan926: Oh, I meant please also change the text to `Display help information.`.

yuchaoran2011: Ok. Now fixed too.

hack/gencerts.sh
```diff
@@ -17,6 +17,61 @@
...
 set -e
+SCRIPT=`basename ${BASH_SOURCE[0]}`
+
+function usage {
+  cat<< EOF
+  Usage: $SCRIPT
+  Options:
+  -h | --help               	   Display help information.
+  -n | --namespace <namespace>   The namespace where the Spark operator is installed.			
+  -p | --pod-container           whether the script is running inside a pod container or manually invoked.
```
liyinan926: Sorry one last comment, I think it's better to rename this to `--in-pod`, and change the description to `Whether the script is running inside a pod or not.`.

yuchaoran2011: Sure. Fixed.

hack/gencerts.sh
```diff
@@ -17,6 +17,61 @@
...
+      if [ -n "$2" ]; then
+        NAMESPACE="$2"
+      else
+        echo "-n or --namespace requires a value."
+        exit 1
+      fi
+      shift 2
+      continue
+      ;;
+      -p|--pod-container) # Whether the script is running inside a pod container or invoked manually
```
mrow4a: Excellent work! One thing from me, do we really need this comment, since there is an explanation above? Please change it to `--in-pod`

hack/gencerts.sh
```diff
@@ -17,6 +17,61 @@
...
+      export IN_POD=true
+      shift 1
+      continue
+      ;;
+      -h|--help)
+      usage
+      exit 0
+      ;;
+      --)              # End of all options.
```
mrow4a: Is it a sanity check?

hack/gencerts.sh
yuchaoran2011: Good point. Yeah I forgot to change it here. Now fixed.

hack/gencerts.sh
yuchaoran2011: Yes

PR #253, opened 2018-08-09
Changed subcommand events to event for sparkctl

liyinan926:
This is to be consistent with naming of the other subcommands.

@mrow4a


mrow4a:
@liyinan926 yes, thanks!

Issue #255, opened 2018-08-12
Automatic setup of exporting driver metrics to Prometheus


liyinan926:
Metrics of the drivers can be exposed through JMX using the `JmxSink`, and then exported to Prometheus through the [Prometheus JMX exporter](https://github.com/prometheus/jmx_exporter). The Prometheus configuration file needed by the JMX exporter can be generated by the operator and automatically mounted into the driver pod through a ConfigMap. Then the annotation `prometheus.io/scrape=true` should make the Prometheus server to scrape the metrics exposed by the driver pods. 

* [x] Implementation.
* [x] Documentation.

Issue #252, opened 2018-08-09
Better integration with Stackdriver Monitoring


liyinan926:
* [x] Automatic configuration for exposing internal Spark metrics to Prometheus: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/255.

andrewgdavis:
Is there a diagram that explains how this works?  Is this just exposing the metrics via a port that can be scraped by prometheus ... 

or do the metrics get aggregated and pushed back to the operator (which is scraped by prometheus)... 

or does it make more sense to make use of the prometheus gateway (pods in a completed state that can no longer be scraped)

PR #266, opened 2018-08-20
Add support for automatic configuration for exposing driver/executor metrics to Prometheus

liyinan926:
For #255 and #252.

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -162,3 +162,37 @@ const (
 	// form the path to the file referred to by HADOOP_TOKEN_FILE_LOCATION.
 	HadoopDelegationTokenFileName = "hadoop.token"
 )
+
+const DefaultMetricsProperties = `
+*.sink.jmx.class=org.apache.spark.metrics.sink.JmxSink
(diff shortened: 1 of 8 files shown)


pkg/config/constants.go
```diff
@@ -162,3 +162,37 @@ const (
...
+    labels:
+      app_namespace: "$1"
+      app_name: "$2"
+  - pattern: metrics<name=(\S+)/(\S+)\.driver\.spark\.streaming\.(\S+)\.(\S+)><>Value
+    name: spark_structured_streaming_driver_$4
+    labels:
+      app_namespace: "$1"
+      app_name: "$2"
+      query_name: "$3"
+  - pattern: metrics<name=(\S+)/(\S+)\.(\S+)\.executor\.(\S+)><>Value
```
skonto: This will miss the ones that are of type count.

skonto:
@liyinan926 also I saw that you use `-javaagent` but I noticed when working with spark submit that the spark driver pod will not terminate and we need this feature: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/269
The jmx exporter has a few threads that dont exit when Spark context stops.

PR #281, opened 2018-09-10
Add support for specifying default spark image for applications in spark operator

mrow4a:
This PR adds support for specifying default spark container images in spark operator, which will be used in all spark applications submitted through that spark operator, unless custom image specified. 

CC @prasanthkothuri


examples/spark-pi-default.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+apiVersion: "sparkoperator.k8s.io/v1alpha1"
+kind: SparkApplication
+metadata:
+  name: spark-pi
+  namespace: default
+spec:
+  type: Scala
+  mode: cluster
+  image: Default
```
mrow4a: @liyinan926 should I add this to all examples as default, and rename this yaml file as `spark-pi-custom` ? Or leave it as it is?

manifest/spark-operator.yaml
```diff
@@ -45,4 +45,5 @@ spec:
         imagePullPolicy: Always
         command: ["/usr/bin/spark-operator"]
         args:
-        - -logtostderr
\ No newline at end of file
+        - -logtostderr
+        - -default-spark-image=gcr.io/spark-operator/spark:v2.3.1
```
mrow4a: @liyinan926 should I add this to all manifests as default, and rename this yaml file as `spark-operator-defaults` ?

mrow4a: Additionaly, this image needs to exists with new change to use this flag - use of this flag with an old spark-operator version gives error during submission

pkg/apis/sparkoperator.k8s.io/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -407,16 +407,6 @@ func (in *SparkApplicationList) DeepCopyObject() runtime.Object {
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *SparkApplicationSpec) DeepCopyInto(out *SparkApplicationSpec) {
 	*out = *in
-	if in.Image != nil {
-		in, out := &in.Image, &out.Image
-		*out = new(string)
-		**out = **in
-	}
```
mrow4a: Not sure if these are needed, since it is no longer string

mrow4a:
@liyinan926 if you could give a feedback on whether this is a good direction to achieve this default unified image

main.go
```diff
@@ -128,9 +133,9 @@ func main() {
 		time.Duration(*resyncInterval)*time.Second,
 		factoryOpts...)
 	applicationController := sparkapplication.NewController(
-		crdClient, kubeClient, apiExtensionsClient, factory, *submissionRunnerThreads, metricConfig, *namespace)
+		crdClient, kubeClient, apiExtensionsClient, factory, *submissionRunnerThreads, appDefaultConfig, metricConfig, *namespace)
```
mrow4a: @liyinan926 I also thought about merging metricConfig and appDefaultConfig - as they are usually representing what wass passed as argument to operator

examples/spark-pi-default.yaml
liyinan926: What if we simply make `image` optional, and if it's not specified, use the default image, instead of using this `Default` value?

main.go
```diff
@@ -63,6 +63,7 @@ var (
 	metricsPort             = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
 	metricsEndpoint         = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
 	metricsPrefix           = flag.String("metrics-prefix", "", "Prefix for the metrics.")
+	defaultSparkImage       = flag.String("default-spark-image", "", "Default unified container image for the Spark application (both on driver and executor)")
```
liyinan926: `both on` -> `for both the`.

main.go
liyinan926: I think we can bundle everything after `factory` together into a single struct named `OperatorConfig`.

manifest/spark-operator.yaml
liyinan926: Essentially users need to first upgrade the operator before they can start leaving `image` empty.

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -54,6 +54,15 @@ const (
 	Always    RestartPolicy = "Always"
 )
 
+// ContainerImage is a string representing registry, image name and image tag of a container for
+// drivers, executors and init containers
+type ContainerImage string
```
liyinan926: I think we can use an empty value as an indication to use the default image.

pkg/apis/sparkoperator.k8s.io/v1alpha1/zz_generated.deepcopy.go
liyinan926: These are automatically generated so we just take whatever changes the code-gen made.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -54,6 +54,15 @@ const (
 	Always    RestartPolicy = "Always"
 )
 
+// ContainerImage is a string representing registry, image name and image tag of a container for
+// drivers, executors and init containers
+type ContainerImage string
```
liyinan926: Ditto.

pkg/controller/scheduledsparkapplication/controller.go
```diff
@@ -257,6 +261,12 @@ func (c *Controller) createSparkApplication(
 		UID:        scheduledApp.UID,
 	})
 	app.ObjectMeta.SetLabels(scheduledApp.GetLabels())
+
+	if app.Spec.Image == v1alpha1.UseDefaultContainerImage {
```
liyinan926: `if app.Spec.Image == nil {...}`.

examples/spark-pi-default.yaml
mrow4a: It is possible, but would require getting rid of this - https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/sparkctl/cmd/create.go#L138 . Sparkctl validates it. Thoughts?

liyinan926: I think we can get rid of that check for now.

pkg/apis/sparkoperator.k8s.io/v1alpha1/zz_generated.deepcopy.go
liyinan926: Run `hack/update-codegen.sh` and `hack/verify-codegen.sh`.

examples/spark-pi-default.yaml
```diff
@@ -0,0 +1,35 @@
...
+ apiVersion: "sparkoperator.k8s.io/v1alpha1"
+kind: SparkApplication
+metadata:
+  name: spark-pi
+  namespace: default
+spec:
+  type: Scala
+  mode: cluster
+  mainClass: org.apache.spark.examples.SparkPi
+  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.11-2.3.1.jar"
```
mrow4a: This requires newest image to be build before this is merged @liyinan926

mrow4a: Additionaly, `sparkctl` would need to be updated to support not specifying `spec.Image`

mrow4a:
@liyinan926 addressed your comments and wrote documentation

mrow4a:
Spark-Pi-Default example requires newest image to be build before this is merged @liyinan926. Old operator templates are ok, as this would not affect it (code is triggered when Image tag is empty in application template)

mrow4a:
@liyinan926 I rebased and added `default-spark-image` only as an custom option, not default in any of `yaml` files.

docs/quick-start-guide.md
```diff
@@ -38,6 +40,44 @@ Now you should see the operator running in the cluster by checking the status of
...
+
+By default, the operator will install the [CustomResourceDefinitions](https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/) for the custom resources it managers. This can be disabled by setting the flag `-install-crds=false`.
+
+The mutating admission webhook is an **optional** component and can be enabled or disabled using the `-enable-webhook` flag, which defaults to `false`.
+
+By default, the operator will manage custom resource objects of the managed CRD types for the whole cluster. It can be configured to manage only the custom resource objects in a specific namespace with the flag `-namespace=<namespace>`
+
+## Using the Mutating Admission Webhook
+
+The Kubernetes Operator for "Apache Spark comes with an optional mutating admission webhook for customizing Spark driver and executor pods based on the specification in `SparkApplication` objects, e.g., mounting user-specified ConfigMaps and volumes, and setting pod affinity/anti-affinity.
```
liyinan926: There's a `"` that should be removed.

docs/quick-start-guide.md
```diff
@@ -38,6 +40,44 @@ Now you should see the operator running in the cluster by checking the status of
...
+
+The webhook requires a X509 certificate for TLS for pod admission requests and responses between the Kubernetes API server and the webhook server running inside the operator. For that, the certificate and key files must be accessible by the webhook server.
+The Spark Operator ships with a tool at `hack/gencerts.sh` for generating the CA and server certificate and putting the certificate and key files into a secret named `spark-webhook-certs` in namespace `sparkoperator`. This secret will be mounted into the Spark Operator pod.
```
liyinan926: Please replace `Spark Operator` with simply `the operator` or `The Kubernetes Operator for Apache Spark` when appropriate. We stopped using the term `Spark Operator` in documentation.

docs/quick-start-guide.md
```diff
@@ -38,6 +40,44 @@ Now you should see the operator running in the cluster by checking the status of
...
+$ kubectl apply -f manifest/spark-operator-with-webhook.yaml
+```
+
+This will create a Deployment named `sparkoperator` and a Service named `spark-webhook` for the webhook in namespace `sparkoperator`.
+
+If the operator is installed via the Helm chart using the default settings (i.e. with webhook enabled), the above steps are all automated for you.
+
+### Default Spark Image
+
+Operator allows to set default unified spark image for driver and executors, using `-default-spark-image=[image]:[tag]` flag. Each submitted spark application will use that image as default, upon customized by the user in the `SparkApplication` template.
```
liyinan926: `Operator` -> `The operator`. Also please use `Spark` instead of `spark` consistently.

main.go
```diff
@@ -63,6 +63,7 @@ var (
 	metricsPort             = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
 	metricsEndpoint         = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
 	metricsPrefix           = flag.String("metrics-prefix", "", "Prefix for the metrics.")
+	defaultSparkImage       = flag.String("default-spark-image", "", "Default unified container image for the Spark application (for both the driver and executor)")
```
liyinan926: I think it should have a default value.

pkg/util/operator_config.go
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package util
+
+type OperatorConfig struct {
+	Namespace               string
+	SubmissionRunnerThreads int
+	UnifiedSparkImage       string
```
liyinan926: You can put all metrics-related field into a `MetricConfig` and embed `MetricConfig` inside the `OperatorConfig`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -340,6 +338,15 @@ func (c *Controller) createSubmission(app *v1alpha1.SparkApplication) error {
 		appStatus.DriverInfo.WebUIPort = port
 	}
 
+	if app.Spec.Image == nil && c.operatorConfig.UnifiedSparkImage == "" {
```
liyinan926: This check and the if block below should be moved into `onAdd`, and ideally we can add a method named `ValidateSparkApplication(app *SparkApplication, defaultSparkImage string)` in the same package as `types.go`, and we can name the file `validation.go`.

main.go
mrow4a: What do you suggest?

liyinan926: I think we can use the same default image as the Helm Chart.

PR #289, opened 2018-09-20
Miscellaneous Prometheus-related fixes

yuchaoran2011:
1. The doc refers to `gcr.io/spark-operator/spark:v2.3.1-gcs-prometheus` but the sample YAML didn't use that image.
2. `spark-docker/Dockerfile` had some issues. Prometheus metrics should be under `/etc/metrics/conf`. If using the current `gcr.io/spark-operator/spark:v2.3.1-gcs-prometheus` image, an error is thrown in the driver pod: 
```
java.io.FileNotFoundException: /etc/metrics/conf/prometheus.yaml (No such file or directory).
```
`gcr.io/spark-operator/spark:v2.3.1-gcs-prometheus` needs to be rebuilt with the updated version in this PR to get it to work.

3. Made base Spark image customizable in `spark-docker/Dockerfile` via an ARG.
4. The doc claims that `spec.monitoring.prometheus.port` in the YAML is optional. It would default to 8090 if not set. But the operator pod would produce the following error and fail to submit the job:
```
spec.monitoring.prometheus.port in body must be of type integer: "null"
```
This PR added the port in the YAML. 

5. As a "complete" example, the YAML should also "exposeExecutorMetrics"


liyinan926:
LGTM. Thanks!

yuchaoran2011:
@liyinan926 Thanks for merging! Could you also rebuild and push `gcr.io/spark-operator/spark:v2.3.1-gcs-prometheus` using the updated  `spark-docker/Dockerfile`? I don't have the permission to push to that gcr repo.

liyinan926:
Sure, will do.

PR #291, opened 2018-09-21
Operator State Management + Ingress Creation

akhurana001:
Ref:  
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/199
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/200


liyinan926:
Wow, thanks @akhurana001 and @kumare3! This is a big one! :).

kumare3:
@liyinan926 PTAL

pkg/controller/sparkapplication/controller.go
```diff
@@ -175,66 +175,30 @@ func (c *Controller) Stop() {
...
-
-	// Delete the driver pod and UI service of the old application. Note that deleting the driver pod kills the
-	// application if it is still running. Skip submitting the new application if cleanup for the old application
-	// failed to avoid potentially running both the old and new applications at the same time.
-	if err := c.deleteDriverAndUIService(oldApp, true); err != nil {
-		glog.Errorf("failed to delete the old driver pod and UI service for SparkApplication %s: %v",
-			oldApp.Name, err)
+	// The spec has changed. This is currently not supported as we can potentially miss this update
+	// and end up in an inconsistent state.
+	if !reflect.DeepEqual(oldApp.Spec, newApp.Spec) {
```
liyinan926: Actually it seems some users are really using this feature, e.g., triggering restart by patching a `SparkApplication`, so I think we should keep it.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,386 +250,461 @@ func (c *Controller) processNextItem() bool {
...
+//|                         |           |       |                         |           |          |          |          |
+//|                         |           |       |                         |           |          |          |          |
+//|                         +-----------+       |                         +-----+-----+          +----------+          |
+//|                                             |                               |                                      |
+//|                                             |                               |                                      |
+//|                                             +-------------------------------+                                      |
+//|                                                                                                                    |
+//+--------------------------------------------------------------------------------------------------------------------+
+func (c *Controller) syncSparkApplication(key string) error {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
```
liyinan926: This method now looks really long and hard to review. Can we break it down into smaller functions, e.g., putting the logic on a per-state basis into individual helper functions?

pkg/controller/sparkapplication/controller.go
liyinan926: I would say let's revert the changes here as I heard a couple of use of this.

akhurana001: Done, I ended up adding a new invalidating state to handle this in the new state machine. Have a look and let me know if there are any questions

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+		var errorMsg string
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			errorMsg = string(exitErr.Stderr)
 		}
-		existingState, ok := status.ExecutorState[update.podName]
-		if ok && isExecutorTerminated(existingState) {
-			return
+		// Already Exists. Do nothing.
+		if strings.Contains(errorMsg, podAlreadyExistsErrorCode) {
+			glog.Warningf("Trying to resubmit an already submitted SparkApplication %s in namespace %s. Error: %s", submission.name, submission.namespace, errorMsg)
```
liyinan926: Should use `%v` for errors.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+	if _, err := cmd.Output(); err != nil {
+		var errorMsg string
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			errorMsg = string(exitErr.Stderr)
 		}
-		existingState, ok := status.ExecutorState[update.podName]
-		if ok && isExecutorTerminated(existingState) {
-			return
+		// Already Exists. Do nothing.
+		if strings.Contains(errorMsg, podAlreadyExistsErrorCode) {
```
liyinan926: How is this gonna happen?

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
-		apiv1.EventTypeNormal,
-		"SparkApplicationRestart",
-		"SparkApplication %s was enqueued for restart",
-		app.Name)
+	return app
 }
 
-func (c *Controller) handleResubmission(app *v1alpha1.SparkApplication, submissionRetries int32) {
-	glog.Infof("Retrying submission of SparkApplication %s", app.Name)
+func (c *Controller) updateAppAndExportMetrics(oldApp, newApp *v1alpha1.SparkApplication) error {
```
liyinan926: There's also the `updateSparkApplicationStatusWithRetries`, why you also do update here?

pkg/controller/sparkapplication/controller.go
```diff
@@ -175,71 +176,83 @@ func (c *Controller) Stop() {
...
-	c.recorder.Eventf(
-		newApp,
-		apiv1.EventTypeNormal,
-		"SparkApplicationUpdated",
-		"SparkApplication %s was updated and enqueued for submission",
-		newApp.Name)
 }
 
 func (c *Controller) onDelete(obj interface{}) {
-	c.dequeue(obj)
```
liyinan926: This should be kept. When an app gets deleted, it should be forgotten by the queue.

pkg/controller/sparkapplication/controller.go
```diff
@@ -175,71 +176,83 @@ func (c *Controller) Stop() {
...
-		if oldApp.Status.AppState.State != newApp.Status.AppState.State && shouldRestart(newApp) {
-			c.handleRestart(newApp)
+	// The spec has changed. This is currently best effort as we can potentially miss updates
+	// and end up in an inconsistent state.
+	if !reflect.DeepEqual(oldApp.Spec, newApp.Spec) {
+		// Force-set the application status to Invalidating which handles clean-up and application re-run.
+		if err := c.updateSparkApplicationStatusWithRetries(newApp, func(status *v1alpha1.SparkApplicationStatus) {
+			status.AppState.State = v1alpha1.InvalidatingState
+		}); err != nil {
+			c.recorder.Eventf(
```
liyinan926: Can we add this into the `recordSparkApplicationEvent`?

pkg/controller/sparkapplication/controller.go
```diff
@@ -175,71 +176,83 @@ func (c *Controller) Stop() {
...
+				newApp,
+				apiv1.EventTypeWarning,
+				"SparkApplicationUpdateFailed",
+				"Failed to process update for SparkApplication %s, Error: %s.",
+				newApp.Name,
+				err)
+			return
+		} else {
+			c.recorder.Eventf(
```
liyinan926: Can we add this into the `recordSparkApplicationEvent`?

pkg/controller/sparkapplication/spark_pod_eventhandler.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if deletedPod == nil {
+		return
+	}
+	s.enqueueSparkAppForUpdate(deletedPod)
+}
+
+func (c *sparkPodEventHandler) enqueueSparkAppForUpdate(pod *apiv1.Pod) {
+	if appKey, ok := createMetaNamespaceKey(pod); ok {
+		glog.V(2).Infof("Enqueuing SparkApp %s", appKey)
```
liyinan926: Ditto. Please use `SparkApplication` consistently in the log messages.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -0,0 +1,72 @@
+package sparkapplication
```
liyinan926: Missing license header.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+}
 
-	app, err := c.getSparkApplication(namespace, name)
-	if err != nil {
-		return err
-	}
+func (c *Controller) getUpdatedAppStatus(app *v1alpha1.SparkApplication) *v1alpha1.SparkApplication {
+
+	// Fetch all the pods for the App.
+	selector, err := labels.NewRequirement(config.SparkAppNameLabel, selection.Equals, []string{app.Name})
```
liyinan926: I feel that it's a bit insufficient to just use the app name as the label selector. To make sure that we just select pods of the current run, I think we should use a label that is unique for each run of an app and also add that into the label selector.

liyinan926:
Also can you rebase the code and run `go fmt` to format the code using go 1.11?

liyinan926:
@akhurana001 is this ready for another round of review?

liyinan926:
> Also I see you already saw https://issues.apache.org/jira/browse/SPARK-25922 which i filed for the SparkApplicationId mis-match. This impacts the SparkApplicationId field in the Status for the CRD as this current logic fetches it from the driver label. Let me know if its possible to get more traction on it.

Yes, this is problematic, I think we can get around it by introducing a per-run application ID generated by the operator and use that as the label selector.

liyinan926:
BTW: looks like there's still conflict. And have you formatted the code by running `go fmt` using Go 1.11.x?

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -231,13 +238,16 @@ type ApplicationStateType string
...
-	NewState              ApplicationStateType = "NEW"
+	NewState              ApplicationStateType = ""
 	SubmittedState        ApplicationStateType = "SUBMITTED"
 	RunningState          ApplicationStateType = "RUNNING"
 	CompletedState        ApplicationStateType = "COMPLETED"
 	FailedState           ApplicationStateType = "FAILED"
 	FailedSubmissionState ApplicationStateType = "SUBMISSION_FAILED"
-	UnknownState          ApplicationStateType = "UNKNOWN"
+	PendingRetryState     ApplicationStateType = "PENDING_RETRY"
+	InvalidatingState     ApplicationStateType = "INVALIDATING"
```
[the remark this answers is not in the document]
liyinan926: Makes sense.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+			}
+			app.Status.SparkApplicationID = currentDriverState.sparkApplicationID
+			if currentDriverState.nodeName != "" {
+				if nodeIP := c.getNodeExternalIP(currentDriverState.nodeName); nodeIP != "" {
+					app.Status.DriverInfo.WebUIAddress = fmt.Sprintf("%s:%d", nodeIP,
+						app.Status.DriverInfo.WebUIPort)
+				}
+			}
+			app.Status.AppState.State = newAppState
+			if app.Status.CompletionTime.IsZero() && !currentDriverState.completionTime.IsZero() {
```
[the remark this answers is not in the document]
liyinan926: Ah, right.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+				appToUpdate = c.submitSparkApplication(appToUpdate)
+			}
+		case v1alpha1.InvalidatingState:
+			// Invalidate the current run and enqueue the SparkApplication for re-execution.
+			if err := c.deleteSparkResources(appToUpdate, true); err != nil {
+				glog.Errorf("failed to delete the driver pod and UI service for deleted SparkApplication %s: %v",
+					appToUpdate.Name, err)
+				return err
 			}
+			appToUpdate.Status.AppState.State = v1alpha1.PendingRetryState
```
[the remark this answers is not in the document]
liyinan926: `PENDING_RERUN` sounds better.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
+// Helper func to determine if we have waited enough to retry the SparkApplication.
+func hasRetryIntervalPassed(retryInterval *int64, attemptsDone int32, lastEventTime metav1.Time) bool {
 
-		if shouldRetrySubmission(app) {
-			submissionRetries++
-			c.handleResubmission(app, submissionRetries)
-		} else {
-			glog.Infof("Not retrying submission of SparkApplication %s", update.name)
-		}
+	if retryInterval == nil || lastEventTime.IsZero() || attemptsDone <= 0 {
```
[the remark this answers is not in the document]
liyinan926: I think we should have sensible defaults for `OnFailureRetryInterval` and `OnSubmissionFailureRetryInterval`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +300,472 @@ func (c *Controller) processNextItem() bool {
...
-		if !errors.IsNotFound(err) {
-			glog.Errorf("failed to get SparkApplication %s in namespace %s from the store: %v", update.appName,
-				update.appNamespace, err)
+		app.Status = v1alpha1.SparkApplicationStatus{
+			AppState: v1alpha1.ApplicationState{
+				State:        v1alpha1.FailedSubmissionState,
+				ErrorMessage: err.Error(),
+			},
+			SubmissionAttempts: app.Status.SubmissionAttempts + 1,
+			SubmissionTime:     metav1.Now(),
```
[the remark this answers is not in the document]
liyinan926: OK, may be let's rename it to `LastSubmissionAttemptTime`?

pkg/controller/sparkapplication/controller.go
liyinan926: Got it.

akhurana001:
> BTW: looks like there's still conflict. And have you formatted the code by running go fmt using Go 1.11.x?

@liyinan926  I had merged  yesterday evening, just did it again  and ran a go fmt using 1.11.2

akhurana001:
> Yes, this is problematic, I think we can get around it by introducing a per-run application ID generated by the operator and use that as the label selector.

This won't solve the use-case we had .  The operator itself doesn't depend on the SparkApplicationID we expose in the CRD and should work fine. 

Spark uses the SparkApplicationID it generates for various use-cases. For e.g.: using it a key to push the event logs to HistoryServer. To fetch these event-logs/ access the history server UI, I was trying to get the SparkApplicationID from the driver label and add it in the CRD status.

liyinan926:
> This won't solve the use-case we had. The operator itself doesn't depend on the SparkApplicationID we expose in the CRD and should work fine. Spark uses the SparkApplicationID it generates for various use-cases. For e.g.: using it a key to push the event logs to HistoryServer. To fetch these event-logs/ access the history server UI, I was trying to get the SparkApplicationID from the driver label and add it in the CRD status.

Right, you are not including the app ID in the label selector when listing pods anyway. I agree this is something that needs to be fixed upstream.

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -43,15 +43,28 @@ const (
...
-// The decision is based on the policy and termination state of the driver pod. An application that fails submission
-// won't be restarted regardless of the policy.
-type RestartPolicy string
+// This completely defines actions to be taken on any kind of Failures during an application run.
+type RestartPolicy struct {
+	Type RestartPolicyType `json:"type,omitempty""`
+
+	// FailureRetries are the number of times to retry a failed application before giving up in a particular case.
+	// This is best effort and actual retry attempts can be >= the value specified due to caching.
+	// These are required if RestartPolicy is onFailure.
```
liyinan926: Nit: `OnFailure` instead of `onFailure`.

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -43,15 +43,28 @@ const (
...
+	OnSubmissionFailureRetries *int32 `json:"onSubmissionFailureRetries,omitempty"`
+	OnFailureRetries           *int32 `json:"onFailureRetries,omitempty"`
+
+	// Interval to wait between successive retries of a failed application.
+	// Required if the RestartPolicy is onFailure or Always.
```
liyinan926: Nit: `OnFailure` instead of `onFailure`.

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -231,13 +238,16 @@ type ApplicationStateType string
...
 const (
-	NewState              ApplicationStateType = "NEW"
+	NewState              ApplicationStateType = ""
 	SubmittedState        ApplicationStateType = "SUBMITTED"
 	RunningState          ApplicationStateType = "RUNNING"
 	CompletedState        ApplicationStateType = "COMPLETED"
 	FailedState           ApplicationStateType = "FAILED"
 	FailedSubmissionState ApplicationStateType = "SUBMISSION_FAILED"
-	UnknownState          ApplicationStateType = "UNKNOWN"
+	PendingRetryState     ApplicationStateType = "PENDING_RETRY"
```
liyinan926: As we discussed, let's use `PENDING_RERUN` for reruns as a result of spec updates?

akhurana001:
> > This won't solve the use-case we had. The operator itself doesn't depend on the SparkApplicationID we expose in the CRD and should work fine. Spark uses the SparkApplicationID it generates for various use-cases. For e.g.: using it a key to push the event logs to HistoryServer. To fetch these event-logs/ access the history server UI, I was trying to get the SparkApplicationID from the driver label and add it in the CRD status.
> 
> Right, you are not including the app ID in the label selector when listing pods anyway. I agree this is something that needs to be fixed upstream.

Yup, I am thinking of adding a logic where we fetch the app-selector label from both driver and executors and prefer the executor label (which is the correct SparkApplicationID in 2.4.0) if they mis-match. This is a hack till its fixed upstream.  Not sure if there is a better way to achieve till then.

pkg/controller/sparkapplication/controller.go
liyinan926: Makes sense.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,464 @@ func (c *Controller) processNextItem() bool {
...
+	podPhase           apiv1.PodPhase // Driver pod phase.
+	completionTime     metav1.Time    // Time the driver completes.
+}
 
-	app, err := c.getSparkApplication(namespace, name)
-	if err != nil {
-		return err
-	}
+func (c *Controller) getUpdatedAppStatus(app *v1alpha1.SparkApplication) v1alpha1.SparkApplicationStatus {
+
```
liyinan926: Nit: unnecessary empty line.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,464 @@ func (c *Controller) processNextItem() bool {
...
+					return err
+				}
+				appToUpdate.Status.AppState.State = v1alpha1.PendingRetryState
+			} else {
+				// App will never be retried. Move to Terminal Completed State.
+				appToUpdate.Status.AppState.State = v1alpha1.CompletedState
+				c.recordSparkApplicationEvent(appToUpdate)
+			}
+		case v1alpha1.FailingState:
+			if !shouldRetry(appToUpdate) {
```
liyinan926: Nit: `shouldRetry(appToUpdate)` is used above, so to be consistent, the if block should also check `shouldRetry(appToUpdate)` here.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,464 @@ func (c *Controller) processNextItem() bool {
...
+			} else if hasRetryIntervalPassed(appToUpdate.Spec.RestartPolicy.OnFailureRetryInterval, appToUpdate.Status.ExecutionAttempts, appToUpdate.Status.CompletionTime) {
+				if err := c.deleteSparkResources(appToUpdate, true); err != nil {
+					glog.Errorf("Failed to delete the driver pod and UI service for deleted SparkApplication %s/%s. Error: %v",
+						appToUpdate.Namespace, appToUpdate.Name, err)
+					return err
+				}
+				appToUpdate.Status.AppState.State = v1alpha1.PendingRetryState
+			}
+		case v1alpha1.FailedSubmissionState:
+			if !shouldRetry(appToUpdate) {
```
liyinan926: Ditto.

pkg/controller/sparkapplication/controller.go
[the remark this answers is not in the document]
liyinan926: As we discussed, this should be `PendingRerunState`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,464 @@ func (c *Controller) processNextItem() bool {
...
+			SubmissionTime:     metav1.Now(),
+		}
+		c.recordSparkApplicationEvent(app)
 
-		// Let's keep the old App status if this is not a valid transition.
-		if original.Status.AppState.State != toUpdate.Status.AppState.State &&
-			!isValidDriverStateTransition(original.Status.AppState.State, toUpdate.Status.AppState.State) {
-			glog.Warningf("Invalid Driver State Transition. From:[%v] To:[%v]", original.Status.AppState.State, toUpdate.Status.AppState.State)
-			toUpdate.Status.AppState = original.Status.AppState
+		// Add driver as a finalizer to prevent SparkApplication deletion till driver is deleted.
```
liyinan926: Can we put this into a utility method corresponding to `remoteFinalizer` and put it into `sparkapp_util.go`?

pkg/controller/sparkapplication/controller.go
liyinan926: The problem is this function doesn't really take care of update conflicts, whereas `updateSparkApplicationStatusWithRetries` does. I think we should combine the two functions in some way to avoid confusion.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,464 @@ func (c *Controller) processNextItem() bool {
...
 }
 
-func (c *Controller) processPodStateUpdates() {
-	for update := range c.podStateReportingChan {
-		switch update.(type) {
-		case *driverStateUpdate:
-			c.processSingleDriverStateUpdate(update.(*driverStateUpdate))
-		case *executorStateUpdate:
-			c.processSingleExecutorStateUpdate(update.(*executorStateUpdate))
+func removeFinalizer(app *v1alpha1.SparkApplication, finalizerToRemove string) *v1alpha1.SparkApplication {
```
liyinan926: Looks like such utility functions can be moved to `sparkapp_util.go`.

liyinan926:
@akhurana001 @kumare3 I will be at KubCcon Shanghai this week, so please expect some delay on PR reviews.

akhurana001:
No worries, thanks for letting us know Yinan.
 I was also out for a couple for days, will try to close out on the
remaining changes this week.

pkg/controller/sparkapplication/controller.go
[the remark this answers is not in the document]
akhurana001: I added the defaults in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/291/commits/50f3c7141b83d1942dc860bc269a46bc167ae9be#diff-29d8bdcd1053149a316e02725c24d194 (pkg/apis/sparkoperator.k8s.io/v1alpha1/defaults.go) but looks like its only invoked if a user uses sparkctl to create the CRD

pkg/controller/sparkapplication/controller.go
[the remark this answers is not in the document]

pkg/controller/sparkapplication/controller.go
akhurana001: I still don’t think we should do retries here since our state-machine is idempotent and eventually consistent i.e. if we process a CRD based on our state-machine transitions and fail to write to etc.d because of update conflicts, the next time we process the CRD (which has an upper bound and at-worst can be equal our informer cache re-sync period), the state-machine will process the CRD correctly and drive it to its desired state.

Secondly, adding retries here can lead to a race-condition with our CRD update processing. For e.g. If a user updates the CRD while a worker is processing the SparkApplication, the onUpdate callback will try to update the CRD status to Invalidating but the worker once its done processing can potentially overwrite this Invalidating state if we add retries here leading to an onUpdate miss.  In the current implementation, the worker overwrite will fail and the Invalidating state will be processed on the next time we process the CRD.  

Let me know what you think on this.

akhurana001:
@liyinan926  I reverted back on the feedback. Let me know if you have any questions. Thanks.

pkg/controller/sparkapplication/controller.go
akhurana001: Also made some minor changes to how we clean up resources on failures/re-run etc. to make sure this is true.

pkg/controller/sparkapplication/controller.go
```diff
@@ -615,61 +613,57 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
...
-				ingress, err := createSparkUIIngress(app, *service, c.ingressUrlFormat, c.kubeClient)
-				if err != nil {
-					glog.Errorf("Failed to create UI Ingress for SparkApplication %s/%s. Error: %v", app.Namespace, app.Name, err)
-				} else {
-					app.Status.DriverInfo.WebUIIngressAddress = ingress.ingressUrl
-					app.Status.DriverInfo.WebUIIngressName = ingress.ingressName
-				}
+	// Add driver as a finalizer to prevent SparkApplication deletion till driver is deleted.
+	app = addFinalizer(app, sparkDriverRole)
+	if app.Spec.Monitoring != nil && app.Spec.Monitoring.Prometheus != nil {
```
liyinan926: Hmm, Prometheus monitoring configuration should be done before submitting the application.

pkg/controller/sparkapplication/controller.go
```diff
@@ -615,61 +613,57 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
...
+		// configPrometheusMonitoring may update app.Spec.
+		configPrometheusMonitoring(app, c.kubeClient)
+	}
+
+	// Create Spark UI Service.
```
liyinan926: It's also better to create the UI service before submitting the application to run, which is currently the case in the master.

pkg/controller/sparkapplication/controller.go
```diff
@@ -289,397 +298,471 @@ func (c *Controller) processNextItem() bool {
...
 
-	// The controller may receive status updates of executors belonging to the previous run of the application and
-	// should ignore such updates. For example, if a user makes some change to a SparkApplication and does 'kube apply'
-	// to update the SparkApplication object on the API server while the application is running, the controller may
-	// still receive status updates of executors of the old application run even after the new run starts.
-	if update.appID != app.Status.AppID {
-		return nil
+	// Try submitting App.
+	submission := newSubmission(submissionCmdArgs, app)
+	sparkHome, present := os.LookupEnv(sparkHomeEnvVar)
```
liyinan926: Can we put the code for executing the `spark-submit` command in a utility function like what we currently have?

akhurana001: Done, created a helper method in submission.go

pkg/controller/sparkapplication/controller.go
akhurana001: I am not clear about what we gain by creating a service upfront.  The service won't even be up till the driver pod is actually running and start listening on the pod. Also the submission can fail and we will need additionally clean-up if the spark-submit failed but we already created the service/ingress.  Let me know if there are any obvious benefits of doing this and I can change this.

pkg/controller/sparkapplication/controller.go
akhurana001: done, I had to clone the App since we modify the Spec which will cause a onUpdate callback and a clean-up/re-run of the Application. Not sure if this is already an issue in master.

pkg/controller/sparkapplication/controller.go
liyinan926: While, there's no real benefit on doing this. But if for some reason the service is not up when the driver is up and running, the UI is not accessible. But the chance of that is probably negligible.

pkg/controller/sparkapplication/submission.go
```diff
@@ -50,6 +54,34 @@ func newSubmission(args []string, app *v1alpha1.SparkApplication) *submission {
 	}
 }
 
+func sparkSubmit(submission *submission) error {
```
liyinan926: Can we rename this to `runSparkSubmit`?

akhurana001: done

pkg/controller/sparkapplication/controller.go
liyinan926: In the current master, only the status section is updated to the API server, so this is not an issue. But this is no longer the case with the changes in this PR as you add a finalizer to the mix.

PR #293, opened 2018-09-21
Adjust documentation, import paths and codegen to avoid case-insensitive import collision

mrow4a:
When following the steps from https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/quick-start-guide.md#build and calling `go generate` and then `go test ./...` there were errors as 

`
../../googlecloudplatform/spark-on-k8s-operator/vendor/k8s.io/apimachinery/pkg/watch/until.go:23:2: case-insensitive import collision: "github.com/googlecloudplatform/spark-on-k8s-operator/vendor/k8s.io/apimachinery/pkg/util/wait" and "github.com/GoogleCloudPlatform/spark-on-k8s-operator/vendor/k8s.io/apimachinery/pkg/util/wait"
`

CC @liyinan926

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,6 @@
 
 
 [[projects]]
-  digest = "1:9702dc153c9bb6ee7ee0587c248b7024700e89e4a7be284faaeeab9da32e1c6b"
   name = "cloud.google.com/go"
   packages = [
     "compute/metadata",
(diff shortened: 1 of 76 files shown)


docs/quick-start-guide.md
```diff
@@ -214,8 +214,8 @@ $ docker build --build-arg SPARK_IMAGE=<your Spark image> -t <image-tag> .
 If you'd like to build/test the spark-operator locally, follow the instructions below:
 
 ```bash
-$ mkdir -p $GOPATH/src/k8s.io
-$ cd $GOPATH/src/k8s.io
+$ mkdir -p $GOPATH/src/github.com/googlecloudplatform
+$ cd $GOPATH/src/github.com/googlecloudplatform
 $ git clone git@github.com:GoogleCloudPlatform/spark-on-k8s-operator.git
```
mrow4a: Here is adjusted and correct doc

mrow4a: the `dep ensure` will not work, only `dep ensure -vendor-only` works

Gopkg.lock
```diff
@@ -2,70 +2,30 @@
...
+    "autorest/date"
   ]
-  pruneopts = ""
   revision = "fc3b03a2d2d1f43fad3007038bd16f044f870722"
   version = "v9.10.0"
 
 [[projects]]
-  branch = "master"
-  digest = "1:a01ad406efc79e2cdf827baa8836eb7a0ceb84bcec7f389a163076748076d4ab"
-  name = "github.com/GoogleCloudPlatform/spark-on-k8s-operator"
```
mrow4a: Now this does not get placed in Gopkg.lock

liyinan926:
This is weird. The correct import paths should start with `github.com/GoogleCloudPlatform/spark-on-k8s-operator/...` instead of the lower-case one. For some reason, `go generate` generates the lower-case one.

mrow4a:
@liyinan926 if you find the better way of solving this, just close this PR with better option. But I am afraid lowercasing all imports might be necessary

mrow4a:
Hmm, I start actually thinking - will it actually not break the imports ? https://github.com/DAddYE/igo/issues/2

liyinan926:
I guess it's more important to be consistent than to favor a particular case.

liyinan926:
Let me dig a bit deeper on this.

mrow4a:
@liyinan926 yes, please go ahead - I think lowercasing as in this PR wont solve the issue

PR #295, opened 2018-09-21
Updated Gopkg.lock

liyinan926:


mrow4a:
is the documentation still valid (`k8s.io`) ?

liyinan926:
Oh, no, the documentation needs to be fixed and the fix you had in the PR that was closed is actually needed.

PR #297, opened 2018-09-25
Fix the namespace for sparkoperator deployment

sarjeet2013:
Fix the namespace for sparkoperator deployment


sarjeet2013:
I signed it!

liyinan926:
The namespace is inconsistent between the Helm [chart](https://github.com/helm/charts/tree/master/incubator/sparkoperator) and the YAML manifests in this repo. The default namespace used by the Helm chart is `spark-operator`, whereas the default one used by the manifests is `sparkoperator`. We cannot just fix the documentation here. It's fine changing the default used in this repo, but then we need to fix it in all the manifests as well.

liyinan926:
Looks like the default namespace in `hack/gencerrts.sh` also needs to be fixed.

PR #298, opened 2018-09-28
Added a developer guide and notes on a code gen issue

liyinan926:
Some client code generated by running `go generate` will have lower-cased import paths starting with prefix `github.com/googlecloudplatform`. This is inconsistent with the camel case prefix `github.com/GoogleCloudPlatform` in other files. This will cause `go build` to fail. The developer guide adds a note on this issue.

@mrow4a


mrow4a:
@liyinan926 should we add a workaround to failing Dockerfile? https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/Dockerfile#L29

PR #312, opened 2018-10-19
Add support for tolerations

samsonov90:
Added support for specifying tolerations for driver/executor pods that allows scheduling on nodes with taints.

### DIFF
--- a/pkg/config/toleration.go
+++ b/pkg/config/toleration.go
@@ -1,0 +1,75 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	apiv1 "k8s.io/api/core/v1"
+
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1alpha1"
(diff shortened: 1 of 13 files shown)


samsonov90:
@liyinan926 please take a look. We need to run Spark driver & executor pods on nodes with taints. And this PR is solving the problem.

pkg/config/toleration.go
```diff
@@ -0,0 +1,78 @@
...
+	for key, value := range annotations {
+		options = append(options, GetExecutorAnnotationOption(key, value))
+	}
+
+	return options, nil
+}
+
+// FindTolerations finds and parses Tolerations in the given annotations.
+func FindTolerations(annotations map[string]string) (map[string]*apiv1.Toleration, error) {
+
```
liyinan926: Additional empty line.

liyinan926: Also looks like you only need the `Toleration` objects themselves, so you can return a `[]*apiv1.Toleration` instead.

samsonov90: @liyinan926 addressed your comments

PR #325, opened 2018-11-20
Refactoring around logging code

liyinan926:
To make the logging message format consistent across the controller code.

@akhurana001


akhurana001:
Looks good 👍

PR #332, opened 2018-12-03
Fix resubmission if app failing

soymo:


liyinan926:
@akhurana001.

pkg/controller/sparkapplication/controller.go
```diff
@@ -504,13 +510,14 @@ func (c *Controller) syncSparkApplication(key string) error {
 				// App will never be retried. Move to Terminal Failure State.
 				appToUpdate.Status.AppState.State = v1alpha1.FailedState
 				c.recordSparkApplicationEvent(appToUpdate)
-			} else if hasRetryIntervalPassed(appToUpdate.Spec.RestartPolicy.OnFailureRetryInterval, appToUpdate.Status.ExecutionAttempts, appToUpdate.Status.CompletionTime) {
+			} else if hasRetryIntervalPassed(appToUpdate.Spec.RestartPolicy.OnFailureRetryInterval, appToUpdate.Status.ExecutionAttempts, appToUpdate.Status.LastSubmissionAttemptTime) {
```
liyinan926: I think it's correct to use `CompletionTime`, which is set if the driver pod phase is `Failed` or `Succeeded`. The application state is set to `FailingState` if the driver pod phase is `Failed`.

liyinan926: If you kill the driver pod, `CompletionTime` wouldn't be set because it's set based on the completion time of the driver pod.

soymo: Ok, so how to keep driver alive? It's particularly noticeable on short living nodes (eg GKE preemptible)

akhurana001: This should be based on `CompetionTime` but to fix this we can potentially set CompetionTime while setting the state as `Failing` in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/332/files#diff-b7eea120d7ac0e8587781b863b7db8e4R364 . We currently do metav1.Now() when we see driver pod phase is `Failed` or `Succeeded` . We will be overloading the term `CompletionTime` thou

pkg/controller/sparkapplication/controller.go
```diff
@@ -354,9 +354,15 @@ func (c *Controller) getUpdatedAppStatus(app *v1alpha1.SparkApplication) v1alpha
 			}
 		} else {
 			glog.Warningf("invalid driver state for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
+			app.Status.AppState.ErrorMessage = fmt.Sprintf("Driver Pod invalid state: %s", currentDriverState.podPhase)
+			app.Status.AppState.State = v1alpha1.FailingState
```
akhurana001: invalid state here essentially means `Unknown` podPhase which can be transient from my understanding in https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase .  Currently this will be retried again after the re-sync interval which I feel is the correct behavior.

Is there a case where the pod can perpetually remain in Unknown ?

pkg/controller/sparkapplication/controller.go
soymo: Ok. I'll do it that way.

akhurana001: `TerminationTime` sounds good 👍

soymo:
@liyinan926 @akhurana001 CompletionTime renamed to TerminationTime. Decision for app resubmission is based on TerminationTime (if app in Failing state) or LastSubmissionAttemptTime (if in FailedSubmission state).
I'm not sure about resetting TerminationTime in PendingRerun state handler.

pkg/apis/sparkoperator.k8s.io/v1alpha1/types.go
```diff
@@ -273,8 +273,8 @@ type SparkApplicationStatus struct {
 	SparkApplicationID string `json:"sparkApplicationId,omitempty"`
 	// LastSubmissionAttemptTime is the time for the last application submission attempt.
 	LastSubmissionAttemptTime metav1.Time `json:"lastSubmissionAttemptTime,omitempty"`
-	// CompletionTime is the time when the application runs to completion if it does.
-	CompletionTime metav1.Time `json:"completionTime,omitempty"`
+	// TerminationTime is the time when the application runs to completion success of fail.
```
liyinan926: `success of fail` -> `with success or failure`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -537,6 +543,7 @@ func (c *Controller) syncSparkApplication(key string) error {
 				// Successfully deleted driver or driver doesn't exist. Remove it from the Finalizer List.
 				appToUpdate = removeFinalizer(appToUpdate, sparkDriverRole)
 				glog.Infof("Submitting SparkApplication %s", key)
+				appToUpdate.Status.TerminationTime = metav1.NewTime(time.Unix(0, 0))
```
liyinan926: You can simply do `appToUpdate.Status.TerminationTime = metav1.Time{}` to clear it.

pkg/controller/sparkapplication/controller.go
```diff
@@ -349,14 +349,19 @@ func (c *Controller) getUpdatedAppStatus(app *v1alpha1.SparkApplication) v1alpha
...
-				app.Status.CompletionTime = currentDriverState.completionTime
+			if app.Status.TerminationTime.IsZero() && !currentDriverState.completionTime.IsZero() {
+				app.Status.TerminationTime = currentDriverState.completionTime
 			}
 		} else {
 			glog.Warningf("invalid driver state for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 		}
 	} else {
 		glog.Warningf("driver not found for SparkApplication: %s/%s", app.Namespace, app.Name)
+		if app.Status.AppState.State == v1alpha1.RunningState && app.Status.TerminationTime.IsZero() {
```
liyinan926: So the idea is that this will trigger a re-run if re-run conditions are met, right?

soymo: Is it possible to miss somehow driver termination and SparkApplication will be in success state?

liyinan926: I'm not sure, but setting the app state to `FailingState` and `TerminationTime` may trigger a re-run with certain `RestartPolicy`.

soymo: And that was my intention. I like to handle case where spark streaming app is running and driver disappears

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -283,7 +283,7 @@ type SparkApplicationStatus struct {
 	// LastSubmissionAttemptTime is the time for the last application submission attempt.
 	LastSubmissionAttemptTime metav1.Time `json:"lastSubmissionAttemptTime,omitempty"`
 	// CompletionTime is the time when the application runs to completion if it does.
-	CompletionTime metav1.Time `json:"completionTime,omitempty"`
```
liyinan926: The comment above also needs to be updated.

soymo:
done

soymo:
done

liyinan926:
For some reason, the diff shows changes from earlier merged PRs. Can you make sure you re-base onto the latest master? Thanks!

soymo:
@liyinan926 done

PR #337, opened 2018-12-06
Minor changes to the state handling logic

liyinan926:
Added a new application state `UnknownState` that corresponds to pod phase `Unknown` and used this new state instead of using an error to indicate an unknown state.


pkg/controller/sparkapplication/controller.go
```diff
@@ -303,7 +303,7 @@ func (c *Controller) getUpdatedAppStatus(app *v1alpha1.SparkApplication) v1alpha
 	}
 
 	if currentDriverState != nil {
-		if newAppState, err := driverPodPhaseToApplicationState(currentDriverState.podPhase); err == nil {
+		if newAppState := driverPodPhaseToApplicationState(currentDriverState.podPhase); newAppState != v1alpha1.UnknownState {
```
akhurana001: You aren't setting the appStatus anywhere to Unknown.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -823,7 +823,7 @@ func TestSyncSparkApp_SubmissionSuccess(t *testing.T) {
 	}
 }
 
-func TestSyncSparkApplication_ExecutingState(t *testing.T) {
+func TestSyncSparkApplication_ExecutorState(t *testing.T) {
```
akhurana001: We can also add a quick test here for Unknown

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller_test.go
liyinan926: Done.

akhurana001:
👍 Looks good

PR #339, opened 2018-12-07
Removed the use of the finalizer

liyinan926:
The operator currently adds a finalizer to a submitted `SparkApplication` object for the purpose of preventing it from being deleted before the driver pod is gone (the operator does try to delete the driver pod when deletion of the `SparkApplication` is requested). The purpose of this is to make sure the driver pod is deleted when the `SparkApplication` gets deleted without relying on an `OwnerReference`. Previously the operator deleted the driver pod and other relevant k8s resources when `onDelete` is called. 

Using a finalizer on a custom resource risks running into this issue: https://github.com/kubernetes/kubernetes/issues/60538 for some versions of Kubernetes. I propose here to remove the use of the finalizer and simply delete the resources when `onDelete` is called. This might risk missing delete events, but the chance is rare. Also given that Kubernetes 1.12 started to support TTL on finished pods (see https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/), using a finalizer is even less appealing for the purpose discussed above. We can set a TTL on the driver pod, which should be garbage collected when the TTL expires, without using an `OwnerReference`. 

@akhurana001

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -219,6 +219,7 @@ func (c *Controller) onDelete(obj interface{}) {
 	}
 
 	if app != nil {
+		c.handleSparkApplicationDeletion(app)
 		c.recorder.Eventf(
 			app,
 			apiv1.EventTypeNormal,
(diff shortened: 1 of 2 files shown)


kumare3:
@liyinan926 Now I understand that there is a bug in k8s 1.10 and before. But, besides that I still feel Finalizer (when supported) is the right solution as you want the deletion to propagate without leaking resources. If we end up using TTL, we may prevent users from launching another SparkApplication with the same name (de-duping during spark-submit).

My proposal, keep finalizer but optionally disable it for some versions? WDYT?

liyinan926:
@kumare3 TTL will only be applicable to pods not deleted in `onDelete`. So in practice it should not be an issue. To me, finalizers (along with initializers) are dangerous beasts and can lead to subtle bugs that are often hard to catch.

kumare3:
@liyinan926 aren't they supposed to be exactly for this scenario where you want a pre-delete cleanup. Also TTL is coming up in the future, it will have bugs at some point, how can we say that it is going to be fool-proof. Also we will need a feature flag to enable it?

liyinan926:
Yes, they are supposed to beg used for that. But the problem is a little bug that causes the finalizer to not be removed can prevent the object from being deleted. Custom resources are more prone to that because of the CRDs. I personally feel that cons of a finalizer outweighs its pros.

kumare3:
@liyinan926 I agree bugs are bad, but isn't it true with anything and shouldn't a stronger testing process be able to circumvent such errors. I am ok with you going ahead and removing the finalizer, its just that we will leak pods and then if somebody runs another SparkApplication with the same name, we will false de-dupe and mark it as complete. I feel this sort of bug is also very subtle and hard to debug.

Spark-Submit is a like an external resource and that is the problem, if we could have passed on the owner references this problem would not have happened. 

Here is my 2 cents
 - I am ok with you pushing this change, but we should probably talk to the Spark guys to see if we can pass owner references :)
- and if so alter spark-submit to allow passing owner references

liyinan926:
Per our discussion on this, I'm going to merge this PR and create a one for deduping left-over driver pod from the previous run of a job using a unique ID for each run.

PR #342, opened 2018-12-09
Fixed webhook

### LINKED ISSUE #335: webhook-service configuration via helm chart impacts mutating admission controller calls
**What happened:**
In an effort to make use of a configMap for metrics.properties the following was specified in the driver spec:
```
  driver:
    configMaps:
      - name: metricsproperties
        path: /home
```
However, the pod was launched without the configMap mounted.   The documentation states that the following:

> The Kubernetes Operator for "Apache Spark comes with an optional mutating admission webhook for customizing Spark driver and executor pods based on the specification in SparkApplication objects, e.g., mounting user-specified ConfigMaps and volumes, and setting pod affinity/anti-affinity, and adding tolerations.

By default the helm chart automatically sets up the webhook.  The old helm chart had the following name for the webhook service hardcoded:
https://github.com/helm/charts/commit/6b0bbba5ce479430949f50016281795ff31d945b#diff-2ecb319330580dabceaf98668ee5b316L125

This hardcoded value can also be found during the webhook-init job that calls /usr/bin/gencerts.sh which looks like this:
```
# Note the CN is the DNS name of the service of the webhook.
openssl req -new -key ${TMP_DIR}/server-key.pem -out ${TMP_DIR}/server.csr -subj "/CN=spark-webhook.${NAMESPACE}.svc" -config ${TMP_DIR}/server.conf
```
Updates to the helm chart break apiserver calls to the webhook service:
https://github.com/helm/charts/commit/6b0bbba5ce479430949f50016281795ff31d945b#diff-63733843ff7febc22d286c29b04fe34eR5
because the templated service name was set to be dynamic:
- If the mutatingwebhookconfiguration does not match the webhook service, the apiserver fails to make calls.
- if the certs that are generated do not match the name of the webhook service, the TLS handshake fails.

**Quick Fix**
Revert to using the hardcoded `spark-webhook` webhook service name and respective TLS cert generated CN `spark-webhook.${NAMESPACE}.svc`

yuchaoran2011:
This PR needs to be merged first and the image `gcr.io/spark-operator/spark-operator:v2.4.0-v1alpha1-latest` needs to be rebuilt and pushed. After that I'll make a PR with these [changes](https://github.com/helm/charts/compare/master...yuchaoran2011:fixwebhook) against helm/charts repo. Because only after the new image is pushed will the helm/charts CI pass on my PR.

I also made the MutatingWebhookConfiguration name configurable. We have a use case where two teams each has an operator instance running in their own namespace. The team who installs the operator later would override the webhook configuration of the team who deployed earlier.

Fixed #335

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,8 +16,8 @@
 
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.0
 
-FROM golang:1.10.2-alpine as builder
-ARG DEP_VERSION="0.4.1"
+FROM golang:1.11.2-alpine as builder
+ARG DEP_VERSION="0.5.0"
 RUN apk add --no-cache bash git
 ADD https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 /usr/bin/dep
 RUN chmod +x /usr/bin/dep
@@ -32,4 +32,6 @@ FROM ${SPARK_IMAGE}
 COPY --from=builder /usr/bin/spark-operator /usr/bin/
 RUN apk add --no-cache openssl curl
 COPY hack/gencerts.sh /usr/bin/
+
 ENTRYPOINT ["/usr/bin/spark-operator"]
+
--- a/hack/gencerts.sh
+++ b/hack/gencerts.sh
@@ -31,11 +31,11 @@ EOF
 }
 
 function parse_arguments {
-  while [ $# -gt 0 ]
+  while [[ $# -gt 0 ]]
   do
     case "$1" in
       -n|--namespace)
-      if [ -n "$2" ]; then
+      if [[ -n "$2" ]]; then
         NAMESPACE="$2"
       else
         echo "-n or --namespace requires a value."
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -16,8 +16,8 @@
 
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.0
 
-FROM golang:1.10.2-alpine as builder
-ARG DEP_VERSION="0.4.1"
+FROM golang:1.11.2-alpine as builder
+ARG DEP_VERSION="0.5.0"
```
skonto: Is this safe enough?

Dockerfile
```diff
@@ -16,8 +16,8 @@
 
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.0
 
-FROM golang:1.10.2-alpine as builder
-ARG DEP_VERSION="0.4.1"
+FROM golang:1.11.2-alpine as builder
```
skonto: Is this safe enough?

hack/gencerts.sh
```diff
@@ -23,18 +23,19 @@ function usage {
   cat<< EOF
   Usage: $SCRIPT
   Options:
-  -h | --help               	   Display help information.
+  -h | --help               	 Display help information.
```
liyinan926: The indention seems off here.

liyinan926:
@yuchaoran2011 the latest image `gcr.io/spark-operator/spark-operator:v2.4.0-v1alpha1-latest` has been pushed.

PR #343, opened 2018-12-11
OpenShift image and fixed bug on running multiple operators/webhooks

yuchaoran2011:
1. Added support for running the containers using a non-root user, which is useful for running as a `restricted` SCC on OpenShift, as also stated in #189. Added a separate Dockerfile `Dockerfile.rh` that builds an OpenShift operator image.
2. Fixed a bug with running multiple instances of operator. Previously, the `MutatingWebhookConfiguration` objects created by the operator did not have a namespace selector, causing multiple webhooks to act on the same Spark pods when each webhook was supposed to only act on pods in the pre-defined namespace. Now with the fix, the chart automatically labels the Spark job namespace properly. (The changes to the chart are coming once this one is merged).
3. Fixed doc to reflect that webhook is now disabled by default in the Helm chart.

This PR was jointly created with my colleague @skonto.


pkg/webhook/webhook.go
```diff
@@ -168,7 +169,14 @@ func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (wh *WebHook) selfRegistration(webhookConfigName string) error {
+func (wh *WebHook) selfRegistration(webhookConfigName string, webhookNamespaceSelectorLabel string) error {
+	namespaceSelector := metav1.LabelSelector{}
```
liyinan926: How about using the value of flag `-namespace` instead if it's set?

entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+# If there is no passwd entry for the container UID, attempt to create one
+if [[ -z "$uidentry" ]] ; then
+    if [[ -w /etc/passwd ]] ; then
+        echo "$myuid:x:$myuid:$mygid:anonymous uid:$SPARK_HOME:/bin/false" >> /etc/passwd
+    else
+        echo "Container ENTRYPOINT failed to add passwd entry for anonymous UID"
+    fi
+fi
+
+exec /sbin/tini -s -- /usr/bin/spark-operator "$@"
```
liyinan926: So this also works with for non-openshift clusters, right?

pkg/webhook/webhook.go
yuchaoran2011: I assume you meant the Spark job namespace that the operator is monitoring. The label needs to be added on that namespace for the webhook to work. Basically there are two scenarios:

1. The `-namespace` flag sets the Spark job namespace to `myNs`. Then the user needs to manually add some label (e.g. `testLabel=stuff`) to the namespace `myNs`. After that the user can install the operator Helm chart by passing this label (I'll create a PR to the chart once this one is merged). Note that if the user forgot to do that during `helm install`, he can also later manually edit the operator-generated MutatingWebhookConfiguration object.

2. The `-namespace` flag is set to empty string, meaning the operator monitors all namespaces. In this case, nothing else needs to be done. Everything remains the same as before because the entire cluster is going to have just one operator deployment. 

In scenario 1, without the labels, things are still going to be fine if only one operator is deployed in the cluster. But as soon as there are two operators deployed, jobs that use the mutating admission webhook is going to fail because two webhooks are going to modify the same Spark pods.

Maybe I should update the documentation with the above?

entrypoint.sh
yuchaoran2011: Yes. I tested on GKE too.

pkg/webhook/webhook.go
yuchaoran2011: That's right. But when `--namespace` is set, the webhook still handles Spark pods in all namespaces, unless a label is set.
When `--namespace` is not set, the user needs to make sure that he doesn't set the label either.

liyinan926: > That's right. But when --namespace is set, the webhook still handles Spark pods in all namespaces, unless a label is set.

That's why I suggested setting the namespace selector here based on the value of `--namespace` so the webhook handles pods in the right namespace.

yuchaoran2011: `--namespace` is the namespace where the operator is installed. The label is supposed to identify the Spark job namespace. The `sparkJobNamespace` setting also only sets the name of the namespace, a separate config option for the label is still needed.

yuchaoran2011: If there's a way to add a label to a namespace in a Helm chart, that will do. But the only way I know of is to include a namespace resource template file in the chart. That's not ideal because Helm shouldn't explicitly manage namespace creation/deletion.

liyinan926: `--namespace` is the namespace the operator watches for Spark applications, not the namespace where the operator is installed. Instead of using a label selector and requiring users to add the same label(s) to the namespace, how about adding some logic in the webhook to filter pods in namespace specified by `--namespace`  if it's set?

yuchaoran2011: I see we are not talking about the same namespace. 
In a command such as `helm install operatorchart --namespace myns`, `--namespace` is where the operator is installed.
The namespace you were referring to was this [guy](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/spark-operator-deployment.yaml#L61), which is the `sparkJobNamespace` setting in the chart.

But I get what you are saying now. I'll take a look at that.

main.go
```diff
@@ -168,7 +170,18 @@ func main() {
 		if err != nil {
 			glog.Fatal(err)
 		}
-		if err = hook.Start(*webhookConfigName); err != nil {
+
+		jobNamespace, err := kubeClient.CoreV1().Namespaces().Get(*namespace, metav1.GetOptions{})
+		namespaceSelector := map[string]string{}
+		namespaceSelector["webhookNamespaceSelectorLabel"] = *namespace
```
yuchaoran2011: Add label `webhookNamespaceSelectorLabel: <sparkJobNamespace>`

pkg/webhook/webhook.go
yuchaoran2011: Done. Please take a look.

main.go
liyinan926: OK, this is not what I was suggesting. I don't think it's a good idea to modify the metadata of a user's namespace. What I was thinking is to check the namespace of each pod in the webhook and simply return a `AdmissionResponse` with `Allowed` set to true if `namespace` is set and  the pod is not in that namespace. This is not ideal, but works. Unfortunately, the `Webhook` API doesn't support selecting namespaces using their names, which to me seems the most natural way of selecting a single namespace.

liyinan926: Another alternative is to get the `Namespace` object with the name specified through `--namespace` in `webhook.go`, and get the labels on the `Namespace`, and use them to construct the namespace selector. If the `Namespace` resource has no labels, fall back to the filtering mechanism I mentioned above.

yuchaoran2011: Yeah if only webhook API supports selecting a namespace using its name, that would be convenient. I'll come back to this PR and finish it after the holiday break.

yuchaoran2011: @liyinan926 I pushed a new commit using your suggested approach. Please take a look. Now no metadata is added/modified. The decision of whether to mutate a pod or not is made solely by comparing the pod's namespace and the pre-set job namespace.

yuchaoran2011: @liyinan926 Have you got a chance to review the changes?

liyinan926: Sorry I haven't got a chance to review. I will try to get to this asap.

pkg/webhook/webhook.go
```diff
@@ -251,7 +257,9 @@ func mutatePods(review *admissionv1beta1.AdmissionReview) *admissionv1beta1.Admi
 
 	response := &admissionv1beta1.AdmissionResponse{Allowed: true}
 
-	if !isSparkPod(pod) {
+	glog.Info("Spark job namespace: ", sparkJobNamespace)
```
liyinan926: Do we need this log?

pkg/webhook/webhook.go
```diff
@@ -114,6 +119,7 @@ func (wh *WebHook) Stop(webhookConfigName string) error {
 }
 
 func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
+	glog.Info("Serving admission request")
```
liyinan926: Can we change this to `glog.V(2).Info()`?

pkg/webhook/webhook.go
```diff
@@ -55,18 +56,22 @@ type WebHook struct {
 	serviceRef *v1beta1.ServiceReference
 }
 
+var sparkJobNamespace string
```
liyinan926: Instead of using a package variable, can we make this a field in `Webhook`?

pkg/webhook/webhook.go
```diff
@@ -251,7 +257,9 @@ func mutatePods(review *admissionv1beta1.AdmissionReview) *admissionv1beta1.Admi
 
 	response := &admissionv1beta1.AdmissionResponse{Allowed: true}
 
-	if !isSparkPod(pod) {
+	glog.Info("Spark job namespace: ", sparkJobNamespace)
+	if !isSparkPod(pod) || !inSparkJobNamespace(review.Request.Namespace) {
+		glog.Info(pod.Name, " in namespace ", review.Request.Namespace, " not mutated")
```
liyinan926: Can we use `glog.V(2)` for this?

pkg/webhook/webhook.go
yuchaoran2011: That was my initial thought too. But in the `New` function in `webhook.go`, it's this line `mux.HandleFunc(path, hook.serve)` that passes the handler function. The signature for the handler function (i.e. `serve` in this case) can't be changed:
```
func(ResponseWriter, *Request)
```
So unless I'm missing something, making the job namespace a field of Webhook object doesn't allow passing it to the handler function.

pkg/webhook/webhook.go
yuchaoran2011: Deleted.

pkg/webhook/webhook.go
yuchaoran2011: Done.

pkg/webhook/webhook.go
liyinan926: If you make it a field of the `Webhook` `struct`, you can access it within `serve` by `wh.sparkJobNamespace`.

yuchaoran2011: Got it. Done.

Dockerfile.rh
```diff
@@ -0,0 +1,50 @@
...
+COPY install_packages.sh /
+RUN --mount=target=/opt/spark/credentials,type=secret,id=credentials,required /install_packages.sh
+RUN rm /install_packages.sh
+
+RUN chmod -R u+x /tmp
+
+RUN apk add --no-cache openssl curl tini
+COPY hack/gencerts.sh /usr/bin/
+COPY entrypoint.sh /usr/bin/
+USER 185
```
liyinan926: What is this?

yuchaoran2011: 185 is a default user ID for the RedHat Openshift image. For example, see this [link](https://lists.openshift.redhat.com/openshift-archives/users/2016-March/msg00283.html).

liyinan926: Can we make it an `ARG` with a default value of 185?

liyinan926:
Also please fix `pkg/webhook/webhook_test.go`.

Dockerfile.rh
yuchaoran2011: Done.

PR #358, opened 2018-12-27
Use a Kubernetes Job to run spark-submit for multi-version support

### LINKED ISSUE #182: Multiple spark versions for breaking changes in spark submit
Suppose a user's cluster has a Spark operator bundled with version 2.3 of Spark Submit. Suppose then we introduce a breaking change in the contract between Spark Submit and the Spark application docker containers. Therefore any Spark application using Spark 2.4 needs to use Spark 2.4's spark submit, and conversely, all applications using Spark 2.3 need to use Spark 2.3's spark submit.

Unfortunately it's not easy right now to support having a single Spark operator deploying both Spark 2.3 and Spark 2.4 applications. If you upgrade the Spark operator's bundled Spark submit version to 2.4, suddenly all applications running Spark 2.3 will break.

There are multiple ways to tackle this problem. First of all, we should be ensuring that the contract between spark submit and the Spark driver it launches is stable. This has been the case for a long time with Spark on YARN.

However, we can be clever and allow for the above use case even if the above contract breaks, as follows:

- Add a `sparkVersion` field to the spec of `SpatkApplication`
- The Spark operator comes bundled with a certain Spark submit version that is compatible with some subset of spark versions of Spark applications
- A given operator will only deploy a `SparkApplication` if the `sparkVersion` declared by that application is compatible with its bundled spark-submit version.

This allows one to implement something like a rolling upgrade strategy for Spark applications run on their cluster. As an example, a cluster administrator can install a Spark 2.4 compatible Spark operator and leave the Spark 2.3 operator running on the cluster. Once the organization has upgraded all Spark applications to Spark 2.4 and higher, the Spark 2.3 operator can be decommissioned.

Tasks:
* [x] Add a new field for specifying the Spark version.
* [x] Use a Kubernetes `Job` to run `spark-submit`.

liyinan926:
Fixes #182.

@akhurana001 @kumare3 This PR changes the submission logic by creating a Kubernetes `Job` to run `spark-submit`. This separates the concern of doing application submission from the operator so the operator is able to support any version of Spark with native k8s integration.

Note that I created this PR against a branch named `multi-version-support` because this is a very significant change and I would like to mark it experimental and have you guys test it thoroughly before making it to the master branch.


pkg/controller/sparkapplication/controller.go
```diff
@@ -476,14 +482,6 @@ func (c *Controller) syncSparkApplication(key string) error {
 			appToUpdate.Status.AppState.State = v1alpha1.PendingRerunState
 			appToUpdate.Status.AppState.ErrorMessage = ""
 		}
-	case v1alpha1.FailedSubmissionState:
-		if !shouldRetry(appToUpdate) {
-			// App will never be retried. Move to terminal FailedState.
-			appToUpdate.Status.AppState.State = v1alpha1.FailedState
-			c.recordSparkApplicationEvent(appToUpdate)
-		} else if hasRetryIntervalPassed(appToUpdate.Spec.RestartPolicy.OnSubmissionFailureRetryInterval, appToUpdate.Status.SubmissionAttempts, appToUpdate.Status.LastSubmissionAttemptTime) {
```
akhurana001: We should remove OnSubmissionFailureRetryInterval since it's not used anywhere as Job doesn't take the retry interval as an input.

pkg/controller/sparkapplication/controller.go
```diff
@@ -447,8 +440,21 @@ func (c *Controller) syncSparkApplication(key string) error {
...
+		if err != nil {
+			return err
+		}
+		if succeeded != nil {
+			if *succeeded {
+				c.createSparkUIResources(appToUpdate)
+				appToUpdate.Status.AppState.State = v1alpha1.SubmittedState
+				appToUpdate.Status.ExecutionAttempts += 1
+			} else {
+				appToUpdate.Status.AppState.State = v1alpha1.FailedSubmissionState
```
akhurana001: Looks like FailedSubmissionState is a terminal state now as we end up in this only after exhausting all the retries. This is a non-backward compatible change since till now only Failed/Completed were terminal states. Does it make sense to deprecate FailedSubmissionState and move this to Failed here to be backward compatible ?

pkg/controller/sparkapplication/controller.go
```diff
@@ -546,63 +542,52 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
...
-	glog.Infof("SparkApplication %s/%s has been submitted", app.Namespace, app.Name)
+	glog.Infof("Submission Job for SparkApplication %s/%s has been created", app.Namespace, app.Name)
 	app.Status = v1alpha1.SparkApplicationStatus{
-		AppState: v1alpha1.ApplicationState{
-			State: v1alpha1.SubmittedState,
-		},
-		SubmissionAttempts:        app.Status.SubmissionAttempts + 1,
-		ExecutionAttempts:         app.Status.ExecutionAttempts + 1,
+		AppState:                  v1alpha1.ApplicationState{State: v1alpha1.PendingSubmissionState},
 		LastSubmissionAttemptTime: metav1.Now(),
```
akhurana001: Because now Job controller drives submission-attempts, LastSubmissionAttemptTime  is probably not what we expect it to be ? Let's remove or rename this ?

pkg/controller/sparkapplication/job.go
```diff
@@ -0,0 +1,117 @@
...
+
+type submissionJobManager struct {
+	kubeClient kubernetes.Interface
+}
+
+func (sjm *submissionJobManager) createSubmissionJob(s *submission) (*batchv1.Job, error) {
+	var image string
+	if s.app.Spec.Image != nil {
+		image = *s.app.Spec.Image
+	} else if s.app.Spec.Driver.Image != nil {
```
akhurana001: As we discussed, using customer image to run spark-submit might be a problem given that Job controller could start multiple pods running spark-submit.

Should we use our own images with a wrapper script over spark-submit ?

pkg/controller/sparkapplication/job.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	// Copy the labels on the SparkApplication to the Job.
+	for key, val := range s.app.Labels {
+		job.Labels[key] = val
+	}
+	return sjm.kubeClient.BatchV1().Jobs(s.app.Namespace).Create(job)
+}
+
+func (sjm *submissionJobManager) getSubmissionJob(app *v1alpha1.SparkApplication) (*batchv1.Job, error) {
+	return sjm.kubeClient.BatchV1().Jobs(app.Namespace).Get(getSubmissionJobName(app), metav1.GetOptions{})
```
akhurana001: We should probably use informers/lister to get status instead of fetching from API server on each re-sync (similar to how we do for SparkApplication/Pods)

akhurana001:
Thanks @liyinan926 for picking this up. Some comments on the PR. Let me know if you have any questions .

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go
liyinan926: Yes, removed.

pkg/controller/sparkapplication/controller.go
liyinan926: We can remove it.

pkg/controller/sparkapplication/job.go
liyinan926: It's the Job's lack of exactly-once semantics that might lead to that issue, not which image is used. Since we set the name of the Kubernetes Job deterministically, this should not be an issue.

pkg/controller/sparkapplication/job.go
liyinan926: We can do that later given the changes are well tested.

pkg/controller/sparkapplication/controller.go
```diff
@@ -546,63 +545,52 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
...
-			AppState: v1alpha1.ApplicationState{
-				State:        v1alpha1.FailedSubmissionState,
-				ErrorMessage: err.Error(),
-			},
-			SubmissionAttempts:        app.Status.SubmissionAttempts + 1,
-			LastSubmissionAttemptTime: metav1.Now(),
+		if !errors.IsAlreadyExists(err) {
+			app.Status = v1alpha1.SparkApplicationStatus{
+				AppState: v1alpha1.ApplicationState{
+					State:        v1alpha1.FailedSubmissionState,
```
akhurana001: Looks like we are still setting the state to FailedSubmissionState at a couple of places in this method. We will need to handle retries for these errors when we fail without even creating a Job and we can't get rid of the logic in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/358/files#diff-b7eea120d7ac0e8587781b863b7db8e4L479 . Sorry I missed this earlier in the last review.

pkg/controller/sparkapplication/job.go
akhurana001: Ok, We won't be able to use this on our production clusters without using informers since its puts a lot of load on the api-server. We should aim to get this in before merging this to the master branch.

pkg/controller/sparkapplication/job.go
akhurana001: Yes, but from my understanding, if the Job ends up creating 2 pods, one of the pod will fail (because spark-submit will fail due to an AlreadyExists error from the kube-api) and that can potentially cause the Job to fail leading to the Operator marking the SparkApplication as Failed when there was a successful spark-submit.

As we discussed in our sync-up, If we use custom spark distribution images with a wrapper over spark-submit, we can identify these cases (when we get an AlreadyExists error from kube-api) and mark the pod as Successful to prevent the issue explained above.

kumare3: @liyinan926 i think I agree with you both, which is weird. I see your point that running customer images would relieve us from the pressure of managing our own distribution images. But, failure in the job execution (which we have seen a few times) due to non-idempotent behavior is confusing for the user. It will essentially result in the Application (Job) failing, when it reality it was executed (or is in progress). This has repercussions on 
- **correctness**: user expect the job to not have started/executed, but it has infact executed and may affect a retry - _like corrupted data_
- **cost attribution**: user expects that the job is not running, but the job executes and ends up used unintended resources. This can be extremely important for companies that run on cloud and every extra compute hour spent affects its bottom line.

Its a hard choice, but I will choose _correctness_ and _predictable cost_ over simplified user interface. Also in this case both the options do not really affect the user experience.

kumare3: On talking more with Anmol, _Pods_ may have the same problem as the _Jobs_ and both the above mentioned problems are true with this too.
the real solution seems to be to make `spark-submit` more reliable and idempotent. Each run of spark-submit should result in success if the pod with that name exists or a predictable exit code.
I thnk predictable exit-code is a better solution and we will handle that in the operator.

On that note, I am ok with you going ahead with this change and hopefully asking Spark to make a change to spark-submit?

kumare3: Also can you document this as part of the PR, and probably we can open an issue with spark and link that in the docs,

pkg/controller/sparkapplication/job.go
liyinan926: Will update this to use an informer.

liyinan926: Done.

pkg/controller/sparkapplication/controller.go
liyinan926: Restored.

akhurana001:
👍  I think the build is failing but otherwise looks good to me.

liyinan926:
> 👍 I think the build is failing but otherwise looks good to me.

Thanks for the review. Will fix the build.

akhurana001:
@liyinan926 any update, can we prioritize/close out on this ?

liyinan926:
@akhurana001 I fixed the build and rebased onto the latest master.

akhurana001:
Thanks @liyinan926  . I will pull the branch and try this out this week. Will update you on how it goes.

manifest/spark-rbac.yaml
```diff
@@ -26,18 +26,23 @@ metadata:
   namespace: default
   name: spark-role
 rules:
+<<<<<<< HEAD
```
akhurana001: merge conflict clean-up

pkg/controller/sparkapplication/job.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	// Copy the labels on the SparkApplication to the Job.
+	for key, val := range s.app.Labels {
+		job.Labels[key] = val
+	}
+	return sjm.kubeClient.BatchV1().Jobs(s.app.Namespace).Create(job)
+}
+
+func (sjm *submissionJobManager) getSubmissionJob(app *v1beta1.SparkApplication) (*batchv1.Job, error) {
+	return sjm.jobLister.Jobs(app.Namespace).Get(app.Name)
```
akhurana001: use getSubmissionJobName

akhurana001:
Hi @liyinan926  

I tried this out and there are a couple of minor issues. Please look at the comments.

manifest/spark-operator-rbac.yaml
```diff
@@ -45,6 +45,9 @@ rules:
 - apiGroups: [""]
   resources: ["events"]
   verbs: ["create", "update", "patch"]
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["create", "get", "update", "delete"]
```
akhurana001: would need list/watch permissions as well for jobs

liyinan926: Done.

manifest/spark-rbac.yaml
liyinan926: Done.

pkg/controller/sparkapplication/job.go
liyinan926: Done.

PR #362, opened 2019-01-11
- added error logging for prometheus configuration issues, otherwise

andrewgdavis:
the code path would fail silently with a misconfiguration
- string format work around for prometheus push gateway issue:
  https://github.com/prometheus/pushgateway/issues/97

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -536,9 +536,9 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
 	// Make a copy since configPrometheusMonitoring may update app.Spec which causes an onUpdate callback.
 	appToSubmit := app.DeepCopy()
 	if appToSubmit.Spec.Monitoring != nil && appToSubmit.Spec.Monitoring.Prometheus != nil {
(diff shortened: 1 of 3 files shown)


andrewgdavis:
I signed it!

pkg/controller/sparkapplication/controller.go
```diff
@@ -536,7 +536,9 @@ func (c *Controller) submitSparkApplication(app *v1alpha1.SparkApplication) *v1a
 	// Make a copy since configPrometheusMonitoring may update app.Spec which causes an onUpdate callback.
 	appToSubmit := app.DeepCopy()
 	if appToSubmit.Spec.Monitoring != nil && appToSubmit.Spec.Monitoring.Prometheus != nil {
-		configPrometheusMonitoring(appToSubmit, c.kubeClient)
+        /* report configuation errors, don't fail silently */
+        err := configPrometheusMonitoring(appToSubmit, c.kubeClient)
```
liyinan926: The err needs to be checked, i.e., `if err := configPrometheusMonitoring(appToSubmit, c.kubeClient); err != nil {glog.Error(err)}`.

liyinan926:
Also please fix the failing tests and run `go fmt` on the code.

PR #377, opened 2019-01-24
Update prometheus configmap after sparkapp was edited

andrewgdavis:


andrewgdavis:
sorry for the noise... gitconfig email address / single sign on and account management conflicts with CLA.

PR #385, opened 2019-01-31
Minor fixes to CRD validation

yuchaoran2011:
1. Upgraded Go to 1.11.5
2. Added required `sparkVersion` in the SparkApplication examples. Otherwise [validation](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/commit/4d6f9a059b86238192421268dc2384f353754383#diff-d1c35bf2dffda85b37b85868c24d9421R157) would fail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.0
 
-FROM golang:1.11.4-alpine as builder
+FROM golang:1.11.5-alpine as builder
 ARG DEP_VERSION="0.5.0"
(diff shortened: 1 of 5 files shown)


yuchaoran2011:
In the CRD [YAML](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/manifest/spark-operator-crds.yaml), only `SparkApplication` requires `sparkVersion` but not `ScheduledSparkApplication`.  You want it to be added to the validation logic too?

liyinan926:
Ah, we need to fix the CRD validation for `ScheduledSparkApplication`.

Issue #381, opened 2019-01-28
Metrics format changed?


razius:
I was trying to setup Prometheus metrics exporting but I couldn't make it work with the current config, it seems that the separator between the namespace and app name could have changed? 

Eg. I removed all the rules from the config (by default anything that doesn't match gets dropped) and I could see metrics again but I got the following help text:

```
# HELP spark_streaming_driver_receivers Attribute exposed for management (metrics<name=spark-popular-products.driver.popular-products.StreamingMetrics.streaming.receivers><>Value)
```
Where `spark` is the namespace and `popular-products` is the app name.

Checking the [match pattern](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/6d495e84d55d5226b3c28c51fe138a3d123d8ebb/spark-docker/conf/prometheus.yaml#L23) I could see that it has a `/` as the separator. I added the rules back and I changed the match pattern to the following and it seems to work again:

```
  - pattern: metrics<name=(\w+)-(\S+)\.driver\.(\S+)\.StreamingMetrics\.streaming\.(\S+)><>Value
```

oliver-tarrant-tessella:
I was having the same issue and changing the separator also seems to have worked for me. It doesn't feel like an ideal solution though as a `-` is allowed in the definitions in an application's yaml file which can cause mismatches in arguments. However, as a fix for now I have found changing each match pattern in the config file to start with `metrics<name=(\S+)-(\S+)` allows the metrics to be exported to prometheus again.

liyinan926:
Do you have `"spark.metrics.namespace": "<namespace>/<name>"` in your `SparkApplication` spec?

razius:
No, it's not filled in.

liyinan926:
OK, can you try adding the following and see if the patterns with a `/` work?

```
spec:
    sparkConf:
        "spark.metrics.namespace": "<your app namespace>/<your app name>"
```

oliver-tarrant-tessella:
I'm afraid that adding `"spark.metrics.namespace": "<my apps namespace>/<my apps name>"` to my spark application's spec has made no difference for me when using the `/` separator.

razius:
Can confirm, nothing changed.

liyinan926:
OK, so looks like `/` is not a valid separator. We will fix the patterns.

yuchaoran2011:
[...]

PR #405, opened 2019-02-13
Various Prometheus-related fixes and improvements

yuchaoran2011:
1. In the Prometheus spark-pi example YAML, added an argument of a large number to the spark-pi job. The purpose is to let the job run longer. Without it, the job finishes quickly, oftentimes without even generating any metrics, because the time is so short.
2. Fixed Prometheus pattern match error. #381 
3. Shortened the suffixes appended to the names of ConfigMaps and volumes. In our applications, we frequently run into an error like: 
```
Invalid value: "<my-custom-name-prefix>-prometheus-config-volume": must be no more than 63 characters
```
The suffix appended by the operator `-prometheus-config-volume` is too long. Making it shorter allows the user to have a longer portion to work with.


yuchaoran2011:
I'll add to this PR a more general solution to always truncate the names to less than 63 characters.

PR #406, opened 2019-02-14
Update test suite to run against arbitrary configuration

skonto:
- Updates the suite so it can use for example an arbitrary namespace and sa.
- fixes several minor issues
Example how it can be used with this patch:
- adds a build script for sparkctl

```
export SPARK_TEST_NAMESPACE=spark
export SPARK_TEST_SERVICE_ACCOUNT=spark-sa
export SPARK_TEST_IMAGE=<image>
go test -v ./test/e2e/ --kubeconfig "$HOME/.kube/config" --operator-image-pullPolicy=Always \
--operator-image=<image> \
--spark-test-namespace=spark --spark-test-image=<image> \
--spark-test-service-account=spark-sa
```

### DIFF
--- a/sparkctl/build.sh
+++ b/sparkctl/build.sh
@@ -13,13 +13,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+SCRIPT=`basename ${BASH_SOURCE[0]}`
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
 set -e
-arg1=$(head -2 /opt/spark/credentials | tail -1)
-arg2=$(head -3 /opt/spark/credentials | tail -1)
-arg3=$(head -1 /opt/spark/credentials | tail -1)
-
-subscription-manager register --username=$arg1 --password=$arg2  --name=docker
-subscription-manager attach --pool=$arg3 && yum install -y openssl
-subscription-manager remove --all
-subscription-manager unregister
-subscription-manager clean
+platforms=("linux:amd64" "darwin:amd64")
+for platform in "${platforms[@]}"
+do
+  GOOS="${platform%%:*}"
+  GOARCH="${platform#*:}"
+  echo $GOOS
+  echo $GOARCH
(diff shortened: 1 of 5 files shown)


skonto:
I signed it!

sparkctl/build.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
liyinan926: Please add the license header similar to the one in other files.

sparkctl/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+  SCRIPT=`basename ${BASH_SOURCE[0]}`
```
liyinan926: The indention seems off in a few places.

test/e2e/basic_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package e2e
 
 import (
+        "os"
```
liyinan926: Indention needs fix.

test/e2e/basic_test.go
```diff
@@ -30,8 +31,12 @@ import (
 	appFramework "github.com/GoogleCloudPlatform/spark-on-k8s-operator/test/e2e/framework"
 )
 
+var namespace = os.Getenv("SPARK_TEST_NAMESPACE")
```
liyinan926: We probably should keep things consistent by using flags instead of env variables.

skonto: it is not easy to pass flags in there.

sparkctl/build.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors.
```
yuchaoran2011: Isn't the header supposed to be the Google one used in other operator code files? "Kubernetes Authors" don't seem right here.

skonto: I just used what the other scripts have, there are two kinds right now.

skonto:
Some random idea I have is to auto-generate these specs via a template so we have a variation of different types of tests.

test/e2e/basic_test.go
skonto: @liyinan926 fixed that ^^

sparkctl/build.sh
liyinan926: No worries, I will fix them later.

PR #409, opened 2019-02-15
Support prometheus custom configuration file

yuchaoran2011:
1. Added an API to pass custom Prometheus configuration file contained in a job image. The current APIs only support passing the contents of a custom Prometheus configuration file, but not the file name. The new API allows the user to bake his config file in the Spark job image and reference it in the job YAML.

2. Automatically truncate the volume name to fit 63 chars. This [PR](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/405) introduced a band-aid solution to shorten the suffixes appended by the operator. Now the operator simply truncates the volume name to be 63 characters if it's longer than 63. The user will no longer to experience a failed job because of this issue.

Both of these issues are what we faced in our internal applications using the operator at Lightbend. We think these improvements will enhance the user experience and improve robustness of the operator. 

3. Changed the separator between the namespace and the rest of the metric name from hyphen to dot. Hyphen is really a poor choice of a separator because both the namespace and the metric name can contain hyphens, causing mismatches in the regex. Dot would solve this problem as K8s resource names are not allowed to contain dots.

### DIFF
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -478,7 +478,7 @@ type PrometheusSpec struct {
 	// Optional.
 	// If not specified, 8090 will be used as the default.
 	Port *int32 `json:"port"`
-	// ConfigFile is the path to the custom Prometheus configuration file provided in the Spark job image.
+	// ConfigFile is the path to the custom Prometheus configuration file provided in the Spark image.
 	// ConfigFile takes precedence over Configuration, which is shown below.
 	ConfigFile *string `json:"configFile,omitempty"`
 	// Configuration is the content of the Prometheus configuration needed by the Prometheus JMX exporter.
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -480,10 +480,10 @@ type PrometheusSpec struct {
 	Port *int32 `json:"port"`
 	// ConfigFile is the path to the custom Prometheus configuration file provided in the Spark job image.
 	// ConfigFile takes precedence over Configuration, which is shown below.
-	ConfigFile string `json:"configFile"`
+	ConfigFile *string `json:"configFile,omitempty"`
 	// Configuration is the content of the Prometheus configuration needed by the Prometheus JMX exporter.
 	// Optional.
 	// If not specified, the content in spark-docker/conf/prometheus.yaml will be used.
 	// Configuration has no effect if ConfigFile is set.
-	Configuration *string `json:"configuration"`
+	Configuration *string `json:"configuration,omitempty"`
 }
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -478,8 +478,12 @@ type PrometheusSpec struct {
 	// Optional.
 	// If not specified, 8090 will be used as the default.
 	Port *int32 `json:"port"`
+	// ConfigFile is the path to the custom Prometheus configuration file provided in the Spark job image.
+	// ConfigFile takes precedence over Configuration, which is shown below.
+	ConfigFile string `json:"configFile"`
 	// Configuration is the content of the Prometheus configuration needed by the Prometheus JMX exporter.
 	// Optional.
 	// If not specified, the content in spark-docker/conf/prometheus.yaml will be used.
+	// Configuration has no effect if ConfigFile is set.
 	Configuration *string `json:"configuration"`
 }
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -177,84 +177,84 @@ const DefaultPrometheusConfiguration = `
 lowercaseOutputName: true
 attrNameSnakeCase: true
 rules:
-  - pattern: metrics<name=(\S+)-(\S+)\.driver\.(BlockManager|DAGScheduler|jvm)\.(\S+)><>Value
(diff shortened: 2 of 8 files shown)


pkg/config/constants.go
```diff
@@ -177,84 +177,84 @@ const DefaultPrometheusConfiguration = `
...
     type: GAUGE
     labels:
       app_namespace: "$1"
       app_name: "$2"
-  - pattern: metrics<name=(\S+)-(\S+)\.driver\.(\S+)\.StreamingMetrics\.streaming\.(\S+)><>Value
+  - pattern: metrics<name=(\S+).(\S+)\.driver\.(\S+)\.StreamingMetrics\.streaming\.(\S+)><>Value
     name: spark_streaming_driver_$4
     type: GAUGE
     labels:
       app_namespace: "$1"
```
skonto: The app_namespace prefix needs to be removed. Spark core metrics does not have this.
The pattern is like this: https://github.com/prometheus/jmx_exporter/blob/master/example_configs/spark.yml#L14

yuchaoran2011: @liyinan926 Are you ok with me removing the namespace prefix from metric names altogether?

liyinan926: I prefer keeping them. Even though Spark core metrics in the linked example don't have namespace, we can always support them by adding the namespace as part of the prefix. It's useful to filter metrics by namespaces.

yuchaoran2011: @liyinan926 Ok. Could you review this PR and get it merged? Thanks!

pkg/webhook/webhook.go
```diff
@@ -256,7 +256,7 @@ func mutatePods(review *admissionv1beta1.AdmissionReview, sparkJobNs string) *ad
 	response := &admissionv1beta1.AdmissionResponse{Allowed: true}
 
 	if !isSparkPod(pod) || !inSparkJobNamespace(review.Request.Namespace, sparkJobNs) {
-		glog.V(2).Info(pod.Name, " in namespace ", review.Request.Namespace, " not mutated")
```
yuchaoran2011: Somehow `pod.Name` is always empty. But `pod.GetObjectMeta().GetName()` works.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -478,8 +478,12 @@ type PrometheusSpec struct {
 	// Optional.
 	// If not specified, 8090 will be used as the default.
 	Port *int32 `json:"port"`
+	// ConfigFile is the path to the custom Prometheus configuration file provided in the Spark job image.
+	// ConfigFile takes precedence over Configuration, which is shown below.
+	ConfigFile string `json:"configFile"`
```
liyinan926: This should have type `*string` to signal it's optional. We also need to add tag `omitempty` for this and the field below.

pkg/controller/sparkapplication/monitoring_config.go
```diff
@@ -39,50 +40,67 @@ const (
...
+			_, updateErr := kubeClient.CoreV1().ConfigMaps(app.Namespace).Update(cm)
+			return updateErr
+		})
 
-	if retryErr != nil {
-		return fmt.Errorf("failed to apply %s in namespace %s: %v", prometheusConfigMapName, app.Namespace, retryErr)
+		if retryErr != nil {
+			return fmt.Errorf("failed to apply %s in namespace %s: %v", prometheusConfigMapName, app.Namespace, retryErr)
+		}
+		javaOption = fmt.Sprintf("-javaagent:%s=%d:%s/%s", app.Spec.Monitoring.Prometheus.JmxExporterJar,
```
liyinan926: Please add an empty line before this line.

pkg/controller/sparkapplication/monitoring_config.go
```diff
@@ -39,50 +40,67 @@ const (
...
+	}
 
-		cm.Data = configMap.Data
-		_, updateErr := kubeClient.CoreV1().ConfigMaps(app.Namespace).Update(cm)
-		return updateErr
-	})
+	configFile := app.Spec.Monitoring.Prometheus.ConfigFile
+	var javaOption string
+	if configFile != "" {
+		glog.V(2).Infof("Overriding the default Prometheus configuration with config file %s in the Spark job image.", configFile)
```
liyinan926: In `in the Spark job image`, I think we can remove job here.

pkg/webhook/webhook.go
liyinan926: Interesting.

pkg/webhook/patch.go
```diff
@@ -206,6 +208,11 @@ func addGeneralConfigMaps(pod *corev1.Pod) []*patchOperation {
 	namesToMountPaths := config.FindGeneralConfigMaps(pod.Annotations)
 	for name, mountPath := range namesToMountPaths {
 		volumeName := name + "-vol"
+		if len(volumeName) > maxNameLength {
+			glog.V(2).Infof("ConfigMap volume name %s is too long. Truncating to length %d.", volumeName, maxNameLength)
```
liyinan926: We can combine this and the log message below into one that's after the truncation and indicates the truncation and the result name.

pkg/controller/sparkapplication/monitoring_config.go
yuchaoran2011: Line 77 is restored.

pkg/config/constants.go
skonto: @liyinan926 it does not make sense to add it. Spark core emits these metrics. How would you modify them? This right now creates problems with parsing.

liyinan926: You can specify prefix for metrics using `spark.metrics.namespace`.

skonto: Yes but the default shouldnt be that IMHO.

PR #416, opened 2019-02-21
Treat an app that has not terminated but with driver not found as failing

liyinan926:


pkg/controller/sparkapplication/controller.go
```diff
@@ -269,13 +269,13 @@ type driverState struct {
...
 
-func (c *Controller) getUpdatedAppStatus(app *v1beta1.SparkApplication) v1beta1.SparkApplicationStatus {
+func (c *Controller) updateAppStatus(app *v1beta1.SparkApplication) {
 	// Fetch all the pods for the application.
 	selector, _ := labels.NewRequirement(config.SparkAppNameLabel, selection.Equals, []string{app.Name})
 	pods, err := c.podLister.Pods(app.Namespace).List(labels.NewSelector().Add(*selector))
 	if err != nil {
 		glog.Errorf("failed to get pods for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
-		return app.Status
+		return
```
skonto: If the pods cannot be listed, that means we return nil here and the main controller logic will finally forget the app in `processNextItem`, is that the idea? Right now it will cause this to loop for ever?

liyinan926: This is a good point. I think we should return the error so it will be retried.

PR #420, opened 2019-02-21
Add InClusterconfig for sparkctl if no kubeconfig provided

sarjeet2013:
Fixes https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/419

### DIFF
--- a/sparkctl/cmd/client.go
+++ b/sparkctl/cmd/client.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
 	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
+	"os"
 )
 
 func buildConfig(kubeConfig string) (*rest.Config, error) {
--- a/sparkctl/cmd/client.go
+++ b/sparkctl/cmd/client.go
@@ -21,14 +21,13 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"os"
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
 	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
-	"os"
 )
 
 func buildConfig(kubeConfig string) (*rest.Config, error) {
-
 	// Check if kubeConfig exist
 	if _, err := os.Stat(kubeConfig); os.IsNotExist(err) {
 		// Try InclusterConfig for sparkctl running in a pod
--- a/sparkctl/cmd/client.go
+++ b/sparkctl/cmd/client.go
@@ -17,11 +17,12 @@ limitations under the License.
 package cmd
 
 import (
+	"os"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"os"
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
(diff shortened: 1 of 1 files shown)


sparkctl/cmd/client.go
```diff
@@ -24,9 +24,22 @@ import (
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
 	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
+	"os"
 )
 
 func buildConfig(kubeConfig string) (*rest.Config, error) {
+
+	// Check if kubeConfig exist
```
liyinan926: The right fix I think should be to change `getKubeConfigPath` in `root.go` to return an empty string when the default path does not exist. `BuildConfigFromFlags` will use `InClusterConfig` if the input path is empty.

sarjeet2013: Right, The `getKubeConfigPath` return the default path if `KUBECONFIG` env is not present. In that case, we can check if the default path exist as well, and return "" if it doesn't exist.

I'll update the PR

sarjeet2013:
@liyinan926 I had it modified as you suggested, but now this is getting the warning log before trying incluster config

sh-4.2# sparkctl list
ERROR: logging before flag.Parse: W0221 19:48:37.844107    1041 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
+------+-------+----------------+-----------------+
| NAME | STATE | SUBMISSION AGE | TERMINATION AGE |
+------+-------+----------------+-----------------+
+------+-------+----------------+-----------------+
This warning is from `BuildConfigFromFlags` where kubeconfig is provided “”

Should we stick to original PR?

sparkctl/cmd/client.go
```diff
@@ -24,9 +24,22 @@ import (
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
 	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
+	"os"
 )
 
 func buildConfig(kubeConfig string) (*rest.Config, error) {
+
```
liyinan926: This empty line is not needed.

sparkctl/cmd/client.go
```diff
@@ -24,9 +24,22 @@ import (
 
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
 	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
+	"os"
```
liyinan926: This should be at the top of the import list with other built-in packages.

sarjeet2013: Fixed

sparkctl/cmd/client.go
sarjeet2013: Done, Removed

sarjeet2013:
@liyinan926 Updated, PTAL. Thanks for the review.

sparkctl/cmd/client.go
```diff
@@ -21,12 +21,24 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"os"
```
liyinan926: Sorry I meant this should be in the group of golang built-in packages.

sarjeet2013: You mean, "os" at the first line in import, then an empty line, and then the rest of the import packages?

sarjeet2013:
```import (
	"os"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	clientset "k8s.io/client-go/kubernetes"
	"k8s.io/client-go/rest"
	"k8s.io/client-go/tools/clientcmd"

	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
	crdclientset "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned"
)```

liyinan926: Yes

sarjeet2013: :) Done

PR #445, opened 2019-03-18
Refactored integration tests

yuchaoran2011:
* Added an integration test for mounting a ConfigMap. This is a common scenario that we used to check manually. Currently webhook is required. But the test can be adapted with minimal changes once pod template is supported in the operator.
* Many minor bug fixes and improvements in the test code.
* Upgraded Kubernetes Go client version to 1.13.4.
* Upgraded dep to 0.5.1

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -75,19 +75,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.11.8"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   name = "k8s.io/apiextensions-apiserver"
-  version = "kubernetes-1.11.8"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.11.8"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.11.8"
+  version = "kubernetes-1.13.4"
 
 [[constraint]]
   branch = "master"
--- a/examples/spark-pi-configmap.yaml
+++ b/examples/spark-pi-configmap.yaml
@@ -29,10 +29,9 @@ spec:
   restartPolicy:
     type: Never
   volumes:
-    - name: "test-volume"
-      hostPath:
-        path: "/tmp"
-        type: Directory
+    - name: config-vol
+      configMap:
+        name: dummy-cm
   driver:
     cores: 0.1
     coreLimit: "200m"
@@ -41,8 +40,8 @@ spec:
       version: 2.4.0
     serviceAccount: spark
     volumeMounts:
-      - name: "test-volume"
-        mountPath: "/tmp"
+      - name: config-vol
+        mountPath: /opt/spark/mycm
   executor:
     cores: 1
     instances: 1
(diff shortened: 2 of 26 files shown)


Gopkg.toml
```diff
@@ -75,19 +75,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.11.8"
+  version = "kubernetes-1.13.4"
```
liyinan926: Why do we want to upgrade to this version?

examples/spark-pi-configmap.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+apiVersion: "sparkoperator.k8s.io/v1beta1"
+kind: SparkApplication
+metadata:
+  name: spark-pi
+  namespace: default
+spec:
+  type: Scala
+  mode: cluster
+  image: "lightbend/spark:2.1.0-OpenShift-2.4.0-rh"
```
liyinan926: This needs to be changed.

examples/spark-pi.yaml
```diff
@@ -22,7 +21,7 @@ metadata:
 spec:
   type: Scala
   mode: cluster
-  image: "gcr.io/spark-operator/spark:v2.4.0"
+  image: "lightbend/spark:2.1.0-OpenShift-2.4.0-rh"
```
liyinan926: Ditto.

sparkctl/cmd/event.go
```diff
@@ -151,7 +153,9 @@ func streamEvents(events watch.Interface, streamSince int64) error {
 		// Start rendering contents of the table without table header as it is already printed
 		table = prepareNewTable()
 		table.SetBorders(tablewriter.Border{Left: true, Top: false, Right: true, Bottom: false})
-		_, err := watch.Until(watchExpire, events, func(ev watch.Event) (bool, error) {
+		ctx := context.Background()
```
liyinan926: Let's use a TODO instead.

manifest/spark-operator-with-webhook.yaml
```diff
@@ -78,7 +79,7 @@ spec:
       - name: main
         image: gcr.io/spark-operator/spark-operator:v2.4.0-v1beta1-latest
         imagePullPolicy: IfNotPresent
-        command: ["/usr/bin/gencerts.sh", "-p"]
+        command: ["/usr/bin/gencerts.sh", "-n", "spark-operator", "-s", "spark-webhook", "-p"]
```
liyinan926: Do we need this change?

Gopkg.toml
yuchaoran2011: There are multiple Kubernetes Go client APIs that have been removed/changed in the new version. Examples include `WorkQueueMetrics` and `Until` that got changed in this PR. And in general, we think it's good to bring the API version used by the operator to a more recent one.

liyinan926: OK, makes sense.

manifest/spark-operator-with-webhook.yaml
yuchaoran2011: You are right. I forgot that they are already defaults in `gencerts.sh`. I've undo this change.

sparkctl/cmd/event.go
yuchaoran2011: Done.

PR #450, opened 2019-03-26
Upgraded client-go to 1.14.0

yuchaoran2011:
The lifecycle test complements the controller unit tests, but from an integration test level.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -75,19 +75,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.13.4"
+  version = "kubernetes-1.14.0"
 
 [[constraint]]
   name = "k8s.io/apiextensions-apiserver"
-  version = "kubernetes-1.13.4"
+  version = "kubernetes-1.14.0"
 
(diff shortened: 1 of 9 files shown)


Gopkg.toml
```diff
@@ -75,19 +75,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.13.4"
+  version = "kubernetes-1.14.0"
```
liyinan926: Do we really need to upgrade to 1.14 given that it's just released?

yuchaoran2011: I guess why not? But if you have any concerns, I'm fine waiting a bit (maybe after 1.14.1 is out).

liyinan926: Yes, let's wait for 1.14 to bake a bit more.

yuchaoran2011: Ok I'll make another PR containing just the added tests. Then we can merge this one later.

yuchaoran2011:
K8s 1.14.2 is out but haven't had time to work on it. Closing it for now. Will re-open when I get to it.

PR #467, opened 2019-04-03
Limit SparkApplication name to 63 characters in JSON schema validation

### LINKED ISSUE #465: Job is not submited if app-name label exceeds 63 characters
@liyinan926 After #459 fix Issue still exists. I guess, now these labels are the problem 
```
--conf spark.kubernetes.driver.label.sparkoperator.k8s.io/app-name
--conf spark.kubernetes.executor.label.sparkoperator.k8s.io/app-name
```
when I remove them, job starts with long name.

amukhopad:
Fixes #465

@liyinan926

### DIFF
--- a/manifest/spark-operator-crds.yaml
+++ b/manifest/spark-operator-crds.yaml
@@ -32,6 +32,12 @@ spec:
   validation:
     openAPIV3Schema:
       properties:
+        metadata:
+          properties:
+            name:
+              type: string
+              minLength: 1
+              maxLength: 63
         spec:
           properties:
             deps:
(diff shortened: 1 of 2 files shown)


manifest/spark-operator-crds.yaml
```diff
@@ -31,6 +31,12 @@ spec:
   scope: Namespaced
   validation:
     openAPIV3Schema:
+      metadata:
```
liyinan926: The correct nesting should be:
```
openAPIV3Schema:
    properties:
        metadata:
            properties:
```

amukhopad: Fixed

PR #473, opened 2019-04-10
Add sparkapplications, scheduledsparkapplications permissions to admin

almariah:
After installations of the operator, the `admin` role will not have any permissions to use `sparkapplications` or `scheduledsparkapplications`.

Using `rbac.authorization.k8s.io/aggregate-to-admin: "true"` annotation with `sparkoperator-aggregate-to-admin` will add these permissions to the admin role. This is very useful when you assign admin role for service accounts or users to allow them to use the operator without any extra RBACs.

### DIFF
--- a/manifest/spark-operator-rbac.yaml
+++ b/manifest/spark-operator-rbac.yaml
@@ -67,3 +67,22 @@ roleRef:
   kind: ClusterRole
   name: sparkoperator
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+  name: sparkoperator-aggregate-to-admin
+rules:
+- apiGroups: ["sparkoperator.k8s.io"]
+  resources: ["sparkapplications", "scheduledsparkapplications"]
+  verbs:
+  - create
+  - delete
+  - deletecollection
+  - get
+  - list
+  - patch
+  - update
+  - watch


almariah:
I signed it!

almariah:
It seems that the build is failing for reasons that is irrelevant to the PR itself

manifest/spark-operator-rbac.yaml
```diff
@@ -67,3 +67,22 @@ roleRef:
...
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+  name: sparkoperator-aggregate-to-admin
+rules:
+- apiGroups: ["sparkoperator.k8s.io"]
```
liyinan926: We actually have a `ClusterRole` above covers `sparkapplications` and `scheduledsparkapplications`. See line 54 above.

almariah: I see. I guess the case if you are managing for example RBACs for teams, users, service accounts, etc, and you attach a an `admin` role in specific namespace for them. Then you expect to have admin access for most of namespace resources except CRDs like sparkapplications, scheduledsparkapplications.

Aggregate that specific role in the PR will add to the admin role (which is defined by default in kubernetes) the ability to do `sparkapplications`, `scheduledsparkapplications`.

almariah: We just got that use case, because we give (for example) users in some specific namespaces  admin role, but they still can't use `sparkapplications`, `scheduledsparkapplications`. By aggregate that role to admin, any admin will be able to use them.

liyinan926: Got it. So what the changes here do is they extend the `cluster-admin` to also include full permissions to manipulate `sparkapplications` and `scheduledsparkapplications`, right?

almariah: Yes, specifically role `admin` which have some predefined permissions (not including `sparkapplications` and `scheduledsparkapplications`). We usually use this role (`admin`) for granting admin access for namespace level. But the one you have mentioned `cluster-admin` already has `[*]` for resources and verbs; there is no need to aggregate spark resources to it.

liyinan926: Got it. Thanks.

Tyler-2:
How is this activated when using the Helm chart? `admin` appears no where inside that documentation.

PR #477, opened 2019-04-19
Fixed mounting of Prometheus configuration ConfigMap

### LINKED ISSUE #468: the operator is broken when using prometheus
the config map for prometheus jmx exporter is not being mounted anymore
as a consequence the driver fail to start when using prometheus

commit `b8c7edca0eeb372fb68089b433d559f721734711` might have broken it

liyinan926:
Fixes #468.


liyinan926:
@yuchaoran2011

PR #481, opened 2019-04-26
Fix scheduler name doesn't work when specified

TommyLike:
#464 patch assumes that the scheduler name of created pod would be empty in default which is not true, this patch fixes it.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -56,11 +56,9 @@ func patchSparkPod(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOperat
 	patchOps = append(patchOps, addTolerations(pod, app)...)
 	patchOps = append(patchOps, addSidecarContainers(pod, app)...)
 
-	if pod.Spec.SchedulerName == "" {
-		op := addSchedulerName(pod, app)
-		if op != nil {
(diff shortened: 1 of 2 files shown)


pkg/webhook/patch.go
```diff
@@ -299,13 +297,13 @@ func addTolerations(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOpera
 
 func addSchedulerName(pod *corev1.Pod, app *v1beta1.SparkApplication) *patchOperation {
 	var schedulerName *string
-	if util.IsDriverPod(pod) {
+	if util.IsDriverPod(pod) && app.Spec.Driver.SchedulerName != nil {
```
liyinan926: You don't need the additional condition here since you check `schedulerName == nil` below.

pkg/webhook/patch.go
```diff
@@ -299,13 +297,13 @@ func addTolerations(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOpera
...
 		schedulerName = app.Spec.Driver.SchedulerName
 	}
-	if util.IsExecutorPod(pod) {
+	if util.IsExecutorPod(pod) && app.Spec.Executor.SchedulerName != nil {
```
liyinan926: Ditto.

PR #486, opened 2019-05-01
Fixed job restart and upgraded Prometheus exporter jar and Go

### LINKED ISSUE #453: SparkApp CR object not updated
We have two SparkApps that are started together and running concurrently. When one of them is updated with a new job image, the CRD object is supposed to enter a PENDING_RERUN state and then re-submitted for execution. But it seems to never get to SUBMITTED state. Instead we saw that the old driver pod was never deleted and the new driver pod was never started. 

We tried a single SparkApp and two SparkApps running serially. Both worked. It's just the aforementioned scenario that failed. Any idea what the cause would be? Does it have anything to do with the number of controller threads? Note that two SparkApps do have different names, so it would be surprising that one of them might interfere/conflict with another.

Below are the logs we saw:

yuchaoran2011:
1. Fixed #453 (Fix is in `controller.go`)
2. Updated Prometheus exporter agent jar version.
3. Updated Go.


pkg/controller/sparkapplication/controller.go
```diff
@@ -494,16 +494,21 @@ func (c *Controller) syncSparkApplication(key string) error {
...
+			return err
+		}
+		glog.V(2).Infof("Invalidating app pod: %s", appToUpdate.Status.DriverInfo.PodName)
 		// Invalidate the current run and enqueue the SparkApplication for re-execution.
 		if err := c.deleteSparkResources(appToUpdate); err != nil {
 			glog.Errorf("failed to delete resources associated with SparkApplication %s/%s: %v",
 				appToUpdate.Namespace, appToUpdate.Name, err)
 			return err
 		}
-		c.clearStatus(&appToUpdate.Status)
```
yuchaoran2011: `clearStatus` resets the driverInfo, which the driver pod name is a part of. In order for `if c.validateSparkResourceDeletion(appToUpdate)` to do its job, the status should be reset only after the check is complete.

liyinan926: `clearStatus` does not clear `DriverInfo` in `Invalidating` state. We should keep this call.

pkg/controller/sparkapplication/controller.go
```diff
@@ -494,16 +494,21 @@ func (c *Controller) syncSparkApplication(key string) error {
...
 		appToUpdate.Status.AppState.State = v1beta1.PendingRerunState
 	case v1beta1.PendingRerunState:
+		glog.V(2).Infof("Sparkapp named %s pending rerun", appToUpdate.Name)
```
liyinan926: Please use `SparkApplication` in logs for consistency.

pkg/controller/sparkapplication/controller.go
```diff
@@ -494,16 +494,21 @@ func (c *Controller) syncSparkApplication(key string) error {
 			appToUpdate = c.submitSparkApplication(appToUpdate)
 		}
 	case v1beta1.InvalidatingState:
+		if err := c.getAndUpdateAppState(appToUpdate); err != nil {
+			return err
+		}
+		glog.V(2).Infof("Invalidating app pod: %s", appToUpdate.Status.DriverInfo.PodName)
```
liyinan926: This log doesn't seem necessary.

pkg/controller/sparkapplication/controller.go
```diff
@@ -494,16 +494,21 @@ func (c *Controller) syncSparkApplication(key string) error {
...
 			glog.Errorf("failed to delete resources associated with SparkApplication %s/%s: %v",
 				appToUpdate.Namespace, appToUpdate.Name, err)
 			return err
 		}
-		c.clearStatus(&appToUpdate.Status)
 		appToUpdate.Status.AppState.State = v1beta1.PendingRerunState
 	case v1beta1.PendingRerunState:
+		glog.V(2).Infof("Sparkapp named %s pending rerun", appToUpdate.Name)
 		if c.validateSparkResourceDeletion(appToUpdate) {
+			glog.V(2).Infof("Resources for sparkapp %s successfully deleted", appToUpdate.Name)
```
liyinan926: Ditto.

pkg/controller/sparkapplication/controller.go
yuchaoran2011: You are right. I mis-read the code. Now added back.

pkg/controller/sparkapplication/controller.go
```diff
@@ -494,6 +494,9 @@ func (c *Controller) syncSparkApplication(key string) error {
 			appToUpdate = c.submitSparkApplication(appToUpdate)
 		}
 	case v1beta1.InvalidatingState:
+		if err := c.getAndUpdateAppState(appToUpdate); err != nil {
```
liyinan926: Put more thoughts into this. Instead of doing this here, which does a lot more things than getting the driver pod name, which is what is essentially needed, how about add logic in `deleteSparkResources` to first list and find the driver pod, and then does the deletion? We can abstract the code for listing and finding the driver pod into a utility method and use it in both `deleteSparkResources` and `getAndUpdateAppState`.

yuchaoran2011: Sounds good to me. I'll make the changes and push them later today.

yuchaoran2011: I refactored the code a little bit so that the original `getAndUpdateAppState` is split into two sub-functions: `getAndUpdateDriverState` and `getAndUpdateExecutorState`. The former is what's needed in `deleteSparkResources`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -362,6 +375,12 @@ func (c *Controller) getAndUpdateAppState(app *v1beta1.SparkApplication) error {
 	return nil
 }
 
+func (c *Controller) getAndUpdateAppState(app *v1beta1.SparkApplication) error {
+	_ = c.getAndUpdateDriverState(app)
+	_ = c.getAndUpdateExecutorState(app)
```
yuchaoran2011: I could have simply written here `c.getAndUpdateExecutorState(app)` without the underscore and assignment. But my editor GoLand complained about "error not handled".

liyinan926: You might want to do:
```
if err := c.getAndUpdateDriverState(app); err != nil {
    return err
}
if err := c.getAndUpdateExecutorState(app); err != nil {
    return err
}
```

pkg/controller/sparkapplication/controller.go
```diff
@@ -336,6 +325,30 @@ func (c *Controller) getAndUpdateAppState(app *v1beta1.SparkApplication) error {
 			app.Status.TerminationTime = metav1.Now()
 		}
 	}
+	return nil
+}
+
+// getAndUpdateExecutorState lists the executor pods of the application
+// and updates the executor state based on the current phase of the pods.
+func (c *Controller) getAndUpdateExecutorState(app *v1beta1.SparkApplication) error {
+	pods, _ := c.getAppPods(app)
```
liyinan926: We need to handle the potential error:
```
pods, err := c.getAppPods(app)
if err != nil {
    return err
}
```

pkg/controller/sparkapplication/controller.go
```diff
@@ -693,10 +709,11 @@ func (c *Controller) getSparkApplication(namespace string, name string) (*v1beta
 
 // Delete the driver pod and optional UI resources (Service/Ingress) created for the application.
 func (c *Controller) deleteSparkResources(app *v1beta1.SparkApplication) error {
-	glog.V(2).Infof("Deleting driver pod named %s", app.Status.DriverInfo.PodName)
+	_ = c.getAndUpdateDriverState(app)
```
liyinan926: Ditto, we should handle the potential error.

pkg/controller/sparkapplication/controller.go
yuchaoran2011: Thanks. Done.

pkg/controller/sparkapplication/controller.go
```diff
@@ -688,9 +719,11 @@ func (c *Controller) getSparkApplication(namespace string, name string) (*v1beta
 
 // Delete the driver pod and optional UI resources (Service/Ingress) created for the application.
 func (c *Controller) deleteSparkResources(app *v1beta1.SparkApplication) error {
+	_ = c.getAndUpdateDriverState(app)
```
liyinan926: We should check the return err of this as well.

pkg/controller/sparkapplication/controller.go
```diff
@@ -292,19 +297,6 @@ func (c *Controller) getAndUpdateAppState(app *v1beta1.SparkApplication) error {
 				currentDriverState.completionTime = metav1.Now()
 			}
```
liyinan926: One more thing, you can add a break here since there should only be one driver pod.

yuchaoran2011: Yep. I should do that. Thanks for reminding me.

pkg/controller/sparkapplication/controller.go
```diff
@@ -292,19 +297,7 @@ func (c *Controller) getAndUpdateAppState(app *v1beta1.SparkApplication) error {
...
-			if !exists || newState != oldState {
-				c.recordExecutorEvent(app, newState, pod.Name)
-			}
-			executorStateMap[pod.Name] = newState
-
-			if executorApplicationID == "" {
-				executorApplicationID = getSparkApplicationID(pod)
-			}
-		}
+		break
```
liyinan926: The break should be within the if block.

yuchaoran2011: lol I have to run and did it in a hurry without looking. I'll fix it when I'm back in an hour.

yuchaoran2011: Fixed.

PR #494, opened 2019-05-13
Expose Driver Container Error Code on Pod Failure

akhurana001:
@liyinan926  PTAL . I think this will be useful to expose reason why the driver pod failed like OOM etc. which we see quite a bit. Let me know if you have any questions/concerns.

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -301,7 +301,8 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) erro
 			// Fetch container ExitCode/Reason if Pod Failed.
 			if pod.Status.Phase == apiv1.PodFailed {
 				if len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].State.Terminated != nil {
-					currentDriverState.err = fmt.Errorf("driver pod failed with ExitCode: %d, Reason: %s", pod.Status.ContainerStatuses[0].State.Terminated.ExitCode, pod.Status.ContainerStatuses[0].State.Terminated.Reason)
+					terminatedState := pod.Status.ContainerStatuses[0].State.Terminated
+					currentDriverState.err = fmt.Errorf("driver pod failed with ExitCode: %d, Reason: %s", terminatedState.ExitCode, terminatedState.Reason)
 				} else {
 					currentDriverState.err = fmt.Errorf("driver container status missing.")
 				}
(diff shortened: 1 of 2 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -296,6 +298,14 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) erro
 			if pod.Status.Phase == apiv1.PodSucceeded || pod.Status.Phase == apiv1.PodFailed {
 				currentDriverState.completionTime = metav1.Now()
 			}
+			// Fetch container ExitCode/Reason if Pod Failed.
+			if pod.Status.Phase == apiv1.PodFailed {
+				if len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].State.Terminated != nil {
```
liyinan926: Looks like it worths creating a variable for `pod.Status.ContainerStatuses[0].State` as its repeated a few times here.

pkg/controller/sparkapplication/controller.go
```diff
@@ -318,6 +328,10 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) erro
 				app.Status.TerminationTime = currentDriverState.completionTime
 			}
 		}
+		// Expose any errors to the users.
+		if currentDriverState.err != nil {
+			app.Status.AppState.ErrorMessage += currentDriverState.err.Error()
```
liyinan926: I think you can just do assign instead of appending here.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -333,7 +333,7 @@ func TestShouldRetry(t *testing.T) {
 	}
 
 	restartPolicyAlways := v1beta1.RestartPolicy{
-		Type:                             v1beta1.Always,
+		Type: v1beta1.Always,
```
liyinan926: The formatting seems off. Can you run `go fmt ./...`?

pkg/controller/sparkapplication/controller.go
akhurana001: Created one for pod.Status.ContainerStatuses[0].State.Terminated  after validating  that the length of ContainerStatuses ?

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller_test.go
akhurana001: done

PR #495, opened 2019-05-14
Mark SparkApplication with no driver as Failed

akhurana001:
We  are seeing some applications stuck forever because the driver pod was never recorded: deleted/evicted somehow before the operator can record it. I think we should just mark the application as failing in these cases and let it retry etc if the driver pod is not found.

I see we went through a couple of iterations on this in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/416/files and https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/commit/c28c593e1168b0cf7afe03941c991a55b0e5a521 . Let me know what your thoughts are on this.  @liyinan926 

**Edit**: Just realized that since we are using the informer cache to fetch the pods, there can still be an edge-case where the next SparkApplication resync after submission (after some seconds) happens before the pod cache is updated.  Ideally the resync period is large enough  (default 30) that this shouldn't happen but we can't guarantee that would hold.  We can solve this by adding a timeout on the lastSubmissionTime field i.e. move to failing if it has been 10 mins since we submitted and we haven't seen the driver pod come up. Let me know what your thoughts are on solving this.


pkg/controller/sparkapplication/controller.go
```diff
@@ -320,14 +320,12 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) erro
...
 	} else {
-		glog.Warningf("driver not found for SparkApplication: %s/%s", app.Namespace, app.Name)
-		// The application has not terminated and has a recorded driver Pod, but no driver Pod was found for it.
-		// This is likely because the driver Pod was deleted. In this case, set the application state to FailingState.
-		if app.Status.TerminationTime.IsZero() && app.Status.DriverInfo.PodName != "" {
-			app.Status.AppState.ErrorMessage = "Driver Pod not found"
-			app.Status.AppState.State = v1beta1.FailingState
-			app.Status.TerminationTime = metav1.Now()
-		}
+		glog.Warningf("driver not found for SparkApplication: %s/%s. Marking SparkApplication as Failing.", app.Namespace, app.Name)
```
liyinan926: This will cause issue for applications that have been submitted but have not started running (no driver pod has been created).

liyinan926:
> We can solve this by adding a timeout on the lastSubmissionTime field i.e. move to failing if it has been 10 mins since we submitted and we haven't seen the driver pod come up. Let me know what your thoughts are on solving this.

This sounds like a good idea.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to get pods for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 	}
+
 	return pods, nil
 }
 
+// Fetch driver pod from APIServer.
+func (c *Controller) getDriverPodFromApiServer(app *v1beta1.SparkApplication) (*apiv1.Pod, error) {
+	selector := labels.SelectorFromSet(map[string]string{config.SparkAppNameLabel: app.Name, config.SparkRoleLabel: config.SparkDriverRole})
```
kumare3: I prefer not doing a list, I think this should be a Get only. As we should know the pod name that exists

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
+	} else if podsList.Size() == 1 {
+		return &podsList.Items[0], nil
+	}
+	return nil, nil
+}
+
 // getAndUpdateDriverState finds the driver pod of the application
 // and updates the driver state based on the current phase of the pod.
 func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) error {
 	pods, err := c.getAppPods(app)
```
kumare3: we should be getting just the driver, this code seems unnecessarily complex. I would do a lookup in the informer cache and if not found go to etcD

akhurana001: This does sound better if we can just do a get on the driverpod name rather than a list/filter using labels but there is still a edge-case where if the user updates the driver podname in the SparkApplication, we can miss terminating the old driver in such a case.

akhurana001:
@liyinan926 To update you on this, @kumare3  and I were discussing yesterday that instead of waiting for some (10?) minutes for the driver pod to show up in cache and marking the app as terminated if it doesn't, another way is to do a fetch the driver pod info from the ApiServer if not found in the cache.  Now we were thinking if we can do a get on the driver pod name instead of list with labels and if there are any edge-cases which we might miss.  Let us know what you think about this approach and if you see any issues.

liyinan926:
@akhurana001 @kumare3 I need to do some research into guarantees provided by the api server regarding read availability of an object after a successful CREATE request.

akhurana001:
> @akhurana001 @kumare3 I need to do some research into guarantees provided by the api server regarding read availability of an object after a successful CREATE request.

Thanks @liyinan926 . I looked at this briefly and from https://github.com/kubernetes/apiserver/blob/master/pkg/storage/cacher/cacher.go#L614 , it looks like it the resource version is not specified while doing a get/list, api-server fetches it from the etc.d which should be the case here. 

Since its been a while,  can we do a  quick sync-up next week to catch-up on the overall state/roadmap for the operator as well as close out this/other changes we were hoping to do in the operator.  I was thinking if we can do this next Tuesday/Wednesday. Let me know what works for you.

liyinan926:
Next Wed 1:30-4:30pm works for me.

pkg/controller/sparkapplication/controller.go
liyinan926: In this case, how about using the driver pod name in the spec, and if it's not set, use the recorded driver pod name in status?

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
 	return pods, nil
 }
 
+// Fetch driver pod from APIServer.
+func (c *Controller) getDriverPodFromApiServer(app *v1beta1.SparkApplication) (*apiv1.Pod, error) {
+	selector := labels.SelectorFromSet(map[string]string{config.SparkAppNameLabel: app.Name, config.SparkRoleLabel: config.SparkDriverRole})
+	listOptions := metav1.ListOptions{LabelSelector: selector.String()}
+	podsList, err := c.kubeClient.CoreV1().Pods(app.Namespace).List(listOptions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get driver pod for SparkApplication from APIServer %s/%s: %v", app.Namespace, app.Name, err)
```
liyinan926: Suggesting changing the error message to `"failed to get driver pod for SparkApplication %s%s from the API server"`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
-					currentDriverState.err = fmt.Errorf("driver container status missing.")
-				}
-			}
+			driverPod = pod
 			break
 		}
 	}
+	if driverPod == nil {
+		// Driver Pod was not found in Informer cache. Try to fetch it from APIServer.
+		glog.Infof("driver not found for SparkApplication: %s/%s. Trying to fetch driver from APIServer.", app.Namespace, app.Name)
```
liyinan926: Suggesting changing the error message to "driver pod not found for SparkApplication %s/%s, trying to fetch it from the API server.".

pkg/controller/sparkapplication/controller.go
liyinan926: This is still an issue even if you try fetching the driver pod from the API server. The driver pod may have not been created yet.

pkg/controller/sparkapplication/controller.go
liyinan926: Agreed.

pkg/controller/sparkapplication/controller.go
```diff
@@ -739,7 +736,6 @@ func (c *Controller) deleteSparkResources(app *v1beta1.SparkApplication) error {
 	if err != nil {
 		return err
 	}
-
```
liyinan926: Please keep the blank line.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -865,9 +865,11 @@ func TestSyncSparkApplication_ExecutingState(t *testing.T) {
 			appName:               "foo-1",
 			oldAppStatus:          v1beta1.SubmittedState,
 			oldExecutorStatus:     map[string]v1beta1.ExecutorState{"exec-1": v1beta1.ExecutorRunningState},
-			expectedAppState:      v1beta1.SubmittedState,
```
liyinan926: Why this change?

pkg/controller/sparkapplication/controller.go
akhurana001: If a user updates the spec to update the driver pod name, we will miss terminating the old driver pod (because the name has now changed). I think this is a slight edge-case and just want to make sure we are aware of this and ok with this.

liyinan926:
I put more thoughts into this issue. I don't think the changes particularly the approach of trying to get the driver pod from the API server really solves the problem you mentioned. If the driver pod somehow gets deleted before it even gets noticed by the controller, it won't be be on the API server either. I do think the right approach should be to set a timeout time for the driver to show up and start running after `spark-submit` is executed, and if the timeout happens, fail the application. I also don't think we should allow updating the driver pod name, which creates a moving target that makes things more than necessarily complicated with little benefit.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
+		// Fetch container ExitCode/Reason if Pod Failed.
+		if driverPod.Status.Phase == apiv1.PodFailed {
+			if len(driverPod.Status.ContainerStatuses) > 0 && driverPod.Status.ContainerStatuses[0].State.Terminated != nil {
+				terminatedState := driverPod.Status.ContainerStatuses[0].State.Terminated
+				app.Status.AppState.ErrorMessage = fmt.Sprintf("driver pod failed with ExitCode: %d, Reason: %s", terminatedState.ExitCode, terminatedState.Reason)
+			} else {
+				app.Status.AppState.ErrorMessage = fmt.Sprintf("driver container status missing.")
+			}
 		}
 		app.Status.AppState.State = newState
```
liyinan926: This could be moved up to after line 314.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
-	if currentDriverState != nil {
-		newState := driverPodPhaseToApplicationState(currentDriverState.podPhase)
+	if driverPod != nil {
+		newState := driverPodPhaseToApplicationState(driverPod.Status.Phase)
 		// Only record a driver event if the application state (derived from the driver pod phase) has changed.
 		if newState != app.Status.AppState.State {
-			c.recordDriverEvent(app, currentDriverState.podPhase, currentDriverState.podName)
+			c.recordDriverEvent(app, driverPod.Status.Phase, driverPod.Name)
 		}
 		if newState != v1beta1.UnknownState {
```
liyinan926: I don't think we really need this check.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
-		// Expose any errors to the users.
-		if currentDriverState.err != nil {
-			app.Status.AppState.ErrorMessage = currentDriverState.err.Error()
+		// Fetch container ExitCode/Reason if Pod Failed.
+		if driverPod.Status.Phase == apiv1.PodFailed {
+			if len(driverPod.Status.ContainerStatuses) > 0 && driverPod.Status.ContainerStatuses[0].State.Terminated != nil {
+				terminatedState := driverPod.Status.ContainerStatuses[0].State.Terminated
+				app.Status.AppState.ErrorMessage = fmt.Sprintf("driver pod failed with ExitCode: %d, Reason: %s", terminatedState.ExitCode, terminatedState.Reason)
+			} else {
+				app.Status.AppState.ErrorMessage = fmt.Sprintf("driver container status missing.")
```
liyinan926: You don't need `fmt.Sprintf` here.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
+			app.Status.DriverInfo.PodName = driverPod.Name
+			app.Status.SparkApplicationID = getSparkApplicationID(driverPod)
+			if driverPod.Spec.NodeName != "" {
+				if nodeIP := c.getNodeIP(driverPod.Spec.NodeName); nodeIP != "" {
 					app.Status.DriverInfo.WebUIAddress = fmt.Sprintf("%s:%d", nodeIP, app.Status.DriverInfo.WebUIPort)
 				}
 			}
-			if app.Status.TerminationTime.IsZero() && !currentDriverState.completionTime.IsZero() {
-				app.Status.TerminationTime = currentDriverState.completionTime
+			if app.Status.TerminationTime.IsZero() && driverPod.Status.Phase == apiv1.PodSucceeded || driverPod.Status.Phase == apiv1.PodFailed {
```
liyinan926: There should be parenthesis around `driverPod.Status.Phase == apiv1.PodSucceeded || driverPod.Status.Phase == apiv1.PodFailed`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,91 +259,88 @@ func (c *Controller) processNextItem() bool {
...
+				app.Status.TerminationTime = metav1.Now()
 			}
 		}
-		// Expose any errors to the users.
-		if currentDriverState.err != nil {
-			app.Status.AppState.ErrorMessage = currentDriverState.err.Error()
+		// Fetch container ExitCode/Reason if Pod Failed.
+		if driverPod.Status.Phase == apiv1.PodFailed {
+			if len(driverPod.Status.ContainerStatuses) > 0 && driverPod.Status.ContainerStatuses[0].State.Terminated != nil {
```
liyinan926: Let's remove `driverPod.Status.ContainerStatuses[0].State.Terminated != nil` from this if check and do the check after line 331 as:
```
if terminatedState != nil {...}
```

PR #496, opened 2019-05-16
`spark-ui-service-annotations` flag to add annotations to SparkUI service to support ambassador

eterna2:
### Background
My k8s cluster uses ambassador as the primary API gateway. Ambassador uses annotations in service instead of of an ingress.

Currently, `spark-on-k8s-operator` only support ingress but not service annotations. This PR adds a `spark-ui-service-annotations` flag which takes in a JSON string which will be used to add annotations to the sparkUI service metadata.

The annotations JSON also supports the following mustache variables:
- `{{$appName}}`
- `{{$serviceName}}`
- `{{$namespace}}`

Also added corresponding test to `sparkui_test.go`.

**Example**
```js
{
  "getambassador.io/config": `|
---
apiVersion: ambassador/v1
kind:  Mapping
name:  sparkui-{{$namespace}}-{{$appName}}
prefix: /{{$appName}}/
service: {{$serviceName}}.{{$namespace}}`
}
```


pkg/controller/sparkapplication/sparkui.go
```diff
@@ -39,16 +39,41 @@ const (
...
 var ingressURLRegex = regexp.MustCompile("{{\\s*[$]appName\\s*}}")
+var sparkUIAnnotationsRegex = regexp.MustCompile("({{\\s*[$]namespace\\s*}}|{{\\s*[$]appName\\s*}}|{{\\s*[$]serviceName\\s*}})")
 
 func getSparkUIingressURL(ingressURLFormat string, appName string) string {
 	return ingressURLRegex.ReplaceAllString(ingressURLFormat, appName)
 }
 
+// getSparkUIServiceAnnotations replace any `{{$appName}}` inside the annotations with the app name.
+func getSparkUIServiceAnnotations(templateAnnotations map[string]string, appName string, serviceName string, namespace string) map[string]string {
+	var annotations = map[string]string{}
```
liyinan926: This can be simplified as `annotations := map[string]string{}`.

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -99,19 +124,29 @@ func createSparkUIIngress(app *v1beta1.SparkApplication, service SparkService, i
 
 func createSparkUIService(
 	app *v1beta1.SparkApplication,
-	kubeClient clientset.Interface) (*SparkService, error) {
+	kubeClient clientset.Interface,
+	serviceMetadata SparkService) (*SparkService, error) {
```
liyinan926: Can this just be the annotations of type `map[string]string` instead?

main.go
```diff
@@ -50,23 +50,24 @@ import (
...
+	webhookCertDir               = flag.String("webhook-cert-dir", "/etc/webhook-certs", "The directory where x509 certificate and key files are stored.")
+	webhookSvcNamespace          = flag.String("webhook-svc-namespace", "spark-operator", "The namespace of the Service for the webhook server.")
+	webhookSvcName               = flag.String("webhook-svc-name", "spark-webhook", "The name of the Service for the webhook server.")
+	webhookPort                  = flag.Int("webhook-port", 8080, "Service port of the webhook server.")
+	enableMetrics                = flag.Bool("enable-metrics", false, "Whether to enable the metrics endpoint.")
+	metricsPort                  = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
+	metricsEndpoint              = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
+	metricsPrefix                = flag.String("metrics-prefix", "", "Prefix for the metrics.")
+	ingressUrlFormat             = flag.String("ingress-url-format", "", "Ingress URL format.")
+	sparkUIServiceAnnotationsStr = flag.String("spark-ui-service-annotations", "{}", "Annotations (as JSON string) for sparkui service.")
```
liyinan926: Is it possible that different applications will need different annotations, even in the same cluster?

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -153,6 +166,40 @@ func TestCreateSparkUIService(t *testing.T) {
 			},
 			expectError: false,
 		},
+		{
+			name: "service with annotations",
+			app:  app1,
+			annotations: map[string]string{
+				"getambassador.io/config": `|
+---
```
liyinan926: Can we define such raw string as a constant at the top to make the indention correct?

main.go
eterna2: Yes and no. For the spark job use case, I can only think of 3 types of annotations I might need:

- pod annotations (driver/executor pods) for kube2iam, kiam, etc (credentials for access control) (already provided in the CRD)

- service annotations (sparkUI) for ambassador which need it to create the envoy proxy (this PR)

- ingress annotations (sparkUI) for other custom ingress controllers (e.g. nginx, kong, alb) also relies on these annotations to create the proxy/entry point. However, u probably use either ambassador or a custom ingress controller (they both serve the same purpose)

PS: This probably means I should also add annotation for the ingress to support pple who use custom ingress controllers.

eterna2: Just reread ur question and realised that u are asking whether annotation should be specified at the CRD (aka diff spark jobs need diff annotations).

I don't think the annotations need to change if it is templated (diff route for diff spark job). However, if we do it at the CRD, then we probably don't need templating. 

But yeah, now that u mentioned it. It might be easier to write it at the CRD. json string as arg is abit clunky. And might not be obvious to the end user, as they might not be the one who setup the k8s spark operator. 

But on the other hand, I might need to teach my data scientists how ambassador works, which may not be desirable, and probably also make the yaml less pure (the annotation is purely a cluster specific setup).

liyinan926: Makes sense.

main.go
```diff
@@ -122,10 +123,15 @@ func main() {
 		}
 	}
 
+	sparkUIServiceAnnotations, err := util.GetAnnotationsFromJSONString(*sparkUIServiceAnnotationsStr)
+	if err != nil {
+		glog.Warningf("failed to unmarshal json string to annotations: %v", err)
+		sparkUIServiceAnnotations = map[string]string{}
+	}
```
liyinan926: Add an empty line after this.

eterna2:
> LGTM. Can you also update the user guide on how to use this?

@liyinan926  Updated user guide and added the newline.

docs/quick-start-guide.md
```diff
@@ -197,10 +197,22 @@ A note about `metrics-labels`: In `Prometheus`, every unique combination of key-
...
-The operator also supports creating an Ingress for the UI. This can be turned on by setting the `ingress-url-format` command-line flag. The `ingress-url-format` should be a template like `{{$appName}}.ingress.cluster.com` and the operator will replace the `{{$appName}}` with the appropriate appName.
+The operator, by default, makes the Spark UI accessible by creating a service of type `NodePort` which exposes the UI via the node running the driver. 
+
+If you are using API gateway that depends on service annotations to expose services (e.g. [Ambassador](https://www.getambassador.io/)), you can configure a templated service annotations by setting the `spark-ui-service-annotations` command-line flag. The `spark-ui-service-annotations` should be a JSON string and supports the following mustache variables:
+- `{{$appName}}`
+- `{{$serviceName}}`
+- `{{$namespace}}`
+
+The following example demonstrate how to config a route to the spark UI pod based on the `appName` with ambassador:
+```bash
```
liyinan926: Can we show the example in a different form that uses part of the deployment yaml?

liyinan926:
It also looks like you need to run `go fmt ./...`.

eterna2:
@liyinan926 

After writing out the deployment example, I find that it seems a bit clunky passing in the annotations as JSON. Works fine for simple key values, but ambassador annotations are yaml. 

Just considering whether are there better ways to do this?

*What do u think of this?*

Write the annotations directly in the sparkoperator pod, and use the arg flag to tell what extract these annotations thru a prefix?
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sparkoperator
spec:
  template:
    spec:
      metadata:
        annotations:
          sparkui-getambassador.io/config: |-
            ---
            apiVersion: ambassador/v1
            kind:  Mapping
            name:  sparkui-{{$namespace}}-{{$appName}}
            prefix: /{{$appName}}/
            service: {{$serviceName}}.{{$namespace}}
      containers:
      - name: sparkoperator
        image: gcr.io/spark-operator/spark-operator:v2.4.0-v1beta1-latest
        imagePullPolicy: Always
        args:
        - -spark-ui-service-annotations-prefix 'sparkui-'
```

liyinan926:
@eterna2 I agree this is a bit clunky. I think we need a thorough design around how the UI service/ingress should be configured and how users define the specifics of the configuration. I'm thinking of this high level approach:

1. Define a CRD named `SparkApplicationClass` that defines parameters and common configuration for a class of applications. Service related configuration can be put into CRs of this new kind.
2. Add an optional field in `SparkApplication` for specifying the name of the `SparkApplicationClass`. 
3. At runtime, the controller fetches the `SparkApplicationClass` (if applicable) and use the information to configure the driver/executor pods and UI service.

eterna2:
> @eterna2 I agree this is a bit clunky. I think we need a thorough design around how the UI service/ingress should be configured and how users define the specifics of the configuration. I'm thinking of this high level approach:
> 
> 1. Define a CRD named `SparkApplicationClass` that defines parameters and common configuration for a class of applications. Service related configuration can be put into CRs of this new kind.
> 2. Add an optional field in `SparkApplication` for specifying the name of the `SparkApplicationClass`.
> 3. At runtime, the controller fetches the `SparkApplicationClass` (if applicable) and use the information to configure the driver/executor pods and UI service.

Yes, I think this is a better idea. Do u need any help?

liyinan926:
@eterna2 Yes, definitely need help on this :).

eterna2:
@liyinan926

Great, how can I start to help?

I assume the new CRD should probably be a subset of SparkApplication + some extras.

The minimal probably should have just the Spark and Hadoop configs + default extra jars (e.g. hadoop-aws sdk) + service/ingress annotations?

liyinan926:
> Great, how can I start to help?

Thanks for being willing to help! How about a google doc that drafts the CRD spec?

> I assume the new CRD should probably be a subset of SparkApplication + some extras.

Pretty much. Basically things that might be shared by a class of applications.

> The minimal probably should have just the Spark and Hadoop configs + default extra jars (e.g. hadoop-aws sdk) + service/ingress annotations?

That sounds reasonable. Probably also default service account for the driver pod.

eterna2:
@liyinan926 

Sorry for taking so long. Was busy with some other stuff.

Here is my draft proposal:
- Use `SparkApplicationSpec` as template in `SparkApplicationClass`
- Adds a `SparkUI` field (`SparkUISpec`) to `SparkApplicationSpec` :
  - `Host`: String
  - `Service`: ObjectMeta
  - `Ingress`: ObjectMeta

https://docs.google.com/document/d/1G9Kvi3ptf4mobrK2I5rPB_uAvFOymOLJsobuuPvCkS8/edit?usp=sharing

liyinan926:
@eterna2 Thanks for putting together the draft! Instead of using `SparkApplicationSpec` as template in `SparkApplicationClass`, I would suggest adding an optional field in `SparkApplicationSpec` to reference a `SparkApplicationClass` that defines common and default configurations for a class of applications. Initially `SparkApplicationClass` contains a `SparkUISpec` or `SparkUIConfig`.

Another thing is instead of using `ObjectMetadata` for `Service` and `Ingress`, we should use a struct that contains the necessary fields, e.g., a `map[string]string` for annotations.

eterna2:
@liyinan926
Hey, I think u have a slight misinterpretation on the design. The `SparkApplicationSpec` for `SparkApplicationClass` is literally the default template for any spark application, i.e. it serves as the defaults for any unspecified fields in  `SparkApplication` and `SchedulesSparkApplication`

`SparkApplication` and `SchedulesSparkApplication` selects the `SparkApplicationClass`  with an annotation.

I originally had 3 ideas:

1. Similar to what u recommended, a basic `SparkApplicationClass` with service and ingress annotations, and some common fields. And a additional field to select the class.

2. A slight modification from 1. A `SparkApplicationClass` with `SparkApplicationSpec` (which essentially holds all the fields) and separate blocks for service and ingress annotations. And use annotations to select the class.

Advantage of annotations via selector field is that there need no change to the `SparkApplication` CRD spec. 

Advantage of `SparkApplicationSpec` instead of manually specifying the "common" fields is that I just need to implement a recursive deep merge instead of a custom map to seed the `SparkApplicationSpec`. This also allows the class to always move along the the application spec, and allow cluster admins more flexibility to define more classes of spark application, i.e. py vs java vs scala, or large jobs vs small jobs, metrics or non-metrics, etc.

3. Is the current proposal where I merge the ingress and service metadata into the `SparkApplicationSpec`. Mostly because I feel that the blocks feels strange to have driver, executor, metrics spec in 1 block, while service and ingress in another.


+1 on `map[string]string` vs `ObjectMetadata`
Just think should I name is service annotation to be explicit.

liyinan926:
> Hey, I think u have a slight misinterpretation on the design. The `SparkApplicationSpec` for `SparkApplicationClass` is literally the default template for any spark application, i.e. it serves as the defaults for any unspecified fields in `SparkApplication` and `SchedulesSparkApplication`.

Makes sense to me. This is sort of a variant of the class/instance model. 

> Advantage of annotations via selector field is that there need no change to the SparkApplication CRD spec.

An optional field is a better choice than annotations for a few reasons: 1) a field is explicit and json schema validation can be applied to a field, 2) deprecation management of fields is much easier with fields than with annotations, and 3) an optional field is backward compatible.

eterna2:
@liyinan926 
I have updated the docs to reflect ur comments:
https://docs.google.com/document/d/1G9Kvi3ptf4mobrK2I5rPB_uAvFOymOLJsobuuPvCkS8/edit?usp=sharing

Removed the annotations, and added a 2 new optional fields to `SparkApplicationSpec`
- `SparkApplicationClass` - name of the `SparkApplicationClass` to inherit
- `SparkUI` - optional configuration for host (aka `ingress_format`), service annotations, and ingress annotations

skonto:
@eterna2 should we support multiple inheritance?

liyinan926:
@eterna2 thanks for updating the doc! It looks good overall. 
@skonto it's technically possible to support multi-class inheritance. However, it's hard to deal with conflicts in different classes. I suggest we start with single-class inheritance.

eterna2:
Great. I will start working on it.

GrigorievNick:
Any updates?

eterna2:
Sorry for the slow progress. Was bogged down with other work.

I will close this PR and track the progress of this CRD in a separate PR #661 

I will find some time this week to rebase the branch, and marked out the parts that I have not completed, so that others can contribute if they want.

PR #504, opened 2019-05-31
Added support for node selector at SparkPod level

jlpedrosa:
Hi!

The following PR enables to have node selector filters at Pod level https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector

This is useful when you want your executors in a different type of nodes than the driver.

### DIFF
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -218,6 +218,8 @@ type SparkApplicationSpec struct {
 	// RestartPolicy defines the policy on if and in which conditions the controller should restart an application.
 	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
 	// NodeSelector is the Kubernetes node selector to be added to the driver and executor pods.
+	// This field is mutually exclusive with nodeSelector at podSpec level (driver or executor).
+	// This field will be deprecated in future versions (at SparkApplicationSpec level).
 	// Optional.
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 	// FailureRetries is the number of times to retry a failed application before giving up.
@@ -391,6 +393,10 @@ type SparkPodSpec struct {
 	// HostNetwork indicates whether to request host networking for the pod or not.
 	// Optional.
 	HostNetwork *bool `json:"hostNetwork,omitempty"`
+	// NodeSelector is the Kubernetes node selector to be added to the driver and executor pods.
+	// This field is mutually exclusive with nodeSelector at SparkApplication level (which will be deprecated).
+	// Optional.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 }
 
 // DriverSpec is specification of the driver.
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -501,7 +501,12 @@ func (c *Controller) syncSparkApplication(key string) error {
 	switch appToUpdate.Status.AppState.State {
 	case v1beta1.NewState:
 		c.recordSparkApplicationEvent(appToUpdate)
-		appToUpdate = c.submitSparkApplication(appToUpdate)
+		if err := c.validateSparkApplication(appToUpdate); err != nil {
+			appToUpdate.Status.AppState.State = v1beta1.FailedState
+			appToUpdate.Status.AppState.ErrorMessage = err.Error()
+		} else {
+			appToUpdate = c.submitSparkApplication(appToUpdate)
+		}
 	case v1beta1.SucceedingState:
 		if !shouldRetry(appToUpdate) {
 			// Application is not subject to retry. Move to terminal CompletedState.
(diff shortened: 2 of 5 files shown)


pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -385,6 +385,9 @@ type SparkPodSpec struct {
 	// Sidecars is a list of sidecar containers that run along side the main Spark container.
 	// Optional.
 	Sidecars []apiv1.Container `json:"sidecars,omitempty"`
+	// NodeSelector is the Kubernetes node selector to be added to the driver and executor pods.
+	// Optional.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
liyinan926: The top-level `SparkApplicationSpec` also has a `NodeSelector` field that applies to both driver and executor pods as the Kubernetes backend doesn't support separate node selector config for driver and executors. This new field should be mutually exclusive to the top-level one. Let's make this clear in the comment both here and in place for the top-level field. And let's add a deprecation comment for the top-level field.

pkg/webhook/patch.go
```diff
@@ -295,6 +296,21 @@ func addTolerations(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOpera
 	return ops
 }
 
+func addNodeSelectors(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOperation {
+	var tolerations map[string]string
```
liyinan926: s/`tolerations`/`nodeSelector`/.

jlpedrosa: Fixed.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
jlpedrosa: The idea that I had in mind it that was additive (if any at Pod level, would be added to what was declared at app spark level). I've pushed the changes in the doc, and also a validator at the controller level. Doing a new transition to failed if the validation is not passed (was not very sure about this part, please double check).

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -218,6 +218,8 @@ type SparkApplicationSpec struct {
 	// RestartPolicy defines the policy on if and in which conditions the controller should restart an application.
 	RestartPolicy RestartPolicy `json:"restartPolicy,omitempty"`
 	// NodeSelector is the Kubernetes node selector to be added to the driver and executor pods.
+	// This field is mutually exclusive with nodeSelector at podSpect level (driver or executor).
```
liyinan926: Typo `podSpect`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -770,6 +775,18 @@ func (c *Controller) deleteSparkResources(app *v1beta1.SparkApplication) error {
 	return nil
 }
 
+func (c *Controller) validateSparkApplication(app *v1beta1.SparkApplication) error {
+
```
liyinan926: Nit: unnecessary empty line.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -321,6 +321,102 @@ func TestSyncSparkApplication_SubmissionFailed(t *testing.T) {
 	assert.Equal(t, int32(2), updatedApp.Status.SubmissionAttempts)
 }
 
+func TestValidateDetectsNodeSelectorSuccessNoSelector(t *testing.T) {
+
```
liyinan926: Nit: unnecessary empty line.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -321,6 +321,102 @@ func TestSyncSparkApplication_SubmissionFailed(t *testing.T) {
...
+			Namespace: "default",
+		},
+	}
+
+	err := ctrl.validateSparkApplication(app)
+	assert.Nil(t, err)
+}
+
+func TestValidateDetectsNodeSelectorSuccessNodeSelectorAtAppLevel(t *testing.T) {
+
```
liyinan926: Ditto.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -321,6 +321,102 @@ func TestSyncSparkApplication_SubmissionFailed(t *testing.T) {
...
+			NodeSelector: map[string]string{"mynode": "mygift"},
+		},
+	}
+
+	err := ctrl.validateSparkApplication(app)
+	assert.Nil(t, err)
+}
+
+func TestValidateDetectsNodeSelectorSuccessNodeSelectorAtPodLevel(t *testing.T) {
+
```
liyinan926: Ditto.

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -321,6 +321,102 @@ func TestSyncSparkApplication_SubmissionFailed(t *testing.T) {
...
+			NodeSelector: map[string]string{"mynode": "mygift"},
+		},
+	}
+
+	err = ctrl.validateSparkApplication(app)
+	assert.Nil(t, err)
+}
+
+func TestValidateDetectsNodeSelectorFailsAppAndPodLevel(t *testing.T) {
+
```
liyinan926: Ditto.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
jlpedrosa: Fixed

pkg/controller/sparkapplication/controller.go
jlpedrosa: fixed

pkg/controller/sparkapplication/controller_test.go
jlpedrosa: fixed

pkg/controller/sparkapplication/controller_test.go
jlpedrosa: fixed

pkg/controller/sparkapplication/controller_test.go
jlpedrosa: fixes

pkg/controller/sparkapplication/controller_test.go
jlpedrosa: fixed

PR #506, opened 2019-05-31
Record driver pod name during submission

liyinan926:
and try getting the driver pod from the API server when it's not found in the informer cache.

@akhurana001 this should solve the issue addressed in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/495.

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -280,15 +280,16 @@ func (c *Controller) getDriverPod(app *v1beta1.SparkApplication) (*apiv1.Pod, er
 		return nil, fmt.Errorf("failed to get driver pod %s: %v", app.Status.DriverInfo.PodName, err)
 	}
 
-	// The driver pod is not in the informer cache, try getting it directly from the API server.
+	// The driver pod was not found in the informer cache, try getting it directly from the API server.
 	pod, err = c.kubeClient.CoreV1().Pods(app.Namespace).Get(app.Status.DriverInfo.PodName, metav1.GetOptions{})
 	if err == nil {
 		return pod, nil
 	}
-	if errors.IsNotFound(err) {
-		return nil, nil
+	if !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get driver pod %s: %v", app.Status.DriverInfo.PodName, err)
 	}
-	return nil, fmt.Errorf("failed to get driver pod %s: %v", app.Status.DriverInfo.PodName, err)
+	// Driver pod was not found on the API server either.
+	return nil, nil
 }
 
 // getAndUpdateDriverState finds the driver pod of the application
@@ -296,7 +297,7 @@ func (c *Controller) getDriverPod(app *v1beta1.SparkApplication) (*apiv1.Pod, er
 func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) error {
 	// Either the driver pod doesn't exist yet or its name has not been updated.
 	if app.Status.DriverInfo.PodName == "" {
-		return nil
+		return fmt.Errorf("empty driver pod name with application state %s", app.Status.AppState.State)
 	}
 
 	driverPod, err := c.getDriverPod(app)
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -45,6 +45,7 @@ import (
 	crdscheme "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/clientset/versioned/scheme"
 	crdinformers "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/informers/externalversions"
 	crdlisters "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/client/listers/sparkoperator.k8s.io/v1beta1"
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/config"
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/util"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -258,102 +259,99 @@ func (c *Controller) processNextItem() bool {
...
+	}
+	return nil, fmt.Errorf("failed to get driver pod %s: %v", app.Status.DriverInfo.PodName, err)
+}
+
 // getAndUpdateDriverState finds the driver pod of the application
 // and updates the driver state based on the current phase of the pod.
 func (c *Controller) getAndUpdateDriverState(app *v1beta1.SparkApplication) error {
-	pods, err := c.getAppPods(app)
+	// Either the driver pod doesn't exist yet or its name has not been updated.
+	if app.Status.DriverInfo.PodName == "" {
```
akhurana001: **minor**: this should ideally not happen as getAndUpdateDriverState should be called only for a submitted/running app. Let's atleast log an error here.

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,102 +259,99 @@ func (c *Controller) processNextItem() bool {
...
+	selector := labels.SelectorFromSet(labels.Set(matchLabels))
 	pods, err := c.podLister.Pods(app.Namespace).List(selector)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get pods for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 	}
 	return pods, nil
 }
 
+func (c *Controller) getDriverPod(app *v1beta1.SparkApplication) (*apiv1.Pod, error) {
+	pod, err := c.podLister.Pods(app.Namespace).Get(app.Status.DriverInfo.PodName)
```
akhurana001: Like this doing a get instead of labels for the driver pod name. 👍

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -38,7 +38,11 @@ func getSparkApplicationID(pod *apiv1.Pod) string {
 	return pod.Labels[config.SparkApplicationSelectorLabel]
 }
 
-func getDefaultDriverPodName(app *v1beta1.SparkApplication) string {
+func getDriverPodName(app *v1beta1.SparkApplication) string {
```
akhurana001: Since driver pod name can be set in sparkConf as well, we probably should check that as well here before defaulting. I came up with something like https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/495/files#diff-d110a4ca56d08f6d4992d143e985ee66R41

pkg/controller/sparkapplication/submission.go
```diff
@@ -96,7 +96,7 @@ func buildSubmissionCommandArgs(app *v1beta1.SparkApplication, submissionID stri
 	args = append(args, "--deploy-mode", string(app.Spec.Mode))
 	args = append(args, "--conf", fmt.Sprintf("spark.kubernetes.namespace=%s", app.Namespace))
 	args = append(args, "--conf", fmt.Sprintf("spark.app.name=%s", app.Name))
-	args = append(args, "--conf", fmt.Sprintf("spark.kubernetes.driver.pod.name=%s", getDefaultDriverPodName(app)))
+	args = append(args, "--conf", fmt.Sprintf("spark.kubernetes.driver.pod.name=%s", driverPodName))
```
akhurana001: we need to remove driver pod name conf explicitly while adding the sparkConfig  to the args as well as in the `addDriverConfOptions` method. I had something like  https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/495/files#diff-8c1e9e387266f24561c52caf907d1793 in mind

pkg/controller/sparkapplication/controller.go
```diff
@@ -258,102 +259,99 @@ func (c *Controller) processNextItem() bool {
...
+	if !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get driver pod %s: %v", app.Status.DriverInfo.PodName, err)
+	}
+
+	// The driver pod is not in the informer cache, try getting it directly from the API server.
+	pod, err = c.kubeClient.CoreV1().Pods(app.Namespace).Get(app.Status.DriverInfo.PodName, metav1.GetOptions{})
+	if err == nil {
+		return pod, nil
+	}
+	if errors.IsNotFound(err) {
```
akhurana001: **minor**: maybe it looks a little cleaner/easier to read if we handle `!errors.IsNotFound(err) ` first similar to how we do it for fetching from informer cache case above.

akhurana001:
> and try getting the driver pod from the API server when it's not found in the informer cache.
> 
> @akhurana001 this should solve the issue addressed in #495.

Thanks for this. I was reworking #495 based on our discussion but we can work on this here and I will close out #495 .

pkg/controller/sparkapplication/submission.go
akhurana001: Also lets use `config.SparkDriverPodNameKey` for the key.

pkg/controller/sparkapplication/sparkapp_util.go
liyinan926: Done.

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/submission.go
liyinan926: Done.

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/submission.go
akhurana001: Please also remove the same from `addDriverConfOptions`  method. lines 261-264 in this file.

akhurana001:
One minor comment. Should be good to go after that 👍

pkg/controller/sparkapplication/submission.go
liyinan926: Done

PR #509, opened 2019-06-05
Add support for requesting GPUs

tkanng:
Hi, this PR adds support for requesting nvidia GPUs, according to https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/426. 

It might be useful for ML programs. And I'm not sure it's good enough, so please help me check it again. 

Thanks!


tkanng:
I signed it!

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -346,6 +346,9 @@ type SparkPodSpec struct {
 	// MemoryOverhead is the amount of off-heap memory to allocate in cluster mode, in MiB unless otherwise specified.
 	// Optional.
 	MemoryOverhead *string `json:"memoryOverhead,omitempty"`
+	// GPU is the number of nvidia.com/gpu to request for the pod
+	// Optional.
+	GPU *int64 `json:"gpu,omitempty"`
```
liyinan926: We can make a struct type that looks like the following to support all GPU types from all vendors:
```go
type GPUSpec struct {
    Name string `json:"name"`
    Quantity int64 `json:"quantity"`
}
```

Then this field becomes:
```go
GPU *GPUSpec `json:"gpu,omitempty"`
```

pkg/webhook/patch_test.go
```diff
@@ -676,6 +678,152 @@ func TestPatchSparkPod_Sidecars(t *testing.T) {
...
+			},
+			Executor: v1beta1.ExecutorSpec{
+				SparkPodSpec: v1beta1.SparkPodSpec{
+					GPU: &gpu,
+				},
+			},
+		},
+	}
+	tests := []testcase{
+		{nil,
```
liyinan926: `nil` should be on a new line.

tkanng:
Thank you for your patience! The latest commit adds `GPUSpec` type and refines corresponding unit tests. Users can specify `gpu` field, just like this:
```yaml
  executor:
   # cores: 1
    instances: 1
   # memory: "512m"
    gpu: 
      name: example.com/gpu 
      quantity: 1
```

pkg/webhook/patch.go
```diff
@@ -18,6 +18,8 @@ package webhook
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
```
liyinan926: Nit: `k8s.io` imports should go immediately under built-in ones.

pkg/webhook/patch_test.go
```diff
@@ -676,6 +678,199 @@ func TestPatchSparkPod_Sidecars(t *testing.T) {
 	assert.Equal(t, "sidecar2", modifiedExecutorPod.Spec.Containers[2].Name)
 }
 
+func TestPatchSparkPod_GPU(t *testing.T) {
+
```
liyinan926: Nit: empty line can be removed.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -499,6 +502,13 @@ type PrometheusSpec struct {
 	Configuration *string `json:"configuration,omitempty"`
 }
 
+type GPUSpec struct {
+	// Name is GPU resource name, such as: nvidia.com/gpu or amd.com/gpu
+	Name string `json:"name"`
+	// Quantity is the number of GPU to request for driver or executor.
```
liyinan926: `number of GPUs`.

tkanng:
Hi, the latest commit added user guide on how to use gpus and refined code format.     :)

gyj0825:
How to add multiple GPU parameters to support bitfusion, like
limits:
   bitfusion.io/gpu-amount: 2
   bitfusion.io/gpu-percent: 50
thanks.

PR #515, opened 2019-06-12
Support live webhook cert reloading and configurable cert location

khogeland:


main.go
```diff
@@ -143,7 +146,7 @@ func main() {
 	var hook *webhook.WebHook
 	if *enableWebhook {
 		var err error
-		hook, err = webhook.New(kubeClient, crInformerFactory, *webhookCertDir, *webhookSvcNamespace, *webhookSvcName, *webhookPort, *namespace)
+		hook, err = webhook.New(kubeClient, crInformerFactory, *webhookServerCertFile, *webhookServerKeyFile, *webhookCACertFile, *webhookCertReloadInterval, *webhookSvcNamespace, *webhookSvcName, *webhookPort, *namespace)
```
liyinan926: Given that we have quite a number of webhook-related config options, I think it makes sense to wrap them all into a struct for example named `WebhookConfig` that has a method to get an instance by parsing flags, e.g., `FromFlags`. And then pass the `WebhookConfig` instance into `New`.

pkg/webhook/certs.go
```diff
@@ -18,25 +18,74 @@ package webhook
 
 import (
 	"crypto/tls"
+	"github.com/golang/glog"
```
liyinan926: This import should go below built-in ones.

pkg/webhook/certs.go
```diff
@@ -18,25 +18,74 @@ package webhook
...
+// certProvider is a container of a X509 certificate file and a corresponding key file for the
 // webhook server, and a CA certificate file for the API server to verify the server certificate.
-type certBundle struct {
+type certProvider struct {
 	serverCertFile string
 	serverKeyFile  string
 	caCertFile     string
+	reloadInterval time.Duration
+	ticker         *time.Ticker
+	stopChannel    *chan interface{}
```
liyinan926: `chan` is already a reference type, so this doesn't need to be a pointer.

pkg/webhook/certs.go
```diff
@@ -18,25 +18,74 @@ package webhook
...
+func (c *certProvider) Stop() {
+	close(*c.stopChannel)
+	c.ticker.Stop()
+}
 
-	return &tls.Config{Certificates: []tls.Certificate{cert}}, nil
+func (c *certProvider) updateCert() {
+	cert, err := tls.LoadX509KeyPair(c.serverCertFile, c.serverKeyFile)
+	if err != nil {
+		glog.Error("Could not reload certificate: ", err)
```
liyinan926: `glog.Errorf("cloud not load certificate %s: %v", c.serverCertFile, err)`.

pkg/webhook/certs.go
```diff
@@ -18,25 +18,74 @@ package webhook
...
+		return
+	}
+	c.currentCert = &cert
```
liyinan926: This only updates `currentCert`. What about making the new certificate effective? I think two things need to be done: 1) update the `MutatingWebhookConfigurations ` with the new cert, and 2) restart the web server to load the new cert.

main.go
khogeland: Agreed. I have another set of webhook enhancements that I think I'm going to roll into the same PR, so that sounds like the best option.

pkg/webhook/certs.go
khogeland: `GetCertificate` in `tls.Config` is a callback used in the TLS handshake step to get the current certificate. Our implementation always returns `currentCert`, so this is all that's needed to make the new cert effective. See:

https://golang.org/pkg/crypto/tls/#Config
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/9c1e54b23f12565e8d10704b4991695951dc1ed4/pkg/webhook/certs.go#L71-L73

liyinan926: Ah, got it. So this handshake also registers the new cert with the API server?

khogeland: The API server doesn't need to be notified of new server certs as the cert chain should roll up to the configured CA cert.

liyinan926: I see. Is this PR tested in production?

khogeland: Yes, we've been running this in prod for a few months.

liyinan926: Cool!

main.go
liyinan926: OK, sounds good to me to roll into another PR.

liyinan926:
@khogeland any update on this PR?

khogeland:
@liyinan926 I'm out of town, probably won't push anything until next Monday.

liyinan926:
@khogeland thanks!

khogeland:
@liyinan926 I've updated to address your comments. The webhook now defines and handles its own command line flags (like glog). This seemed preferable to splitting the logic between main.go and webhook.go, but please let me know if you disagree.

PR #516, opened 2019-06-12
Add support for strict webhook failure policy

khogeland:
Allows users to enable a strict failure policy in the webhook config. Requires that a namespace selector be set for the webhook, as an operator pod that is malfunctioning or a lack of healthy operator instances could cause a cluster-wide outage (and prevent the operator itself from being rescheduled).


pkg/webhook/webhook.go
```diff
@@ -122,10 +143,12 @@ func (wh *WebHook) Start(webhookConfigName string) error {
 
 // Stop deregisters itself with the API server and stops the admission webhook server.
 func (wh *WebHook) Stop(webhookConfigName string) error {
-	if err := wh.selfDeregistration(webhookConfigName); err != nil {
-		return err
+	if wh.failurePolicy != arv1beta1.Fail {
```
liyinan926: Can you explain why this change is needed?

pkg/webhook/webhook.go
```diff
@@ -91,6 +99,19 @@ func New(
 		cert:              cert,
 		serviceRef:        serviceRef,
 		sparkJobNamespace: jobNamespace,
+		failurePolicy:     failurePolicy,
+	}
+
+	if webhookNamespaceSelector != "" {
+		kv := strings.SplitN(webhookNamespaceSelector, "=", 2)
```
liyinan926: Should we allow more multi-selector, e.g., `key1=val1,key2=val2`?

pkg/webhook/webhook.go
khogeland: I'll add a comment. If the webhook deregisters itself, some pods may be scheduled without hitting the webhook (for example, if the webhook gets deregistered while a driver is creating executors).

pkg/webhook/webhook.go
khogeland: I think it's best to keep this as simple until given a reason otherwise. If filtering on multiple values is needed, users can use e.g. `foo=sparkoperatorv123` instead of `sparkoperator=yes,sparkoperatorversion=123`.

PR #517, opened 2019-06-13
Improve webhook error handling

khogeland:
- Ensures all errors (internal and bad input) have a properly formatted `AdmissionReview` response
- Returns internal errors in `mutatePods` to the client and API server instead of logging/swallowing

(FYI, 4xx response codes are still not properly handled by the API server, but that's been fixed recently. They'll continue to return 500 to the client until 1.16.)


pkg/webhook/webhook.go
```diff
@@ -138,34 +138,40 @@ func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
...
 	} else {
-		reviewResponse = mutatePods(review, wh.lister, wh.sparkJobNamespace)
+		if review.Request.Resource == podResource {
+			reviewResponse, err = mutatePods(review, wh.lister, wh.sparkJobNamespace)
+			if err != nil {
+				internalError(w, err)
+				return
+			}
+		} else {
+			denyRequest(w, fmt.Sprintf("Unexpected resource type: %v", review.Request.Resource.String()), http.StatusUnsupportedMediaType)
```
liyinan926: We use lower-case for the first letter of error messages throughout the code base. Better to be consistent.

pkg/webhook/webhook.go
```diff
@@ -178,13 +184,39 @@ func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
...
+		internalError(w, err)
 		return
 	}
 	if _, err := w.Write(resp); err != nil {
+		internalError(w, err)
+	}
+}
+
+func internalError(w http.ResponseWriter, err error) {
+	glog.Errorf("Internal error: %v", err)
```
liyinan926: Ditto. Lower case for the first letter.

khogeland:
@liyinan926 updated

PR #511, opened 2019-06-09
Support leader election mechanism for Spark operator HA

tkanng:
Hi, this PR adds support for leader election mechanism for Spark operator HA by utilizing [leaderelection](https://github.com/kubernetes/client-go/tree/master/tools/leaderelection), according to https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/458.

PR #518, opened 2019-06-13
Implement leader election for HA

khogeland:
This is a duplicate of #511, but we already had our own implementation so... 🤷‍♂ Please feel free to close and take any insights over to that PR. A few differences:

- Configurable leader election values (timing + resource location)
- Avoids calling `selfDeregistration` in the webhook to avoid downtime (the other instances of the operator can still fulfill requests)


khogeland:
A thought: it might be cleaner to get rid of deregistration altogether, and instead set an `OwnerReference` on it to the Spark operator deployment. That avoids having to disable deregistration here and in #516, and since the default policy is to ignore errors, it shouldn't visibly affect the current behavior.

liyinan926:
@khogeland that sounds like a good idea.

ringtail:
@liyinan926 @khogeland Hi,What's the status of this issue.

khogeland:
@ringtail You could just use/merge my branch, we've been running this patch in prod for a long time. It's up to Li whether to merge into this repo. 🤷‍♂

main.go
```diff
@@ -50,23 +54,30 @@ import (
...
+	webhookPort                 = flag.Int("webhook-port", 8080, "Service port of the webhook server.")
+	enableMetrics               = flag.Bool("enable-metrics", false, "Whether to enable the metrics endpoint.")
+	metricsPort                 = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
+	metricsEndpoint             = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
+	metricsPrefix               = flag.String("metrics-prefix", "", "Prefix for the metrics.")
+	ingressUrlFormat            = flag.String("ingress-url-format", "", "Ingress URL format.")
+	enableLeaderElection        = flag.Bool("leader-election", false, "Enable Spark operator leader election.")
+	leaderElectionLockNamespace = flag.String("leader-election-lock-namespace", "spark-operator", "Namespace in which to create the ConfigMap for leader election.")
+	leaderElectionLockName      = flag.String("leader-election-lock-name", "spark-operator-lock", "Name of the ConfigMap for leader election.")
+
```
liyinan926: Unnecessary empty line.

main.go
```diff
@@ -84,23 +95,51 @@ func main() {
...
+	startCh := make(chan struct{}, 1)
+
+	if *enableLeaderElection {
+		hostname, err := os.Hostname()
+		if err != nil {
+			glog.Fatal(err)
+		}
+		resourceLock, err := resourcelock.New(resourcelock.ConfigMapsResourceLock, *leaderElectionLockNamespace, *leaderElectionLockName, kubeClient.CoreV1(), resourcelock.ResourceLockConfig{
+			Identity: hostname,
+			// TODO: This is a workaround for a nil dereference in client-go. This line can be removed when that dependency is updated.
```
liyinan926: Any pointer to some github issue that explains the issue?

main.go
```diff
@@ -133,17 +185,10 @@ func main() {
...
-	}
-	if err = scheduledApplicationController.Start(*controllerThreads, stopCh); err != nil {
-		glog.Fatal(err)
-	}
-
 	var hook *webhook.WebHook
 	if *enableWebhook {
 		var err error
-		hook, err = webhook.New(kubeClient, crInformerFactory, *webhookCertDir, *webhookSvcNamespace, *webhookSvcName, *webhookPort, *namespace)
+		hook, err = webhook.New(kubeClient, crInformerFactory, *webhookCertDir, *webhookSvcNamespace, *webhookSvcName, *webhookPort, *namespace, !*enableLeaderElection)
```
liyinan926: Please add a comment regarding `!*enableLeaderElection`.

main.go
```diff
@@ -84,23 +95,51 @@ func main() {
...
+
+	stopCh := make(chan struct{}, 1)
+	startCh := make(chan struct{}, 1)
+
+	if *enableLeaderElection {
+		hostname, err := os.Hostname()
+		if err != nil {
+			glog.Fatal(err)
+		}
+		resourceLock, err := resourcelock.New(resourcelock.ConfigMapsResourceLock, *leaderElectionLockNamespace, *leaderElectionLockName, kubeClient.CoreV1(), resourcelock.ResourceLockConfig{
```
liyinan926: Do we need to remove the lock upon exiting?

main.go
```diff
@@ -153,11 +198,25 @@ func main() {
 		}
 	}
 
-	signalCh := make(chan os.Signal, 1)
-	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM)
-	<-signalCh
+	if *enableLeaderElection {
```
liyinan926: Should this be moved up to be right after the elector is up and running?

manifest/spark-operator-rbac.yaml
```diff
@@ -34,7 +34,10 @@ rules:
   resources: ["pods"]
   verbs: ["*"]
 - apiGroups: [""]
-  resources: ["services", "configmaps", "secrets"]
+  resources: ["configmaps"]
```
liyinan926: Why this change?

main.go
khogeland: I don't think so, that could cause an unnecessary re-election when a non-leader pod is restarted. And when the leader loses its lease, it restarts, so that could cause some thrashing.

main.go
khogeland: The webhook should definitely be started regardless of leadership, because it's stateless and having multiple replicas makes it more reliable. I think the metrics thread should probably only start after winning the election, but if the webhook starts producing metrics in the future it'd have to change back to this order.

manifest/spark-operator-rbac.yaml
khogeland: The lock object is a ConfigMap, the leader election code needed some additional perms to work.

main.go
khogeland: It was fixed a while ago, the client-go dependency is just a couple versions behind: https://github.com/kubernetes/client-go/blame/b831b8de7155117e51afaffeb647007a756ddc92/tools/leaderelection/resourcelock/configmaplock.go#L96-L98

main.go
liyinan926: Got it.

main.go
liyinan926: Which version of `client-go` has the fix?

main.go
liyinan926: Makes sense.

main.go
khogeland: `kubernetes-1.14.0-alpha.1` and later, looks like

main.go
khogeland: Okay, I'll reorder, and since the metric config will need to be passed into the webhook if it ever starts producing metrics, it'll be really obvious that the ordering needs to be changed again.

main.go
liyinan926: In that case, should we actually supply a real `EventRecorder`, like the one used by the `SparkApplication` controller?

liyinan926:
LGTM overall with one remaining comment on the use of a `FakeRecorder`.

main.go
liyinan926: I think we can also update the version of `client-go` and other k8s libs.

liyinan926:
Merging this to master. Will take care of the `FakeRecorder` later.

IVAN-bd:
@khogeland Is there any docs telling about how to deploy HA operators?

PR #519, opened 2019-06-13
Added DNS config option for the driver and executor pods

jlpedrosa:
This PRs adds the option to specify DNSConfig as defined in [k8s api](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#poddnsconfig-v1-core)  for the Driver and the Executor pods. 

It adds the objects in the api and modifies the weebhook

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -378,6 +378,24 @@ spec:
       ...
 ```
 
+### Using DNS Settings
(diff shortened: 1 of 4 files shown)


jlpedrosa:
Hi @liyinan926 
Done, let me know if there would be anything else.

PR #520, opened 2019-06-15
Replace NodePort Service with ClusterIP

akhurana001:
@liyinan926 Removing NodePort Service as we discussed.  I have updated `WebUIAddress` under Status in CRD to `<CLUSTER-IP>:<PORT>` which will be accessible only from within the cluster.

As expected, this is non-backward compatible. I am not sure if any other SparkOperator user might be depending on the node-port service to access the UI.

Let me know your thoughts on this.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -147,7 +147,7 @@ A `DriverInfo` captures information about the driver pod and the Spark web UI ru
 | ------------- | ------------- |
 | `WebUIServiceName` | Name of the service for the Spark web UI. |
 | `WebUIPort` | Port on which the Spark web UI runs on the Node. |
-| `WebUIAddress` | Address to access the web UI from outside the cluster via the Node. |
+| `WebUIAddress` | Address to access the web UI from within the cluster. |
 | `WebUIIngressName` | Name of the ingress for the Spark web UI. |
 | `WebUIIngressAddress` | Address to access the web UI via the Ingress. |
 | `PodName` | Name of the driver pod. |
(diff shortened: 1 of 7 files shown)


liyinan926:
@eterna2 I believe you need the `NodePort` service for your Ambassador based approach for accessing the UI?

eterna2:
@liyinan926

ClusterIP is fine as well. Ambassador is pretty flexible. Ambassador 0.7 just released their own CRD. So we can do it via CRD or annotations. But I will still be working on the SparkApplicationClass.

Progress is a bit slow cuz I had some production issues, and spend the last few weeks on that.

alexakr:
Hi,
What was the reason of replacing NodePort service to ClusterIP?
Maybe it should be controllable by helm installation.

akhurana001:
@alexakr NodePort have some limitations and are not preferred to be used in production/at-scale. For e.g. NodePort are limited to one service/port in the cluster, require a public-ip on the node to be accessible and I think we were also making excessive calls to APIServer to get the updated node public-ip etc. which can cause issues at scale.

alexakr:
OK, but NodePort as well as LoadBalancer are legitimate options to expose services onto an external IP address. Exposing ClusterIP requires ingress controller installation and during development we would like to avoid deal with ingress controller just to access the UI.
I think that one option should not be replaced with another, but rather allowing both options, and this can be achieved by helm parameter. So, is it possible to do that?

akhurana001:
Got it, adding @liyinan926 to see what his thoughts are on this and if other users have also raised this concern.

FYI, you can also do a port-forward on the driver to access the UI to mitigate this during development.

PR #527, opened 2019-06-21
Reset ExecutionAttempts/SubmissionAttempts properly

liyinan926:
Currently we don't reset `ExecutionAttempts` upon a successful run of an application. This cause the counter to quickly reach the retry limit if it has previously failed several times. For example, assuming that an application failed 3 times in a row, and then succeeded once. If the retry limit on failures is 3, then the application will not be retried the next time it fails as its `ExecutionAttempts` has already exceeded 3. The semantics of ``ExecutionAttempts`, however, is a counter of failed attempts, and its value is used to determine if the controller should keep retrying running an application in case of failure, as the [code](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L421) shows. `The same goes for `SubmissionAttempts`. This PR fixes the bug by resetting `ExecutionAttempts` and `SubmissionAttempts` upon a successful run and submission of an application.

@akhurana001

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -526,7 +526,6 @@ func (c *Controller) syncSparkApplication(key string) error {
 					appToUpdate.Namespace, appToUpdate.Name, err)
 				return err
 			}
-			appToUpdate.Status.AppState.ErrorMessage = ""
 			appToUpdate.Status.AppState.State = v1beta1.PendingRerunState
 		}
 	case v1beta1.FailedSubmissionState:
@@ -905,10 +904,14 @@ func (c *Controller) clearStatus(status *v1beta1.SparkApplicationStatus) {
 		status.ExecutionAttempts = 0
 		status.LastSubmissionAttemptTime = metav1.Time{}
 		status.TerminationTime = metav1.Time{}
+		status.AppState.ErrorMessage = ""
 		status.ExecutorState = nil
 	} else if status.AppState.State == v1beta1.PendingRerunState {
 		status.SparkApplicationID = ""
+		status.SubmissionAttempts = 0
+		status.LastSubmissionAttemptTime = metav1.Time{}
 		status.DriverInfo = v1beta1.DriverInfo{}
+		status.AppState.ErrorMessage = ""
 		status.ExecutorState = nil
 	}
 }
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -295,9 +295,11 @@ type SparkApplicationStatus struct {
 	AppState ApplicationState `json:"applicationState,omitempty"`
 	// ExecutorState records the state of executors by executor Pod names.
 	ExecutorState map[string]ExecutorState `json:"executorState,omitempty"`
-	// ExecutionAttempts is the total number of attempts made to run a submitted Spark App to successful completion.
+	// ExecutionAttempts is the total number of attempts to run a submitted application to completion.
+	// Incremented upon each attempted run of the application and reset upon invalidation.
 	ExecutionAttempts int32 `json:"executionAttempts,omitempty"`
-	// SubmissionAttempts is the total number of submission attempts made to submit a Spark App.
+	// SubmissionAttempts is the total number of attempts to submit an application to run.
+	// Incremented upon each attempted submission of the application and reset upon invalidation and rerun.
 	SubmissionAttempts int32 `json:"submissionAttempts,omitempty"`
 }
 


akhurana001:
I think the right way to achieve this is to reset these fields in the `PendingRerunState` which we should go through for every re-run. I see we already clear some of these in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/527/files#diff-b7eea120d7ac0e8587781b863b7db8e4L909 similar to what we seem to be doing for `InvalidatingState`  Let me know if you think that might not work.

liyinan926:
@akhurana001 resetting them in `PendingRerun` state means resetting them even after each consecutive failed run, which is certainly not what the semantic of these counters are.

akhurana001:
> resetting them in `PendingRerun` state means resetting them even after each consecutive failed run, which is certainly not what the semantic of these counters are.

My bad, that's correct for `ExecutionAttempts` although resetting `SubmissionAttempts` in `PendingRerun` should be fine. 

> Currently we don't reset ExecutionAttempts upon a successful run of an application. This cause the counter to quickly reach the retry limit if it has previously failed several times. For example, assuming that an application failed 3 times in a row, and then succeeded once. If the retry limit on failures is 3, then the application will not be retried the next time it fails as its ExecutionAttempts has already exceeded 3. 

Thinking more about this, this case shouldn't happen. The application is only restarted after a successful run if the `restartPolicy` is `Always` and if so, any future restarts decisions won't take the `ExecutionAttempts` into account as per https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L414 

Overall, I believe `SubmissionAttempts` should still capture what happened within a run of an application so it will be better if we reset in `PendingRerun` before we start a new run of the application.

For `ExecutionAttempts` , as I mentioned above, I don't think its causing any issues right now and is probably useful to understand the total runs (success + failures) for the application but yes, some future `restartPolicy` changes might end up breaking it. I am not sure if this required now but I be fine if you strongly believe that we should be resetting this after each successful run.

liyinan926:
> Thinking more about this, this case shouldn't happen. The application is only restarted after a successful run if the restartPolicy is Always and if so, any future restarts decisions won't take the ExecutionAttempts into account as per https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L414

`ExecutionAttempt` is checked against the failure retry limit when the `restartPolicy` is `OnFailure` so it does matter.

akhurana001:
> `ExecutionAttempt` is checked against the failure retry limit when the `restartPolicy` is `OnFailure` so it does matter.

Yes, I was just pointing out that a case like `Failed -> Failed -> Success -> Failed -> Failed -> ....` can only happen if the `restartPolicy` is `Always` so  `ExecutionAttempt` check won't happen for such a case and I don't think its causing any issues right now but I am fine if you strongly believe that we should be resetting it.

pkg/controller/sparkapplication/controller.go
```diff
@@ -503,6 +503,10 @@ func (c *Controller) syncSparkApplication(key string) error {
 			appToUpdate = c.submitSparkApplication(appToUpdate)
 		}
 	case v1beta1.SucceedingState:
+		if appToUpdate.Status.ExecutionAttempts > 1 {
```
akhurana001: This check should be `>=1` . I don't think we even need this check given that we are in `SuceedingState`

liyinan926:
> I was just pointing out that a case like `Failed -> Failed -> Success -> Failed -> Failed -> ....` can only happen if the `restartPolicy` is `Always` so `ExecutionAttempt` check won't happen for such a case and I don't think its causing any issues right now but I am fine if you strongly believe that we should be resetting it.

Yes, normally this won't cause the issue. However, the user may update the app spec and trigger a rerun. The rerun may fail and not get retried if `ExecutionAttempts` is not cleared.

pkg/controller/sparkapplication/controller.go
liyinan926: We don't need to check. Fixed.

liyinan926:
> Overall, I believe SubmissionAttempts should still capture what happened within a run of an application so it will be better if we reset in PendingRerun before we start a new run of the application.

Done.

pkg/controller/sparkapplication/controller.go
```diff
@@ -643,7 +644,7 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 		DriverInfo: v1beta1.DriverInfo{
 			PodName: driverPodName,
 		},
-		SubmissionAttempts:        app.Status.SubmissionAttempts + 1,
+		SubmissionAttempts:        0, // Reset in case of a successfuly submission.
```
akhurana001: Given that we are now resetting in `PendingRerun` we don't need this anymore. Can we please revert this ?

This way, SubmissionAttempts will still reflect correct attempt count for a single run of the application without impacting any retry behavior.

akhurana001:
One minor comment. Looks good to me otherwise 👍

pkg/controller/sparkapplication/controller.go
liyinan926: Just like we reset `ExecutionAttempts` upon a successful run, I think the right semantic is to reset `SubmissionAttempts` upon a successful submission as `SubmissionAttempts` is really the number of consecutive failed submission attempts.

PR #529, opened 2019-06-24
Added support for overriding command and args

jlpedrosa:
Hi
This patch adds the option to override the command and args for the driver and executor.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -409,7 +409,7 @@ spec:
         - name: edns0
 ```
 ### Overriding args and commands
-A `SparkApplication` can override the command and arguments provided to both the executor and the driver. `.spec.driver.args` and `.spec.executor.args` for arguments and `.spec.driver.commands` and `.spec.driver.commands`. Four fields are an array of strings, following the semantics and behaviour described at the [official specs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#container-v1-core)
+A `SparkApplication` can override the command and arguments provided to both the executor and the driver. `.spec.driver.args` and `.spec.executor.args` for arguments and `.spec.driver.command` and `.spec.driver.command`. Four fields are an array of strings, following the semantics and behaviour described at the [official specs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#container-v1-core)
 
 
 Note that the mutating admission webhook is needed to use this feature. Please refer to the 
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -408,6 +408,9 @@ spec:
           value: "2"
         - name: edns0
 ```
+### Overriding args and commands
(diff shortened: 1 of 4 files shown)


pkg/webhook/patch.go
```diff
@@ -182,6 +183,42 @@ func addEnvironmentVariable(pod *corev1.Pod, envName, envValue string) patchOper
 	return patchOperation{Op: "add", Path: path, Value: value}
 }
 
+func addCommandArgs(pod *corev1.Pod, app *v1beta1.SparkApplication) []patchOperation {
+	var podSpec *v1beta1.SparkPodSpec
```
liyinan926: This can be moved to after line 198.

jlpedrosa: @liyinan926 

Fixed, please double check.

liyinan926:
What are some use cases of overriding commands and arguments of driver/executors?

jlpedrosa:
Hi @liyinan926 
When using [virtial kubelet](https://github.com/virtual-kubelet/virtual-kubelet) the feature to get the ip based on env variable fromField of the pod spec does not work because is not supported in all cloud providers. So I need to override the entry point and args to actually be able to asing the ip based on commands that can be executed in that platform and populate the env field.

docs/user-guide.md
```diff
@@ -408,6 +408,9 @@ spec:
           value: "2"
         - name: edns0
 ```
+### Overriding args and commands
+A `SparkApplication` can override the command and arguments provided to both the executor and the driver. `.spec.driver.args` and `.spec.executor.args` for arguments and `.spec.driver.commands` and `.spec.driver.commands`. Four fields are an array of strings, following the semantics and behaviour described at the [official specs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#container-v1-core)
```
liyinan926: Typo: `.spec.driver.command` and `.spec.executor.command`.

docs/user-guide.md
```diff
@@ -408,6 +408,9 @@ spec:
           value: "2"
         - name: edns0
 ```
+### Overriding args and commands
```
liyinan926: Capital case of the first letter of each word.

docs/user-guide.md
jlpedrosa: Ok!

docs/user-guide.md
jlpedrosa: Ah, ok, I was trying to aim for the name in the yaml/json. Changed.

pkg/webhook/patch.go
```diff
@@ -182,6 +183,43 @@ func addEnvironmentVariable(pod *corev1.Pod, envName, envValue string) patchOper
...
+	// Find the driver or executor container in the pod.
+	for ; i < len(pod.Spec.Containers); i++ {
+		if pod.Spec.Containers[i].Name == sparkDriverContainerName ||
+			pod.Spec.Containers[i].Name == sparkExecutorContainerName {
+			break
+		}
+	}
+
+	path := fmt.Sprintf("/spec/containers/%d", i)
+	var podSpec *v1beta1.SparkPodSpec
```
liyinan926: Looks like the Travis build was not completed successfully, please do some dummy changes to trigger another build, e.g., add an empty line before 198 and remove the empty line after 198.

jlpedrosa: @liyinan926 
Done.

PR #530, opened 2019-06-24
Allows arbitrary JSON patch operations for the driver and the executors.

jlpedrosa:
Hi

This PR adds the capability of applying arbitrary json patches on the executor and driver pod.
This enables the use case of virtual kubelet which does not support env variables fromField pod definitions.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -409,6 +409,10 @@ spec:
         - name: edns0
 ```
 
+### Arbitrary json modification
+Allows aplying arbitrary json patches to the pods definitions created by spark submit. Enables removal/modification of settings that may be required in some edge cases.
+
+
 Note that the mutating admission webhook is needed to use this feature. Please refer to the 
 [Quick Start Guide](quick-start-guide.md) on how to enable the mutating admission webhook.
 
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -404,7 +404,7 @@ type SparkPodSpec struct {
 	DNSConfig *apiv1.PodDNSConfig `json:"dnsConfig,omitempty"`
 	// JSONPatchOperations aribitraty set of json patch operations to be applied to the resulting pods before admission.
 	// Optional
-	JSONPatchOperations []PatchOperation
+	JSONPatchOperations []PatchOperation `json:"patchOps,omitempty"`
 }
 
 // DriverSpec is specification of the driver.
--- a/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
@@ -402,6 +402,9 @@ type SparkPodSpec struct {
 	// DnsConfig dns settings for the pod, following the Kubernetes specifications.
 	// Optional.
 	DNSConfig *apiv1.PodDNSConfig `json:"dnsConfig,omitempty"`
+	// JSONPatchOperations aribitraty set of json patch operations to be applied to the resulting pods before admission.
+	// Optional
+	JSONPatchOperations []PatchOperation
 }
 
 // DriverSpec is specification of the driver.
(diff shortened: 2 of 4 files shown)


jlpedrosa:
Rebased, history re-writen.

liyinan926:
Interesting. Can you give some examples (yamls) of how this is used for the Virtual Kubelet case? This sounds to be a really powerful and advanced feature that can open up a lot of possibilities and risks.

jlpedrosa:
Hi @liyinan926 

Let me explain, virtual kubelet does not support env variables obtained via “fieldRef”. So the goal of this patch is to be able to delete it via the patch and then have custom args and commands to be able to specify custom script to be able so set the POD_IP env variable.

I agree, it's pretty advanced feature with a lot of possibilities and risks. This is the error message provided by virtual kubelet. 

```
level=warning msg="requeuing [[\"default/hellospark2-1557482540117-driver\" \"failed to sync pod \\\"default/hellospark2-1557482540117-driver\\\" in the provider: unsupported fieldPath: status.podIP\"]]
```
so the idea would be something like:

```yaml
spec:
  driver:  
      ...  
  executor:
    patchOps:
    - Op: "remove"
      Path: "/spec/containers/0/env"
```
Arguably we could root the patch operation to the container directly instead of the pod, to avoid the /0/ which I'd say is not the nicest option, but we'd loose use cases. What do you think?

liyinan926:
Hmm, so you need to get the pod IP into your driver or executor containers. But Virtual Kubelet doesn's support `fromFieldRef` for environment variables. So instead you need to modify the entrypoint cmd of the container as a workaround to do that? If your goal is just to support what `fromFieldRef` is able to do, how about adding support for that in the  webhook when the k8s provider does not support that? This and the PR #529 are both too powerful and bring both a lot of flexibilities and risks. I would suggest expanding the webhook to address tour need instead of the approach in this and the other PR.

jlpedrosa:
I'm not sure if I am following you, or maybe I did not express myself correctly. What do yo mean by 

> about adding support for that in the webhook when the k8s provider does not support that? 

The problem is that field that is not supported, is also avoiding the pod to get scheduled at all. The message says warning, but is actually, pod rejected. In some layer/place I need to actually remove the  ```fromFieldRef``` in order to get scheduled.

I was trying to add a feature that is generic and makes sense by itself, but I am absolutely open to implement it however you consider more appropriate. Not sure if you could explain more how you want it or if you changed your mind and you find this one ok.

Thanks!

liyinan926:
Sorry I didn't make it clear. I was thinking of the following approach: the operator supports what envsubst does, i.e., replacing variables in environment variables with real values. We can have a set of pre-defined variables that the operator supports, like podIP. Not sure if this works for your use case or not.

Regarding PR #529, I understand better now why that's needed. After stepping back and giving it more thought, I do think it brings a lot of risks. The ability to inject arbitrary commands also means the ability to do things that are not desirable.

jlpedrosa:
@liyinan926 
The problem is that with the pod definition generated by spark (or even using templates from 3.0 master branch), The container gets rejected because the fieldRef. So indeed I could generate other images for the ip address, but I'd need in any case this one  to eliminate the env variables generated by spark.

skonto:
Me and @yuchaoran2011 are working on moving the operator to the controller-runtime and also as a next step to support the pod template feature internally, with the intention of removing the webhook thing that causes several issues eg. olm support difficulties, mutation happens too late so defaults may pollute your spec since other plugins may run etc. I think we should still support this but I dont see it as a generic case. Its a corner case, so I believe webhook should be optional at best but much better if you have your custom webhook implemented outside of the operator to make things work with the virtual kubelet.
It is not ok to me to open such capabilities to the end user as it was not for the pod template to be exposed to the end-user via the crd (discussed this with @liyinan926 before).

jlpedrosa:
Hi @skonto 

I think more than virtual kubelet problem is the users problems. What I mean is that as the backend of the virtual kubelet (the container runtime) does not support that feature, only few dark workarounds could be implemented. In any case I understand your point of view.

liyinan926:
Closing this as discussed above this is really beyond what the operator is supposed to support.

PR #532, opened 2019-06-24
Revert "Added support for overriding command and args"

liyinan926:
Reverts GoogleCloudPlatform/spark-on-k8s-operator#529

@jlpedrosa Like what I said in #530. I think we should focus on coming up with an approach that enables your use case with Virtual Kubelet, while introducing minimum risks. Sorry I haven't thought thoroughly about this PR, which IMO brings a lot of risks just for enabling a very specific case.


jlpedrosa:
Hi @liyinan926 

The IP in the case of virtual kubelet is not managed by k8s, it's assigned directly by the cloud provider.
This means we can only assign the IP at the container runtime (that is why overriding the command and args), and the moment we try to schedule the pod. The next issue is that obtain the ip at the container will be cloud specific. For instance, azure ACI, has two nics, which is not natively supported by k8s, that is why we went for override command and args to be able to get the IP for that env.

Can you think of any other way to support this scenario radther than command args + json patch?

liyinan926:
@jlpedrosa How about an init container that resolves the IP and write the IP to a file on a emptyDir volume shared with the main Spark container?

liyinan926:
The other alternative is you create your custom Spark Dockerfile that has the command you need, instead of using the official Dockerfile.

PR #543, opened 2019-07-03
Webhook enhancements

khogeland:
Bundled to avoid merge conflicts

Subsumes #515 #516 #517:
- Periodic cert reloading
- Improved error reporting
- Strict error handling


pkg/webhook/webhook.go
```diff
@@ -60,74 +59,127 @@ type WebHook struct {
...
+		certProvider:      cert,
 		serviceRef:        serviceRef,
 		sparkJobNamespace: jobNamespace,
+		failurePolicy:     arv1beta1.Ignore,
+	}
+	if userConfig.webhookFailOnError {
+		if userConfig.webhookNamespaceSelector == "" {
+			glog.Fatal("webhook-namespace-selector must be set when webhook-fail-on-error is true.")
+		} else {
+			kv := strings.SplitN(userConfig.webhookNamespaceSelector, "=", 2)
```
liyinan926: I think we should also support specification of multiple selector entries, e.g., `key1=val1,key2=val2`.

pkg/webhook/certs.go
```diff
@@ -19,24 +19,74 @@ package webhook
...
+			}
+		}
+	}()
+}
 
-	return &tls.Config{Certificates: []tls.Certificate{cert}}, nil
+func (c *certProvider) tlsConfig() *tls.Config {
+	return &tls.Config{
+		GetCertificate: func(ch *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return c.currentCert, nil
```
liyinan926: Do we need to use a mutex to guard `currentCert`, which may be read in `tlsConfig` and written in `updateCert`?

pkg/webhook/webhook.go
```diff
@@ -60,74 +59,127 @@ type WebHook struct {
...
 		lister:            informerFactory.Sparkoperator().V1beta1().SparkApplications().Lister(),
-		cert:              cert,
+		certProvider:      cert,
 		serviceRef:        serviceRef,
 		sparkJobNamespace: jobNamespace,
+		failurePolicy:     arv1beta1.Ignore,
+	}
+	if userConfig.webhookFailOnError {
+		if userConfig.webhookNamespaceSelector == "" {
+			glog.Fatal("webhook-namespace-selector must be set when webhook-fail-on-error is true.")
```
liyinan926: Remind me why is this a requirement?

khogeland: Because the operator going down would cause a cluster-wide outage. And if the operator pod were rescheduled, it would required manual intervention (because it would have a circular dependency on itself to start up).

pkg/webhook/certs.go
khogeland: It's probably a good idea, since this behavior is technically undefined (although AFAICT it's safe on all modern architectures).

pkg/webhook/webhook.go
liyinan926: Got it. Instead of using `glog.Fatal`, please return an error instead. The caller of `New` will do `glog.Fatal` on the return error anyway.

liyinan926:
@khogeland can you address the remaining comments? It looks good otherwise.

khogeland:
@liyinan926 Addressed your above comments

pkg/webhook/webhook.go
```diff
@@ -254,61 +348,47 @@ func (wh *WebHook) selfDeregistration(webhookConfigName string) error {
 func mutatePods(
 	review *admissionv1beta1.AdmissionReview,
 	lister crdlisters.SparkApplicationLister,
-	sparkJobNs string) *admissionv1beta1.AdmissionResponse {
+	sparkJobNs string) (*admissionv1beta1.AdmissionResponse, error) {
 	if review.Request.Resource != podResource {
```
liyinan926: Looks like we don't need this check here any more as the same check is added above.

pkg/webhook/webhook.go
```diff
@@ -178,13 +247,39 @@ func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
...
+	resp, err := json.Marshal(response)
+	if err != nil {
 		glog.Error(err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(code)
+	_, err = w.Write(resp)
+	if err != nil {
+		glog.Errorf("Failed to write response body: %v", err)
```
liyinan926: Nit: lower case for error messages.

pkg/webhook/webhook.go
```diff
@@ -138,34 +201,40 @@ func (wh *WebHook) serve(w http.ResponseWriter, r *http.Request) {
...
 
 	var reviewResponse *admissionv1beta1.AdmissionResponse
 	review := &admissionv1beta1.AdmissionReview{}
 	deserializer := codecs.UniversalDeserializer()
 	if _, _, err := deserializer.Decode(body, nil, review); err != nil {
-		glog.Error(err)
-		reviewResponse = toAdmissionResponse(err)
+		internalError(w, err)
+		return
 	} else {
```
liyinan926: Nit: this `else` is no longer needed with the `return` above.

pkg/webhook/webhook.go
```diff
@@ -60,74 +59,138 @@ type WebHook struct {
...
-		cert:              cert,
+		certProvider:      cert,
 		serviceRef:        serviceRef,
 		sparkJobNamespace: jobNamespace,
+		failurePolicy:     arv1beta1.Ignore,
+	}
+	if userConfig.webhookFailOnError {
+		if userConfig.webhookNamespaceSelector == "" {
+			return nil, fmt.Errorf("webhook-namespace-selector must be set when webhook-fail-on-error is true.")
+		} else {
```
liyinan926: Nit: this `else` is no longer needed with the `return` above.

PR #553, opened 2019-07-13
Multi Version Support via Job:  Add Tests + other changes 

akhurana001:
Some changes for the multi-version branch (Ref: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/526):
-  Fix Controller tests & add tests for SubmissionJobManager.
-  Create interface for SubmissionJobManager and move `submissionCmdArgs` generation to within SubmissionJobManager.
- Add default resource limits for the Job pods. We use resourceQuotas on clusters which cause failures unless resource requests/limits are set on the job pod.
- Remove OnSubmissionFailureRetryInterval from v1beta1 spec.

I also see that now `FailedSubmissionState`  is terminal and we don't move it to `FailedState` or retry it even if the retryPolicy is `Always` . This is a behavior change compared to master, we can chat more if we want to update this.

PTAL @liyinan926  when you get a chance.

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -424,6 +424,11 @@ func shouldRetry(app *v1beta1.SparkApplication) bool {
 				return true
 			}
 		}
+	case v1beta1.FailedSubmissionState:
+		// We retry only if the RestartPolicy is Always. The Submission Job already retries upto the OnSubmissionFailureRetries specified.
+		if app.Spec.RestartPolicy.Type == v1beta1.Always {
+			return true
+		}
 	}
 	return false
 }
@@ -555,7 +560,16 @@ func (c *Controller) syncSparkApplication(key string) error {
 			appToUpdate.Status.AppState.State = v1beta1.PendingRerunState
 		}
 	case v1beta1.FailedSubmissionState:
-		c.recordSparkApplicationEvent(appToUpdate)
+		// Submission Job terminated in failure, check if the application needs to be retried.
+		if !shouldRetry(appToUpdate) {
+			// Application is not subject to retry. Move to terminal FailedState.
+			appToUpdate.Status.AppState.State = v1beta1.FailedState
+			c.recordSparkApplicationEvent(appToUpdate)
+		} else {
+			// Application is subject to retry. Move to PendingRerunState.
+			appToUpdate.Status.AppState.ErrorMessage = ""
+			appToUpdate.Status.AppState.State = v1beta1.PendingRerunState
+		}
 	case v1beta1.InvalidatingState:
 		// Invalidate the current run and enqueue the SparkApplication for re-submission.
 		if err := c.deleteSparkResources(appToUpdate); err != nil {
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/google/uuid"
 	"golang.org/x/time/rate"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -35,7 +34,6 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	batchv1 "k8s.io/client-go/listers/batch/v1"
 	v1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
@@ -72,9 +70,8 @@ type Controller struct {
 	metrics           *sparkAppMetrics
 	applicationLister crdlisters.SparkApplicationLister
 	podLister         v1.PodLister
-	jobLister         batchv1.JobLister
 	ingressURLFormat  string
-	subJobManager     *submissionJobManager
+	subJobManager     submissionJobManagerIface
 }
 
 // NewController creates a new Controller.
@@ -115,7 +112,6 @@ func newSparkApplicationController(
 		recorder:         eventRecorder,
 		queue:            queue,
 		ingressURLFormat: ingressURLFormat,
-		subJobManager:    &submissionJobManager{kubeClient: kubeClient},
 	}
 
 	if metricsConfig != nil {
@@ -141,7 +137,6 @@ func newSparkApplicationController(
 	controller.podLister = podsInformer.Lister()
 
 	jobLister := informerFactory.Batch().V1().Jobs().Lister()
-	controller.jobLister = jobLister
 	controller.subJobManager = &submissionJobManager{kubeClient: kubeClient, jobLister: jobLister}
 
 	controller.cacheSynced = func() bool {
@@ -504,9 +499,7 @@ func (c *Controller) syncSparkApplication(key string) error {
 	case v1beta1.PendingSubmissionState:
 		// Check the status of the submission Job and set the application status accordingly.
 		succeeded, completionTime, err := c.subJobManager.hasJobSucceeded(appToUpdate)
-		if err != nil {
-			return err
-		}
+
 		if succeeded != nil {
 			if *succeeded {
 				c.createSparkUIResources(appToUpdate)
@@ -515,12 +508,18 @@ func (c *Controller) syncSparkApplication(key string) error {
 				if completionTime != nil {
 					appToUpdate.Status.SubmissionTime = *completionTime
 				}
+				c.recordSparkApplicationEvent(appToUpdate)
 			} else {
 				// Since we delegate submission retries to the Kubernetes Job controller, the fact that the
 				// submission Job failed means all the submission attempts failed. So we set the application
 				// state to FailedSubmission, which is a terminal state.
 				appToUpdate.Status.AppState.State = v1beta1.FailedSubmissionState
+				appToUpdate.Status.AppState.ErrorMessage = err.Error()
+				c.recordSparkApplicationEvent(appToUpdate)
 			}
+		} else if err != nil {
+			// Received an error trying to query the status of the Job.
+			return err
 		}
 	case v1beta1.SucceedingState:
 		// The current run of the application has completed, check if it needs to be restarted.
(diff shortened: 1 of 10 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -72,9 +70,8 @@ type Controller struct {
 	metrics           *sparkAppMetrics
 	applicationLister crdlisters.SparkApplicationLister
 	podLister         v1.PodLister
-	jobLister         batchv1.JobLister
 	ingressURLFormat  string
-	subJobManager     *submissionJobManager
+	subJobManager     submissionJobManagerIface
```
liyinan926: nit: let's name this interface `submissionJobManager` and the real implementation `realSubmissionJobManager` and the fake one for testing `fakeSubmissionJobManager`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -504,9 +499,7 @@ func (c *Controller) syncSparkApplication(key string) error {
 	case v1beta1.PendingSubmissionState:
 		// Check the status of the submission Job and set the application status accordingly.
 		succeeded, completionTime, err := c.subJobManager.hasJobSucceeded(appToUpdate)
-		if err != nil {
-			return err
```
liyinan926: I prefer handling errors first.

pkg/controller/sparkapplication/controller.go
```diff
@@ -515,12 +508,18 @@ func (c *Controller) syncSparkApplication(key string) error {
 				if completionTime != nil {
 					appToUpdate.Status.SubmissionTime = *completionTime
 				}
+				c.recordSparkApplicationEvent(appToUpdate)
 			} else {
 				// Since we delegate submission retries to the Kubernetes Job controller, the fact that the
 				// submission Job failed means all the submission attempts failed. So we set the application
 				// state to FailedSubmission, which is a terminal state.
 				appToUpdate.Status.AppState.State = v1beta1.FailedSubmissionState
+				appToUpdate.Status.AppState.ErrorMessage = err.Error()
```
liyinan926: `err` may be `nil` and cause an panic here.

pkg/controller/sparkapplication/job.go
```diff
@@ -30,62 +35,96 @@ import (
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/config"
 )
 
+const (
+	sparkSubmitPodMemory = "1000Mi"
+	sparkSubmitPodCpu    = "1024m"
+)
+
+type submissionJobManagerIface interface {
+	createSubmissionJob(app *v1beta1.SparkApplication) (*string, *string, error)
```
liyinan926: Please change the return types to `string, string, error` instead.

pkg/controller/sparkapplication/job.go
```diff
@@ -18,8 +18,13 @@ package sparkapplication
 
 import (
 	"fmt"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"k8s.io/api/core/v1"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/resource"
```
liyinan926: The order of imports needs fix.

pkg/controller/sparkapplication/job.go
```diff
@@ -30,62 +35,96 @@ import (
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/config"
 )
 
+const (
+	sparkSubmitPodMemory = "1000Mi"
```
liyinan926: The default resource requests are too high for a simple submit job.

pkg/controller/sparkapplication/job.go
```diff
@@ -30,62 +35,96 @@ import (
...
+	sparkSubmitPodCpu    = "1024m"
+)
+
+type submissionJobManagerIface interface {
+	createSubmissionJob(app *v1beta1.SparkApplication) (*string, *string, error)
+	deleteSubmissionJob(app *v1beta1.SparkApplication) error
+	getSubmissionJob(app *v1beta1.SparkApplication) (*batchv1.Job, error)
+	hasJobSucceeded(app *v1beta1.SparkApplication) (*bool, *metav1.Time, error)
```
liyinan926: Use `bool` instead of `*bool`.

pkg/controller/sparkapplication/job_test.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+	return &submissionJobManager{
+		jobLister:  lister,
+		kubeClient: kubeClient,
+	}
+}
+
+func TestGetSubmissionJob(t *testing.T) {
+
```
liyinan926: Unnecessary empty line.

pkg/controller/sparkapplication/job_test.go
```diff
@@ -0,0 +1,222 @@
...
+package sparkapplication
+
+import (
+	"k8s.io/api/core/v1"
+	"os"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
liyinan926: Order of imports needs fix.

pkg/controller/sparkapplication/job.go
```diff
@@ -30,62 +35,96 @@ import (
...
-	for key, val := range s.app.Labels {
+	for key, val := range app.Labels {
 		job.Labels[key] = val
 	}
-	return sjm.kubeClient.BatchV1().Jobs(s.app.Namespace).Create(job)
+	_, err = sjm.kubeClient.BatchV1().Jobs(app.Namespace).Create(job)
+	if err != nil {
+		return nil, nil, err
+	}
+	// Submission Job Created successfully.
```
liyinan926: Unnecessary comment.

liyinan926:
> I also see that now FailedSubmissionState is terminal and we don't move it to FailedState or retry it even if the retryPolicy is Always . This is a behavior change compared to master, we can chat more if we want to update this.

Yes, we need to fix that.

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/controller.go
akhurana001: The method generates errors in two separate cases: when succeeded is nil (actual error trying to query job status) or when succeeded is false (error as to why the job failed). Hence this looks like a more cleaner way to handle it to me.

pkg/controller/sparkapplication/controller.go
akhurana001: shouldn't be but happy to add a nil check

pkg/controller/sparkapplication/job.go
akhurana001: done

pkg/controller/sparkapplication/job.go
akhurana001: done

pkg/controller/sparkapplication/job.go
akhurana001: reduced to 250Mi/250m. right now, i pass the same defaults for both requests and limits. let me know if you think we should set a higher limit

pkg/controller/sparkapplication/job.go
akhurana001: This already existed and we need a pointer as we need to return a nil value if we weren't able to query the status at all ?

pkg/controller/sparkapplication/job_test.go
akhurana001: removed

pkg/controller/sparkapplication/job_test.go
akhurana001: done

pkg/controller/sparkapplication/job.go
akhurana001: done

akhurana001:
> > I also see that now FailedSubmissionState is terminal and we don’t move it to FailedState or retry it even if the retryPolicy is Always . This is a behavior change compared to master, we can chat more if we want to update this.

> Yes, we need to fix that.

K8s Jobs don’t support retrying forever even when the restartPolicy is `onFailure` so if we want to retry submission forever, we would need to handle this on our end.

Since we don’t take in a separate `OnSubmissionFailureRetryInterval` anymore (K8s jobs do a default backoff) or know about how many times a submission failure has occurred, I was thinking of just doing a constant wait time (maybe 60 seconds) and move the app to `PendingRerunState` if the restartPolicy is `Always` on the SparkApplication ?

A cleaner way probably is to handle all retries in the operator and let users set `OnSubmissionFailureRetryInterval` as before.  We can pass in a `backofflimit` of 0 (or 1 https://github.com/kubernetes/kubernetes/issues/67848 ?) at the Job level although there will be some overhead and we can probably use pods at that point .

Let me know what your thoughts are on this.

pkg/controller/sparkapplication/job.go
liyinan926: The default is still too high IMO. Let's set a low request but a higher limit.

pkg/controller/sparkapplication/controller.go
liyinan926: Got it. Can you add some comments about the semantics here?

pkg/controller/sparkapplication/job.go
```diff
@@ -30,85 +36,121 @@ import (
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/config"
 )
 
-type submissionJobManager struct {
+const (
+	sparkSubmitPodMemory = "256Mi"
```
liyinan926: I think we can use a lower memory request.

pkg/controller/sparkapplication/controller.go

pkg/controller/sparkapplication/job.go
akhurana001: lowered request cpu/memory to 100m/100Mi  . limits are 250m/256Mi

akhurana001:
@liyinan926 let me know what your thoughts are on handling `always` restartPolicy 
 for submission job: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/553#issuecomment-517042482

liyinan926:
In case of a `FailedSubmission` and a retry policy of `OnFailure` or `Always`, we can just change the state from `FailedSubmission` to `Failing`.

Can you run `go fmt ./...` to fix the formatting of the code?

akhurana001:
> In case of a `FailedSubmission` and a retry policy of `OnFailure` or `Always`, we can just change the state from `FailedSubmission` to `Failing`.
> 
> Can you run `go fmt ./...` to fix the formatting of the code?

On `FailedSubmission`,  I made the changes to move to `PendingRerun`  only if the restartPolicy is `Always`. If the restartPolicy is `OnFailure` the Job should have already handled the retries and we just move it to  the terminal `FailedState`. This is same as the current master implementation: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L431

Also ran `go fmt` as well.

PR #563, opened 2019-08-09
Update Dockerfiles after #562

Reamer:
With this changes users should be able to build spark-operator again with provided Dockerfiles.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,16 +17,11 @@
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.0
 
 FROM golang:1.12.5-alpine as builder
-ARG DEP_VERSION="0.5.3"
-RUN apk add --no-cache bash git
-ADD https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 /usr/bin/dep
-RUN chmod +x /usr/bin/dep
 
+RUN apk add --no-cache git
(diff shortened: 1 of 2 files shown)


Reamer:
@googlebot I signed it!

Dockerfile
```diff
@@ -17,16 +17,11 @@
...
-ADD https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 /usr/bin/dep
-RUN chmod +x /usr/bin/dep
 
+RUN apk add --no-cache git
 WORKDIR ${GOPATH}/src/github.com/GoogleCloudPlatform/spark-on-k8s-operator
-COPY Gopkg.toml Gopkg.lock ./
-RUN dep ensure -vendor-only
 COPY . ./
-RUN go generate && CGO_ENABLED=0 GOOS=linux go build -o /usr/bin/spark-operator
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux go build -o /usr/bin/spark-operator
```
liyinan926: I think we still need `go generate`.

liyinan926: Just realized `go generate` is broken now due to the change in #562. The script used to perform code generation is not included even if we do `go mod vendor`.

liyinan926:
#562 was reverted in #564 because it broke `go generate`. I will redo it once I figure out how to handle `go generate` with go modules. So this PR can be closed for now.

Issue #550, opened 2019-07-10
Integration spark-operator with Volcano


k82cn:
We're evaluating spark-operator and would like to integrate with [Volcano](http://github.com/volcano-sh/volcano) for following features:

1. Resource sharing between spark job by DRF
1. Preemption based on resource sharing, e.g. drf
1. Delay pod creation to reduce resource consumption from driver
1. Queue (on-going)

What we're going to do is to create PodGroup for each SparkApplication; the PodGroup can not be created outside operator as 1. it's better to make SparkApplcation as its owner reference for GC 2. PodGroup's `minResource` will help to avoid one driver with one executor, and the `minResources` comes from SparkApplication.

k82cn:
If that's ok, I'm going to open PR for that :)

liyinan926:
@k82cn will get back to you when I get back to work next week.

k82cn:
sure, "see" you next week :)

liyinan926:
@k82cn I have a couple of questions/comments:
* Is a `PodGroup` only for the executors or both the driver and executors? How does it work if it's the latter where the executor pods come after the driver pod starts running?
* Creating a `PodGroup` should be optional and can be enabled/disabled through a flag.
* If we do this, there'll be a dependency on the Volcano controller and CRDs. We need a way to check the presence of the Volcano controller and CRDs and install them when necessary, if the feature is enabled.

k82cn:
> Is a PodGroup only for the executors or both the driver and executors? 

For spark-operator, prefer to use the same PodGroup for both driver and executors.

> How does it work if it's the latter where the executor pods come after the driver pod starts running?

We will provide `minResource` in PodGroup to support DelayPodCreation, and scheduler support minAvaialbe is 1, total pod is more than 1. As PodGroups in scheduler will share resource by DRF, so SparkApplication (driver + executor). And some other features that supported by volcano :)

> Creating a PodGroup should be optional and can be enabled/disabled through a flag.

Yes, it's controlled by flag.

> If we do this, there'll be a dependency on the Volcano controller and CRDs.
> We need a way to check the presence of the Volcano controller and CRDs and install them when necessary, if the feature is enabled.

I think it's not required: if no volcano there, we just give a log/event when creating PodGroup;
the behaviour is similar to feature disabled.

liyinan926:
[...]

PR #544, opened 2019-07-04
Resource quota enforcement webhook

khogeland:
Presented for comment :eyes: Will update with documentation. I've based this PR on one of my other PR branches that this depends on, so here is the isolated diff: https://github.com/khogeland/spark-on-k8s-operator/compare/webhook-enhancements...khogeland:resource-quota-enforcement

This is a validating webhook for enforcing Resource Quotas at the SparkApplication level. This serves two main purposes: preventing deadlocks ([see this JIRA item](https://issues.apache.org/jira/browse/SPARK-24105?focusedCommentId=16797719&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16797719)) and providing early backpressure to users who are hitting their quota limits. 

It operates in basically the same way that the k8s pod-level quota controller does, using an informer to keep a running tally of (expected) resource usage in the background, with the validating webhook comparing the requested resources to the most recent usage. It calculates the expected usage for a given SparkApplication using the same logic that Spark uses internally, and also counts non-Spark pods in each namespace to allow for simple mixed workloads.

FWIW, this has been tested in several high-throughput clusters (although is unfortunately disabled due to a bug in the k8s version we're on).

PR #567, opened 2019-08-13
Support volcano podgroup

TommyLike:
The basic idea would be create podgroup for executor pods when:
1. the feature is enabled when application starts up.
2. there is podgroup crd in this cluster.
3. the spark application app is configured to use `volcano` for executor pods.

And these below are the additional things will be handled during sync spark application
1. create podgroup for the executor pods.
2. update executor pod's annotation to include "scheduling.k8s.io/group-name": "podgroup-name"

For #550 

@liyinan926 @k82cn please leave your comments on this.


main.go
```diff
@@ -50,23 +51,24 @@ import (
...
+	webhookCertDir          = flag.String("webhook-cert-dir", "/etc/webhook-certs", "The directory where x509 certificate and key files are stored.")
+	webhookSvcNamespace     = flag.String("webhook-svc-namespace", "spark-operator", "The namespace of the Service for the webhook server.")
+	webhookSvcName          = flag.String("webhook-svc-name", "spark-webhook", "The name of the Service for the webhook server.")
+	webhookPort             = flag.Int("webhook-port", 8080, "Service port of the webhook server.")
+	enableMetrics           = flag.Bool("enable-metrics", false, "Whether to enable the metrics endpoint.")
+	metricsPort             = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
+	metricsEndpoint         = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
+	metricsPrefix           = flag.String("metrics-prefix", "", "Prefix for the metrics.")
+	ingressUrlFormat        = flag.String("ingress-url-format", "", "Ingress URL format.")
+	enableVolcanoScheduling = flag.Bool("enable-volcano-scheduling", false, "Whether enable scheduling executor pods via volcano.")
```
k82cn: prefer `--enable-batch-scheduler`, so spark-operator can integrate with other scheduler later.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -526,6 +526,10 @@ type GPUSpec struct {
 func (s *SparkApplication) PrometheusMonitoringEnabled() bool {
 	return s.Spec.Monitoring != nil && s.Spec.Monitoring.Prometheus != nil
 }
+// VolcanoSchedulingRequired returns true if these executor pods are scheduled by volcano
+func (s *SparkApplication) VolcanoSchedulingRequired() bool {
+	return s.Spec.Executor.SchedulerName != nil && *(s.Spec.Executor.SchedulerName) == "volcano"
```
k82cn: maybe we can change `--enable-volcano-scheduling` to `--batch-scheduler-name`, so we can check executor's schedulerName against batch scheduler name.

pkg/controller/sparkapplication/controller.go
```diff
@@ -672,9 +683,54 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 			}
 		}
 	}
+
+	//Use PodGroup to schedule spark application executor pods if required
+	if c.volcanoClient != nil && app.VolcanoSchedulingRequired() {
+		if app.Spec.Executor.Annotations == nil {
+			app.Spec.Executor.Annotations = make(map[string]string)
+		}
+		if _, ok := app.Spec.Executor.Annotations["scheduling.k8s.io/group-name"]; !ok {
```
k82cn: I'm thinking to also include driver into PodGroup.

liyinan926: The problem is the executors will be created only after the driver is started. Does `PodGroup` works in this case?

liyinan926: Please define `scheduling.k8s.io/group-name` as a constant.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
liyinan926: That's a good point.

pkg/controller/sparkapplication/controller.go
```diff
@@ -672,9 +683,54 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
...
+			if pgName := c.createPodGroupIfNeeded(app); pgName != "" {
+				app.Spec.Executor.Annotations["scheduling.k8s.io/group-name"] = pgName
+			}
+		}
+	}
 	return app
 }
 
+func (c *Controller) createPodGroupIfNeeded(app *v1beta1.SparkApplication) string {
```
liyinan926: How about return a `(string, error)` instead and change `glog.Warningf` below to `return "", fmt.Errorf(...)`, and then let's the caller handle the error?

pkg/controller/sparkapplication/controller.go
```diff
@@ -672,9 +683,54 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
...
+		Spec: v1alpha2.PodGroupSpec{
+			MinMember: *app.Spec.Executor.Instances,
+		},
+	}
+
+	if pg, err := c.volcanoClient.SchedulingV1alpha2().PodGroups(app.Namespace).Create(&podGroup); err != nil {
+		glog.Warningf(
+			"Unable to create PodGroup with error: %s. Abandon schedule executor pods via volcano.", err)
+		return ""
+	} else {
```
liyinan926: `else` is not needed.

pkg/controller/sparkapplication/controller.go
k82cn: yes, podgroup can help on that: set `minNumber` to `1` to start driver, and set `minResoruce` to other value, e,g. 1 driver with 3 executors; the admission of Volcano will only admit the `PodGroup`s that total of `minResource` is less than `cluster resource * threshold (1.2 by default)`. That'll help to reducing the chance that driver pod consume so many resources :)

k82cn: we already have this const in volcano, please help to reuse that const :)

TommyLike:
@liyinan926 @k82cn refactored most of the codes according to your comments, will follow up on testing and trivial fixes tomorrow :)

pkg/controller/sparkapplication/batchscheduler/factory.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"k8s.io/client-go/rest"
+
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/controller/sparkapplication/batchscheduler/interface"
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/controller/sparkapplication/batchscheduler/volcano"
+)
+
+type schedulerInitializeFunc func(config *rest.Config) schedulerinterface.BatchScheduler
+
+var manageMutex sync.Mutex
```
liyinan926: Put both into the same var block to tell that the map is guarded by the mutex, e.g.,:

```
var (
  mutex sync.Mutex
  schedulers map[string]schedulerInitializeFunc
)
```

main.go
```diff
@@ -112,12 +113,11 @@ func main() {
...
-	if *enableVolcanoScheduling {
-		var err error
-		volcanoClient, err = volcanoclient.NewForConfig(config)
-		if err != nil {
-			glog.Fatal(err)
+	var batchScheduler schedulerinterface.BatchScheduler
+	if *batchSchedulerName != "" {
+		batchScheduler = batchscheduler.InitializeBatchScheduler(*batchSchedulerName, config)
+		if batchScheduler == nil {
+			glog.Fatal("Batch scheduler %s failed to initialize.", batchSchedulerName)
```
liyinan926: Let's change the error message to `no batch scheduler named %s registered`.

pkg/controller/sparkapplication/batchscheduler/factory.go
```diff
@@ -0,0 +1,38 @@
...
+	registerBatchScheduler(volcano.GetPluginName(), volcano.New)
+}
+
+func registerBatchScheduler(name string, iniFunc schedulerInitializeFunc) {
+	manageMutex.Lock()
+	defer manageMutex.Unlock()
+	schedulerContainers[name] = iniFunc
+}
+
+func InitializeBatchScheduler(name string, config *rest.Config) schedulerinterface.BatchScheduler {
```
liyinan926: Let's call this `GetBatchScheduler`.

pkg/controller/sparkapplication/batchscheduler/interface/interface.go
```diff
@@ -0,0 +1,12 @@
+package schedulerinterface
+
+import "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
+
+type BatchScheduler interface {
+	Name() string
+
+	ShouldSchedule(app *v1beta1.SparkApplication) bool
+
+	OnSubmitSparkApplication(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error)
```
liyinan926: Let's rename this to `OnSparkApplicationSubmitted`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -683,52 +694,11 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
...
-	}
-
-	if pg, err := c.volcanoClient.SchedulingV1alpha2().PodGroups(app.Namespace).Create(&podGroup); err != nil {
-		glog.Warningf(
-			"Unable to create PodGroup with error: %s. Abandon schedule executor pods via volcano.", err)
-		return ""
-	} else {
-		return pg.Name
-	}
+func (c *Controller) scheduleViaBatchScheduler(app *v1beta1.SparkApplication) bool {
```
liyinan926: Let's rename this to `shouldDoBatchScheduling`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -650,6 +652,15 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 		return app
 	}
 
+	// Use batch scheduler to perform app submit tasks
+	if c.scheduleViaBatchScheduler(app) {
+		var err error
+		app, err = c.batchScheduler.OnSubmitSparkApplication(app)
+		if err != nil {
+			glog.Errorf("Failed to process batch scheduler OnSubmitSparkApplication with error %v", err)
```
liyinan926: Let's change the error message to `failed to apply batch scheduling to SparkApplication %s: %v`.

pkg/controller/sparkapplication/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if app.Spec.Mode == v1beta1.ClientMode {
+		return v.syncPodGroupForClientAPP(app)
+	} else if app.Spec.Mode == v1beta1.ClusterMode {
+		return v.syncPodGroupForClusterAPP(app)
+	}
+	return app, nil
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupForClientAPP(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
```
liyinan926: `syncPodGroupInClientMode`

pkg/controller/sparkapplication/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+	//We only care about the executor pods in client mode
+	if _, ok := app.Spec.Executor.Annotations[v1alpha2.GroupNameAnnotationKey]; !ok {
+		if err := v.syncPodGroup(app, *app.Spec.Executor.Instances); err == nil {
+			app.Spec.Executor.Annotations[v1alpha2.GroupNameAnnotationKey] = v.getAppPodGroupName(app)
+		}
+	}
+	return app, nil
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupForClusterAPP(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
```
liyinan926: `syncPodGroupInClusterMode`

pkg/controller/sparkapplication/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+func (v *VolcanoBatchScheduler) OnSparkDriverPodScheduled(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
+	// Update PodGroup MinMember if required
+	if app.Spec.Mode == v1beta1.ClusterMode {
+		return app, v.syncPodGroup(app, 1+*(app.Spec.Executor.Instances))
+	}
+	return app, nil
+}
+
+func (v *VolcanoBatchScheduler) getAppPodGroupName(app *v1beta1.SparkApplication) string {
+	return fmt.Sprintf("sparkapplication-%s-podgroup", app.Name)
```
liyinan926: Let's remove the prefix `sparkapplication` to reduce the chance of the name being too long.

pkg/controller/sparkapplication/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+	if _, ok := app.Spec.Executor.Annotations[v1alpha2.GroupNameAnnotationKey]; !ok {
+		if err := v.syncPodGroup(app, *app.Spec.Executor.Instances); err == nil {
+			app.Spec.Executor.Annotations[v1alpha2.GroupNameAnnotationKey] = v.getAppPodGroupName(app)
+		}
+	}
+	return app, nil
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupForClusterAPP(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
+	//We need both mark Driver and Executor when submitting
```
liyinan926: The comment should explain why the group size is initially 1 and how it's adjusted once the driver pod gets scheduled.

pkg/controller/sparkapplication/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,162 @@
...
+			"Unable to sync PodGroup with error: %s. Abandon schedule pods via volcano.", err)
+	}
+	return err
+}
+
+func New(config *rest.Config) schedulerinterface.BatchScheduler {
+
+	vkClient, err := volcanoclient.NewForConfig(config)
+	if err != nil {
+		glog.Errorf("Unable to initialize volcano client with error %v", err)
```
liyinan926: We typically use lower case for error messages and error messages like `failed to ...: %v`. In this case `failed to initialize the valcano client: %v`.

pkg/controller/sparkapplication/controller.go
liyinan926: Other than that, will Valcano guarantee that it will only admit a `PodGroup` when there exists enough resources in the cluster to run the `minSize` number of pods in the group? If not, the value of creating a `PodGroup` that includes the driver pod is much less. And will Valcano reserve the needed resources to run the pods?

k82cn: > Other than that, will Valcano guarantee that it will only admit a PodGroup when there exists enough resources in the cluster to run the `minSize` number of pods in the group? 

Almost yes :) If `total minSize` is larger than `cluster resource * threshold`, both driver and executor pod will not admit. Threshold was introduced because admission does not mean schedulable, e.g. node selector; it give a balance between admission and schedulable.

> And will Valcano reserve the needed resources to run the pods?

hm.... we don-t reserve resources but we will schedule next job until `minResources` was fitted.

TommyLike:
@liyinan926 @k82cn , Now user can configure batch scheduler plugin for spark operator.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,184 @@
...
+			return nil, err
+		}
+	}
+	return app, nil
+}
+
+func (v *VolcanoBatchScheduler) OnSparkDriverPodScheduled(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
+	// Update PodGroup MinMember if required
+	if app.Spec.Mode == v1beta1.ClusterMode {
+		return app, v.syncPodGroup(app, 1+*(app.Spec.Executor.Instances))
```
k82cn: why set this? btw, where do we set `minResources`?

TommyLike: hmm, catch up with your conversion now.

TommyLike: ok, this is what I infer from your conversation, 
1. For client Mode: We will schedule jobs when  `pod resource > minMember and free resource > resource(executor * executerInstance ) `
2. For driver Mode: We will schedule jobs when  `pod resource > 1 (for driver)  and free resource >resource(executor * executerInstance + driver)`

@k82cn  Do I get the correct conclusion?

k82cn:
LGTM overall :)

pkg/controller/sparkapplication/controller.go
```diff
@@ -639,6 +644,16 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 		return app
 	}
 
+	// Use batch scheduler to perform app submit tasks
```
liyinan926: Why not perform the optional batch scheduling before the application gets submitted? This way, we don't need to rely on an update to the `SparkApplicationSpec` to trigger a re-submission to make the batch scheduling effective.

liyinan926: If batch scheduling needs to be applied after the application gets submitted, but before the pods get scheduled, the webhook is the right place to apply the batch scheduling settings.

TommyLike: Based on the testing, it would be a long process before actual creating the pods, the pods can be patched with the correct annotation. But as you said. there could be a competition. 
what about?
1. Invoking spark batch scheduler before submitting and update the spark app.
or 2. Adding this logic into webhook which doesn't sound a proper place to me.

TommyLike: This patch has been updated to the direction of option 1.

liyinan926: I like option 2. With option 1, you need to make sure that wherever `submitSparkApplication` gets called, the logic of applying batch scheduling gets added. With option 2, there's no need to change the `SparkApplicationSpec`, and the webhook is a central place for applying batch scheduling mutations directly to pods.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	return newApp, nil
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupInClientMode(app *v1beta1.SparkApplication) (*v1beta1.SparkApplication, error) {
+	//We only care about the executor pods in client mode
+	newApp := app.DeepCopy()
+	if _, ok := newApp.Spec.Executor.Annotations[v1alpha2.GroupNameAnnotationKey]; !ok {
+		//Only executor resource will be considered.
+		if err := v.syncPodGroup(newApp, *newApp.Spec.Executor.Instances, getExecutorRequestResource(app)); err == nil {
```
khogeland: Shouldn't `MinMember` be 1, not the total number of executors? If you require they all be scheduled, you can't start the application early, increasing overall execution time. This also doesn't allow fair scheduling between applications.

docs/volcano.md
```diff
@@ -0,0 +1,75 @@
...
+  volumes:
+    - name: "test-volume"
+      hostPath:
+        path: "/tmp"
+        type: Directory
+  driver:
+    cores: 0.1
+    coreLimit: "200m"
+    memory: "512m"
+    schedulerName: "volcano"     #Note: the scheduler name must be specified with `volcano`
```
khogeland: How do the pods themselves get set to use the Volcano scheduler? Is there a matching Spark PR, a mutating webhook, or something I'm missing?

khogeland:
This appears to serve a similar purpose to https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/544, mostly in guaranteeing that all components of an application can be scheduled. What are the benefits over that PR, that would justify the complex extra dependency?

Volcano seems really interesting to me for its gang scheduling capabilities, but as long as the driver requests its own executors, those can't be taken advantage of :disappointed:.

k82cn:
> This appears to serve a similar purpose to #544, mostly in guaranteeing that all components of an application can be scheduled. What are the benefits over that PR, that would justify the complex extra dependency?

The major differentiation is that: this PR will dependent on scheduler to share resource between spark application instead of Quota, because 1. admitted by Quota does not mean scheduled 2. Quota is hard to support fair-share, priority and other algorithm.

> Volcano seems really interesting to me for its gang scheduling capabilities, 

gang scheduling is only one plugins of Volcano, please refer to https://github.com/volcano-sh/volcano/tree/master/docs/samples/kubecon-2019-china for others features. And we're also working several other features https://github.com/volcano-sh/volcano/issues/244 :)

> but as long as the driver requests its own executors, those can't be taken advantage of 😞.

We did not only dependent on gang scheduling in this pr, it also combine other features, e.g. DelayPodCreation, Fair-share (DRF), preemption and so on :) We already have a test data to show the advantage of Volcano, will share it later.

pkg/batchscheduler/volcano/volcano_scheduler.go
k82cn: +1, 1 is better for fair-share :)

docs/volcano.md
TommyLike: We already have logic to patch the scheduler name into pod's scheduler attribute,  I uploaded this file by mistake, The document will be uploaded when the codes get merged first.

pkg/controller/sparkapplication/controller.go
TommyLike: @liyinan926  thanks, patch updated heading to the option 2

pkg/batchscheduler/volcano/volcano_scheduler.go
TommyLike: Sure.

main.go
```diff
@@ -70,6 +72,7 @@ var (
 	leaderElectionLeaseDuration = flag.Duration("leader-election-lease-duration", 15*time.Second, "Leader election lease duration.")
 	leaderElectionRenewDeadline = flag.Duration("leader-election-renew-deadline", 14*time.Second, "Leader election renew deadline.")
 	leaderElectionRetryPeriod   = flag.Duration("leader-election-retry-period", 4*time.Second, "Leader election retry period.")
+	batchSchedulerName          = flag.String("batchSchedulerName", "", "Use specified scheduler for pods' batch scheduling.")
```
liyinan926: Name of the flag should not use camel case. Please change it to `batch-scheduler-name`.

pkg/batchscheduler/interface/interface.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/util"
+)
+
+type BatchScheduler interface {
+	Name() string
+
+	ShouldSchedule(app *v1beta1.SparkApplication) bool
+	PatchApplicationPod(pod *corev1.Pod, app *v1beta1.SparkApplication) []util.PatchOperation
+	BeforeSubmitSparkApplication(app *v1beta1.SparkApplication) error
```
liyinan926: Let's rename this to `BeforeApplicationSubmission`.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+	if len(pod.Annotations) != 0 {
+		if _, ok := pod.Annotations[v1alpha2.GroupNameAnnotationKey]; ok {
+			return nil
+		}
+		pod.Annotations[v1alpha2.GroupNameAnnotationKey] = v.getAppPodGroupName(app)
+		value = pod.Annotations
+	} else {
+		value = map[string]string{v1alpha2.GroupNameAnnotationKey: v.getAppPodGroupName(app)}
+	}
+	return &util.PatchOperation{Op: "replace", Path: path, Value: value}
```
liyinan926: Why `replace`?

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+
+		_, err = v.volcanoClient.SchedulingV1alpha2().PodGroups(app.Namespace).Create(&podGroup)
+	} else {
+		if pg.Spec.MinMember != size {
+			pg.Spec.MinMember = size
+			_, err = v.volcanoClient.SchedulingV1alpha2().PodGroups(app.Namespace).Update(pg)
+		}
+	}
+	if err != nil {
+		glog.Errorf(
```
liyinan926: Instead of logging, simply return a `fmt.Errorf`.

pkg/client/clientset/versioned/clientset.go
```diff
@@ -32,8 +32,6 @@ type Interface interface {
 	Discovery() discovery.DiscoveryInterface
 	SparkoperatorV1alpha1() sparkoperatorv1alpha1.SparkoperatorV1alpha1Interface
 	SparkoperatorV1beta1() sparkoperatorv1beta1.SparkoperatorV1beta1Interface
-	// Deprecated: please explicitly pick a version if possible.
-	Sparkoperator() sparkoperatorv1beta1.SparkoperatorV1beta1Interface
```
liyinan926: It looks like you are using an older version of the code generator.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+			"Unable to sync PodGroup with error: %s. Abandon schedule pods via volcano.", err)
+	}
+	return err
+}
+
+func New(config *rest.Config, webhookEnabled bool) schedulerinterface.BatchScheduler {
+
+	if !webhookEnabled {
+		glog.Error("failed to initialize volcano client, webhook enable is required")
```
liyinan926: `failed to initialize the volcano scheduler as it requires the webhook to be enabled`

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+	}
+	if err != nil {
+		glog.Errorf(
+			"Unable to sync PodGroup with error: %s. Abandon schedule pods via volcano.", err)
+	}
+	return err
+}
+
+func New(config *rest.Config, webhookEnabled bool) schedulerinterface.BatchScheduler {
```
liyinan926: Instead of logging errors, let this method return `(schedulerinterface.BatchScheduler, error)`.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+		if value, err := resource.ParseQuantity(*app.Spec.Executor.Memory); err == nil {
+			minResource[corev1.ResourceMemory] = value
+		}
+	}
+	if app.Spec.Executor.MemoryOverhead != nil {
+		if value, err := resource.ParseQuantity(*app.Spec.Executor.MemoryOverhead); err == nil {
+			if existing, ok := minResource[corev1.ResourceMemory]; ok {
+				existing.Add(value)
+				minResource[corev1.ResourceMemory] = existing
+			} else {
```
liyinan926: If a base memory request doesn't exist, I think we should ignore the overhead.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+				minResource[corev1.ResourceMemory] = value
+			}
+		}
+	}
+
+	return minResource
+}
+
+func sumResourceList(list []corev1.ResourceList) corev1.ResourceList {
+
```
liyinan926: For consistency in code style, please remove such kind of empty lines.

pkg/controller/sparkapplication/controller.go
```diff
@@ -617,6 +622,15 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 		return app
 	}
 
+	// Use batch scheduler to perform scheduling task before submitting.
+	if c.shouldDoBatchScheduling(app) {
+		err := c.batchScheduler.BeforeSubmitSparkApplication(app)
+		if err != nil {
+			glog.Errorf("Failed to process batch scheduler BeforeSubmitSparkApplication with error %v", err)
```
liyinan926: Error messages should start with a lower-case letter with no period at the end.

liyinan926:
I have the same concern with @khogeland: we cannot leverage the gang scheduling support in Volcano, which to me is one of the most important features of Volcano.

@k82cn can you share some stats on how Volcano improves Spark app scheduling, even without gang scheduling?

k82cn:
> can you share some stats on how Volcano improves Spark app scheduling, even without gang scheduling?

Some major features for Spark application:

1. DelayPodCreation/Enqueue: in Spark Application, we set `minMember` to `1` which disable gang-scheduling; when we set `minResources`, Volcano will en-queue part of SparkApplication to make sure the drivers will not consume all resources: `total minResources` < `cluster resources` * threshold (1.2 by default)
1. DRF: dominant resource fairness algorithm, which will help spark application share resource with each other
1. Queue: similar to Queue in YARN, Role in Mesos
1. Preemption/Reclaim: the preemption/reclaim in Volcano are built on plugins, which means also support fair-share based preemption, e.g. DRF
1. Job priority: Spark Application level priority

And several features are on the way, e.g. [Fair-share scheduling of namespace cross queues](https://github.com/volcano-sh/volcano/issues/147)

pkg/batchscheduler/volcano/volcano_scheduler.go
TommyLike: The annotation has been replaced by a whole collection.

TommyLike:
@liyinan926  done!

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupInClientMode(app *v1beta1.SparkApplication) error {
+	//Only executor resource will be considered.
+	return v.syncPodGroup(app, 1, getExecutorRequestResource(app))
+}
+
+func (v *VolcanoBatchScheduler) syncPodGroupInClusterMode(app *v1beta1.SparkApplication) error {
+	totalResource := sumResourceList([]corev1.ResourceList{getExecutorRequestResource(app), getDriverRequestResource(app)})
+	//NOTE: In cluster mode, the initial size of PodGroup is set to 1 in order to schedule driver pod first.
```
liyinan926: Are the driver pod and executors pods considered to be in the same group? If so, should the minSize of the group be 2 (1 driver pod plus one executor pod)?

pkg/batchscheduler/interface/interface.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
+	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/util"
+)
+
+type BatchScheduler interface {
+	Name() string
+
+	ShouldSchedule(app *v1beta1.SparkApplication) bool
+	PatchApplicationPod(pod *corev1.Pod, app *v1beta1.SparkApplication) []util.PatchOperation
```
liyinan926: I just realized that with the recent changes (only setting `minSize` before submitting the application), we can do everything before submitting the application without triggering an application rerun. To add the annotation without triggering a rerun, simply put it into the `Annotations` field of `SparkApplicationSpec` and use the modified `SparkApplication` to submit the application. This change to `SparkApplicationSpec` won't be persisted to etcd as we only update the status section of a `SparkApplication`. We use a similar approach when applying Prometheus monitoring configs in `submitSparkApplication`. So let's remove this `PatchApplicationPod` method and changes to the webhook.

liyinan926: Let's also rename `BeforeApplicationSubmission` to `DoBatchSchedulingOnSubmission`.

liyinan926:
Thanks for this PR! I'm wondering what's your plan for follow-ups after this PR? I am thinking of queue support being on thing. We might also need some API change, e.g., adding a new optional field for specifying the job queue name.

TommyLike:
> Thanks for this PR! I'm wondering what's your plan for follow-ups after this PR? I am thinking of queue support being on thing. We might also need some API change, e.g., adding a new optional field for specifying the job queue name.

I have thought about this, we may add a new map attribute for the spark application, something like `BatchSchedulerOption` in the future.

pkg/batchscheduler/volcano/volcano_scheduler.go
TommyLike: The Minresource stands for the Min pods that volcano will start the podgroup. if it sets to 2, the driver will never been started, since there is not coming any executor pods.

k82cn:
>  I'm wondering what's your plan for follow-ups after this PR? I am thinking of queue support being on thing. We might also need some API change, e.g., adding a new optional field for specifying the job queue name.

We'll continue to maintain the integration with Volcano, and follow Spark-Operator's style to add new filed if necessary. Prefer to open issue to discuss when introducing new field.

TommyLike:
@liyinan926 done

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -0,0 +1,282 @@
...
+			pg.Spec.MinMember = size
+			_, err = v.volcanoClient.SchedulingV1alpha2().PodGroups(app.Namespace).Update(pg)
+		}
+	}
+	return fmt.Errorf("failed to sync PodGroup with error: %s. Abandon schedule pods via volcano", err)
+}
+
+func New(config *rest.Config, webhookEnabled bool) (schedulerinterface.BatchScheduler, error) {
+
+	if !webhookEnabled {
```
liyinan926: Looks like we don't need this check and `webhookEnabled` any more.

TommyLike: How about the scheduler name for the pods?

liyinan926: Why do we need to consider the scheduler name here? I meant the check on `webhookEnabled` is no longer needed since we don't use the webhook to apply the batch scheduling mutations, so this check doesn't make sense any more.

TommyLike: Volcano only focus on the pod that configured with schedulername `volcano`. and webhook is responsible for passing that value from SparkApplication to the driver/executor pods.

liyinan926: Adding annotations is done without the webhook. Spark on k8s has Spark config options for adding driver/executor annotations.

liyinan926: Got it. This check makes sense. However, I think the check applies to all batch schedulers and should be done in `main` to make sure when `batch-scheduler-name` is not empty, `enable-webhook` is set to `true`.

liyinan926:
I also have another concern, although we don't need to solve it in this PR. So we use a command-line flag to select/enable a batch scheduler from the set of pre-registered schedulers. Individual applications can also specify a scheduler for the driver and/or executors. So essentially, we are limiting the operator to only allow one batch scheduler at runtime, even though there might be more than one registered.

1. Remove the command-line flag.
2. Use the scheduler name in the application specification to decide which scheduler to use, and in case no scheduler with the given name is registered, fail the application.

TommyLike:
> I also have another concern, although we don't necessarily need to resolve in this PR. So we use a command-line flag to select/enable a batch scheduler from the set of pre-registered schedulers. However, individual applications can also specify a scheduler for the driver and/or executors. Essentially, we are limiting the operator to only allow one batch scheduler at runtime, even though there might be more than one registered. Below is what I propose.
> 
> 1. Remove the command-line flag. Or allow the flag to select a set of batch schedulers.
> 2. Use the scheduler name in the application specification to decide which scheduler to use, and in case no scheduler with the given name is registered, skip applying batch scheduling configuration.

Will do in another PR.

TommyLike:
@liyinan926, thanks, all of the following up will be tracked on #583

pkg/controller/sparkapplication/controller.go
```diff
@@ -675,6 +691,10 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 	return app
 }
 
+func (c *Controller) shouldDoBatchScheduling(app *v1beta1.SparkApplication) bool {
+	return c.batchScheduler != nil && c.batchScheduler.ShouldSchedule(app)
```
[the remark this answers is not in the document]
TommyLike: @khogeland batchScheduler it's an interface not a struct.

ggaaooppeenngg:
Any document about this PR?

PR #573, opened 2019-08-16
Updated metrics prefix in documentation

ramyarajasekaran:


liyinan926:
I didn't realize you were talking about this prefix. We cannot change the prefix for two reasons: 1) the workqueue metrics are standard ones and the current prefix is fine, and 2) the prefix might be used and replied upon by users.

ramyarajasekaran:
Oh okay. Sounds good!

PR #576, opened 2019-08-21
sparkapplication status will report succeed/fail after driver contain…

jinxingwang:
…er in spark driver has terminated even there is long running sidecar container in spark driver pod

### DIFF
--- a/pkg/controller/sparkapplication/sparkapp_util.go
+++ b/pkg/controller/sparkapplication/sparkapp_util.go
@@ -95,8 +95,14 @@ func driverStateToApplicationState(podStatus apiv1.PodStatus) v1beta1.Applicatio
 		// TODO: upcoming Kubernenetes feature will make this code redundant
 		// https://github.com/kubernetes/enhancements/issues/753
 		for _, c := range podStatus.ContainerStatuses {
-			if c.Name == config.SparkDriverContainerName && c.State.Terminated != nil && c.State.Terminated.ExitCode == 0 {
-				return v1beta1.SucceedingState
+			if c.Name == config.SparkDriverContainerName {
+				if c.State.Terminated != nil {
+					if c.State.Terminated.ExitCode == 0 {
+						return v1beta1.SucceedingState
+					}
+					return v1beta1.FailingState
+				}
+				break
 			}
 		}
 		return v1beta1.RunningState
--- a/pkg/controller/sparkapplication/sparkapp_util.go
+++ b/pkg/controller/sparkapplication/sparkapp_util.go
@@ -95,15 +95,8 @@ func driverStateToApplicationState(podStatus apiv1.PodStatus) v1beta1.Applicatio
 		// TODO: upcoming Kubernenetes feature will make this code redundant
 		// https://github.com/kubernetes/enhancements/issues/753
 		for _, c := range podStatus.ContainerStatuses {
-			if (c.Name == config.SparkDriverContainerName) {
-				if c.State.Terminated != nil {
-					if (c.State.Terminated.ExitCode == 0) {
-						return v1beta1.SucceedingState
-					} else {
-						return v1beta1.FailingState
-					}
-				}
-				break;
+			if c.Name == config.SparkDriverContainerName && c.State.Terminated != nil && c.State.Terminated.ExitCode == 0 {
+				return v1beta1.SucceedingState
 			}
 		}
 		return v1beta1.RunningState
--- a/pkg/controller/sparkapplication/sparkapp_util.go
+++ b/pkg/controller/sparkapplication/sparkapp_util.go
@@ -87,11 +87,25 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
 	return executorState == v1beta1.ExecutorCompletedState || executorState == v1beta1.ExecutorFailedState
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -87,11 +87,25 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
...
-	switch podPhase {
+func driverStateToApplicationState(podStatus apiv1.PodStatus) v1beta1.ApplicationStateType {
+	switch podStatus.Phase {
 	case apiv1.PodPending:
 		return v1beta1.SubmittedState
 	case apiv1.PodRunning:
+		// TODO: upcoming Kubernenetes feature will make this code redundant
+		// https://github.com/kubernetes/enhancements/issues/753
+		for _, c := range podStatus.ContainerStatuses {
+			if (c.Name == config.SparkDriverContainerName) {
```
liyinan926: There's no need to put parenthesis around a if condition in Golang.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -87,11 +87,25 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
...
+				if c.State.Terminated != nil {
```
liyinan926: Looks like you can combine this condition with the one above into one:

```
if c.Name == config.SparkDriverContainerName && c.State.Terminated != nil {
    ...
}
```

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -87,11 +87,25 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
...
+					if (c.State.Terminated.ExitCode == 0) {
+						return v1beta1.SucceedingState
+					} else {
+						return v1beta1.FailingState
+					}
+				}
+				break;
```
liyinan926: No need to add a semicolon at the end. Actually the `break` is not needed.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -87,11 +87,25 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
...
 		return v1beta1.SubmittedState
 	case apiv1.PodRunning:
+		// TODO: upcoming Kubernenetes feature will make this code redundant
+		// https://github.com/kubernetes/enhancements/issues/753
+		for _, c := range podStatus.ContainerStatuses {
+			if (c.Name == config.SparkDriverContainerName) {
+				if c.State.Terminated != nil {
+					if (c.State.Terminated.ExitCode == 0) {
+						return v1beta1.SucceedingState
+					} else {
```
liyinan926: The `else` is not needed here.

jinxingwang:
@liyinan926 addressed your comments. please review again. thanks!

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -87,11 +87,18 @@ func isExecutorTerminated(executorState v1beta1.ExecutorState) bool {
...
+func driverStateToApplicationState(podStatus apiv1.PodStatus) v1beta1.ApplicationStateType {
+	switch podStatus.Phase {
 	case apiv1.PodPending:
 		return v1beta1.SubmittedState
 	case apiv1.PodRunning:
+		// TODO: upcoming Kubernenetes feature will make this code redundant
+		// https://github.com/kubernetes/enhancements/issues/753
+		for _, c := range podStatus.ContainerStatuses {
+			if c.Name == config.SparkDriverContainerName && c.State.Terminated != nil && c.State.Terminated.ExitCode == 0 {
+				return v1beta1.SucceedingState
```
liyinan926: You still need to check the case of `c.State.Terminated.ExitCode != 0` and return `v1beta1.FailingState` in that case.

jinxingwang: If the Terminated.ExitCode != 0 on one of the containers, ist the pod will mark as failed.

liyinan926: The `PodPhase` stays `Running` if the sidecar container is still running.

jinxingwang: Thanks for pointing it out.

pkg/controller/sparkapplication/sparkapp_util.go
jinxingwang: I think this break is needed because I only want to look the driver container after I find driver container, I don't need to continue for loop.

liyinan926: We always add sidecar containers after the main Spark container so you really only need to check the status of the first container if there's more than one container. I would change the for loop to the following:

```
if len(pod.Status.ContainerStatuses) > 1 {
  state := pod.Status.ContainerStatuses[0].State
  if state.Terminated == nil {
    return v1beta1.RunningState
  }
  if state.Terminated.ExitCode == 0 {
    return v1beta1.SucceedingState
  }
  return v1beta1.FailingState
}
```

jinxingwang: I see. I think I will keep the name-checking approach. since people can write their webhook that mess the container orders. this way feel safer. thanks

PR #578, opened 2019-08-22
Generate CRD specs, bump to v1beta2

khogeland:
The current, manually-defined CRD validation is really unsafe, which we just learned of due to an outage:

- Since k8s will ignore unknown fields, if all fields aren't listed, values could be of the wrong type
- Fields that are listed but are missing type annotations can be replaced by values of the wrong type
- Deserialization errors will apparently break the Lister (and therefore the entire controller) because it deserializes the entire SparkApplicationList at once.

This PR replaces the manually written CRD specs with ones generated by the [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen) tool (part of Kubebuilder). It also gets rid of CRD self-registration, because there is no CRD Go code generated (I explored bundling the generated YAML, but it would be a hack and doesn't seem that useful).

Also bumped the version because this is a backwards-incompatible change:
- I changed the `Cores` field from `float32` to `int32`, because Spark [parses it as an integer](https://github.com/apache/spark/blob/cbad616d4cb0c58993a88df14b5e30778c7f7e85/core/src/main/scala/org/apache/spark/deploy/rest/StandaloneRestServer.scala#L176), and controller-gen [doesn't support floats](https://github.com/kubernetes-sigs/controller-tools/issues/245)
- I assume that the stricter validation, while not being an API change, may reject some applications that previously worked

Existing validation has been migrated over, with the exception of the `metadata.name` field length, because I can't put annotations on a field from k8s. I'd like to figure out how to get that validation back, currently it'll fail at spark-submit.

I'm pretty sure I missed a few things when bumping the version, and some docs are now out of date, so I'll clean that up asap. Sorry I'm dumping another large PR on you without warning!


liyinan926:
Please rebase. I will need to cut a final release for `v1beta1` before being able to review and merge this one.

liyinan926:
Can you rebase this?

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -0,0 +1,569 @@
...
+	// +kubebuilder:validation:Minimum=1
+	MaxSimultaneousDownloads *int32 `json:"maxSimultaneousDownloads,omitempty"`
+}
+
+// SparkPodSpec defines common things that can be customized for a Spark driver or executor pod.
+// TODO: investigate if we should use v1.PodSpec and limit what can be set instead.
+type SparkPodSpec struct {
+	// Cores is the number of CPU cores to request for the pod.
+	// Optional.
+	// +kubebuilder:validation:Minimum=0
```
liyinan926: Can we do `+kubebuilder:validation:Minimum=1`?

khogeland: Ah, right, `ExclusiveMinimum` isn't necessary now that this is an integer. 👍

README.md
```diff
@@ -14,7 +14,7 @@
...
 
-**If you are currently using the `v1alpha1` version of the APIs in your manifests, please update them to use the `v1beta1` version by changing `apiVersion: "sparkoperator.k8s.io/v1alpha1"` to `apiVersion: "sparkoperator.k8s.io/v1beta1"`. You will also need to delete the `v1alpha1` version of the CustomResourceDefinitions named `sparkapplications.sparkoperator.k8s.io` and `scheduledsparkapplications.sparkoperator.k8s.io`, and replace them with the `v1beta1` version either by installing the latest version of the operator or by running `kubectl create -f manifest/spark-operator-crds.yaml`.**
+**If you are currently using the `v1alpha1` or `v1beta1` version of the APIs in your manifests, please update them to use the `v1beta2` version by changing `apiVersion: "sparkoperator.k8s.io/<version>"` to `apiVersion: "sparkoperator.k8s.io/v1beta1"`. You will also need to delete the `previous` version of the CustomResourceDefinitions named `sparkapplications.sparkoperator.k8s.io` and `scheduledsparkapplications.sparkoperator.k8s.io`, and replace them with the `v1beta2` version either by installing the latest version of the operator or by running `kubectl create -f manifest/crds`.**
```
liyinan926: `to v1beta2`.

liyinan926:
@khogeland have you been running this version in production?

khogeland:
We haven't, it seems too far-reaching (i.e. too likely to turn into a merge conflict nightmare) to maintain in an internal fork.

liyinan926:
All right. Please fix the remaining typo and I will merge this.

liyinan926:
Any more commits coming? If no, will merge soon.

khogeland:
Nope, that should be the last of them.

Issue #583, opened 2019-08-26
Follow up for batch scheduler support changes.


TommyLike:
There are some leftovers from patch #567, including.

- [x] 1. configure update for supporting different schedulers. see [comment](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/567#issuecomment-524406290)
- [x] 2. add document for how to use and configure
- [x] 3. add helm support for scheduler.
- [ ] 4. More testing.

liyinan926:
@TommyLike do you have more PRs coming for this? If so, I will wait for those before cutting a new release for `v1beta2`.

TommyLike:
@liyinan926 oh, please do the release, I will add them later.

liyinan926:
@TommyLike are we pretty much done on Volcano integration?

TommyLike:
@liyinan926  it looks good to me now, will add more pr if there is any issues/feature requests in the future. But those would not bound to this issue, so will close this.

PR #585, opened 2019-08-27
Support batch scheduler manager

TommyLike:
For #583, Now user can configure different batch schedulers and specify using different batch scheduler in Spark Application.


pkg/batchscheduler/scheduler_manager.go
```diff
@@ -0,0 +1,84 @@
...
+
+func GetRegisteredNames() []string {
+	var pluginNames []string
+	for key := range schedulerContainers {
+		pluginNames = append(pluginNames, key)
+	}
+	return pluginNames
+}
+
+type SchedulerManagerImpl struct {
```
liyinan926: I don't think we really need an interface for `SchedulerManager`. Also `schedulerContainers` can be a field of `SchedulerManager` instead of a package-level thing, and `GetRegisteredNames` can be a method of `SchedulerManager`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -691,8 +692,16 @@ func (c *Controller) submitSparkApplication(app *v1beta1.SparkApplication) *v1be
 	return app
 }
 
-func (c *Controller) shouldDoBatchScheduling(app *v1beta1.SparkApplication) bool {
-	return c.batchScheduler != nil && c.batchScheduler.ShouldSchedule(app)
+func (c *Controller) shouldDoBatchScheduling(app *v1beta1.SparkApplication) (bool, schedulerinterface.BatchScheduler) {
+	if c.batchSchedulerMgr == nil || app.Spec.BatchScheduler == nil || *app.Spec.BatchScheduler == "" {
```
liyinan926: What if `app.Spec.BatchScheduler` is empty, but `app.Spec.Driver.SchedulerName` or `app.Spec.Executor.SchedulerName` is set?

pkg/batchscheduler/scheduler_manager.go
```diff
@@ -0,0 +1,84 @@
...
+	batch.Lock()
+	defer batch.Unlock()
+
+	if plugin, existed := batch.plugins[schedulerName]; existed && plugin != nil {
+		return plugin, nil
+	} else if existed && plugin == nil {
+		return nil, fmt.Errorf(
+			"failed to get scheduler plugin %s, previous initialization has failed", schedulerName)
+	} else {
+		if plugin, err := iniFunc(batch.config); err != nil {
```
liyinan926: I don't think we need to add an entry in this case. We can use the absence of a scheduler as the signal instead.

pkg/controller/sparkapplication/controller.go
TommyLike: Since we have `BatchScheduler` supported in SparkApplication's Spec. I personally would prefer to have only one entrance for end user to trigger batch scheduling. and for the configuration of driver and executor scheduler, they still can utilize them to make SparkApplication scheduled via other scheduler.

pkg/batchscheduler/scheduler_manager.go
TommyLike: There are three different cases.
1. scheduler exists. //Normal case
2. scheduler exists but it's none. //Abnormal case. We have this scheduler supported, but failed to initialize before.
3. scheduler not exists, //Normal case, we will initialize this now.
Which one are you saying we don't need to consider about?

liyinan926: I meant we don't really treat the case that the scheduler failed to be initialized differently than the case that the scheduler simply is not registered. In both cases, the scheduler doesn't exist in the registry.

TommyLike: Just want to save the time of creating scheduler repeatly.

liyinan926: OK, that makes sense.

pkg/batchscheduler/scheduler_manager.go
```diff
@@ -0,0 +1,84 @@
...
+			batch.plugins[schedulerName] = nil
+			return nil, err
+		} else {
```
liyinan926: This `else` is not needed.

TommyLike: No, the plugin variable is defined in if statement.

liyinan926: Oh, you are right.

TommyLike:
@k82cn, please take a review :)

pkg/batchscheduler/scheduler_manager.go
k82cn: +1, it seems a `Factory` is enough :)

k82cn:
LGTM overall :)

pkg/batchscheduler/scheduler_manager.go
TommyLike: @liyinan926 I don't have a favor we should or should not use interface here, updated in latest commits , but the package level function are kept since it's used before object initialized. see
```
fmt.Sprintf("Enable batch schedulers for pods' scheduling, the available batch schedulers are: (%s).", strings.Join(batchscheduler.GetRegisteredNames(), ",")))
```

liyinan926: I think having an interface only makes sense if we expect to have more than one possible implementations for `SchedulerManager`, which doesn't seem the case. So I think using a `struct` is sufficient.

PR #606, opened 2019-09-09
Add batchSchedulerOptions support

TommyLike:
For #583, add `BatchSchedulerOptions` to have end user to have the ability to fine-grained control how to batch scheduling.


pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -244,6 +244,9 @@ type SparkApplicationSpec struct {
 	// BatchScheduler configures which batch scheduler will be used for scheduling
 	// Optional.
 	BatchScheduler *string `json:"batchScheduler,omitempty"`
+	// BatchSchedulerOptions provides fine-grained control on how to batch scheduling.
+	// Optional.
+	BatchSchedulerOptions map[string]string `json:"batchSchedulerOptions,omitempty"`
```
liyinan926: Instead of using a map, I suggest creating a `construct` that have dedicated fields for the configuration properties. For example, something like the following:

```
type BatchSchedulerConfiguration struct {
  Queue string `json:"queue,omitempty"`
}

type SparkApplicationSpec struct {
  ...
  BatchSchedulerConfig *BatchSchedulerConfiguration `json:"batchSchedulerConfig,omitempty"`
  ...
}

``` 

This way, it's much easier to perform schema validation.

liyinan926:
Also I'm thinking of merging PR #578 to bump the API version to `v1beta2`. So please wait until that PR gets merged to make the recommended changes. Thanks!

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
TommyLike: There would be a bunch of attributes if new batch schedulers coming in.

liyinan926: That's fine. It's better to use explicit fields than key-value pairs.

liyinan926:
@TommyLike #578 has been merged. Please update this. Thanks!

TommyLike:
@

> @TommyLike #578 has been merged. Please update this. Thanks!

sure！

PR #608, opened 2019-09-10
Improve documentation for SparkJobNamespace

michaeljmarshall:
After following the `quick-start-guide.md` to start using this project, I noticed an inconsistency between the documentation on the `SparkJobNamespace` and its implementation. This PR seeks to fix the inconsistency and improve upon the existing documentation. In writing this PR, I assume that the implementation is correct and that the documentation ought to reflect the implementation.

The specific issue relates to both the `sparkJobNamespace`'s default value and the helm chart's behavior regarding the `sparkJobNamespace`.

The documentation currently states that the default value for the `sparkJobNamespace` is `default`. As shown in the helm chart's [values.yaml](https://github.com/helm/charts/blob/master/incubator/sparkoperator/values.yaml#L16) file, the default is the empty string (`""`). This PR fixes this discrepancy.

I added documentation detailing the behaviors of the spark-operator for the default and non default values of the `sparkJobNamespace`. For example, the spark operator's [main.go](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/403cf0a7e8dced3b81782ce6b48d824b85b69cd6/main.go#L62) file states that the Spark Operator will filter events based on a flagged namespace. The helm [chart](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/spark-operator-deployment.yaml#L66) provides the `sparkJobNamespace` as the flagged namespace.

Additionally, the current documentation discusses the creation of service accounts and associated RBAC. When the `sparkJobNamespace` is the default value, helm creates a service account in the namespace where the spark-operator is to be deployed, see [spark-serviceaccount.yaml](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/spark-serviceaccount.yaml). However, the [spark-rbac.yaml](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/spark-rbac.yaml#L1) explicitly excludes all content when the `sparkJobNamespace` is the default value. When the value is set to something other than the default, both the service account and the RBAC are created in the helm chart. The documentation now reflects this behavior.

Although, it seems to me that it would make more sense in the case of the default `sparkJobNamespace` to either also create the RBAC in the spark-operator's namespace as is currently done for the service account, or to also prevent the creation of a service account as is currently done for the RBAC. Based on the current implementation, the helm chart creates a useless service account. Since changing the helm implementation will require a PR against the helm chart, I think it'd be helpful to gather consensus here, before I submit a PR to that repo.


docs/quick-start-guide.md
```diff
@@ -155,6 +158,14 @@ $ helm upgrade <YOUR-HELM-RELEASE-NAME> --set operatorImageName=org/image --set
 
 Refer to the Helm [documentation](https://docs.helm.sh/helm/#helm-upgrade) for more details on `helm upgrade`.
 
+## About the Spark Job Namespace
+
+The default SparkJobNamespace is `""`, as defined in the Helm chart's [README](https://github.com/helm/charts/blob/master/incubator/sparkoperator/README.md). Note that in the [Kubernetes apimachinery](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apimachinery) project, the constants `NamespaceAll` and `NamespaceNone` are both defined as the empty string. In this case, the empty string represents `NamespaceAll`. When set to `""`, the Spark Operator supports deploying `SparkApplications` to all namespaces. The Helm chart will create a service account in the namespace where the spark-operator is deployed, but Helm skips setting up the RBAC used to give permission to the service account. In order to deploy jobs, you will need to ensure the driver pod's service account meets the criteria described in the [service accounts for driver pods](#about-the-service-account-for-driver-pods) section.
```
liyinan926: Two things:
1. `SparkJobNamespace` -> `Spark Job Namespace` to be consistent. Then let's add a note after `Spark Job Namespace` that mentions the `sparkJobNamespace` parameter defined in the Helm chart (with ` quotes).
2. `but Helm skips setting up the RBAC used to give permission to the service account`. I would rephrase this as: `but Helm skips setting up the RBAC for driver pods of your Spark applications to be able to manipulate executor pods`.

michaeljmarshall: I added a commit to address both points.

PR #612, opened 2019-09-12
Makefile for sparkctl

avnerl:
Makefile for building sparkctl
mainly for dev people with no golang knowledge/setup

please consider uploading the binaries to github releases


avnerl:
also, we cant use:
SPARK_OPERATOR_GOPATH=$GOPATH/src/github.com/GoogleCloudPlatform/spark-on-k8s-operator
as this interpolates at runtime with the host machine and we actually need the docker image internal go path

PR #615, opened 2019-09-13
Added TTL for SparkApplications

### LINKED ISSUE #603: Add support for TTL for SparkApplications
Looks like a way to automatically garbage collect completed `SparkApplication`s is useful for some use cases. This can be achieved by adding an optional TTL field to `SparkApplicationSpec` that specifies the expiration time after the application terminates. The operator is responsible for garbage collecting applications with their TTL expired.

liyinan926:
Fixes #603.


khogeland:
Hey, have you tested this in a live environment? I think [this check](https://github.com/liyinan926/spark-on-k8s-operator/blob/fca3a466e7ab6fba38d3ecd2971eaa8f41e8b14b/pkg/controller/sparkapplication/controller.go#L187-L191) will prevent GC from working if you're relying on resync

liyinan926:
Ah, your are right. That check needs to be removed for this to work.

PR #617, opened 2019-09-14
Support backwards compatibility with v1beta1

khogeland:
To ease migration from v1beta1 to v1beta2, this PR adds backwards compatibility by rounding up the previously-float `Cores` field.

To migrate with minimal downtime:
- Add `v1beta2` to the old CRDs' `versions` lists
- Deploy the new image
- Switch all apps to `v1beta2`
- Replace the old CRDs with the new CRDs

I'm not sure if this should be merged or maybe released as a transitional image, but I'm leaving it here as a pointer for anyone else who needs to figure out how to upgrade in production.

### DIFF
--- a/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
@@ -47,15 +47,15 @@ const (
 
 // For backwards compatibility with v1beta1, must be able to unmarshal from float64 and marshal to int32
 // This is achieved by rounding up previous float64-type fields
-type cores int32
+type Cores int32
 
-func (c *cores) UnmarshalJSON(data []byte) error {
+func (c *Cores) UnmarshalJSON(data []byte) error {
 	var v float64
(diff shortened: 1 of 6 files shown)


pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -364,8 +381,9 @@ type Dependencies struct {
 type SparkPodSpec struct {
 	// Cores is the number of CPU cores to request for the pod.
 	// Optional.
-	// +kubebuilder:validation:Minimum=1
-	Cores *int32 `json:"cores,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:ExclusiveMinimum=true
```
khogeland: Bizarrely, just setting `Minimum=1` didn't work. With a value of 1, it failed validation with `should be greater than 1` (seems like a bug?)

liyinan926:
Looks like we probably don't need this one. Closing this.

PR #618, opened 2019-09-14
Check application expiration in OnUpdate

liyinan926:
@khogeland


khogeland:
Couldn't the TTL be checked here to fail the condition? The check was added to avoid filling the controller queue with unchanged applications.

liyinan926:
> Couldn't the TTL be checked here to fail the condition? The check was added to avoid filling the controller queue with unchanged applications.

The check is actually kind of anti-pattern as it may lead to missing updates if the informer fails right after receiving the update but before calling `onUpdate`.

liyinan926:
Well, looks like controller-time also uses this as one of the predicates: https://github.com/kubernetes-sigs/controller-runtime/blob/d7467fcf9e1c5817231f2768ffd626a244f2221b/pkg/predicate/predicate.go#L98. So I guess this is a set practice. I have updated the PR to check app expiration on `OnUpdate`.

PR #624, opened 2019-09-18
Derive driver pod name if it's not found in status

liyinan926:


liyinan926:
@khogeland can you take a look?

pkg/controller/sparkapplication/controller.go
```diff
@@ -783,12 +783,16 @@ func (c *Controller) getSparkApplication(namespace string, name string) (*v1beta
...
 func (c *Controller) deleteSparkResources(app *v1beta2.SparkApplication) error {
 	driverPodName := app.Status.DriverInfo.PodName
-	if driverPodName != "" {
-		glog.V(2).Infof("Deleting pod %s in namespace %s", driverPodName, app.Namespace)
-		err := c.kubeClient.CoreV1().Pods(app.Namespace).Delete(driverPodName, metav1.NewDeleteOptions(0))
-		if err != nil && !errors.IsNotFound(err) {
-			return err
-		}
+	// Derive the driver pod name in case the driver pod name was not recorded in the status,
+	// which could happen if the status update right after submission failed.
```
khogeland: This fix looks like an objective improvement, but I don't understand the race here. How can the controller function exit without having updated the status to include the pod name, unless it encounters an error after spark-submit?

khogeland:
I think the same fix applies to `validateSparkResourceDeletion`

pkg/controller/sparkapplication/controller.go
liyinan926: Yes, that could happen. Although we do retries for status update, it may fail all attempts due to connection issue to the api server for example. The app runs into a situation that the driver pod has been created and running but the driver pod name is absent in status.

liyinan926:
> I  think the same fix applies to validateSparkResourceDeletion

Good catch. Done.

PR #629, opened 2019-09-20
Enumerate string spec on Python version

coetzeevs:
**Context:**
Running Python3 on a K8s cluster kept throwing: 
```
spec.pythonVersion in body must be of type string: "integer"
spec.pythonVersion in body should be one of [2 3]
```
Conflicting message - the list of possible values was listed as integers but it expects a string. 

**Changes:**
Simply make the list of possible values be string i.s.o. integers.

**Checks:**
Made the changes locally, applied the CRDs on a K8s cluster, and was able to deploy a SparkApplication resource successfully without getting any errors.

### DIFF
--- a/manifest/crds/sparkoperator.k8s.io_scheduledsparkapplications.yaml
+++ b/manifest/crds/sparkoperator.k8s.io_scheduledsparkapplications.yaml
@@ -1918,8 +1918,8 @@ spec:
                   type: object
                 pythonVersion:
                   enum:
-                  - 2
-                  - 3
+                  - "2"
+                  - "3"
                   type: string
                 restartPolicy:
                   properties:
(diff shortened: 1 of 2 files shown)


michaeljmarshall:
@liyinan926 - Since the CRDs are now generated, this change should have been made to the annotation found here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L247.

The next time the `controller-gen` is run, these changes will be erased.

liyinan926:
@michaeljmarshall yep, you are right. Will fix that.

liyinan926:
The annotation has been fixed in #630.

PR #645, opened 2019-10-02
Fix v1beta2: Use UpdateStatus to update subresource

akhurana001:
@liyinan926 here is a quick fix for https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/644 .  Haven't  extensively tested but the sample examples run fine with this change . Also didn't test the `ScheduledSparkApplication`.

I also ran into `validation failure list:
status.terminationTime in body must be of type string: "null"` errors.  Haven't looked much into this but removing the validation did fix it.  I think K8s also started supporting `nullable: true` in 1.14 which should work as well.


manifest/crds/sparkoperator.k8s.io_sparkapplications.yaml
```diff
@@ -2563,9 +2563,6 @@ spec:
               type: integer
             submissionID:
               type: string
-            terminationTime:
```
liyinan926: Why this change?

liyinan926: Didn't see your message above. Removing the check is fine for now before we can add `nullable: true`. I think we need to remove this check for all `metav1.Time` field in both `SparkApplication` and `ScheduledSparkApplication`.

liyinan926: Actually a quick fix is to add `omitempty`.

liyinan926: It seems that `controller-gen` supports the `+nullable` tag. Can we try that?

liyinan926: Looks like `nullable` support is only available beginning 1.14. So I guess for now we need to remove this check. Actually I don't even think validation on `status` makes any sense for now. Let's remove validation for `status` all together.

akhurana001: Yes, nullable support starts with 1.14.  Agreed, we can remove the validations for status. Let me do that

liyinan926:
@khogeland FYI not sure if you ran into the issue this PR tries to address or not.

manifest/crds/sparkoperator.k8s.io_sparkapplications.yaml
akhurana001: I did remove the validations manually. I don't see any usable marker in https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/validation.go which would achieve the same ? Let me know if there is one.

liyinan926: See https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/validation.go#L163.

akhurana001: Not sure on https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/validation.go#L163 . I think this is the same `+nullable` tag which sets `nullable:true` in the CRD. It will let the status field be null but  if the `status` field is not null, it will still validate the types and we will run into the same `terminationTime` issue. Maybe I am missing something here.

liyinan926: Yes, that's right. But for `status` as a whole, it's type is `object`, and I think they fixed the issue with empty `status`. We can still apply `+nullable` to those `metav1.Time` fields to solve the issue.

akhurana001: `+nullable` will need 1.14+ support as we discussed before. I can do that in another PR if you think that's the right way.

liyinan926: Yes, I meant for now we just remove validation for status, and in the future, we should add it back and use `+nullable` on those time fields.

PR #646, opened 2019-10-02
 Updated the Prerequisite section of README.md

liyinan926:
@akhurana001

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,16 +23,7 @@ on how to enable the webhook.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


akhurana001:
looks good. Thanks for getting to this.  I will rebase the other PR once this is checked this.

liyinan926:
@akhurana001 thanks for the quick review! Actually I just realized `nullable` is not supported before 1.14. I made a comment on your PR, and I think we should just remove validation for `status` all together for now.

PR #694, opened 2019-11-14
Full support of k8s environment variables for SparkApplication through webhook

urihoenig:
Current support of environment variables is only for standard key-value through spark-submit.
K8S can support more complex envrionment variables like fetching values from other fields and secrets.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -58,6 +58,7 @@ func patchSparkPod(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperat
 	patchOps = append(patchOps, addHostNetwork(pod, app)...)
 	patchOps = append(patchOps, addNodeSelectors(pod, app)...)
 	patchOps = append(patchOps, addDNSConfig(pod, app)...)
+	patchOps = append(patchOps, addEnvVars(pod, app)...)
 
 	op := addSchedulerName(pod, app)
 	if op != nil {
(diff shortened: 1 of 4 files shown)


urihoenig:
@googlebot I signed it!

urihoenig:
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/695

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
```diff
@@ -365,6 +365,9 @@ type SparkPodSpec struct {
 	// Secrets carries information of secrets to add to the pod.
 	// Optional.
 	Secrets []SecretInfo `json:"secrets,omitempty"`
+	// Env carries the environment variables to add to the pod.
+	// Optional.
+	Env []apiv1.EnvVar `json:"env,omitempty"`
 	// EnvVars carries the environment variables to add to the pod.
 	// Optional.
 	EnvVars map[string]string `json:"envVars,omitempty"`
```
liyinan926: We probably should mark this field `Deprecated` and deprecate it in the next API version.

pkg/webhook/patch.go
```diff
@@ -158,6 +159,48 @@ func addVolumeMount(pod *corev1.Pod, mount corev1.VolumeMount) patchOperation {
...
+
+	var value interface{}
+	var patchOps []patchOperation
+
+	first := false
+	if len(pod.Spec.Containers[i].Env) == 0 {
+		first = true
+	}
+
+	for _, envVar := range envVars {
```
liyinan926: This reminds me that we are **not** doing the correct thing for `addTolerations` and maybe some others. We should fix those.

pkg/webhook/patch_test.go
```diff
@@ -1108,6 +1108,98 @@ func TestPatchSparkPod_HostNetwork(t *testing.T) {
 	}
 }
 
+func TestPatchSparkPod_Env(t *testing.T) {
+	drvEnvKey := "TEST_DRV_ENV_VAR_KEY"
```
liyinan926: Please use more than one env vars both the driver and executor pods.

pkg/webhook/patch.go
liyinan926: Created https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/697 to fix those.

pkg/apis/sparkoperator.k8s.io/v1beta1/types.go
urihoenig: Done

pkg/webhook/patch_test.go
urihoenig: Done

liyinan926:
FYI: Just realized that this will also make the field `envSecretKeyRefs` obsolete.

urihoenig:
@liyinan926 - I thought about it - my code supports adding environment variables only through the webhook. The existing code allows adding them directly from the operator.
Are you sure that the existing behavior should be marked as deprecated? Shouldn't we allow both behaviors?

liyinan926:
> I thought about it - my code supports adding environment variables only through the webhook. The existing code allows adding them directly from the operator.
Are you sure that the existing behavior should be marked as deprecated? Shouldn't we allow both behaviors?

The problem is that both ways achieve the same goal. It doesn't matter if they are fundamentally different implementations, as long as they do the same thing. Further, if a user happenly defines the same environment variable with different values in both ways, the result will be confusing because we don't have validation for this.

urihoenig:
> > I thought about it - my code supports adding environment variables only through the webhook. The existing code allows adding them directly from the operator.
> > Are you sure that the existing behavior should be marked as deprecated? Shouldn't we allow both behaviors?
> 
> The problem is that both ways achieve the same goal. It doesn't matter if they are fundamentally different implementations, as long as they do the same thing. Further, if a user happenly defines the same environment variable with different values in both ways, the result will be confusing because we don't have validation for this.

Sounds right. Thanks. I've marked also the envSecretKeyRefs as deprecated.

sb2nov:
@liyinan926 is there an ETA on when would the next release for the operator be ?

liyinan926:
The next release is expected to be based on Spark 2.4.5, which should be out soon.

sb2nov:
Thanks.

jkleckner:
Noting here that there isn't anything in the documentation for this way of setting individual env vars such as described here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/785#issuecomment-580573895

Issue #703, opened 2019-11-22
Add sparkVersion in examples/spark-pi-schedule.yaml


echang0929:
When I applied this example, I got error:

[...]

PR #708, opened 2019-11-29
Add sparkVersion in spark-pi-scheduled example

Jeffwan:
Address #703 

`sparkVersion` is a required field in `ScheduledSparkApplication`. Without the field, user will see error like beflow

```
The ScheduledSparkApplication "spark-pi-scheduled" is invalid:
......
validation failure list:
spec.template.sparkVersion in body is required
```


echang0929:
Thanks, Jeff

PR #723, opened 2019-12-11
Create builds with spark:2.4.5-SNAPSHOT

jkleckner:
This patch set builds the operator using the spark image:
- gcr.io/spark-operator/spark:2.4.5-SNAPSHOT

This is needed as documented in this issue:
- https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/669

The underlying problem is that openjdk on alpine has become unsupported
and the spark project has moved to openjdk:8-jdk-slim which is ubuntu
based as described in this issue:
- https://issues.apache.org/jira/browse/SPARK-28938

Without this, the GKE service accounts or workload identity won't
properly authorize.  This was not observed with the given examples
because they explicitly pass in credentials rather than relying on
workload identity or service accounts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,10 +34,6 @@ RUN apt-get update \
     && apt-get install -y openssl curl tini \
     && rm -rf /var/lib/apt/lists/*
 COPY hack/gencerts.sh /usr/bin/
-ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
-RUN chmod +x /tini \
-    && mv /tini /sbin/tini
 
 COPY entrypoint.sh /usr/bin/
 ENTRYPOINT ["/usr/bin/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.4
+ARG SPARK_IMAGE=gcr.io/spark-operator/spark:2.4.5-SNAPSHOT
 
 FROM golang:1.12.5-alpine as builder
 ARG DEP_VERSION="0.5.3"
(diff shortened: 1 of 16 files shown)


jkleckner:
@googlebot I signed it!

docs/user-guide.md
```diff
@@ -64,9 +64,9 @@ metadata:
 spec:
   type: Scala
   mode: cluster
-  image: gcr.io/spark/spark:v2.4.4
+  image: gcr.io/spark/spark:v.4.5-SNAPSHOT
```
aboyett: Change `v.4.5-SNAPSHOT` to `2.4.5-SNAPSHOT`

Dockerfile
```diff
@@ -30,8 +30,14 @@ RUN go generate && CGO_ENABLED=0 GOOS=linux go build -o /usr/bin/spark-operator
...
 COPY --from=builder /usr/bin/spark-operator /usr/bin/
-RUN apk add --no-cache openssl curl tini
+RUN apt-get update \
+    && apt-get install -y openssl curl tini \
+    && rm -rf /var/lib/apt/lists/*
 COPY hack/gencerts.sh /usr/bin/
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini \
+    && mv /tini /sbin/tini
```
aboyett: `tini` is being installed through apt-get and the Debian Buster base distro already includes tini v0.18.0. These lines could be removed and `entrypoint.sh` fixed up to use the `/usr/bin/tini` path provided by the Debian tini package

docs/user-guide.md
```diff
@@ -539,7 +539,7 @@ spec:
   template:
     type: Scala
     mode: cluster
-    image: gcr.io/spark/spark:v2.4.4
+    image: gcr.io/spark/spark:v2.4.5-SNAPSHOT
```
aboyett: `v2.4.5-SNAPSHOT` -> `2.4.5-SNAPSHOT`

docs/user-guide.md
jkleckner: Thanks.

Dockerfile
jkleckner: Thanks.

docs/user-guide.md
jkleckner: And it shouldn't include the v for the snapshot build.

liyinan926:
I have built and pushed `docker pull gcr.io/spark-operator/spark:v2.4.5-SNAPSHOT`.

jkleckner:
I think #739 is preferred.

jkleckner:
Closing this vs #739

PR #707, opened 2019-11-29
Add volumes support for Spark scratch space spark.local.dir

Jeffwan:
This PR helps to mount `HostPath` and `PersistentVolumeClaim` volume for Spark scratch space. Part of #702 

Upstream support was added here and this is included in Spark-3.0.0 (preview)
https://issues.apache.org/jira/browse/SPARK-28042
https://github.com/apache/spark/pull/24879/files

There're two different implementations I considered. 
1. Handle local dir differently and don't use webhook
2. Use webhook patch volume

`spark-local-dir-1` will be created automatically by spark before pods get patched if we don't specify volume for the scratch space. If we choose 2 option, we have to remove this volume and attach new volumes. What's more, SPARK_LOCAL_DIR env needs to be updated as well. I think the first option is more like a native way. 

Let me know if it makes sense and If there's more elegant way, I'd love to make the change.

We can  ask user specify in `SparkApplicationSpec` if they like to use `hostPath`, but this won't give the flexibility like `persistentVolumeClaim` support and we don't know if user want to use multiple disk for scratch space to increase overall IOPS. Need to consider tradeoff here. 
/cc @liyinan926

PR #729, opened 2019-12-12
Propagate volumeMouns to spark-submit job + higher spark-submit memory limit

urihoenig:
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/728 - Volume mounts

As for the memory limit - it wasn't enough for the image we use.


pkg/controller/sparkapplication/job.go
```diff
@@ -103,9 +103,11 @@ func (sjm *realSubmissionJobManager) createSubmissionJob(app *v1beta2.SparkAppli
 			BackoffLimit: app.Spec.RestartPolicy.OnSubmissionFailureRetries,
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
+					Volumes:          app.Spec.Volumes,
```
liyinan926: We probably should limit the `Volumes` to only those appearing on `Driver.VolumeMounts`.

liyinan926: And we should also not include volumes for scratch space. See #707.

urihoenig:
Not needed. I'll open a separate PR for the memory limit

PR #745, opened 2019-12-19
Fix entrypoint.sh to /usr/bin/tini

jkleckner:
Somehow this got lost in the updates.
The debian tini install puts it into /usr/bin/tini

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -24,4 +24,4 @@ if [[ -z "$uidentry" ]] ; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


aboyett:
LGTM.

This was noted in the review #723 and fixed in a subsequent changeset on that branch, but then lost when #723 was replaced by #739.

jkleckner:
FYI, this version was tested in development yesterday with this fix.
registry.gitlab.com/jkleckner/spark-on-k8s-operator/spark-operator:v1beta2-28-gddf0250_v2.4.4-145-g15bad8d008

PR #751, opened 2019-12-20
Updated GCP specific image and user guide

liyinan926:


liyinan926:
@jkremser please take a look, and particularly the changes to the Dockerfiles. Thanks!

PR #832, opened 2020-03-08
Fix for #826 and some refactoring

### LINKED ISSUE #826: Race Condition to Cause Driver Pod Not Found
I am relatively new to go-client and k8s. But in the past few days I spent a lot of time trying to see what’s going on there, regarding [issue795](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/795)

### This section explains briefly a behavior why spark driver restarts

When I submit a SparkApplication, I do not set the restartPolicy. There are other default specs that cause the spark driver to restart similar to this. Spark operator will delete the submitted spark driver pod because its local cache (from Informer) is different from the API server in terms of the SparkApplication spec (local cache has an updated default restartPolicy - default, compared to the API server’s restartPolicy - “”). So normally, you will see the following pod behavior for spark driver pod.

Pu-spark-driver                 0/1     Pending             0          0s
Pu-spark-driver                 0/1     Pending             0          0s
Pu-spark-driver                 0/1     ContainerCreating   0          0s
Pu-spark-driver                 0/1     Terminating         0          1s
Pu-spark-driver                 0/1     Terminating         0          1s
Pu-spark-driver                 0/1     Pending             0          0s
Pu-spark-driver                 0/1     Pending             0          0s
Pu-spark-driver                 0/1     ContainerCreating   0          0s
Pu-spark-driver                 0/1     ContainerCreating   0          2s
Pu-spark-driver                 1/1     Running             0          3s

I think this behavior is also observed by other people in the spark-operator channel.

### This section explains the race condition (as least that’s what I believe, correct me if I am wrong)

A failed SparkApplication has following state transition parsed from the spark operator log. This log is reproducible and the order is verified to be like this across all my failed SparkApplications. 

{ } -> {SUBMITTED }
{SUBMITTED } -> {SUBMITTED }
{INVALIDATING } -> {PENDING_RERUN }
{SUBMITTED } -> {FAILING Driver Pod not found}
{FAILING Driver Pod not found} -> {FAILED Driver Pod not found}
{FAILED Driver Pod not found} -> {FAILED Driver Pod not found}

The {INVALIDATING } -> {PENDING_RERUN } is the state transition after spark operator deletes the driver pod. However, the next state transition is {SUBMITTED } -> {FAILING Driver Pod not found}. Which does not make sense because the previous state should be {PENDING_RERUN}, should not be {SUBMITTED}. This is where my theory of race condition comes into the picture. If you check the code, you will find the states showed above in the log refer to cached states and submitted state to API server.

So basically for this state transition, {INVALIDATING} is the local cache state, and the spark operator is sending {PENDING_RERUN} state to the API server, using the rest API to api server, not via informer. It does not update the local cache.
{INVALIDATING } -> {PENDING_RERUN }

And in the next state transition, it reads from the local cache, where the state is still {SUBMITTED }. In this case, it checks if the driver pod is alive. It could not find the driver pod, so it transits the state to {FAILING Driver Pod not found}. And the following state transition from {FAILING Driver Pod not found} -> {FAILED Driver Pod not found} make senses. This concludes my analysis.

liyinan926:
This PR fixed #826 by moving the place of applying defaults from the `onAdd` callback to the `submitSparkApplication` such that the defaults only get applied to a copy of the `SparkApplication` resource when submitting the application, instead of to the resource in the cache as received by `onAdd`. This avoids the changes of defaulting from being sent back to the API server and causing the application to be invalidated and resubmitted.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -627,7 +627,7 @@ Below is an example that shows how to configure the metric system to expose metr
 
 ```yaml
 spec:
-  dep:
+  deps:
     jars:
     - http://central.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.11.0/jmx_prometheus_javaagent-0.11.0.jar
   monitoring:
@@ -689,7 +689,7 @@ spec:
   timeToLiveSeconds: 3600
 ```
 
-Note that this feature requires that informer cache resync to be enabled, which is true by default with a resync internal of 30 seconds. You can change the resync interval by setting the flag `-resync-interval=<internval>`.
+Note that this feature requires that informer cache resync to be enabled, which is true by default with a resync internal of 30 seconds. You can change the resync interval by setting the flag `-resync-interval=<interval>`.
 
 ## Running Spark Applications on a Schedule using a ScheduledSparkApplication
 
--- a/pkg/batchscheduler/interface/interface.go
+++ b/pkg/batchscheduler/interface/interface.go
@@ -24,5 +24,5 @@ type BatchScheduler interface {
 	Name() string
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
-	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) (*v1beta2.SparkApplication, error)
+	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
 }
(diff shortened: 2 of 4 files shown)


vincentye38:
We have the issue similar to this, but happening when ApplicationState is in RUNNIG.  The state update sequence is:
RUNNING -> SECCEEDING
RUNNING -> FAILING (driver pod not found)

The ApplicationState ended up in Failed state, even the spark job completed successfully.
Any idea if it's the same issue of default value being copied into copy of crd resource,  but in different execution path?

PR #846, opened 2020-03-25
golang:1.14.1-alpine is necessary related to issue 845

cytar:
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/845


cytar:
@googlebot I signed it!

Issue #848, opened 2020-03-26
plaintext http redirection even with spark.ssl.ui.enabled


cytar:
Hello,

as we work with my team with the spark UI, we are facing the following problem:
- we use the ability of the spark operator to create an ingress
- we enable ssl for spark UI with sparkConf parameters spark.ssl.ui.enabled and spark.ssl.ui.port

but when accessing the UI we saw that after accessing ingress in SSL, we face an HTTP plaintext redirection.
In our case, because we have a proxy authorizing only SSL between clients and UI, it's failing after redirection.

cytar:
To solve this issue we propose what we have successfully tested: 
Adding the ability to use 2 new sparkConf parameters:
- spark.ui.ingress.class to specify if one use nginx ingress controller or another one
- spark.ui.ingress.forcessl that specify if we want to force the full SSL mode

If not used, it's ignored, if its used, in our case with :
  sparkConf:
    "spark.ui.ingress.class": "nginx"
    "spark.ui.ingress.forcessl: "true"

it's adding specific annotations to ingress:
  annotations:
    kubernetes.io/ingressclass: nginx
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"

In all cases it's working.
So i'm proposing the following Pull Request => https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/849

PR #849, opened 2020-03-26
Add ingress annotations to force ssl (related to issue #848)

cytar:
related to issue #848


pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -70,6 +71,19 @@ func getResourceLabels(app *v1beta2.SparkApplication) map[string]string {
 	return labels
 }
 
+func getResourceAnnotations(app *v1beta2.SparkApplication) map[string]string {
+	annotations := map[string]string{}
+
+	if app.Spec.SparkConf["spark.ui.ingress.class"] == "nginx" {
```
liyinan926: Instead of defining these Spark config options, how about the following: make a struct type field that encapsulates all UI related config options, e.g.,:

```
type UIConfig struct {
  IngressClass *string `json:"ingressClass,omitempty"`
  EnableSSL bool `json:"enableSSL,omitempty"`
  ForceSSL bool `json:"forceSSL,omitempty"`
}
```

And then add an optional field of type `UIConfig` in `SparkApplicationSpec`.

AungkokoAungko:
All checks have passed

pkg/controller/sparkapplication/sparkapp_util.go
cytar: Hello. Thanks for your answer.
Based on your advice, i made this new version, declaring a UIConfig struct.
All my tests are working fine, with or without the use of this struct, and whatever the values of its parameters are. But i only handle the nginx ingress Class because its the most used and the only one i use myself. For those who need to adapt for another class, they will only have to append the sparkapplication.getResourceAnnotations function.

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -273,6 +280,8 @@ type SparkApplicationSpec struct {
 	// BatchSchedulerOptions provides fine-grained control on how to batch scheduling.
 	// +optional
 	BatchSchedulerOptions *BatchSchedulerConfiguration `json:"batchSchedulerOptions,omitempty"`
+	// Define specific sparkUI config parameters like classs used for ingress creation
+	UIConfig *UIConfig `json:"uiConfig,omitempty"`
```
liyinan926, line 284: Add `// +optional` for the field.

cytar: done

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -72,6 +72,24 @@ func getResourceLabels(app *v1beta2.SparkApplication) map[string]string {
...
 }
 
+func getResourceAnnotations(app *v1beta2.SparkApplication) map[string]string {
+	annotations := map[string]string{}
+
+	if app.Spec.UIConfig != nil {
+		if app.Spec.UIConfig.IngressClass != nil {
+			if *app.Spec.UIConfig.IngressClass == "nginx" {
+				annotations["kubernetes.io/ingressclass"] = "nginx"
+				if &app.Spec.UIConfig.EnableSSL != nil && &app.Spec.UIConfig.ForceSSL != nil {
```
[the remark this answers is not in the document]
cytar: right

pkg/controller/sparkapplication/sparkapp_util_test.go
```diff
@@ -57,3 +58,23 @@ func TestPrintStatus(t *testing.T) {
...
+func Test_getResourceAnnotations(t *testing.T) {
+	type args struct {
+		app *v1beta2.SparkApplication
+	}
+	tests := []struct {
+		name string
+		args args
+		want map[string]string
+	}{
+		// TODO: Add test cases.
```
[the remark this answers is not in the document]
cytar: no problem, its done with 3 cases.

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -232,3 +232,32 @@ func TestCreateSparkUIIngress(t *testing.T) {
...
+		ingressURLFormat string
+		kubeClient       clientset.Interface
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    *SparkIngress
+		wantErr bool
+	}{
+		// TODO: Add test cases.
```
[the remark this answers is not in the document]
cytar: as TestCreateSparkUIIngress function already exists, i appended it with a new case and cleaned up my own test function.

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -162,6 +162,13 @@ type ScheduledSparkApplicationList struct {
 	Items           []ScheduledSparkApplication `json:"items,omitempty"`
 }
 
+// Specific SparkUI config parameters
+type UIConfig struct {
+	IngressClass *string `json:"ingressClass,omitempty"`
```
liyinan926, line 167: What is each field for? Can you add comments for the new fields?

liyinan926, line 167: What does it mean for `IngressClass` to be `nil`?

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -180,55 +180,90 @@ func TestCreateSparkUIIngress(t *testing.T) {
...
+	app2 := &v1beta2.SparkApplication{
+		Spec: v1beta2.SparkApplicationSpec{
+			UIConfig: &v1beta2.UIConfig{IngressClass: &ingressclass_nginx},
+		},
 	}
 
-	if sparkIngress.ingressName != expectedIngress.ingressName {
-		t.Errorf("Ingress name wanted %s got %s", expectedIngress.ingressName, sparkIngress.ingressName)
+	// list of tests related to app list
+	type args struct {
```
liyinan926, line 192: Why do you need a `struct` for a single arg?

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
cytar: As any json item is omitempty, its covering the case where UIconfig is defined but not IngressClass

pkg/controller/sparkapplication/sparkui_test.go
cytar: I find it easier to use a generic args struct that can be extended, instead of building a code per case. For example, it can be easily extended. Tell me if you prefer that i only use a more simple object only adapted to my own test.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -72,6 +72,22 @@ func getResourceLabels(app *v1beta2.SparkApplication) map[string]string {
...
 }
 
+func getResourceAnnotations(app *v1beta2.SparkApplication) map[string]string {
+	annotations := map[string]string{}
+
+	if app.Spec.UIConfig != nil {
+		if app.Spec.UIConfig.IngressClass != nil {
+			if *app.Spec.UIConfig.IngressClass == "nginx" {
+				annotations["kubernetes.io/ingressclass"] = "nginx"
+				if app.Spec.UIConfig.EnableSSL && app.Spec.UIConfig.ForceSSL {
```
liyinan926, line 82: Will enabling and forcing SSL only apply to ingress backed by `nginx`?

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -162,6 +162,16 @@ type ScheduledSparkApplicationList struct {
 	Items           []ScheduledSparkApplication `json:"items,omitempty"`
 }
 
+// Specific SparkUI config parameters
+type UIConfig struct {
+	// Ingress class used for k8s annotation: kubernetes.io/ingressclass
+	IngressClass *string `json:"ingressClass,omitempty"`
```
liyinan926, line 168: Golang comment for fields should start with the name of the field, like `IngressClass defines the value for annotation ...`. Ditto for the fields below. Also please add `// +optional` for the fields since they all look optional.

pkg/controller/sparkapplication/sparkapp_util.go
cytar: I'm not sure, i only know that other IngressClasses exist, but i don't know how they behave. But i opened the door with this function, and anyone that well know other cases will be able to append it.

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
cytar: i adapted the comments. Thanks for your help.

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN chmod +x /usr/bin/dep
 
 WORKDIR ${GOPATH}/src/github.com/GoogleCloudPlatform/spark-on-k8s-operator
 COPY Gopkg.toml Gopkg.lock ./
-RUN dep ensure -vendor-only
+RUN dep ensure -v -vendor-only
```
liyinan926, line 27: Why turn on verbose mode?

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -162,6 +162,19 @@ type ScheduledSparkApplicationList struct {
...
+type UIConfig struct {
+	// Ingressclass defines the class of ingress used for k8s annotation: kubernetes.io/ingressclass
+	// +optional
+	IngressClass *string `json:"ingressClass,omitempty"`
+	// EnableSSL is necessary to specify that k8s ingress Will use SSL
+	// +optional
+	EnableSSL bool `json:"enableSSL,omitempty"`
+	// EnableSSL combined with ForceSSL will set Ingress k8s annotation: nginx.ingress.kubernetes.io/force-ssl-redirect to true
+	// +optional
+	ForceSSL bool `json:"forceSSL,omitempty"`
```
liyinan926, line 175: Will `forceSSL` ever be false if `enableSSL` is true? For example, if `ingressClass` is `nginx`, is it likely that `forceSSL` is false when `enableSSL` is true? I still don't quite understand the relationship between the two boolean fields. Looking at this more, I tend to think that it might make sense to reorganize `UIConfig` (renamed to `SparkUIIngressConfig`) like the following:

```go
// SparkUIIngressConfig is for Ingress configuration for the Spark driver UI.
type struct SparkUIIngressConfig {
  // Looks like an Ingress must reference an IngressClass, so this should not be an optional field. 
  IngressClass string `json:"ingressClass"`
  Nginx *NginxIngressConfig `json:"nginxConfig,omitempty"`
}

// NginxIngressConfig is for configuring the NGINX Ingress.
type NginxIngressConfig struct {
  EnableSSL bool `json:"enableSSL,omitempty"`
  ForceSSLRedirect bool `json:"forceSSLRedirect,omitempty"`
}
```

PR #852, opened 2020-03-27
Add job start latency metrics and add namespace tag in metrics

huskysun:
This PR adds job start latency metric. The job start latency is defined as the time since the sparkapp k8s object is created till when the job is in a running or terminal state. There are two metrics introduced in this change, one is `spark_app_start_latency_microseconds`, which is of Prometheus Summary type, and the other one is `spark_app_start_latency_seconds`, which is of Histogram type. The summary one is to give the user accurate latency values in different percentiles (e.g. 0.5, 0.9. 0.99), thus it's in microseconds. The histogram one is to give the user an option to configure the bucket boundaries, see how the job latency values fall into different histogram buckets, and thus can pick appropriate service job start latency SLA out of it, and thus it's in seconds for easier configuration.

This PR also adds configurability to include `namespace` as a tag in the metrics, to make metrics more useful in a multi-tenant k8s cluster.

I also corrected the state machine diagram in `controller.go` during implementation.

### DIFF
--- a/main.go
+++ b/main.go
@@ -78,6 +78,10 @@ var (
 func main() {
 	var metricsLabels util.ArrayFlags
 	flag.Var(&metricsLabels, "metrics-labels", "Labels for the metrics")
+	var metricsJobStartLatencyBuckets util.HistogramBuckets = util.DefaultJobStartLatencyBuckets
+	flag.Var(&metricsJobStartLatencyBuckets, "metrics-job-start-latency-buckets",
+		"Comma-separated boundary values (in seconds) for the job start latency histogram bucket; "+
+			"it accepts any numerical values that can be parsed into a 64-bit floating point")
 	flag.Parse()
 
 	// Create the client config. Use kubeConfig if given, otherwise assume in-cluster.
@@ -159,10 +163,11 @@ func main() {
 	var metricConfig *util.MetricConfig
 	if *enableMetrics {
 		metricConfig = &util.MetricConfig{
-			MetricsEndpoint: *metricsEndpoint,
-			MetricsPort:     *metricsPort,
-			MetricsPrefix:   *metricsPrefix,
-			MetricsLabels:   metricsLabels,
+			MetricsEndpoint:               *metricsEndpoint,
+			MetricsPort:                   *metricsPort,
+			MetricsPrefix:                 *metricsPrefix,
+			MetricsLabels:                 metricsLabels,
+			MetricsJobStartLatencyBuckets: metricsJobStartLatencyBuckets,
 		}
 
 		glog.Info("Enabling metrics collecting and exporting to Prometheus")
--- a/main.go
+++ b/main.go
@@ -60,10 +60,6 @@ var (
 	namespace                      = flag.String("namespace", apiv1.NamespaceAll, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset.")
 	enableWebhook                  = flag.Bool("enable-webhook", false, "Whether to enable the mutating admission webhook for admitting and patching Spark pods.")
 	enableResourceQuotaEnforcement = flag.Bool("enable-resource-quota-enforcement", false, "Whether to enable ResourceQuota enforcement for SparkApplication resources. Requires the webhook to be enabled.")
-	enableMetrics                  = flag.Bool("enable-metrics", false, "Whether to enable the metrics endpoint.")
-	metricsPort                    = flag.String("metrics-port", "10254", "Port for the metrics endpoint.")
-	metricsEndpoint                = flag.String("metrics-endpoint", "/metrics", "Metrics endpoint.")
-	metricsPrefix                  = flag.String("metrics-prefix", "", "Prefix for the metrics.")
 	ingressURLFormat               = flag.String("ingress-url-format", "", "Ingress URL format.")
 	enableLeaderElection           = flag.Bool("leader-election", false, "Enable Spark operator leader election.")
 	leaderElectionLockNamespace    = flag.String("leader-election-lock-namespace", "spark-operator", "Namespace in which to create the ConfigMap for leader election.")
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -78,6 +78,10 @@ var (
 func main() {
 	var metricsLabels util.ArrayFlags
```
liyinan926: We should move all flag variables into the `var` block above.

pkg/controller/sparkapplication/controller.go
```diff
@@ -122,7 +122,7 @@ func newSparkApplicationController(
 	}
 
 	if metricsConfig != nil {
-		controller.metrics = newSparkAppMetrics(metricsConfig.MetricsPrefix, metricsConfig.MetricsLabels)
+		controller.metrics = newSparkAppMetrics(metricsConfig.MetricsPrefix, metricsConfig.MetricsLabels, metricsConfig.MetricsJobStartLatencyBuckets)
```
liyinan926: Looks like we really should just pass in the `metricsConfig`.

pkg/util/histogram_buckets.go
```diff
@@ -0,0 +1,43 @@
...
+type HistogramBuckets []float64
+
+func (hb *HistogramBuckets) String() string {
+	return fmt.Sprint(*hb)
+}
+
+func (hb *HistogramBuckets) Set(value string) error {
+	*hb = nil
+	for _, boundaryStr := range strings.Split(value, ",") {
+		boundary, err := strconv.ParseFloat(boundaryStr, 64)
```
liyinan926: You probably want to trim `boundaryStr` just in case users add spaces in the bucket flag value.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -196,6 +218,18 @@ func (sm *sparkAppMetrics) exportMetrics(oldApp, newApp *v1beta2.SparkApplicatio
 		}
 	}
 
+	// In the event that state transitions happened too quickly and the spark app skipped the RUNNING state, the job
```
liyinan926: Do we really need this? Generally Kubernetes controller should not be programmed against a state machine. I know we are already doing this in the job state handling logic, but putting in more logic that works against a state machine makes the code even more fragile and harder to read. I think we should try to gradually reduce reliance on state transitions.

main.go
huskysun: Sure, will move it

pkg/controller/sparkapplication/controller.go
huskysun: Sounds good, will fix it in the next commit

pkg/util/histogram_buckets.go
huskysun: Sounds good, will do it

pkg/controller/sparkapplication/sparkapp_metrics.go
huskysun: @liyinan926  There is a concern that if the work queue is huge and it takes a long time to process the next item for a certain sparkapp after it ran spark-submit on it, the driver pod would have already gone through the `PodRunning` state and be in `PodSucceeded` or `PodFailed`, then the sparkapp state will skip `RUNNING` and transit from `SUBMITTED` to `FAILING`/`SUCCEEDING` directly. This is unlikely in normal operator, but when the queue gets piled up, the metric will lose those data.

liyinan926: The problem is that in this case, there's no way you can get an accurate measure of the latency.

huskysun: Yes, but at least there would be a data point indicating that the job has already started. Even the value is just an upper bound and it's not accurate, I think it's still valuable to have the data, otherwise it might trigger a lot more false positive alerts if user sets up a pager on this metric.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -196,6 +218,18 @@ func (sm *sparkAppMetrics) exportMetrics(oldApp, newApp *v1beta2.SparkApplicatio
...
+	// start latency should still be captured.
+	// Note: There is an edge case that a Submitted state can go directly to a Failing state if the driver pod is
+	//       deleted. This is very unlikely if not being done intentionally, so we choose not to handle it.
+	if newState != oldState {
+		if (newState == v1beta2.FailingState || newState == v1beta2.SucceedingState) && oldState != v1beta2.RunningState {
```
liyinan926, line 226: The condition after `&&` probably should be `oldState == v1beta2.SubmittedState` as I don't think it makes sense to update the metric if the oldState is `PendingRerun` for example.

pkg/controller/sparkapplication/sparkapp_metrics.go
```diff
@@ -196,6 +218,18 @@ func (sm *sparkAppMetrics) exportMetrics(oldApp, newApp *v1beta2.SparkApplicatio
...
+			// TODO: remove this log once we've gathered some data in prod fleets.
+			glog.Infof("Calculating job start latency metrics for edge case transition from %v to %v in app %v in namespace %v.", oldState, newState, newApp.Name, newApp.Namespace)
```
liyinan926, line 228: Can we change the level to `2` for this log?

pkg/controller/sparkapplication/sparkapp_metrics.go
huskysun: Sure, that sounds good to me

pkg/controller/sparkapplication/sparkapp_metrics.go
huskysun: Sure

huskysun:
Hi @liyinan926 I've made changes according to your review and tested it. Please review again. Thanks!

PR #865, opened 2020-04-08
Re-enable scheduled app unit test (TestSyncScheduledSparkApplication_Forbid)

duongnt:
Apparently if we don't explicitly set the namespace of the sparkApp, it will be "" and thus causing the unit test to fail.

### DIFF
--- a/pkg/controller/scheduledsparkapplication/controller_test.go
+++ b/pkg/controller/scheduledsparkapplication/controller_test.go
@@ -176,9 +176,6 @@ func TestSyncScheduledSparkApplication_Allow(t *testing.T) {
 }
 
 func TestSyncScheduledSparkApplication_Forbid(t *testing.T) {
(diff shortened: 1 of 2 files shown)


duongnt:
@googlebot I signed it!

liyinan926:
I have never seen unit test failures in `scheduledsparkapplication/controller_test.go`.

pkg/controller/scheduledsparkapplication/controller_test.go
```diff
@@ -176,9 +176,6 @@ func TestSyncScheduledSparkApplication_Allow(t *testing.T) {
 }
 
 func TestSyncScheduledSparkApplication_Forbid(t *testing.T) {
-	// TODO: figure out why the test fails and remove this.
-	t.Skip()
```
duongnt, line 180: @liyinan926 this unit test has been skipped and I was trying to enable it.

Issue #841, opened 2020-03-21
Termination time is not reported when using sidecars


jrj-d:
## The bug

The changes from PR #576 introduce a "shortcut" in the behavior of the sparkapplication FSM and cause termination time not being reported in the sparkapplication resource.

To be more precise, `status.terminationTime` is `nil` when the sparkapplication is finished.

## Context

* Spark operator 1.0.1 & 1.1.1, probably 1.1.0 too
* Kubernetes:
```
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-gke.27", GitCommit:"145f9e21a4515947d6fb10819e5a336aff1b6959", GitTreeState:"clean", BuildDate:"2020-02-21T18:01:40Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}
```

## The reason

Here is the normal FSM flow for a sparkapplication without sidecar **and** for a sparkapplication with sidecar before PR #576 :
```
Events:
  Type    Reason                     Age   From            Message
  ----    ------                     ----  ----            -------
  Normal  SparkApplicationAdded      35s   spark-operator  SparkApplication pyspark-pi was added, enqueuing it for submission
  Normal  SparkApplicationSubmitted  32s   spark-operator  SparkApplication pyspark-pi was submitted successfully
  Normal  SparkDriverRunning         30s   spark-operator  Driver pyspark-pi-driver is running
  Normal  SparkExecutorPending       23s   spark-operator  Executor pyspark-pi-4c5b8270fd5abd06-exec-1 is pending
  Normal  SparkExecutorRunning       22s   spark-operator  Executor pyspark-pi-4c5b8270fd5abd06-exec-1 is running
  Normal  SparkDriverCompleted       14s   spark-operator  Driver pyspark-pi-driver completed
  Normal  SparkApplicationCompleted  14s   spark-operator  SparkApplication pyspark-pi completed
```

PR #576 makes the application state turn to `CompletedState` (via `SucceedingState`) once the driver container is terminated, regardless of the status of sidecars and thus regardless of the status of the pod.

**For most use cases of sidecars**, the driver container finishes before the sidecars. So effectively, PR #576 makes the sparkapplication turn to `CompletedState` before the pod is terminated.

[...]

PR #867, opened 2020-04-10
update code to fix error on termination time when using sidecar

ImpSy:
Hello,

This PR is related to issue #841 

Following the discussion on this issue, i've implemented `option 2`

Specifically, I've introduced a new State Machine specific to the Driver that allow finer grain detection of application end when using sidecar containers
This is also compatible with  kubernetes sidecar implementation on version 1.18

I've also updated the tests to cover sidecar cases

PS: for the review the commit are  scoped
bfacfe9 is logic only
dbd2f21 is to fix and update tests

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -324,16 +324,17 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta2.SparkApplication) erro
 	}
 
 	app.Status.SparkApplicationID = getSparkApplicationID(driverPod)
+	driverState := podStatusToDriverState(driverPod.Status)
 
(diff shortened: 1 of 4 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -324,29 +324,30 @@ func (c *Controller) getAndUpdateDriverState(app *v1beta2.SparkApplication) erro
 	}
 
 	app.Status.SparkApplicationID = getSparkApplicationID(driverPod)
+	driverState := podStatusToDriverState(driverPod.Status)
 
-	if driverPod.Status.Phase == apiv1.PodSucceeded || driverPod.Status.Phase == apiv1.PodFailed {
+	if isDriverTerminated(driverState) {
```
liyinan926, line 329: nit: `hasDriverTerminated`.

ImpSy: Fixed

PR #876, opened 2020-04-17
Fix for #868

### LINKED ISSUE #868: Linear search does not account for the case when the container is not found
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2e4559eac0f593398d5e19ef85bfc6a5df032033/pkg/webhook/patch.go#L167

When the container not found with the name, hardcoded into the configuration, it should log an error that it was not able to patch the Pod `spec`. Currently, this results in an NPE and panic. In addition, provide a hint to the users that the problem around container naming may be fixed using `spark.kubernetes.executor.podTemplate`.

liyinan926:
Fixes #868.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -141,7 +141,11 @@ func addVolumes(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperation
 				ops = append(ops, addVolume(pod, v))
 				addedVolumeMap[m.Name] = v
 			}
-			ops = append(ops, addVolumeMount(pod, m))
+			vmPatchOp := addVolumeMount(pod, m)
+			if vmPatchOp == nil {
+				return nil
+			}
+			ops = append(ops, *vmPatchOp)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/webhook/patch.go
```diff
@@ -162,14 +166,11 @@ func addVolume(pod *corev1.Pod, volume corev1.Volume) patchOperation {
...
-	// Find the driver or executor container in the pod.
-	for ; i < len(pod.Spec.Containers); i++ {
-		if pod.Spec.Containers[i].Name == config.SparkDriverContainerName ||
-			pod.Spec.Containers[i].Name == config.SparkExecutorContainerName {
-			break
-		}
+func addVolumeMount(pod *corev1.Pod, mount corev1.VolumeMount) *patchOperation {
+	i := findContainer(pod)
+	if i < 0 {
+		glog.Warningf("not able to add VolumeMount %s as Spark container was not found in pod %s", mount.Name, pod.Name)
```
zzvara, line 172: I think others who might depend on podTemplates may find it useful to have the container name said out in the warning-log, which is "executor" for Spark 2 and "spark-kubernetes-executor" for Spark 3. It may help to debug problems of this kind. Verbosity in case of error is always useful. :-)

liyinan926: We don't parse the `sparkVersion` due to the lack of validation on the field and a proper parser. This is why in `findContainer` we match executor container with one of the two possible container names. Because of that, it's no easy way of knowing which container is missing.

PR #880, opened 2020-04-20
Cache Sync before Worker Threads are started

akhurana001:


akhurana001:
PTAL @liyinan926

pkg/controller/sparkapplication/controller.go
```diff
@@ -152,17 +152,19 @@ func newSparkApplicationController(
 
 // Start starts the Controller by registering a watcher for SparkApplication objects.
 func (c *Controller) Start(workers int, stopCh <-chan struct{}) error {
+
```
liyinan926, line 155: Nit: unnecessary empty line.

akhurana001: removed

PR #914, opened 2020-05-15
Support changing port in ui exposition

AliGouta:
This PR makes it possible to change and customize:

- Service object : By allowing to specify a `port` different from `targetport`
- Ingress object : By allow adding annotations + adding tls specification (optional)

By default only the port `4040` or a port specified in `spark.ui.port `is exposed with a basic configuration of the Service and Ingress objects. 

It turns out, in several companies spark applications run in some private network which is usually accessible with limited number of ports exposed + using some ingress controllers to redirect trafic within the private network. For instance, we may have only ports `443` and `80` exposed to the ingress controller. These ports could never be defined in the spark configuration since they are below than 1024. Though the spark app fails at the start time. A common solution, is to expose a specific port at the ingress object which then redirects to the Service. The service listens on the same port used by the ingress object. Then redirects internally using `targetport` which might be `4040` for instance. Hence `port` and `targetport` should be different.

Added sepcification in the CRD:

```
            sparkUIOptions:
              properties:
                ingressAnnotations:
                  additionalProperties:
                    type: string
                  type: object
                ingressTLS:
                  items:
                    properties:
                      hosts:
                        items:
                          type: string
                        type: array
                      secretName:
                        type: string
                    type: object
                  type: array
                servicePort:
                  format: int32
                  type: integer
              required:
              - servicePort
              type: object
```

Example in SparkApplication specification:

```
  sparkUIOptions:(OPTIONAL)
    servicePort: 80            => (Optional)
    ingressAnnotations:    => (OPTIONAL)
      "kubernetes.io/ingress.class": "nginx"
    ingressTLS:                  => (OPTIONAL)
      - hosts:
          - "logs-normalizer.zzz.yyy.xxx.company.com"
        serviceName: ""
```

After deploying the app:

Service:
```
logs-normalizer-ui-svc:

apiVersion: v1
kind: Service
...
spec:
  clusterIP: xxx.yy.zzz.245
  ports:
  - name: spark-driver-ui-port
    port: 80
    protocol: TCP
    targetPort: 4040
  selector:
    spark-role: driver
    sparkoperator.k8s.io/app-name: logs-normalizer
  sessionAffinity: None
  type: ClusterIP
```

Ingress:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: nginx
...
...
spec:
  rules:
  - host: logs-normalizer.zzz.yyy.xxx.company.com
    http:
      paths:
      - backend:
          serviceName: logs-normalizer-ui-svc
          servicePort: 80
  tls:
  - hosts:
    - logs-normalizer.zzz.yyy.xxx.company.com
```


pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -284,6 +288,19 @@ type BatchSchedulerConfiguration struct {
 	PriorityClassName *string `json:"priorityClassName,omitempty"`
 }
 
+// Specific SparkUI config parameters
```
liyinan926, line 291: Can we rename this to `SparkUIConfiguration`, which is easier to understand?

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -85,6 +88,12 @@ func createSparkUIIngress(app *v1beta2.SparkApplication, service SparkService, i
 			}},
 		},
 	}
+	if len(getIngressResourceAnnotations(app)) != 0 {
```
liyinan926, line 91: Can you store the result of `getIngressResourceAnnotations` in a variable?

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -85,6 +88,12 @@ func createSparkUIIngress(app *v1beta2.SparkApplication, service SparkService, i
...
+		ingress.ObjectMeta.Annotations = getIngressResourceAnnotations(app)
+	}
+	if len(getIngressTlsHosts(app)) != 0 {
```
liyinan926, line 94: Ditto, can you store the result of `getIngressTlsHosts ` in a variable?

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -94,16 +103,24 @@ func createSparkUIIngress(app *v1beta2.SparkApplication, service SparkService, i
...
+		annotations: ingress.Annotations,
+		ingressTLS:  ingress.Spec.TLS,
 	}, nil
 }
 
 func createSparkUIService(
 	app *v1beta2.SparkApplication,
 	kubeClient clientset.Interface) (*SparkService, error) {
-	portStr := getUITargetPort(app)
+	portStr := getUIServicePort(app)
```
liyinan926, line 114: Can `getUIServicePort` return a integer port instead?

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
AliGouta: changed as you suggested + I changed ExpositionOptions to SparkUIOptions

pkg/controller/sparkapplication/sparkui.go
AliGouta: done

pkg/controller/sparkapplication/sparkui.go
AliGouta: done

pkg/controller/sparkapplication/sparkui.go
AliGouta: done + I did the same to getUITargetPort

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
```diff
@@ -284,6 +288,19 @@ type BatchSchedulerConfiguration struct {
...
+type SparkUIConfiguration struct {
+	// ServicePort allows configuring the port at service level that might be different from the targetPort. This field is mandatory if SparkUIOptions is called.
+	// TargetPort should be the same as the one defined in spark.ui.port
```
liyinan926, line 294: Missing `+optional`.

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -137,17 +159,30 @@ func createSparkUIService(
...
 // in Spec.SparkConf if it is present, otherwise the default port is returned.
 // Note that we don't attempt to get the port from Spec.SparkConfigMap.
-func getUITargetPort(app *v1beta2.SparkApplication) string {
+func getUITargetPort(app *v1beta2.SparkApplication) (int, error) {
 	port, ok := app.Spec.SparkConf[sparkUIPortConfigurationKey]
 	if ok {
-		return port
+		return strconv.Atoi(port)
+	}
+	return strconv.Atoi(defaultSparkWebUIPort)
```
liyinan926, line 175: We can change `defaultSparkWebUIPort` to an `int32`.

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -18,6 +18,8 @@ package sparkapplication
 
 import (
 	"fmt"
+	extensions "k8s.io/api/extensions/v1beta1"
```
liyinan926, line 21: These imports should go under the golang built-in ones.

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -126,13 +128,39 @@ func TestCreateSparkUIService(t *testing.T) {
 			SparkApplicationID: "foo-3",
 		},
 	}
+	var appPort int32 = 80
+	app4 := &v1beta2.SparkApplication{
+
```
liyinan926, line 133: Unnecessary empty line here.

AliGouta: done

pkg/controller/sparkapplication/sparkui_test.go
AliGouta: done

pkg/controller/sparkapplication/sparkui.go
AliGouta: done, I also changed the return type of `getUIServicePort` and `getUITargetPort` to int32

pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
AliGouta: done

PR #917, opened 2020-05-17
Rework the Dockerfile to not hardcode amd64

holdenk:


Dockerfile
```diff
@@ -17,10 +17,9 @@
 ARG SPARK_IMAGE=gcr.io/spark-operator/spark:v2.4.5
 
 FROM golang:1.14.1-alpine as builder
-ARG DEP_VERSION="0.5.4"
-RUN apk add --no-cache bash git
-ADD https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 /usr/bin/dep
-RUN chmod +x /usr/bin/dep
+ARG DEP_RELEASE_TAG="v0.5.4"
```
liyinan926, line 20: Just realized we recently changed to use gomodule and removed `Gopkg.toml` and `Gopkg.lock`.

liyinan926: FYI: I created #919 to fix the Dockerfiles.

Issue #934, opened 2020-05-29
containerPort is not added to driver and executor pods when specifying monitoring in app spec


AliGouta:
Jmx exporter works well and it does export metrics **locally** that might be scrapped by prometheus. However, there is no `containerPort` added to the pods after starting the application so that Prometheus may scrap these metrics through a `Service/ServiceMonitor` for instance.

**To reproduce, add the following in the application spec:**

```
  monitoring:
    exposeDriverMetrics: true
    exposeExecutorMetrics: true
    prometheus:
      jmxExporterJar: "/prometheus/jmx_prometheus_javaagent-0.12.0.jar"
      port: 9999
```

**On executor pods, the exposed ports are:**

```
    name: executor
    ports:
    - containerPort: 7079
      name: blockmanager
      protocol: TCP
    resources:
```

**On Driver pod, the exposed ports are:**

```
    name: spark-kubernetes-driver
    ports:
    - containerPort: 7078
      name: driver-rpc-port
      protocol: TCP
    - containerPort: 7079
      name: blockmanager
      protocol: TCP
    - containerPort: 4040
      name: spark-ui
      protocol: TCP
    resources:
```

**Exepected behaviour is to add the prometheus exporter port to the pod as a containerPort:**

```
    name: spark-kubernetes-driver
    ports:
    ....
    - containerPort: 9999
      protocol: TCP
    resources:
```

A workaround would be to add a sidecar container. But this is not proper.

A PR is on the way to fix this (already tested) :).

liyinan926:
Related: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/476, https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/825.

krish512:
Did something get changed? I still don't see the prometheus port getting creating even after passing the prometheus options to SparkApplication,
```
monitoring:
    exposeDriverMetrics: true
    exposeExecutorMetrics: true
    prometheus:
      jmxExporterJar: /opt/spark/jars/jmx_prometheus_javaagent-0.15.0.jar
      port: 8090
```

I've ensured my jar file is present in /opt/spark/jars/jmx_prometheus_javaagent-0.15.0.jar location

PR #935, opened 2020-05-29
Fixed issue (#934)

AliGouta:
The issue is related to [#934](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/934). It allows to properly exposing the prometheus exporter to a SeriveMonitor or to Prometheus to scrape metrics. Up to now, we can only see the metrics generated locally but not exposed outside the pod running the JMX exporter.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -186,6 +186,25 @@ func addVolumeMount(pod *corev1.Pod, mount corev1.VolumeMount) *patchOperation {
 	return &patchOperation{Op: "add", Path: path, Value: value}
 }
 
+func addPort(pod *corev1.Pod, containerPort corev1.ContainerPort) *patchOperation {
+	i := findContainer(pod)
+	if i < 0 {
+		glog.Warningf("not able to add containerPort %d as Spark container was not found in pod %s", containerPort.ContainerPort, pod.Name)
+		return nil
+	}
+	path := fmt.Sprintf("/spec/containers/%d/ports", i)
+	var value interface{}
+	if len(pod.Spec.Containers[i].Ports) == 0 {
+		value = []corev1.ContainerPort{containerPort}
+	} else {
+		path += "/-"
+		value = containerPort
+	}
+	pod.Spec.Containers[i].Ports = append(pod.Spec.Containers[i].Ports, containerPort)
+
+	return &patchOperation{Op: "add", Path: path, Value: value}
+}
+
 func addEnvVars(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperation {
 	var envVars []corev1.EnvVar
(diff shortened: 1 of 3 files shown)


pkg/webhook/patch.go
```diff
@@ -360,15 +379,31 @@ func addPrometheusConfigMap(pod *corev1.Pod, app *v1beta2.SparkApplication) []pa
 	name := config.GetPrometheusConfigMapName(app)
 	volumeName := name + "-vol"
 	mountPath := config.PrometheusConfigMapMountPath
+	port := config.DefaultPrometheusJavaAgentPort
```
liyinan926, line 382: Can we rename the method from `addPrometheusConfigMap` to `getPrometheusConfigPatches` given it's doing more than just adding the ConfigMap?

pkg/webhook/patch.go
```diff
@@ -360,15 +379,31 @@ func addPrometheusConfigMap(pod *corev1.Pod, app *v1beta2.SparkApplication) []pa
...
+	patchOps = append(patchOps, *portPatchOp)
 	return patchOps
 }
 
+func addContainerPort(pod *corev1.Pod, port int32, protocol string) *patchOperation {
+	containerPort := corev1.ContainerPort{
+		ContainerPort: port,
+		Protocol:      corev1.Protocol(protocol),
+	}
+	return addPort(pod, containerPort)
```
liyinan926, line 404: Looks like you can just combine `addContainerPort` and `addPort` into `addContainerPort`.

pkg/webhook/patch.go
```diff
@@ -360,15 +379,31 @@ func addPrometheusConfigMap(pod *corev1.Pod, app *v1beta2.SparkApplication) []pa
...
+	if app.Spec.Monitoring.Prometheus.Port != nil {
+		port = *app.Spec.Monitoring.Prometheus.Port
+	}
+	protocol := config.DefaultPrometheusPortProtocol
+
 	patchOps = append(patchOps, addConfigMapVolume(pod, name, volumeName))
 	vmPatchOp := addConfigMapVolumeMount(pod, volumeName, mountPath)
-	if vmPatchOp == nil {
+	portPatchOp := addContainerPort(pod, port, protocol)
+	if vmPatchOp == nil || portPatchOp == nil {
```
liyinan926, line 391: Please separate the two conditions out and handle them separately.

pkg/webhook/patch.go
AliGouta: Agress. Done.

pkg/webhook/patch.go
AliGouta: Done

pkg/webhook/patch.go
AliGouta: Done.

PR #940, opened 2020-06-03
Populate Spark Application labels to Driver and Executors

gongx:
Populate Spark Application labels to Driver and Executors

The original discussion thread: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/905.


gongx:
@googlebot I signed it!

Issue #942, opened 2020-06-09
Unit test to be fixed


AliGouta:
The last commit leads to some unstable behaviour when running tests in `submission_test.go`.
Tests fail sometimes in the following [assertion](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/submission_test.go#L415). And several times run fine. I suspect that is related to the following [map](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/submission.go#L271) : 

`driverLabels := make(map[string]string)`

However, labels seems to be not inserted in order all the time in slice `driverOptions` [herein](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/submission_test.go#L452), eventhough we call `sort.Strings(executorOptions)` [here](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/submission_test.go#L456). 

This makes `driverOptions[3]` in this [line](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/submission_test.go#L461) not matching the assertion since we do not have a guarantee on the order how labels are inserted in the slice. A PR is on the way.

PR #943, opened 2020-06-09
(issue #942) Fix unit tests in submission_test

AliGouta:
related to issue [#942](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/942).

### DIFF
--- a/pkg/controller/sparkapplication/submission_test.go
+++ b/pkg/controller/sparkapplication/submission_test.go
@@ -19,6 +19,7 @@ package sparkapplication
 import (
 	"fmt"
 	"github.com/google/uuid"
+	"reflect"
 	"sort"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/controller/sparkapplication/submission_test.go
```diff
@@ -398,22 +398,40 @@ func TestPopulateLabels_Driver_Executor(t *testing.T) {
 		t.Fatal(err)
 	}
 	assert.Equal(t, 5, len(driverOptions))
-	assert.Equal(t, fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "app-name", "spark-test"), driverOptions[0])
-	assert.Equal(t, fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "launched-by-spark-operator", strconv.FormatBool(true)), driverOptions[1])
-	assert.Equal(t, fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "submission-id", submissionID), driverOptions[2])
-	assert.Equal(t, fmt.Sprintf(SparkDriverLabelTemplate, AppLabelKey, AppLabelValue), driverOptions[3])
-	assert.Equal(t, fmt.Sprintf(SparkDriverLabelTemplate, DriverLabelKey, DriverLabelValue), driverOptions[4])
+	ExpectedDriverLabels := []string{
```
liyinan926, line 401: Variables should start with a lower case letter.

pkg/controller/sparkapplication/submission_test.go
```diff
@@ -398,22 +398,40 @@ func TestPopulateLabels_Driver_Executor(t *testing.T) {
...
+		fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "app-name", "spark-test"),
+		fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "launched-by-spark-operator", strconv.FormatBool(true)),
+		fmt.Sprintf(SparkDriverLabelAnnotationTemplate, "submission-id", submissionID),
+		fmt.Sprintf(SparkDriverLabelTemplate, AppLabelKey, AppLabelValue),
+		fmt.Sprintf(SparkDriverLabelTemplate, DriverLabelKey, DriverLabelValue),
+	}
+
+	for _, label := range driverOptions {
```
liyinan926, line 409: You don't need to do this. The built-in function `reflect.DeepEqual` can be used to check map equality.

pkg/controller/sparkapplication/submission_test.go
AliGouta: Done.

pkg/controller/sparkapplication/submission_test.go
AliGouta: `driverOptions` is a slice: `driverOptions, err := addDriverConfOptions(app, submissionID)`. `reflect.DeepEqual` is indeed a good option however, I am not sure it fits what we need. Elements within `driverOptions` may not be inserted in a sorted way since within the function `addDriverConfOptions` we have a non sorted map `driverLabels`. 

Though, back to our test we may have `driverOptions` and `expectedDriverLabels` that have the same labels, however these labels might be sorted in a different way which makes IMHO `reflect.DeepEqual` not appropriate. Please, feel free to propose another alternative.

liyinan926: You could sort both slices before calling `DeepEqual`.

AliGouta: right :) I changed it.

Issue #929, opened 2020-05-27
Filtering on namespaces the controller listens to


AliGouta:
We need to have several instances of the operator. Each one listen Events from a different set of Namespaces. For example, imagine we want to have an instance of the operator that listens to namespaces dedicated to Dev environment and another one for staging environment. Both share the same cluster. Today, this is not supported as we either specify one or all namespaces. Looking at the code, we see two potential options, the first one seems hard to be implemented and even not sure that it is doable, the second option should be easier:

**1st option:**
```
There might be something to be done here:

	eventBroadcaster := record.NewBroadcaster()
	eventBroadcaster.StartLogging(glog.V(2).Infof)
	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{
		Interface: kubeClient.CoreV1().Events(namespace),
	})
	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, EventSource{Component: "spark-operator"})
```

**2nd option:**

```
We modify onAdd, onUpdate and onDelete functions:

	crdInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    controller.onAdd,
		UpdateFunc: controller.onUpdate,
		DeleteFunc: controller.onDelete,
	})
```

and prevent enqueuing the event if the namespace is not in the range of namespaces defined as a property while configuring the operator (a regex for example). More precisely, we change the following function:

```
func (c *Controller) onAdd(obj interface{}) {
	app := obj.(*v1beta2.SparkApplication)
	v1beta2.SetSparkApplicationDefaults(app)
	glog.Infof("SparkApplication %s/%s was added, enqueueing it for submission", app.Namespace, app.Name)
	c.enqueue(app)
}

and add something like: 
        if app.Namespace not_in ListOfDesiredNamespaces
           do_not_enqueue (do_nothing)
```

We did not really find a solution for option one, there is no something similar to `predicate.Funcs` in `controller-runtime` where we can call `WithEventFilter`. Though, we think we can implement the behaviour of `predicates` by changing the behaviour of `onAdd, onUpdate and onDelete functions` in spark operator.

We want to have your opinion before starting the implementation. This feature would be really interesting for us and respond to many cases where we need to seperate environments within the same kubernetes cluster.

PR #944, opened 2020-06-10
(#929) support filtering events by namespace

AliGouta:
Related to [#929](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/929).

In this PR we make it possible to users to specify on which namespace(s) the operator would run/manage spark applications. The k8s client, allows listing/watching either on one single namespace or all namespaces. Though, we use a `FilteringResourceEventHandler` handler to filter on a specific namespaces using compiled regexp supplied to the operator. This is similar to the behavior of `predicates` in `kubebuilder`. This fits well the requirement if we want to separate logically our kubernetes cluster assuming that it must run both `dev` and `uat` environments within the same cluster. Or, if we want to test different releases of the operator within the same cluster. Hence, running two instances of the same operator with different `namespace-filter` option becomes possible (i.e. one to cover *-uat and the other *-dev).

If the PR is fine, I can update the documentation as well and explains how we can have seperate instances of the operator running within the same cluster and make sure that the webhook1 of operator1 does not handle mutation of Pods handled by operator2.

If `namespace-filter` is not specified or `namespace` is specified, then we should have the same behavior as before.

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ var (
 	controllerThreads              = flag.Int("controller-threads", 10, "Number of worker threads used by the SparkApplication controller.")
 	resyncInterval                 = flag.Int("resync-interval", 30, "Informer resync interval in seconds.")
 	namespace                      = flag.String("namespace", apiv1.NamespaceAll, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset.")
+	namespaceFilter                = flag.String("namespace-filter", "", "A comma-separated list of regexps to manage specific Kubernetes namespaces. Ignored if namespace if specified")
 	enableWebhook                  = flag.Bool("enable-webhook", false, "Whether to enable the mutating admission webhook for admitting and patching Spark pods.")
 	enableResourceQuotaEnforcement = flag.Bool("enable-resource-quota-enforcement", false, "Whether to enable ResourceQuota enforcement for SparkApplication resources. Requires the webhook to be enabled.")
 	ingressURLFormat               = flag.String("ingress-url-format", "", "Ingress URL format.")
@@ -162,6 +163,11 @@ func main() {
 	crInformerFactory := buildCustomResourceInformerFactory(crClient)
 	podInformerFactory := buildPodInformerFactory(kubeClient)
 
+	var namespaceFilterConfig = util.NamespaceFilterConfig{
+		Namespace:       *namespace,
+		NamespaceFilter: *namespaceFilter,
+	}
+
 	var metricConfig *util.MetricConfig
 	if *enableMetrics {
 		metricConfig = &util.MetricConfig{
@@ -177,7 +183,7 @@ func main() {
 	}
 
 	applicationController := sparkapplication.NewController(
-		crClient, kubeClient, crInformerFactory, podInformerFactory, metricConfig, *namespace, *ingressURLFormat, batchSchedulerMgr)
+		crClient, kubeClient, crInformerFactory, podInformerFactory, metricConfig, *namespace, &namespaceFilterConfig, *ingressURLFormat, batchSchedulerMgr)
 	scheduledApplicationController := scheduledsparkapplication.NewController(
 		crClient, kubeClient, apiExtensionsClient, crInformerFactory, clock.RealClock{})
 
--- a/main.go
+++ b/main.go
@@ -57,6 +57,7 @@ var (
 	resyncInterval                 = flag.Int("resync-interval", 30, "Informer resync interval in seconds.")
 	namespace                      = flag.String("namespace", apiv1.NamespaceAll, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset.")
 	namespaceFilter                = flag.String("namespace-filter", "", "A comma-separated list of regexps to manage specific Kubernetes namespaces. Ignored if namespace if specified")
+	labelsFilter                   = flag.String("labels-filter", "", "A comma-separated list of key=value labels to filter events during Watch and List.")
 	enableWebhook                  = flag.Bool("enable-webhook", false, "Whether to enable the mutating admission webhook for admitting and patching Spark pods.")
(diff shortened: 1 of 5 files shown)


pkg/controller/sparkapplication/controller.go
```diff
@@ -125,22 +128,27 @@ func newSparkApplicationController(
 		controller.metrics = newSparkAppMetrics(metricsConfig)
 		controller.metrics.registerMetrics()
 	}
+	namespaceFilter := util.GetNamespaceFilter(namespaceFilterConfig)
 
 	crdInformer := crdInformerFactory.Sparkoperator().V1beta2().SparkApplications()
-	crdInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+	crdEventHandlerFuncs := cache.ResourceEventHandlerFuncs{
```
liyinan926, line 134: You construct this within `getEventResourceHandlerFuncs`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -125,22 +128,27 @@ func newSparkApplicationController(
...
 		AddFunc:    controller.onAdd,
 		UpdateFunc: controller.onUpdate,
 		DeleteFunc: controller.onDelete,
-	})
+	}
+	crdEventResourceHandlerFuncs := getEventResourceHandlerFuncs(crdEventHandlerFuncs, namespaceFilter)
```
liyinan926, line 139: Please rename both the variable and function names to use `ResourceEvent` instead of `EventResource`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -150,6 +158,34 @@ func newSparkApplicationController(
 	return controller
 }
 
+func getEventResourceHandlerFuncs(handlerFuncs cache.ResourceEventHandlerFuncs, namespaceFilter *regexp.Regexp) cache.ResourceEventHandler {
+	var resourceEventHandlerFuncs cache.ResourceEventHandler
+
+	if namespaceFilter == nil {
```
liyinan926, line 164: You can simplify the logic of this function:

```
if namespaceFilter == nil {
    return handlerFuncs
}

return cache.FilteringResourceEventHandler{
    FilterFunc: filterSparkAndPodEventsByNamespace(*namespaceFilter),
    Handler:    handlerFuncs,
}
```

main.go
```diff
@@ -162,6 +163,11 @@ func main() {
 	crInformerFactory := buildCustomResourceInformerFactory(crClient)
 	podInformerFactory := buildPodInformerFactory(kubeClient)
 
+	var namespaceFilterConfig = util.NamespaceFilterConfig{
```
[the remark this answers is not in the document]
AliGouta: Done.

main.go
```diff
@@ -177,7 +183,7 @@ func main() {
 	}
 
 	applicationController := sparkapplication.NewController(
-		crClient, kubeClient, crInformerFactory, podInformerFactory, metricConfig, *namespace, *ingressURLFormat, batchSchedulerMgr)
+		crClient, kubeClient, crInformerFactory, podInformerFactory, metricConfig, *namespace, &namespaceFilterConfig, *ingressURLFormat, batchSchedulerMgr)
```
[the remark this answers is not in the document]
AliGouta: Done.

pkg/controller/sparkapplication/controller.go
AliGouta: Done.

pkg/controller/sparkapplication/controller.go
AliGouta: Done.

AliGouta:
Regarding your comment : "instead of using a regex filter, can't we just filter by some label(s) on the namespaces? ... you can attach a label to the namespaces to indicate which environment they belong to...". 
This would be really interesting. However, to make sure I correctly understood :

1- Do you mean "a label we attach to a namespace" is like doing:

`kubectl label namespaces spark-dev  deploy-spark-dev=true`

or a label we attach to objects within these namespaces ?

2- Is it a kind of filter where we filter events with `FilteringResourceEventHandler` or at the creation of the informer with the function `tweakListOptionsFunc` ? I tried the latter with the label attached to the namespace `(i.e. deploy-spark-dev=true)`: It did not work and the controller do to catches anything. However, when setting the label (deploy-spark-dev=true) on the SparkApplication/Pod objects. Then, it works and this should be fine for me since I can automate and attach this label to all spark deployments where I have a namespace that matches : *-dev.

This being said, I updated the PR with a new` labels-filter `option, where I can filter on objects with custom labels. Does this sounds good for you ? Shall we keep either on `labels-filter` or `namespace-filter` or keep on both and then let the end-user choosing what fits his requirement ? Both seems to be fine to me. As I said, I will update the documentation with respect to what we will agree on. Thanks !

liyinan926:
> 1- Do you mean "a label we attach to a namespace" or a label we attach to objects within these namespaces ?

Labels attached to the namespace resources.

> 2- Is it a kind of filter where we filter events with FilteringResourceEventHandler or at the creation of the informer with the function tweakListOptionsFunc ? I tried the latter with the label attached to the namespace (i.e. deploy-spark-dev=true): It did not work and the controller do to catches anything. However, when setting the label (deploy-spark-dev=true) on the SparkApplication/Pod objects. Then, it works and this should be fine for me since I can automate and attach this label to all spark deployments where I have a namespace that matches : *-dev.

If this is the case, putting the labels on the `SparkApplication` objects is fine as long as it serves the filtering purpose.

> This being said, I updated the PR with a newlabels-filteroption, where I can filter on objects with custom labels. Does this sounds good for you ? Shall we keep either on labels-filter or namespace-filter or keep on both and then let the end-user choosing what fits his requirement ? Both seems to be fine to me. As I said, I will update the documentation with respect to what we will agree on. Thanks !

I don't think we need both the label-filter and namespace-filter. If the label-filter serves your purpose, we can keep that and drop the namespace filter.

AliGouta:
I will close this pull request since it is a duplicate. The other PR is more simple.

PR #952, opened 2020-06-22
support filtering resources on custom labels

AliGouta:
Related to [#929](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/929).

As discussed in a previous PR. This allows to define some custom labels on which the operator will list and watch resources. For instance, this is useful if we want to run multiple instances of the operator.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -44,6 +44,7 @@ The Kubernetes Operator for Apache Spark ships with a command-line tool called `
 * [Running Spark Applications on a Schedule using a ScheduledSparkApplication](#running-spark-applications-on-a-schedule-using-a-scheduledsparkapplication)
 * [Enabling Leader Election for High Availability](#enabling-leader-election-for-high-availability)
 * [Enabling Resource Quota Enforcement](#enabling-resource-quota-enforcement)
+* [Running Multiple Instances Of The Operator Within The Same K8s Cluster](#running-multiple-instances-of-the-operator-within-the-same-k8s-cluster)
 * [Customizing the Operator](#customizing-the-operator)
 
 ## Using a SparkApplication
@@ -751,6 +752,31 @@ The operator supports a high-availability (HA) mode, in which there can be more 
 The Spark Operator provides limited support for resource quota enforcement using a validating webhook. It will count the resources of non-terminal-phase SparkApplications and Pods, and determine whether a requested SparkApplication will fit given the remaining resources. ResourceQuota scope selectors are not supported, any ResourceQuota object that does not match the entire namespace will be ignored. Like the native Pod quota enforcement, current usage is updated asynchronously, so some overscheduling is possible.
 
 If you are running Spark applications in namespaces that are subject to resource quota constraints, consider enabling this feature to avoid driver resource starvation. Quota enforcement can be enabled with the command line arguments `-enable-resource-quota-enforcement=true`. It is recommended to also set `-webhook-fail-on-error=true`.
+
+## Running Multiple Instances Of The Operator Within The Same K8s Cluster
+
+If you need to run multiple instances of the operator within the same k8s cluster. Therefore, you need to make sure that the running instances should not compete for the same custom resources or pods. You can achieve this:
+
+Either:
+* By specifying a different `namespace` flag for each instance of the operator.
+
+Or if you want your operator to watch specific resources that may exist in different namespaces:
+
+* You need to add custom labels on resources by defining for each instance of the operator a different set of labels in `-label-selector-filter (e.g. env=dev,app-type=spark)`.
+* Run different `webhook` instances by specifying different `-webhook-config-name` flag for each deployment of the operator.
+* Specify different `webhook-svc-name` and/or `webhook-svc-namespace` for each instance of the operator. 
+* Edit the job that generates the certificates `webhook-init` by specifying the namespace and the service name of each instance of the operator, `e.g. command: ["/usr/bin/gencerts.sh", "-n", "ns-op1", "-s", "spark-op1-webhook", "-p"]`. Where `spark-op1-webhook` should match what you have specified in `webhook-svc-name`. For instance, if you use the following [helm chart](https://github.com/helm/charts/tree/master/incubator/sparkoperator) to deploy the operator you may specify for each instance of the operator a different `--namespace` and `--name-template` arguments to make sure you generate a different certificate for each instance, e.g:
+```
+helm install spark-op1 incubator/sparkoperator --namespace ns-op1
+helm install spark-op2 incubator/sparkoperator --namespace ns-op2
+```
+Will run 2 `webhook-init` jobs. Each job executes respectively the command:
+```
+command: ["/usr/bin/gencerts.sh", "-n", "ns-op1", "-s", "spark-op1-webhook", "-p"`]
+command: ["/usr/bin/gencerts.sh", "-n", "ns-op2", "-s", "spark-op2-webhook", "-p"`]
+```
+
+* Although resources are already filtered with respect to the specified labels on resources. You may also specify different labels in `-webhook-namespace-selector` and attach these labels to the namepsaces on which you want the webhook to listen to.
 
 ## Customizing the Operator
 
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ var (
 	controllerThreads              = flag.Int("controller-threads", 10, "Number of worker threads used by the SparkApplication controller.")
 	resyncInterval                 = flag.Int("resync-interval", 30, "Informer resync interval in seconds.")
 	namespace                      = flag.String("namespace", apiv1.NamespaceAll, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset.")
+	labelSelectorFilter            = flag.String("label-selector-filter", "", "A comma-separated list of key=value, or key labels to filter resources during watch and list based on the specified labels.")
 	enableWebhook                  = flag.Bool("enable-webhook", false, "Whether to enable the mutating admission webhook for admitting and patching Spark pods.")
 	enableResourceQuotaEnforcement = flag.Bool("enable-resource-quota-enforcement", false, "Whether to enable ResourceQuota enforcement for SparkApplication resources. Requires the webhook to be enabled.")
 	ingressURLFormat               = flag.String("ingress-url-format", "", "Ingress URL format.")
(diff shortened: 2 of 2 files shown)


docs/user-guide.md
```diff
@@ -752,6 +753,14 @@ The Spark Operator provides limited support for resource quota enforcement using
 
 If you are running Spark applications in namespaces that are subject to resource quota constraints, consider enabling this feature to avoid driver resource starvation. Quota enforcement can be enabled with the command line arguments `-enable-resource-quota-enforcement=true`. It is recommended to also set `-webhook-fail-on-error=true`.
 
+## Running multiple instances of the operator within the same k8s cluster
+
+If you need to run multiple instances of the operator within the same k8s cluster. Therefore, you need to make sure that the running instances should not compete for the same custom resources or pods. You can achieve this:
+
+* By specifying custom labels on resources by defining for each instance of the operator a different `-label-selector-filter (e.g. env=dev,app-type=spark)` option for custom resources such as `SparkApplication`.
+* Running different `webhook` instances by specifying different `-webhook-svc-name` option for each deployment of the operator.
```
liyinan926, line 761: We should also mention the `namespace` flag for specifying the namespace to watch for objects and `webhook-svc-namespace` for specifying the namespace of the webhook service resource.

main.go
```diff
@@ -56,6 +56,7 @@ var (
 	controllerThreads              = flag.Int("controller-threads", 10, "Number of worker threads used by the SparkApplication controller.")
 	resyncInterval                 = flag.Int("resync-interval", 30, "Informer resync interval in seconds.")
 	namespace                      = flag.String("namespace", apiv1.NamespaceAll, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset.")
+	labelSelectorFilter            = flag.String("label-selector-filter", "", "A comma-separated list of key=value, or key labels to filter events when watching and listing resources.")
```
liyinan926, line 59: The flag description mentioned `filter events`, which seems irrelevant as it is for filtering resources based on labels.

main.go
```diff
@@ -265,6 +272,9 @@ func buildPodInformerFactory(kubeClient clientset.Interface) informers.SharedInf
 	}
 	tweakListOptionsFunc := func(options *metav1.ListOptions) {
 		options.LabelSelector = fmt.Sprintf("%s,%s", operatorConfig.SparkRoleLabel, operatorConfig.LaunchedBySparkOperatorLabel)
+		if len(*labelSelectorFilter) > 0 {
+			options.LabelSelector = options.LabelSelector + *labelSelectorFilter
```
liyinan926, line 276: Missing comma in between selectors.

docs/user-guide.md
```diff
@@ -752,6 +753,14 @@ The Spark Operator provides limited support for resource quota enforcement using
 
 If you are running Spark applications in namespaces that are subject to resource quota constraints, consider enabling this feature to avoid driver resource starvation. Quota enforcement can be enabled with the command line arguments `-enable-resource-quota-enforcement=true`. It is recommended to also set `-webhook-fail-on-error=true`.
 
+## Running multiple instances of the operator within the same k8s cluster
+
+If you need to run multiple instances of the operator within the same k8s cluster. Therefore, you need to make sure that the running instances should not compete for the same custom resources or pods. You can achieve this:
+
+* By specifying custom labels on resources by defining for each instance of the operator a different `-label-selector-filter (e.g. env=dev,app-type=spark)` option for custom resources such as `SparkApplication`.
```
liyinan926, line 760: An alternative is to use the flag `namespace` to limit the operator to watching resources in a single namespace.

docs/user-guide.md
AliGouta: Actually, I propose 2 options (the second one is more interesting IMHO):
- We specify one single `namespace` (this is doable by specifying the flag namespace)
- Suppose we have different namespaces for environment dev, then : We add label `env:dev` to resources within these namespaces. This makes the operator listening to all namespaces + filter according to the specified labels.

main.go
AliGouta: done.

main.go
AliGouta: My bad ! Done.

docs/user-guide.md
AliGouta: As discussed above. It is ok for you ?

docs/user-guide.md
```diff
@@ -752,6 +753,21 @@ The Spark Operator provides limited support for resource quota enforcement using
 
 If you are running Spark applications in namespaces that are subject to resource quota constraints, consider enabling this feature to avoid driver resource starvation. Quota enforcement can be enabled with the command line arguments `-enable-resource-quota-enforcement=true`. It is recommended to also set `-webhook-fail-on-error=true`.
 
+## Running multiple instances of the operator within the same k8s cluster
```
liyinan926, line 756: Please use the capital case for the first letter of the words.

docs/user-guide.md
```diff
@@ -752,6 +753,21 @@ The Spark Operator provides limited support for resource quota enforcement using
...
+
+If you need to run multiple instances of the operator within the same k8s cluster. Therefore, you need to make sure that the running instances should not compete for the same custom resources or pods. You can achieve this:
+
+Either:
+* By specifying a different `namespace` flag for each instance of the operator.
+
+OR if you want your operator to watch specific resources that may exist on different namespaces:
```
liyinan926, line 763: `on different namespaces ` -> `in different namespaces`.

docs/user-guide.md
```diff
@@ -752,6 +753,21 @@ The Spark Operator provides limited support for resource quota enforcement using
...
+
+* You need to add custom labels on resources by defining for each instance of the operator a different set of labels in `-label-selector-filter (e.g. env=dev,app-type=spark)`.
+* Run different `webhook` instances by specifying different `-webhook-config-name` flag for each deployment of the operator.
+* Specify different `webhook-svc-name` and/or `webhook-svc-namespace` for each instance of the operator. 
+* Edit the job that generates the certificates `webhook-init` by specifying the namespace and the service name of each instance of the operator, `e.g. "/usr/bin/gencerts.sh", "-n", "ns-op1", "-s", "spark-webhook-op1", "-p"`. Where `spark-webhook-op1` should match what you have specified in `webhook-svc-name`.
+* (Optional) Although resources are already filtred with respect to the specified labels on resources. You may also specify different labels in `-webhook-namespace-selector` and attach these labels to the namepsaces on which you want the webhook to listen to.
```
liyinan926, line 769: This reminds me the following question: given that `webhook-namespace-selector` allows the webhook to filter namespaces by labels on them, can we do the same for custom resources and pods and filter them by labels on namespaces also, instead of having a different semantic of filtering? This way, it's less confusing to users.

AliGouta: I investigated this option. But it seems to be not possible or at least not supported natively. The type webhook proposes a [namespaceSelector](https://github.com/kubernetes/api/blob/master/admissionregistration/v1beta1/types.go#L234) which allows filtering natively on namespaces using labels. All we have to do is to specify the list of labels in `-webhook-namespace-selector`. This changes with regard to the custom resources/pods and I don't see a straight solution to do filtering based on lables attached to namsepaces. I tried to set labels on pods/crs the same as the ones set on the namespace. But pods/crs does not inherit theses labels and no filtering happens.

The cleanest way IMHO, Filtering should be done using `informers.WithNamespace` and/or `informers.WithTweakListOptions`. The latter exposes a list of possible options to filter on, which we can find [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L335). Labels gives a clean and finer way to filter resources even within the namespace itself. For this reason and as per my understanding, we really need to seperate filtering semantic: NamespaceSelector on namespace for the webhook and labelSelector on resources for crs/pods.

docs/user-guide.md
AliGouta: Done.

docs/user-guide.md
AliGouta: Done.

docs/user-guide.md
```diff
@@ -752,6 +753,21 @@ The Spark Operator provides limited support for resource quota enforcement using
...
+
+Either:
+* By specifying a different `namespace` flag for each instance of the operator.
+
+Or if you want your operator to watch specific resources that may exist in different namespaces:
+
+* You need to add custom labels on resources by defining for each instance of the operator a different set of labels in `-label-selector-filter (e.g. env=dev,app-type=spark)`.
+* Run different `webhook` instances by specifying different `-webhook-config-name` flag for each deployment of the operator.
+* Specify different `webhook-svc-name` and/or `webhook-svc-namespace` for each instance of the operator. 
+* Edit the job that generates the certificates `webhook-init` by specifying the namespace and the service name of each instance of the operator, `e.g. "/usr/bin/gencerts.sh", "-n", "ns-op1", "-s", "spark-webhook-op1", "-p"`. Where `spark-webhook-op1` should match what you have specified in `webhook-svc-name`.
```
liyinan926, line 768: It's worth mentioning how this can be achieved when using Helm to install the operator as both the namespace and name are parameterized in the Helm template that calls `gencerts.sh`.

AliGouta: Ok, I proposed somthing. We are used to Kustomize to deploy on GKE. I hope what I proposed makes it clearer.

docs/user-guide.md
```diff
@@ -752,6 +753,31 @@ The Spark Operator provides limited support for resource quota enforcement using
...
+
+* You need to add custom labels on resources by defining for each instance of the operator a different set of labels in `-label-selector-filter (e.g. env=dev,app-type=spark)`.
+* Run different `webhook` instances by specifying different `-webhook-config-name` flag for each deployment of the operator.
+* Specify different `webhook-svc-name` and/or `webhook-svc-namespace` for each instance of the operator. 
+* Edit the job that generates the certificates `webhook-init` by specifying the namespace and the service name of each instance of the operator, `e.g. command: ["/usr/bin/gencerts.sh", "-n", "ns-op1", "-s", "spark-op1-webhook", "-p"]`. Where `spark-op1-webhook` should match what you have specified in `webhook-svc-name`. For instance, if you use the following [helm chart](https://github.com/helm/charts/tree/master/incubator/sparkoperator) to deploy the operator you may specify for each instance of the operator a different `--namespace` and `--name-template` arguments to make sure you generate a different certificate for each instance, e.g:
+```
+helm [install/template] incubator/sparkoperator --namespace ns-op1 --name-template spark-op1
```
liyinan926, line 770: I don't see the option `--name-template`. Is it the same as `.Release.Name`?

AliGouta: It is in Helm3. I am not really used to helm. I like more Kustomize. I changed the doc. But Yes, the goal is to fill `.Release.Name`. Now, I think the doc is better. Please, let me know if we need to change it.

PR #990, opened 2020-07-29
Allow any tag not just annotated tags in gitlab builds

jkleckner:
Change the use of git describe to allow any tag including non-annotated
tags.


jkleckner:
This was built in gitlab here: https://gitlab.com/jkleckner/spark-on-k8s-operator/-/jobs/660699732

PR #1020, opened 2020-09-16
Add PUBG to who is using

jacobhjkim:
Hi, [PUBG](https://careers.pubg.com/#/en/) has been using spark-on-k8s-operator.


jacobhjkim:
@googlebot I signed it!

PR #1033, opened 2020-09-29
Add Stitch Fix under who-is-using

nssalian:
In this PR:
1. Added Stitch Fix to the who-is-using doc.
2. Fixed some spacing nits for the doc itself to be consistent.


nssalian:
@googlebot I signed it!

Issue #842, opened 2020-03-21
Client mode examples?


mattdornfeld:
I see that client mode is an option when submitting a SparkApplication, but I can't seem to find an example of client mode being used. When I try to set client mode myself on the spark-pi example I get an error. Any working examples of creating a SparkApplication in client mode? I'm hoping to connect a Jupyter notebook to a created Spark cluster.

liyinan926:
Although `client` mode is an option, we have never figured out what it means for the operator to support the client mode, so it's currently not usable.

Bole0331:
Hi, can you help me understand why even when I set deploy-mode as "cluster" it's still shown as "client" model in my Spark-UI environments?

Issue #580, opened 2019-08-23
Cannot set deploy-mode cluster


AceHack:
It always uses client no matter what I put in SparkApplication for mode

liyinan926:
The driver pod launched in the cluster mode runs `spark-submit` in client mode internally to run the driver program. That's why you will always see the `client` mode being used if you check the command run in the driver container. The client mode is not supported by the operator.

AceHack:
Thank you

PR #1034, opened 2020-10-01
Adding details on deploy mode to user guide

psschwei:
This PR adds some minor details about `.spec.deployMode` to the [User Guide](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/user-guide.md). The goal is to bring info from issues #580  and #842 into the main documentation.

It also fixes a broken link in the design docs.


liyinan926 (approved):
LGTM. Thanks!

docs/user-guide.md
```diff
@@ -75,6 +76,11 @@ spec:
   mainApplicationFile: local:///opt/spark/examples/jars/spark-examples_2.12-3.0.0.jar
 ```
 
+### Specifying Deployment Mode
+
+A `SparkApplication` should set `.spec.deployMode` to `cluster`, as `client` is not currently implemented. The driver pod will then run `spark-submit` in `client` mode internally to run the driver program. Additional details of how `SparkApplication`s are run can be found in the [design documentation](design.md#architecture).
```
rachitchauhan43, line 81: @psschwei : Shouldn't the field be `.spec.mode` instead of `.spec.deployMode` ?
Looking at examples here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/user-guide.md#writing-a-sparkapplication-spec

psschwei: It's been over two years since I've used the spark operator, so I really couldn't say.

Issue #889, opened 2020-04-25
Please support istio with auto inject


AceHack:
When istio is set to auto inject the executors fail to talk to the driver and jobs never finish.  Please add first-class support for istio.

ringtail:
@AceHack It's a little different in Istio. Istio don't support headless kubernetes svc by default. @liyinan926 Should we expose how to create the driver svc ?

AceHack:
Istio does support headless services as of 1.4
https://github.com/istio/istio/issues/7495

Istio does not work with job like things in Kubernetes
https://github.com/istio/istio/issues/6324

The problem is spark would need to use something like envoy pre-flight to properly wait for envoy to come up and then shut down pre-flight when the spark application ends.

https://github.com/monzo/envoy-preflight

We are using envoy-preflight for all other job in Kubernetes.

dmoore62:
One work around that we tested is configuring the application to run outside the istio service mesh.

AceHack:
Running outside the service mesh is not really a valid work around we want all the observability a service mesh brings with our spark jobs.

dmoore62:
```Working with the upstream community, we have added a configuration workaround for Zookeeper along with Casssandra, Elasticsearch, Redis and Apache NiFi.  I am sure there are additional applications that are not compatible with sidecars.  Please inform the community if you know of any.```

https://www.cncf.io/blog/2020/10/26/service-mesh-is-still-hard/

We have yet to really dig into enhancements in Spark 3, but we may need to accept that we are apart of the "applications that are not compatible with sidecars."

afilipchik:
hi, what is exactly incompatible? we added an istio waiter to the image and executor starts. But then fails with:
`20/11/19 07:55:58 INFO TransportClientFactory: Successfully created connection to spark.spark.svc/10.X.X.X:7078 after 99 ms (0 ms spent in bootstraps)
20/11/19 07:55:58 ERROR TransportResponseHandler: Still have 1 requests outstanding when connection from  spark.spark.svc/10.X.X.X:7078 is closed
Exception in thread "main" java.lang.reflect.UndeclaredThrowableException
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1713)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:64)
[...]

PR #1050, opened 2020-10-26
[ISSUE][#889] add example to run job outside of service mesh

dmoore62:


liyinan926 (approved):

examples/spark-pi-istio.yaml
```diff
@@ -0,0 +1,58 @@
+#
```
liyinan926, line 1: Should the file be named `spark-pi-without-istio`?

vara-bonthu:
@dmoore62, please rename the file to `spark-pi-without-istio.yaml` and add some notes at the beginning of the file explaining that this example excludes the Istio sidecars.

PR #1054, opened 2020-11-03
Propagate Spark app's priorityClassName to pods' spec

pgillet:
See https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/821 .

Currently, the priority info is only at the `SparkApplication` resource level. This ensures the scheduling order while enqueueing the Spark applications but does not allow for job preemption, as the driver and executor pods all end up with the same priority (no priority actually - 0).

This PR is a straightforward solution: we simply propagate `spec.batchSchedulerOptions.priorityClassName`, if any, to `spec.priorityClassName` in the driver and executor pods (exactly as the priority is propagated to the `PodGroup` for Volcano).

I see no reason to implement it differently:
- The priority should apply to the Spark application as a whole, that means the driver and its executor pods
- As a consequence, the `SparkApplication` only must hold the priority info
- The user is still forbidden to specify a `priorityClassName` directly in the driver or executor spec. This prevents inconsistencies like an executor's priority which would be higher than that of its driver, etc.

Note also that the field `priorityClassName` is valid whether with Volcano or with the default Kubernetes scheduler, contrary to what is suggested in the doc. 

Hope this will help those who needed it.

**Subsidiary question**: I am not proficient with Helm. How can I build the chart to test it locally?

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -526,7 +526,7 @@ func addSchedulerName(pod *corev1.Pod, app *v1beta2.SparkApplication) *patchOper
 func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) *patchOperation {
 	var priorityClassName *string
 
-	if app.Spec.BatchSchedulerOptions.PriorityClassName != nil {
+	if app.Spec.BatchSchedulerOptions != nil {
 		priorityClassName = app.Spec.BatchSchedulerOptions.PriorityClassName
 	}
 
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -68,6 +68,11 @@ func patchSparkPod(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperat
 		patchOps = append(patchOps, *op)
 	}
 
+	op = addPriorityClassName(pod, app)
+	if op != nil {
+		patchOps = append(patchOps, *op)
+	}
+
 	if pod.Spec.Affinity == nil {
 		op := addAffinity(pod, app)
 		if op != nil {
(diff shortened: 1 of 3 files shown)


pkg/webhook/patch.go
```diff
@@ -518,6 +523,19 @@ func addSchedulerName(pod *corev1.Pod, app *v1beta2.SparkApplication) *patchOper
 	return &patchOperation{Op: "add", Path: "/spec/schedulerName", Value: *schedulerName}
 }
 
+func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) *patchOperation {
+	var priorityClassName *string
+
+	if app.Spec.BatchSchedulerOptions.PriorityClassName != nil {
```
liyinan926, line 529: Looks like this if block is not needed and you can directly do:

```
priorityClassName := app.Spec.BatchSchedulerOptions.PriorityClassName
```

since `priorityClassName` is checked below.

liyinan926:
It also looks like the Travis build failed. Please fix the build.

pkg/webhook/patch.go
pgillet: My bad! We must test the existence of block `batchSchedulerOptions` before getting the `priorityClassName`.

liyinan926 (approved):
LGTM

HHK1:
@pgillet I'm trying to use the priorityClassName feature, but I'm getting an error when submitting the spark-app:

```
Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://172.20.0.1/api/v1/namespaces/spark-apps/pods. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. pods "sparkpi5-20210311-151901-umbra-driver" is forbidden: the integer value of priority (0) must not be provided in pod spec; priority admission controller computed 300 from the given PriorityClass name.
```

It seems that the operator submits an app with a priority 0, which is conflicting with the priority given by the priorityClassName. Any idea on how I could troubleshoot this ?

PR #1057, opened 2020-11-10
Default Logic: Add SparkConf check

akhurana001:
Only default Driver/Executor resources if not set both in Driver/Executor Spec and in SparkConf.

Ref: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1055 for details


liyinan926 (approved):
LGTM

PR #1059, opened 2020-11-11
Update volcano-integration.md

fisache:
webhook have to be enabled for volcano scheduler because it needs `mutating admission webhook`


liyinan926:
Maybe I'm missing something, but does using the Volcano scheduler need the webhook?

fisache:
@liyinan926 for injection of scheduler name `volcano` to driver/executor pod, it seem to use `mutating admission webhook`.

liyinan926 (approved):

PR #1064, opened 2020-11-15
Add incrmntal to who-is-using.md

scravy:
This adds incrmntal to the list of users of spark-on-k8s-operator. We are using the operator in production on EKS to operate spark without the needs for Hadoop and using cloud filesystems directly (using s3a output committer).

Best


scravy:
@googlebot I signed it!

PR #1072, opened 2020-11-18
Apply total resource when BatchSchedulerOption is not nil

fisache:
when `BatchSchedulerOption` is nil, requested resource make the operator pod to error state with error message
```
Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
```

The attributes could be nil because volcano batch scheduler accepts empty scheduler queue in `BatchSchedulerOptions`.

### DIFF
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -77,11 +77,8 @@ func (v *VolcanoBatchScheduler) syncPodGroupInClientMode(app *v1beta2.SparkAppli
 	if _, ok := app.Spec.Executor.Annotations[v1beta1.KubeGroupNameAnnotationKey]; !ok {
 		totalResource := getExecutorRequestResource(app)
 
-		if app.Spec.BatchSchedulerOptions != nil {
-			requestResources := app.Spec.BatchSchedulerOptions.Resources
-			if len(requestResources) > 0 {
-				totalResource = requestResources
-			}
+		if app.Spec.BatchSchedulerOptions != nil && len(app.Spec.BatchSchedulerOptions.Resources) > 0 {
+			totalResource = app.Spec.BatchSchedulerOptions.Resources
 		}
 		if err := v.syncPodGroup(app, 1, totalResource); err == nil {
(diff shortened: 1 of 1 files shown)


pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -75,12 +75,13 @@ func (v *VolcanoBatchScheduler) DoBatchSchedulingOnSubmission(app *v1beta2.Spark
...
 	if _, ok := app.Spec.Executor.Annotations[v1beta1.KubeGroupNameAnnotationKey]; !ok {
-		var totalResource corev1.ResourceList
-		requestResources := app.Spec.BatchSchedulerOptions.Resources
-		if len(requestResources) > 0 {
-			totalResource = requestResources
-		} else {
-			totalResource = getExecutorRequestResource(app)
+		totalResource := getExecutorRequestResource(app)
+
+		if app.Spec.BatchSchedulerOptions != nil {
```
liyinan926, line 80: You can do the following:

```
if app.Spec.BatchSchedulerOptions != nil && len(app.Spec.BatchSchedulerOptions.Resources) > 0 {
  totalResource = app.Spec.BatchSchedulerOptions.Resources
}
```

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -97,9 +98,12 @@ func (v *VolcanoBatchScheduler) syncPodGroupInClusterMode(app *v1beta2.SparkAppl
 	if _, ok := app.Spec.Driver.Annotations[v1beta1.KubeGroupNameAnnotationKey]; !ok {
 		//Both driver and executor resource will be considered.
 		totalResource := sumResourceList([]corev1.ResourceList{getExecutorRequestResource(app), getDriverRequestResource(app)})
-		requestResources := app.Spec.BatchSchedulerOptions.Resources
-		if len(requestResources) > 0 {
-			totalResource = requestResources
+
+		if app.Spec.BatchSchedulerOptions != nil {
```
liyinan926, line 102: Ditto.

pkg/batchscheduler/volcano/volcano_scheduler.go
fisache: @liyinan926 thanks it looks more clear

liyinan926 (approved):

PR #1079, opened 2020-11-20
Add scheduler func for clearing batch scheduling on completed

fisache:
Related to issue https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1062

The spark application injected with the volcano scheduler does not delete the pod group when the state of the spark application is changed to a completed or failed.
It can only be removed from the volcano queue by spark application deletion command or TTL timeout.
This will cause new pod groups to be scheduled to get stuck(pending) in the queue until spark application removed with ownerReference.
In my opinion, it would be better to remove the pod group when the spark application is changed to a completed or failed state.

### DIFF
--- a/pkg/batchscheduler/interface/interface.go
+++ b/pkg/batchscheduler/interface/interface.go
@@ -25,4 +25,5 @@ type BatchScheduler interface {
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
 	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
+	ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error
 }
--- a/pkg/batchscheduler/interface/interface.go
+++ b/pkg/batchscheduler/interface/interface.go
@@ -25,5 +25,5 @@ type BatchScheduler interface {
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
 	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
-	CleanupOnCompleted(app *v1beta2.SparkApplication) error
+	CleanupOnCompletion(app *v1beta2.SparkApplication) error
 }
--- a/pkg/batchscheduler/interface/interface.go
+++ b/pkg/batchscheduler/interface/interface.go
@@ -25,5 +25,5 @@ type BatchScheduler interface {
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
 	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
-	ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error
+	CleanupOnCompleted(app *v1beta2.SparkApplication) error
 }
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -160,6 +160,15 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
+func (v *VolcanoBatchScheduler) ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error {
+	// remove pod group
+	podGroupName := v.getAppPodGroupName(app)
+	if err := v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+	return nil
+}
+
 func New(config *rest.Config) (schedulerinterface.BatchScheduler, error) {
 	vkClient, err := volcanoclient.NewForConfig(config)
 	if err != nil {
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -160,13 +160,10 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
-func (v *VolcanoBatchScheduler) ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error {
-	// remove pod group
+func (v *VolcanoBatchScheduler) CleanupOnCompleted(app *v1beta2.SparkApplication) error {
 	podGroupName := v.getAppPodGroupName(app)
-	if err := v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{}); err != nil {
-		return err
-	}
-	return nil
+	//Remove pod group for Spark Application
+	return v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
 }
 
 func New(config *rest.Config) (schedulerinterface.BatchScheduler, error) {
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -160,7 +160,7 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
-func (v *VolcanoBatchScheduler) CleanupOnCompleted(app *v1beta2.SparkApplication) error {
+func (v *VolcanoBatchScheduler) CleanupOnCompletion(app *v1beta2.SparkApplication) error {
 	podGroupName := v.getAppPodGroupName(app)
 	//Remove pod group for Spark Application
 	return v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -162,8 +162,11 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 
 func (v *VolcanoBatchScheduler) CleanupOnCompletion(app *v1beta2.SparkApplication) error {
 	podGroupName := v.getAppPodGroupName(app)
-	//Remove pod group for Spark Application
-	return v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
+	err := v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	return nil
 }
 
 func New(config *rest.Config) (schedulerinterface.BatchScheduler, error) {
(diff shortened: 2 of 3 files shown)


fisache:
if want to add some test codes for volcano scheduler, I will write it :)

pkg/batchscheduler/interface/interface.go
```diff
@@ -25,4 +25,5 @@ type BatchScheduler interface {
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
 	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
+	ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error
```
liyinan926, line 28: Nit: `CleanupOnCompletion`.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -160,6 +160,15 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
+func (v *VolcanoBatchScheduler) ClearBatchSchedulingOnCompleted(app *v1beta2.SparkApplication) error {
+	// remove pod group
+	podGroupName := v.getAppPodGroupName(app)
+	if err := v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{}); err != nil {
```
liyinan926, line 166: You can do:

```
return v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
```

pkg/controller/sparkapplication/controller.go
```diff
@@ -597,6 +597,11 @@ func (c *Controller) syncSparkApplication(key string) error {
 			glog.Errorf("failed to update SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 			return err
 		}
+
+		if err := c.clearBatchScheduling(app, appCopy); err != nil {
+			glog.Errorf("failed to clean up batch scheduling %s/%s: %v", app.Namespace, app.Name, err)
```
liyinan926, line 602: `failed to clean up batch scheduling config for SparkApplication`

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1006,19 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
 
 	return false
 }
+
+// Clean up batch scheduler if use
```
liyinan926, line 1010: `Clean up batch scheduling config if applicable.`.

fisache:
@liyinan926 Thanks! Update code for your comments

fisache:
@liyinan926 PTAL

pkg/controller/sparkapplication/controller.go
```diff
@@ -597,6 +597,11 @@ func (c *Controller) syncSparkApplication(key string) error {
 			glog.Errorf("failed to update SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 			return err
 		}
+
+		if err := c.cleanUpBatchScheduling(app, appCopy); err != nil {
```
liyinan926, line 601: Just realized that we have a method for cleaning up application resources like the driver UI service. We should add deletion of this resource to that method.

liyinan926 (commented):
Sorry for the delay in review.

pkg/controller/sparkapplication/controller.go
fisache: Thanks for review!
I wonder if the function you are talking about is [deleteSparkResources](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/controller.go#L804).
As far as I understand, this function is called before the Spark application retry, which is called when not normal (invalidating, failed submission, etc.).
In the case of cleanup, the podgroup is deleted when it is in completed and failed state. This means that the spark application no longer restarts. If spark operator deletes the podgroup in the state of invalidating or failed submission, the spark application will not be queued to the volcano and will not be restarted until re-queued.
Please comment if it is different from what I understood.

liyinan926: Makes sense.

pkg/batchscheduler/interface/interface.go
```diff
@@ -25,4 +25,5 @@ type BatchScheduler interface {
 
 	ShouldSchedule(app *v1beta2.SparkApplication) bool
 	DoBatchSchedulingOnSubmission(app *v1beta2.SparkApplication) error
+	CleanupOnCompleted(app *v1beta2.SparkApplication) error
```
liyinan926, line 28: To be consistent, let's rename this to `CleanupOnCompletion`.

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1006,19 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
 
 	return false
 }
+
+// Clean up batch scheduling config if applicable.
+func (c *Controller) cleanUpBatchScheduling(oldApp, newApp *v1beta2.SparkApplication) error {
```
liyinan926, line 1011: Please rename this to `cleanUpOnTermination`, which sounds more generic and can be extended later to handle deletion of other resources, e.g., the UI service. Please update the comments also.

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1006,19 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
...
+	if needScheduling, scheduler := c.shouldDoBatchScheduling(newApp); needScheduling {
+		state := newApp.Status.AppState.State
+		// If new app state is completed or failed, no more app is running,
+		// and only once needs to clean up on completed or failed
+		if (state == v1beta2.CompletedState || state == v1beta2.FailedState) &&
```
liyinan926, line 1016: I suggest moving the state check to the caller of this method.

pkg/batchscheduler/interface/interface.go
fisache: Done

pkg/controller/sparkapplication/controller.go
fisache: fix it

pkg/controller/sparkapplication/controller.go
fisache: thanks, I moved it to caller

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1009,17 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
 
 	return false
 }
+
+// Clean up resources such as batch scheduler config if applicable.
```
liyinan926, line 1013: Comment needs to be updated.

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1009,17 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
...
+func (c *Controller) cleanUpOnTermination(oldApp, newApp *v1beta2.SparkApplication) error {
+	if needScheduling, scheduler := c.shouldDoBatchScheduling(newApp); needScheduling {
+		// If new app state is completed or failed, no more app is running,
```
liyinan926, line 1016: Comment needs update.

pkg/controller/sparkapplication/controller.go
fisache: updated it
`Clean up when the spark application is terminated.`

pkg/controller/sparkapplication/controller.go
fisache: updated it
`batch schduler is cleaned up only when state is changed to completion state`

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -160,6 +160,12 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
+func (v *VolcanoBatchScheduler) CleanupOnCompletion(app *v1beta2.SparkApplication) error {
+	podGroupName := v.getAppPodGroupName(app)
+	//Remove pod group for Spark Application
+	return v.volcanoClient.SchedulingV1beta1().PodGroups(app.Namespace).Delete(podGroupName, &metav1.DeleteOptions{})
```
liyinan926, line 166: We should ignore the error if it's NotFound.

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -160,6 +160,12 @@ func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size
 	return nil
 }
 
+func (v *VolcanoBatchScheduler) CleanupOnCompletion(app *v1beta2.SparkApplication) error {
+	podGroupName := v.getAppPodGroupName(app)
+	//Remove pod group for Spark Application
```
liyinan926, line 165: This comment is not needed.

pkg/controller/sparkapplication/controller.go
```diff
@@ -1001,3 +1009,16 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
 
 	return false
 }
+
+// Clean up when the spark application is terminated.
+func (c *Controller) cleanUpOnTermination(oldApp, newApp *v1beta2.SparkApplication) error {
+	if needScheduling, scheduler := c.shouldDoBatchScheduling(newApp); needScheduling {
+		// batch schduler is cleaned up only when state is changed to completion state
+		if newApp.Status.AppState.State != oldApp.Status.AppState.State {
```
liyinan926, line 1017: This check is not needed if deletion is idempotent, i.e., if we ignore NotFound error above.

pkg/batchscheduler/volcano/volcano_scheduler.go
fisache: remove it

pkg/controller/sparkapplication/controller.go
fisache: looks reasonable
updated to
```go
if needScheduling, scheduler := c.shouldDoBatchScheduling(newApp); needScheduling {
	if err := scheduler.CleanupOnCompletion(newApp); err != nil && !errors.IsNotFound(err) {
		return err
	}
}
```

pkg/controller/sparkapplication/controller.go
```diff
@@ -1013,11 +1013,8 @@ func (c *Controller) hasApplicationExpired(app *v1beta2.SparkApplication) bool {
 // Clean up when the spark application is terminated.
 func (c *Controller) cleanUpOnTermination(oldApp, newApp *v1beta2.SparkApplication) error {
 	if needScheduling, scheduler := c.shouldDoBatchScheduling(newApp); needScheduling {
-		// batch schduler is cleaned up only when state is changed to completion state
-		if newApp.Status.AppState.State != oldApp.Status.AppState.State {
-			if err := scheduler.CleanupOnCompletion(newApp); err != nil {
-				return err
-			}
+		if err := scheduler.CleanupOnCompletion(newApp); err != nil && !errors.IsNotFound(err) {
```
liyinan926, line 1016: Instead of dealing with the NotFound error here, why not return a nil if so in CleanupOnCompletion?

pkg/batchscheduler/volcano/volcano_scheduler.go
fisache: I saw this comments at now, I will update code that ignores NotFound error here

pkg/controller/sparkapplication/controller.go
fisache: you're right :)
Updated not found checking in CleanUpOnCompletion

liyinan926 (approved):
LGTM

PR #986, opened 2020-07-24
Manually fix the CRD validation issue in k8s 1.18

liyinan926:
Fixes https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/921.

PR #1084, opened 2020-11-24
Fix crds to add defaults for k8s 1.18+

jkleckner:
PR #1045 [1] appears to have accidentally reverted #986 [2] which added
defaults to the protocol field to work with versions of k8s 1.18+

This PR also draws on https://github.com/kudobuilder/operators/pull/297 [3]

[1] https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1045

[2] https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/986

[3] https://github.com/kudobuilder/operators/pull/297

### DIFF
--- a/charts/spark-operator-chart/crds/sparkoperator.k8s.io_scheduledsparkapplications.yaml
+++ b/charts/spark-operator-chart/crds/sparkoperator.k8s.io_scheduledsparkapplications.yaml
@@ -1,3 +1,5 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
@@ -793,6 +795,7 @@ spec:
                                   type: string
                               required:
                               - containerPort
+                              - protocol
                               type: object
                             type: array
                             x-kubernetes-list-map-keys:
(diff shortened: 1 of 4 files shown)


ordukhanian:
Hello,
Could you please release a new version of GCP Operator with this fix? 
We are using the previous helm chart which is now deprecated, and would like to upgrade.

jkleckner:
In the meantime, you can clone the repo and install using something like (adjust namespaces to your liking).
Because of [functional changes](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1085),
I actually think that the 0.8.6 chart version might also make sense to bundle in the repo.
```
helm \
  install \
  spark-operator \
  /path-to-your-clone/github.com/GoogleCloudPlatform/spark-on-k8s-operator/charts/spark-operator-chart \
  --namespace spark-operator \
  --set        sparkJobNamespace=spark \
  --set        enableWebhook=true \
  --set        webhook.enable=true \
  --set        webhookPort=443 \
  --set-string enableMetrics=true
```

PR #1090, opened 2020-12-01
Fix the role name to make rolebinding work

zhaohc10:


zhaohc10:
@ googlebot I signed it!

zhaohc10:
> googlebot I signed it!

@ googlebot I signed it!

liyinan926:
Thanks for the PR. Looks like this has been fixed in #1098.

PR #1099, opened 2020-12-03
Add prometheus containr port name

fisache:
`PodMonitor` is one of [prometheus operator](https://github.com/prometheus-operator/prometheus-operator) CRDs, and it collects metrics from pod selector.
For collecting, it requires port name of metrics
https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmetricsendpoint
port | Name of the pod port this endpoint refers to. Mutually exclusive with targetPort. | string
-- | -- | --

then, I have added metrics port name when mutating pod

it's example of `PodMonitor`
```yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: spark-pod-monitor
  labels:
    foo: bar
spec:
  selector:
    matchLabels:
      foo: bar
  podMetricsEndpoints:
  - port: metrics
    interval: 10s
```

### DIFF
--- a/charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml
+++ b/charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml
@@ -3600,6 +3600,8 @@ spec:
                       maximum: 49151
                       minimum: 1024
                       type: integer
+                    portName:
+                      type: string
                   required:
                   - jmxExporterJar
                   type: object
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -375,6 +375,11 @@ func getPrometheusConfigPatches(pod *corev1.Pod, app *v1beta2.SparkApplication) 
 		port = *app.Spec.Monitoring.Prometheus.Port
 	}
 	protocol := config.DefaultPrometheusPortProtocol
+	portName := config.DefaultPrometheusPortName
+	if app.Spec.Monitoring.Prometheus.PortName != nil {
+		portName = *app.Spec.Monitoring.Prometheus.PortName
+	}
+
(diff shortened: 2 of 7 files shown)


pkg/webhook/patch.go
```diff
@@ -401,6 +401,7 @@ func addContainerPort(pod *corev1.Pod, port int32, protocol string) *patchOperat
 
 	path := fmt.Sprintf("/spec/containers/%d/ports", i)
 	containerPort := corev1.ContainerPort{
+		Name:          "metrics",
```
liyinan926, line 404: The function `addContainerPort` is a generic one and not specific to the Prometheus port. What you probably should do is pass in the port name and specify the port name in the caller.

fisache: @liyinan926 
Hello,
There is a suggestion for this. https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1101
After your review is made, I will fix it.
Thanks!

fisache: @liyinan926 Added port name from caller

liyinan926 (approved):
LGTM

charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml
```diff
@@ -3600,6 +3600,8 @@ spec:
                       maximum: 49151
                       minimum: 1024
                       type: integer
+                    portName:
```
liyinan926, line 3603: This is also needed in the `ScheduledSparkApplication` CRD.

fisache: @liyinan926 Thanks, I updated it

PR #1100, opened 2020-12-03
Add configuration for SparkUI service type

### LINKED ISSUE #1096: Support configuring Service type for Spark UI
The Spark UI services are currently hardcoded to be ClusterIP types. Our Ingress controller requires services to be of type NodePort or LoadBalancer in order for our external load balancer to reach them. In order to support more Ingress implementations, the service type should be configurable.

jutley:
Fixes #1096. 

I also found that my controller-gen version behaved differently than the developer documentation described. I had to provide an additional configuration to use the v1beta1 CustomResourceDefinition version. I updated the docs and the generated CRDs accordingly.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -38,7 +38,7 @@ To update the auto-generated code, run the following command. (This step is only
 $ hack/update-codegen.sh
 ```
 
-To update the auto-generated CRD definitions, run the following command:
+To update the auto-generated CRD definitions, run the following command. After doing so, you must update the list of required fields under each `ports` field to add the `protocol` field to the list. Skipping this step will make the CRDs incompatible with Kubernetes v1.18+.
 
 ```bash
 $ GO111MODULE=off go get -u sigs.k8s.io/controller-tools/cmd/controller-gen
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -42,7 +42,7 @@ To update the auto-generated CRD definitions, run the following command:
 
 ```bash
 $ GO111MODULE=off go get -u sigs.k8s.io/controller-tools/cmd/controller-gen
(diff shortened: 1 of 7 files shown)


liyinan926:
Apparently this PR will revert some changes made in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1084. This is probably due to the version of controller-gen used.

jkleckner:
And to expand on the previous comment, losing the protocol defaults from #1084 will make the chart not installable with k8s versions v1.18+

jutley:
I'm struggling to understand how these CRDs are intended to be maintained. controller-gen generated the files in this PR (after I explicitly asked for the v1beta1 CRDs). This project is currently using v1.16 APIs, so I wouldn't necessarily expect the generation to gracefully handle v1.18 features.

It looks like #1084 manually updated the CRDs? This is just bad practice. The CRDs should either be generated or manually configured. Not both. I'll see if I can find a way to get the generated objects to include these fields.

jutley:
Seems in the upstream library, `protocol` is documented as required, but annotated as optional.
https://github.com/kubernetes/kubernetes/blob/v1.19.4/pkg/apis/core/types.go#L1759-L1761

liyinan926 (approved):
LGTM.

jkleckner:
@jutley Thank you for running this to ground.

nssalian:
Has this been merged @liyinan926 ?
I don't see 
```
func getUIServiceType(app *v1beta2.SparkApplication) apiv1.ServiceType {
	if app.Spec.SparkUIOptions != nil && app.Spec.SparkUIOptions.ServiceType != nil {
		return *app.Spec.SparkUIOptions.ServiceType
	}
	return apiv1.ServiceTypeClusterIP
}
``` in the master branch.

PR #1107, opened 2020-12-08
[WIP] k8s.io version up to v0.18.3

fisache:
- Add context to k8s and crd API templates
- Volcano client version up to v1.1.0
- Prometheus client version up to v1.0.0

In this version, `context` must be entered as a parameter to call k8s client.
Currently I used `context.TODO()`, if any better opinions on this, please review

I'm using kubernetes v1.19.2 and I have checked this operator working.
But I need more check running in older cluster.
Let me know any checklist according to the version change, I will test it.


fisache:
I have tested on kubernetes v1.13.0 launched by minikube about spark application, volcano queue and prometheus annotaion working.
And this worked as expected.

Issue #1076, opened 2020-11-19
Executor is dead but kubectl describe still shows as RUNNING


FloraZhang:
Hi expert,

I noticed sometimes there's a mismatch between executor state shown in "kubectl describe sparkapplications" and the actual executor state. In a recent helm deploy, for example, the executor1 is already dead, but it's still showing as running in "kubectl describe sparkapplications": 

```
[root@k8s-node1 ~]# kubectl describe sparkapplications.sparkoperator.k8s.io spark-test-app
Status:
  Application State:
    State:  RUNNING
  Driver Info:
    Pod Name:             spark-test-app-driver
    Web UI Address:       10.233.15.219:40450
    Web UI Port:          40450
    Web UI Service Name:  spark-test-app-ui-svc
  Execution Attempts:     3
  Executor State:
    spark-test-app-evaluator-1605721320126-exec-1:  RUNNING
    spark-test-app-evaluator-1605774718513-exec-2:  RUNNING
```

Could it be a similar issue like this one? https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/689

Thanks,
Wenjing

fisache:
it's resolved at https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1074

FloraZhang:
@fisache Thank you. Good to know that. I'll grab the latest operator and give it another try.

jystephan:
Hello guys -- we ran into the same issue today at [Data Mechanics](https://www.datamechanics.co).

@fisache are you sure that https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1074 fixes it?

In this case, the application is running, and there are executors that appear are running, even though they're actually dead. In #1074, the logic only applies when the application is completed. So it's not clear to me that it fixes this issue.

Thank you for your help (and thanks for #1074).

fisache:
@jystephan hello,
Actually, the state sync by spark-operator is triggered not only application state changed but only pod states including driver and executors changed(add, update, delete).
Even though the app is COMPLETED with running state of executor pods, the executor pod is deleted eventually, it's triggered again with the app is completed state.

If any question or suggestion, please reply about that.
Thanks for your comment

jrj-d:
Hey @fisache, chiming in. Thanks for your work!
I work with @jystephan .

In this case we had:
- in the status section of the `sparkapplication`:
    - app marked as `RUNNING`
[...]

PR #1111, opened 2020-12-10
update executor status if pod is lost while app is still running

ImpSy:
Hello,

This PR is related to issue #1076

Following the discussion on this issue, it seems that we can change the status of lost pods to `unknown` while the app is still `running`
Then once the driver pod is terminated, we fall back to the current implementation


liyinan926 (approved):
LGTM. Thanks for the fix!

PR #1146, opened 2021-01-26
Include Gojek in who-is-using.md

pradithya:
We, at Gojek, are using spark operator for performing distributed batch prediction in [Merlin](https://github.com/gojek/merlin). It has been working quite well and having strong adoption internally.


liyinan926 (approved):
Thanks!

PR #1165, opened 2021-02-18
Always open the prometheus port on the pod if prometheus is enabled

grantatspothero:
Previously the port would not be opened if in container configuration was used.  Thus, if you used custom configuration then prometheus would not be able to scrape the metrics port. 

Now we will always open the port if prometheus is enabled, and only mount volumes if custom configuration is not used.


grantatspothero:
@googlebot I signed it!

liyinan926 (commented):
Are existing unit tests (if any) covering the changes?

PR #1193, opened 2021-03-05
correct hook annotation for webhook job

chaudhryfaisal:
correct name forhook is `pre-upgrade` and not `pre-update` https://helm.sh/docs/topics/charts_hooks/


liyinan926 (approved):
Thanks for fixing!

PR #1203, opened 2021-03-16
fix: Remove priority from the spec when using priority class

HHK1:
By default pods created by kube will have a priority of 0 set in their
spec, or the value of the PriorityClass acting as a `globalDefault`.
When using the `batchSchedulerOptions.priorityClassName` the spec was
mutated by adding the `priorityClassName` to the spec, but the
`priority` was left unchanged.

This caused an error 
```
the integer value of priority (0) must not be
provided in pod spec; priority admission controller computed X from
the given PriorityClass name.
```
(where X is the priority associated to the PrioritClass kube object)

The patcher now properly remove the existing priority when a
priorityClassName is applied, letting kube apply the correct priority to 
the executor and driver pods.

See https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1054


liyinan926:
Can you run `go fmt` to format your changes?

HHK1:
@liyinan926 Done

ImpSy (approved):
I've tested it on a real spark app, works as intended

PR #1206, opened 2021-03-23
added --s3-force-path-style to force path style URLs for S3 objects

chaudhryfaisal:
added --s3-force-path-style to choose whether to force path style URLs for S3 objects


liyinan926 (approved):
LGTM. Thanks!

PR #1207, opened 2021-03-24
Allow custom bucket path

bribroder:
Presently, anything specified in the sparkctl parameter `--upload-to` which is not parsed as part of the `host` component of a URL gets silently dropped and a hard-coded value is used instead:

`s3://test.bucket.io/a_prefix_for_my_app`  is parsed down to `test.bucket.io` and then `/spark-app-dependencies` gets added on the end.

This PR allows specifying a separate prefix component at the user's discretion with the --upload-prefix parameter:
```--upload-to test.bucket.io --upload-prefix a_prefix_for_my_app```

### DIFF
--- a/sparkctl/cmd/create.go
+++ b/sparkctl/cmd/create.go
@@ -40,8 +40,8 @@ import (
 )
 
 const bufferSize = 1024
-const rootPath = "spark-app-dependencies"
 
+var RootPath string
 var UploadToPath string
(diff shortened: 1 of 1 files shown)


bribroder:
@googlebot I signed it!

liyinan926 (approved):
LGTM. Thanks!

PR #1209, opened 2021-03-24
Fix the error of hostAliases when there are more than 2 hostnames

cdmikechen:
If there are more than 2 hostnames in `hostAliases` path, `spec/hostAliases` will only show the first and many last hostname. Like this:
```yaml
hostAliases:
- ip: "10.1.1.1"
  hostnames:
  - "hostname1"
- ip: "10.1.1.3"
  hostnames:
  - "hostname3"
- ip: "10.1.1.3"
  hostnames:
  - "hostname3"
```
Now fix this problem.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -796,5 +796,8 @@ func addHostAliases(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOpera
 	}
 
 	var ops []patchOperation
-	return append(ops, patchOperation{Op: "add", Path: "/spec/hostAliases", Value: hostAliases})
+	if len(hostAliases) > 0 {
+		ops = append(ops, patchOperation{Op: "add", Path: "/spec/hostAliases", Value: hostAliases})
+	}
+	return ops
 }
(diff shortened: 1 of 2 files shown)


cdmikechen:
@googlebot I signed it!

pkg/webhook/patch.go
```diff
@@ -795,19 +795,9 @@ func addHostAliases(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOpera
...
 	var ops []patchOperation
-	for _, v := range hostAliases {
-		if first {
-			ops = append(ops, patchOperation{Op: "add", Path: "/spec/hostAliases", Value: []corev1.HostAlias{v}})
-			first = false
-		} else {
-			ops = append(ops, patchOperation{Op: "add", Path: "/spec/hostAliases/-", Value: &v})
-		}
+	if len(hostAliases) > 0 {
+		ops = append(ops, patchOperation{Op: "add", Path: "/spec/hostAliases", Value: hostAliases})
```
liyinan926, line 800: If the pod already has some host alias, we should use `/spec/hostAliases/-` for the path.

cdmikechen:
@googlebot I fixed it.

cdmikechen:
@liyinan926  Hi~ Thanks for your review.
I've changed my code, and run test in my computer. But I've seen CI build failed. Can you help me find the cause of the specific error, please ? I haven't found the connection between my code and the CI build failed.

cdmikechen:
@liyinan926 Hi~ CI have passed by PR https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1210
I've merge this change to codes.

liyinan926 (approved):
LGTM

PR #1210, opened 2021-03-25
remove multiple prefixes for 'p'

chaudhryfaisal:
after merging https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1207 and https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1206 there are multiple prefixes for `p` causing build to fail, this PR removes `p` prefix for `s3-force-path-style`


liyinan926 (approved):
LGTM

PR #1236, opened 2021-04-21
fix: remove preemptionPolicy when priority class name is used

### LINKED ISSUE #1233: Can't set a priorityClassName with a priority defining a premptionPolicy
This issue only happens for kube 1.19 and later.

A PriorityClass can now define a `preemptionPolicy` to prevent from preempting already running resources. See https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#example-non-preempting-priorityclass

Specifying a `batchSchedulerOptions.priorityClassName` results in an incorrect patch IF the associated PriorityClass has defined a `preemptionPolicy`: Because by default a pod will have a `PreemptionPolicy` of `PreemptLowerPriority`, there is a conflict between the `preemptionPolicy` defined by the pod spec, and the one defined by the `priorityClassName`, resulting in the following error:

```
the string value of PreemptionPolicy (PreemptLowerPriority) must not be provided in pod spec; priority admission controller computed Never from the given PriorityClass name.
```

I've already solved a similar conflict that was with the `priority` of the pod spec that defaulted to 0 in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1203

I plan to do the same here: modify the patch to add a patchOperation `remove` for `"/spec/preemptionPolicy"`. Caveat here is I think it should only be removed when the priority class defines a `preemptionPolicy`, otherwise the default one on the pod spec should be left untouched. 

Here is my plan to implement this:

1. In the webhook, modify the mutatePods to fetch the PriorityClass object if the `batchSchedulerOptions.priorityClassName` was provided
2. In the patch.go use the PriorityClass if it exists to determine if the preemptionPolicy should be removed from the pod spec.

HHK1:
Fixes #1233 

Starting at kubernetes 1.19, a PriorityClass can now define a
PreemptionPolicy. By default a pod will have a PreemptionPolicy set to
PreemptLowerPriority.

If the PriorityClass defines a different PreemptionPolicy, there will be
a conflict, so the patcher has to remove the PreemptionPolicy on the
pod. By default a PriorityClass will have a PreemptionPolicy set to
PreemptLowerPriority too, so it's safe to always remove it from the pod
when the priorityClassName is specified for the sparkapplication.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -560,6 +560,7 @@ func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) []patc
 
 		if pod.Spec.Priority != nil {
 			ops = append(ops, patchOperation{Op: "remove", Path: "/spec/priority"})
+			ops = append(ops, patchOperation{Op: "remove", Path: "/spec/preemptionPolicy"})
 		}
 	}
 
--- a/pkg/webhook/patch_test.go
+++ b/pkg/webhook/patch_test.go
@@ -823,6 +823,7 @@ func TestPatchSparkPod_PriorityClassName(t *testing.T) {
 	assert.Equal(t, priorityClassName, modifiedDriverPod.Spec.PriorityClassName)
 
 	var defaultPriority int32 = 0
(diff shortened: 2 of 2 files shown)


liyinan926:
Thanks for the PR.

Issue #1239, opened 2021-04-23
Passing custom service port name for spark UI


marcozov:
Whenever a SparkApplication is created, a corresponding service is created as well in order to expose the spark UI.
The resulting `spec` looks like this:
```
spec:
  clusterIP: ...
  ports:
  - name: spark-driver-ui-port
    port: 4040
    protocol: TCP
    targetPort: 4040
  selector:
    spark-role: driver
    sparkoperator.k8s.io/app-name: spark-app-name
  sessionAffinity: None
  type: ClusterIP
```
Most of these fields can be customized, including the port. However, the resulting port name is always going to be `spark-driver-ui-port`, as it is hardcoded: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/pkg/controller/sparkapplication/sparkui.go#L153 .

While in general this is not an issue, there are some cases where it would be nice to have control over that field.
For instance, if the spark application driver is part of a service mesh, the traffic forwarded to the UI is redirected through an Istio ingress gateway, and the access has to be restricted according to some [authorization policies](https://istio.io/latest/docs/reference/config/security/authorization-policy/) (so, the spark driver would run with an Istio sidecar container), the easiest way to treat traffic as proper HTTP is via [explicit protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/). And if protocol is not selected properly, any RBAC attempt at the network level will cause all the incoming traffic to be rejected (and Istio does not seem to be able to select protocol properly and it will just discard incoming traffic because it will be treated as TCP).

PR #1240, opened 2021-04-23
[#1239] Custom service port name for spark application UI

marcozov:
A proposal for solving https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1239 .
This PR makes it possible to specify the service port name for exposing the spark UI.

### DIFF
--- a/pkg/controller/sparkapplication/sparkui.go
+++ b/pkg/controller/sparkapplication/sparkui.go
@@ -36,8 +36,9 @@ import (
 )
 
 const (
-	sparkUIPortConfigurationKey       = "spark.ui.port"
-	defaultSparkWebUIPort       int32 = 4040
+	sparkUIPortConfigurationKey        = "spark.ui.port"
+	defaultSparkWebUIPort       int32  = 4040
+	defaultSparkWebUIPortName   string = "spark-driver-ui-port"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controller/sparkapplication/sparkui.go
```diff
@@ -204,3 +211,15 @@ func getUIServicePort(app *v1beta2.SparkApplication) (int32, error) {
 	}
 	return defaultSparkWebUIPort, nil
 }
+
+func getUIServicePortName(app *v1beta2.SparkApplication) (string, error) {
+
```
liyinan926, line 216: This empty line can be removed.

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -204,3 +211,15 @@ func getUIServicePort(app *v1beta2.SparkApplication) (int32, error) {
 	}
 	return defaultSparkWebUIPort, nil
 }
+
+func getUIServicePortName(app *v1beta2.SparkApplication) (string, error) {
```
liyinan926, line 215: Looks like this method does not need the second return value of type `error`.

liyinan926 (approved):
LGTM. Thanks!

PR #1244, opened 2021-04-27
Update who-is-using.md: Fossil is using Spark Operator for Production

duyet:


liyinan926:
Thanks for the PR and using the operator in production!

PR #1246, opened 2021-04-29
fix: do not remove preemptionPolicy in patcher when not present

HHK1:
Introduced a bug in 3090b2abb80a6703a24423b76d144911a6891e2d. I
incorrectly assumed the preemptionPolicy was always set to a default on
pods, which is not the case in K8s versions prior to 1.19.


jrj-d (approved):
Looks good to me!

liyinan926 (approved):
LGTM

PR #1254, opened 2021-05-11
Add DiDi to who-is-using.md

Run-Lin:


Run-Lin:
@googlebot I fixed it

PR #1260, opened 2021-05-15
 Added enable-ui-service flag to disable UI service

### LINKED ISSUE #1031: Removal Of Spark History service [ui-svc]
When spark job is created it creates a k8s service for Spark UI. 
Is there any configuration by which we can stop this service from getting created ?

Steps to reproduce:
    - Run a spark job
    - Wait for it to complete and see ```kubectl get svc -n <namespace>```

Thanks in Advance

sairamankumar2:
This Pull request tries to provide users an option to disable UI service due to following reasons.
1.  Spark Driver service contains the same port that is already exposed by UI service.
2. A global way to disable the service, as service is a limited resource in a Kubernetes cluster (Limit of 5000 per namespace and 10,000 per cluster). 

Adding [link](https://docs.openshift.com/container-platform/4.7/scalability_and_performance/planning-your-environment-according-to-object-maximums.html#cluster-maximums-major-releases_object-limits) to OpenShift maximums as I was not able to identify Kubernetes service maximum

This PR also closes #1031

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -87,7 +87,8 @@ All charts linted successfully
 | image.repository | string | `"gcr.io/spark-operator/spark-operator"` | Image repository |
 | image.tag | string | `""` | Overrides the image tag whose default is the chart appVersion. |
 | imagePullSecrets | list | `[]` | Image pull secrets |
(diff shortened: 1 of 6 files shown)


sairamankumar2:
@googlebot I fixed it.

PR #1261, opened 2021-05-16
Enable UI service flag for disabling UI service

### LINKED ISSUE #1031: Removal Of Spark History service [ui-svc]

sairamankumar2:
This Pull request tries to provide users an option to disable UI service due to following reasons.

Spark Driver service contains the same port that is already exposed by UI service.
A global way to disable the service, as service is a limited resource in a Kubernetes cluster (Limit of 5000 per namespace and 10,000 per cluster).
Adding link to OpenShift maximums as I was not able to identify Kubernetes service maximum

This PR also closes #1031

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.0
+version: 1.1.1
 appVersion: v1beta2-1.2.3-3.1.1
 keywords:
   - spark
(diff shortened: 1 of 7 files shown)


liyinan926 (commented):
Thanks for the PR! Can you update the Chart version?

sairamankumar2:
Hi @liyinan926,
Have updated the charts version from 1.1.0 to 1.1.1.

liyinan926 (approved):

PR #1272, opened 2021-05-25
add crds drift check between chart/ and manifest/

ImpSy:
Hello,

This PR address the issue of crds drift between chart/ and manifest/ encountered in #1262 
I've also delete redundant element in `.github/workflows/main.yaml`

This PR comes with an open question: 
Should we keep Travis-ci, since everything that it does is also check by GitHub workflow ?


liyinan926 (approved):
Thanks for the PR!

Issue #1263, opened 2021-05-17
Ease debug of Driver and Executor failure


ImpSy:
# Context

With the workload we're currently running we encounter various error on the driver or executor pods
But debug is almost impossible currently because k8s log aren't useful and neither is the CRD 

# Propositions
To enhance debug of those issue I see 2 different ways:  one of them feeling not really perfect and a cleaner one that works only if we release a new version of the API (`v1beta3`)
### 1) `recordDriverEvent` and `recordExecutorEvent` now allow variadic parameters
We could then add new context to failures
This solve partially the problem as it doesn't persist the errorMessage in the CRD but would be a good start
### 2) Change `DriverState` and `ExecutorState` to a struct like `ApplicationState`
This would break the API and need to be release as `v1beta3`
This would be a lot cleaner and persist the error message in the CRDs while also logging the error easily 

The new state would be something like:
```go
type ExecutorStateType string

// Different states an executor may have.
const (
	ExecutorPendingState   ExecutorStateType = "PENDING"
	ExecutorRunningState   ExecutorStateType = "RUNNING"
	ExecutorCompletedState ExecutorStateType = "COMPLETED"
	ExecutorFailedState    ExecutorStateType = "FAILED"
	ExecutorUnknownState   ExecutorStateType = "UNKNOWN"
)

// ExecutorState tells the current state of an executor and an error message in case of failures.
type ExecutorState struct {
	State        ExecutorStateType `json:"state"`
	ErrorMessage string            `json:"errorMessage,omitempty"`
}
```

ImpSy:
Hello @liyinan926,

Are either of those choices OK with you ?

PR #1280, opened 2021-06-07
add failure context to recordExecutorEvent

ImpSy:
Hello,

Since they were no traction on this issue: #1263

I've decided to implement solution 1 for executors

If i need to change anything do not hesitate

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -374,7 +374,12 @@ func (c *Controller) getAndUpdateExecutorState(app *v1beta2.SparkApplication) er
 			oldState, exists := app.Status.ExecutorState[pod.Name]
 			// Only record an executor event if the executor state is new or it has changed.
 			if !exists || newState != oldState {
-				c.recordExecutorEvent(app, newState, pod.Name)
+				if newState == v1beta2.ExecutorFailedState {
+					execContainerState := getExecutorContainerTerminatedState(pod.Status)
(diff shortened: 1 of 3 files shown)


jrj-d:
@liyinan926 can review this PR or tell us who else we should ask for a review?

pkg/controller/sparkapplication/controller.go
```diff
@@ -972,18 +977,18 @@ func (c *Controller) recordDriverEvent(app *v1beta2.SparkApplication, phase v1be
 	}
 }
 
-func (c *Controller) recordExecutorEvent(app *v1beta2.SparkApplication, state v1beta2.ExecutorState, name string) {
+func (c *Controller) recordExecutorEvent(app *v1beta2.SparkApplication, state v1beta2.ExecutorState, args ...interface{}) {
 	switch state {
 	case v1beta2.ExecutorCompletedState:
-		c.recorder.Eventf(app, apiv1.EventTypeNormal, "SparkExecutorCompleted", "Executor %s completed", name)
```
liyinan926, line 978: How does the output look like for `args`?

ImpSy: Same as before we already use `c.recorder.Eventf` which already accept variadics parameters

per doc:
```
Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{})
```

We already use something similar in `recordSparkApplicationEvent` for `v1beta2.FailedState`

liyinan926 (approved):

PR #1284, opened 2021-06-20
Update pythonVersion to fix example

stratus:
spark 3.1.1 (specified in image) doesn't work with Python 2. 
This change updates pythonVersion from 2 to 3.

### DIFF
--- a/examples/spark-py-pi.yaml
+++ b/examples/spark-py-pi.yaml
@@ -23,7 +23,7 @@ metadata:
   namespace: default
 spec:
   type: Python
-  pythonVersion: "2"
+  pythonVersion: "3"
   mode: cluster
   image: "gcr.io/spark-operator/spark-py:v3.1.1"
   imagePullPolicy: Always


stratus:
FWIW, I've tested this on a Mac as follows (w/o the sed call I'd get the python version related error):

kind create cluster --name spark-foo
kubectl cluster-info --context kind-spark-foo

git clone https://github.com/GoogleCloudPlatform/spark-on-k8s-operator
cd spark-on-k8s-operator
sed -i '' 's/pythonVersion: \"2\"/pythonVersion: \"3\"/' examples/spark-py-pi.yaml

helm repo add spark-operator https://googlecloudplatform.github.io/spark-on-k8s-operator
helm install my-release spark-operator/spark-operator \
  --namespace spark-operator --create-namespace \
  --set webhook.enable=true --set image.tag=latest

kubectl apply -f manifest/spark-rbac.yaml
kubectl apply -f examples/spark-py-pi.yaml

liyinan926 (approved):

PR #1286, opened 2021-06-23
Add AWS S3 dependencies to spark-operator image

korjek:
This adds the ability to use s3 path for spark dependencies without creating custom spark-operator images

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,5 +42,11 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 COPY hack/gencerts.sh /usr/bin/
 
+# Setup dependencies for AWS S3 access
+ADD https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-aws/2.7.4/hadoop-aws-2.7.4.jar $SPARK_HOME/jars
+RUN chmod 644 $SPARK_HOME/jars/hadoop-aws-2.7.4.jar
+ADD https://repo1.maven.org/maven2/com/amazonaws/aws-java-sdk/1.7.4/aws-java-sdk-1.7.4.jar $SPARK_HOME/jars
+RUN  chmod 644 $SPARK_HOME/jars/aws-java-sdk-1.7.4.jar
+
 COPY entrypoint.sh /usr/bin/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -42,5 +42,11 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 COPY hack/gencerts.sh /usr/bin/
 
+# Setup dependencies for AWS S3 access
```
liyinan926, line 45: Can we create a separate copy named `Dockerfile.aws` and make the changes in that copy?

korjek: @liyinan926 the idea was to have this support in the original Dockerfile so ppl do not need to build images by themselves. 

Also, a new Docker file may start to diverge from the original one as time flies and this is also unwanted effect.

korjek: @liyinan926 can we proceed this way?

wanshicheng:
I tried to execute with your pr, then I found the error like this:
com.amazonaws.services.s3.model.AmazonS3Exception: Status Code: 400, AWS Service: Amazon S3, AWS Request ID: null, AWS Error Code: null, AWS Error Message: Bad Request
So, how to edit yaml?

korjek:
> I tried to execute with your pr, then I found the error like this: com.amazonaws.services.s3.model.AmazonS3Exception: Status Code: 400, AWS Service: Amazon S3, AWS Request ID: null, AWS Error Code: null, AWS Error Message: Bad Request So, how to edit yaml?

not sure which yaml you are asking about. I guess 400 means bad request was issued.

joanjiao2016 (approved):

joanjiao2016 (approved):

dcharbonnier:
Not sure this is a good idea, then one will need kafka, ...
This image should be part of your personal ci not the one of the operator, you image can take this one as base.

sshah90:
> Not sure this is a good idea, then one will need kafka, ... This image should be part of your personal ci not the one of the operator, you image can take this one as base.

100% agree. We shouldn't load packages for all user as those packages can be unnecessary for them.

korjek:
I see, thanx for the feedback. Closing this.

PR #1288, opened 2021-06-29
Add user defined pod labels

pdrastil:
This adds the ability to define user labels on Spark operator pod. Main use case is integration with external logging solution. Please let me know if I should bump chart version in the PR or if you manage release cycle for the chart. Thanks

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.4
+version: 1.1.5
 appVersion: v1beta2-1.2.3-3.1.1
 keywords:
   - spark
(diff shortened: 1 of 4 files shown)


pdrastil:
@googlebot I signed it!

liyinan926 (commented):
Can you update the Chart readme and also bump the chart version?

pdrastil:
Updated README and minor chart version.

pdrastil:
@liyinan926 Hi, can you please check what's with your test cluster that is used for pre-commit checks? It seems that scheduler cannot find any nodes for spark-operator.

charts/spark-operator-chart/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.3
+version: 1.1.4
```
liyinan926, line 4: Can you bump this to 1.1.5?

pdrastil: Done

Issue #1287, opened 2021-06-26
failed to run spark-submit for SparkApplication namespace/job-timestamp: signal: killed


zzvara:
Spark operator deployed as (latest version as of this speaking):

```
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: spark-operator-development
  namespace: development
spec:
  chart:
    repository: https://googlecloudplatform.github.io/spark-on-k8s-operator
    name: spark-operator
    version: 1.1.3
  releaseName: spark-operator-development
  values:
    image:
      pullPolicy: Always
    webhook:
      enable: true
      namespaceSelector: "name=development"
    sparkJobNamespace: development
    resyncInterval: 30
    metrics:
      enable: false
    controllerThreads: 1
    batchScheduler:
      enable: true
    resources:
      requests:
        memory: 90Mi
        cpu: 25m
      limits:
        memory: 90Mi
        cpu: 25m
```

Spark operator starts as:
```
        - '-v=2'
        - '-logtostderr'
        - '-namespace=development'
        - '-enable-ui-service=true'
        - '-ingress-url-format='
        - '-controller-threads=1'
        - '-resync-interval=30'
        - '-enable-batch-scheduler=true'
        - '-label-selector-filter='
        - '-enable-webhook=true'
        - '-webhook-svc-namespace=development'
        - '-webhook-port=8080'
        - '-webhook-svc-name=spark-operator-development-webhook'
        - '-webhook-config-name=spark-operator-development-webhook-config'
        - '-webhook-namespace-selector=name=development'
        - '-enable-resource-quota-enforcement=false'
```

Application deployed as:

```
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  annotations:
    helm.fluxcd.io/antecedent: development:helmrelease/redacted
  creationTimestamp: "2021-06-26T18:22:25Z"
  generation: 1
  name: redacted-job
  namespace: development
  resourceVersion: "307711966"
  uid: ffc987aa-e56c-43e9-a65e-87ab3c5dec4f
spec:
  driver:
    cores: 1
    env:
    - name: TZ
      value: UTC
    memory: 2500m
    memoryLimit: 2500m
    serviceAccount: spark-operator-development
    volumeMounts:
    - mountPath: /opt/spark/configuration
      name: configuration
  executor:
    cores: 1
    env:
    - name: TZ
      value: UTC
[...]

PR #1292, opened 2021-07-01
Explain "signal: kill" errors during submission

zzvara:
Added comment that better describes the issue as in:
failed to run spark-submit for SparkApplication namespace/job-timestamp: signal: killed #1287


zzvara:
@googlebot I signed it!

PR #1304, opened 2021-07-16
add pprof to spark operator

goodluckbot:
We are seeing some performance issue with spark operator when the cluster is heavy loaded with spark jobs. Add pprof to find the details

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -143,6 +145,12 @@ func main() {
 
 	glog.Info("Starting the Spark Operator")
 
+	go func() {
```
liyinan926, line 148: Should we have a flag to control whether to turn this on?

ChenYi015:
@goodluckbot Thanks for the contribution! The pprof feature has been added by #2164, will close this PR.

PR #1313, opened 2021-07-30
Add hook deletion policy for spark-operator service account

pdrastil:
This PR relates to https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1155 that added pre-install hook for Argo CD. Mentioned PR fixed the issue with the webhooks, however service account is deleted and recreated by the Argo on every sync. This effectively deletes service account token used by spark-operator pod and results in loosing RBAC permissions for operated Spark pods. New annotation with helm hook deletion policy will ensure that service account is not recreated.


pdrastil:
Most likely related issue https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1299

Issue #885, opened 2020-04-22
custom serviceAccount for executor


cwei-bgl:
Hi thanks for the work, spark-on-k8s-operator really makes thing easier.

It seems the operator currently does not support specifying a custom serviceAccount for executors. I tried both serviceAccount and serviceAccountName fields on executor node, which did not work and it was always being set to 'default'. I need a custom serviceAccount for executors so that they can access AWS resources. Alternatively, at least executors use the same serviceAccount from driver so that security can be controlled to a reasonable level. Thanks.

liyinan926:
The operator doesn't support custom service account for the executor pods. Spark on Kubernetes will support it in the future in 3.1 (see https://github.com/apache/spark/pull/27034). We can definitely add support for this.

cwei-bgl:
Great! thanks @liyinan926

batCoder95:
> The operator doesn't support custom service account for the executor pods. Spark on Kubernetes will support it in the future in 3.1 (see [apache/spark#27034](https://github.com/apache/spark/pull/27034)). We can definitely add support for this.

@liyinan926 - The pull request (27034) in above reply is in state closed. Does that mean that a new version has been released with custom service account support ? Checking this in relation to the issue: [IAM Role for Driver and Executor Pods? #1002](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1002)

bbenzikry:
Issue can be closed :)
@batCoder95 This is now available via #1322

PR #1322, opened 2021-08-16
Add support for executor service account

bbenzikry:
This is a follow-up to #885, passes executor service account to spark.kubernetes.authenticate.executor.serviceAccountName

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -117,6 +117,8 @@ const (
 	// SparkDriverServiceAccountName is the Spark configuration key for specifying name of the Kubernetes service
 	// account used by the driver pod.
 	SparkDriverServiceAccountName = "spark.kubernetes.authenticate.driver.serviceAccountName"
+	// account used by the executor pod.
+	SparkExecutorAccountName = "spark.kubernetes.authenticate.executor.serviceAccountName"
 	// SparkInitContainerImage is the Spark configuration key for specifying a custom init-container image.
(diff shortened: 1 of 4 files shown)


pkg/config/constants.go
```diff
@@ -117,6 +117,8 @@ const (
 	// SparkDriverServiceAccountName is the Spark configuration key for specifying name of the Kubernetes service
 	// account used by the driver pod.
 	SparkDriverServiceAccountName = "spark.kubernetes.authenticate.driver.serviceAccountName"
+	// account used by the executor pod.
+	SparkExecutorAccountName = "spark.kubernetes.authenticate.executor.serviceAccountName"
```
liyinan926, line 121: Is this config option available in Spark on Kubernetes?

bbenzikry: Seems so 

https://github.com/apache/spark/blob/79ea014c68d0930e1b9ac313ef4edaaec1e63882/resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/Config.scala#L103

https://github.com/apache/spark/blob/79ea014c68d0930e1b9ac313ef4edaaec1e63882/resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/Config.scala#L150

liyinan926: OK, looks like the doc was not updated.

liyinan926 (approved):
LGTM. Thanks!

PR #1325, opened 2021-08-18
fix NPE introduce on #1280

ImpSy:
Hi,

Following testing on our side,
We've detected that the change I introduced on #1280 introduced NPE on executor failure on Spark3

This PR fix the executor on Spark3 and avoid NPE if we can't find the executor container
In case of NPE we show `Executor {{exec-name}} failed with ExitCode: -1, Reason: Unknown (Container not Found)`


bbenzikry:
Thanks for this, just ran into it yesterday

PR #1259, opened 2021-05-13
Add {{$appName}} {{$appNamespace}} variables support for SparkConf

ZPerling:
```
ingressUrlFormat "xxxdomain.com/{{$appNamespace}}/{{$appName}}"
```
In the case, webui worked must be set the `spark.ui.proxyBase=/{{$appNamespace}}/{{$appName}}`, otherwise the page is messed up,
but now the spark.ui.proxyBase can not use {{$appName}} and {{$appNamespace}} variables, so I create the PR.

PR #1330, opened 2021-08-20
#1329 update the operator to allow subpaths to be used with the spark ui ingress.

### LINKED ISSUE #1329: Using an Ingress created by the operator with subPaths fail
I've been using the spark-on-k8s-operator for a while, and I've recently tried to use the `ingressUrlFormat` option to configure Ingresses for me.

I've noticed it is only really designed for subdomains, rather than subpaths. For example, configuring using the helm chart.
`.Values.ingressUrlFormat: "spark.example.com/sparkjobs/{{$appName}}` would produce the ingress, but the spark ui would not load, and instead redirects to "spark.example.com/jobs/.

TomHellier:
Slightly restructure the way the ingressURL is created, and the variable which is passed around, and configure the Ingress manifest to work with subpaths

Fixes #1329 

This allows me to use the following ingressUrlFormat `spark.example.com/sparkjobs/{{$appName}}`, and the url 
spark.example.com/sparkjobs/spark-pi-1629482011066826445 now displays the spark ui correctly.


TomHellier:
@liyinan926 would you mind having a look at this when you get a chance?

TomHellier:
I think this also resolves #1259 . The integration test appears to have failed because of some intermittent problem, but I cant seem to trigger a re-run.

pkg/controller/sparkapplication/controller.go
```diff
@@ -657,6 +657,40 @@ func (c *Controller) submitSparkApplication(app *v1beta2.SparkApplication) *v1be
...
+			app.Status.DriverInfo.WebUIServiceName = service.serviceName
+			app.Status.DriverInfo.WebUIPort = service.servicePort
+			app.Status.DriverInfo.WebUIAddress = fmt.Sprintf("%s:%d", service.serviceIP, app.Status.DriverInfo.WebUIPort)
+			// Create UI Ingress if ingress-format is set.
+			if c.ingressURLFormat != "" {
+				// We are going to want to use an ingress url.
+				ingressURL, err := getSparkUIingressURL(c.ingressURLFormat, app.GetName(), app.GetNamespace())
+				if err != nil {
+					glog.Errorf("failed to run get ingress url %s/%s: %v", app.Namespace, app.Name, err)
+				}
```
liyinan926, line 674: An `else` should be used here as the code below should only be run if err == nil.

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -62,26 +74,19 @@ type SparkService struct {
...
-	if err != nil {
-		return nil, err
-	}
-	if parsedURL.Scheme == "" {
-		//url does not contain any scheme, adding http:// so url.Parse can function correctly
-		parsedURL, err = url.Parse("http://" + ingressURL)
-		if err != nil {
-			return nil, err
-		}
+	var ingressURLPath = ingressURL.Path
```
liyinan926, line 86: ingressURLPath := ingressURI.Path

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -62,26 +74,19 @@ type SparkService struct {
...
+	// If we're serving on a subpath, we need to ensure we create capture groups
+	if ingressURL.Path != "" && ingressURL.Path != "/" {
```
liyinan926, line 88: Use `ingressURLPath ` here.

pkg/controller/sparkapplication/sparkui.go
```diff
@@ -115,11 +120,19 @@ func createSparkUIIngress(app *v1beta2.SparkApplication, service SparkService, i
 	if len(ingressResourceAnnotations) != 0 {
 		ingress.ObjectMeta.Annotations = ingressResourceAnnotations
 	}
+
+	// If we're serving on a subpath, we need to ensure we use the capture groups
+	if ingressURL.Path != "" && ingressURL.Path != "/" {
```
liyinan926, line 125: Ditto, use `ingressURLPath `.

pkg/controller/sparkapplication/sparkui_test.go
```diff
@@ -542,7 +554,10 @@ func TestCreateSparkUIIngress(t *testing.T) {
 			app:  app1,
 			expectedIngress: SparkIngress{
 				ingressName: fmt.Sprintf("%s-ui-ingress", app1.GetName()),
-				ingressURL:  "ingress.clusterName.com/" + app1.GetNamespace() + "/" + app1.GetName(),
+				ingressURL:  parseURLAndAssertError("ingress.clusterName.com/"+app1.GetNamespace()+"/"+app1.GetName(), t),
```
tekumara, line 557: do you know if this will also work when there is no hostname, eg:
```
ingressUrlFormat="/{{$appNamespace}}/{{$appName}}"
```

tekumara: I tried this in the test and looks like it does! 🎉

pkg/controller/sparkapplication/controller.go
```diff
@@ -657,6 +657,40 @@ func (c *Controller) submitSparkApplication(app *v1beta2.SparkApplication) *v1be
...
+				// need to ensure the spark.ui variables are configured correctly if a subPath is used.
+				if ingressURL.Path != "" {
+					if app.Spec.SparkConf == nil {
+						app.Spec.SparkConf = make(map[string]string)
+					}
+					app.Spec.SparkConf["spark.ui.proxyBase"] = ingressURL.Path
+					app.Spec.SparkConf["spark.ui.proxyRedirectUri"] = "/"
```
tekumara, line 681: might be worth covering this in a test?

TomHellier: I've written a test which covers this (sort of)... however, it looks like when I'm writing to SparkConf here, the contents of SparkConf isn't written back to the kubernetes resource. I'm unsure whether it is worth syncing this contents back to the SparkApplication when this has been done

tekumara (approved):
🥳

PR #1350, opened 2021-09-15
Feature/sa annotations

moskitone:
Custom serviceAccount annotations. Can be used for Workload Identity.


GuusDeGraeve:
Thanks @moskitone, we also need this and I don't want to create my own helm chart just because I need to be able to change the SA annotations. Hoping for a merge soon.

PR #1358, opened 2021-09-27
fix running metrics for application deletion

### LINKED ISSUE #1354: Metric spark_app_running_count is not updated when application CR is deleted
When a CR of a long running application (e.g. streaming) is deleted, operator deletes all of the application pods but does not update the value of `spark_app_running_count `. 

As per documentation: 

> Additionally, these metrics are best-effort for the current operator run and will be reset on an operator restart. Also some of these metrics are generated by listening to pod state updates for the driver/executors and deleting the pods outside the operator might lead to incorrect metric values for some of these metrics.

But deleting the application CR falls into "inside the operator" category, I think - so the metrics `spark_app_running_count`  should get updated in this case.

Aakcht:
Decrease `running` metrics for the case of application deletion.

closes #1354

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -442,6 +442,7 @@ func (c *Controller) getAndUpdateAppState(app *v1beta2.SparkApplication) error {
 }
 
 func (c *Controller) handleSparkApplicationDeletion(app *v1beta2.SparkApplication) {
+	c.metrics.exportMetricsOnDelete(app)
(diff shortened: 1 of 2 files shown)


Aakcht:
@googlebot I fixed it.

Aakcht:
@googlebot I fixed it.

Aakcht:
Looks like `cla/google — All necessary CLAs are signed` check is fixed now, please tell me if anything else is needed.

PR #1367, opened 2021-10-12
Make manifests usable by Kustomize

karpoftea:
1. Group manifests by purpose
2. Add kustomization.yaml for each group to benefit for users that prefer Kustomize 
3. Move examples to appropriate folder


karpoftea:
@googlebot I fixed it.

liyinan926:
I think somehow we need to disable one of the check for CRD drift between `manifest` and the chart folder.

PR #1377, opened 2021-10-28
fix container name in addsecuritycontext patch

lybavsky:


pkg/webhook/patch.go
```diff
@@ -613,14 +613,9 @@ func addSecurityContext(pod *corev1.Pod, app *v1beta2.SparkApplication) *patchOp
...
-	// Find the driver/executor container in the pod.
-	for ; i < len(pod.Spec.Containers); i++ {
-		if pod.Spec.Containers[i].Name == config.SparkDriverContainerName || pod.Spec.Containers[i].Name == config.SparkExecutorContainerName {
-			break
-		}
-	}
-	if i == len(pod.Spec.Containers) {
+	i := findContainer(pod)
+
+	if i == -1 {
```
liyinan926, line 618: Nit: can we use `i < 0` here instead?

lybavsky: I think yes, anyway findContainer returns only non-negative values if container found in pod, or -1 if not.

liyinan926: Yes, but I think it's safer to check `i < 0`.

lybavsky: Ok, sure,this is acceptable

lybavsky: Fixed!

PR #1384, opened 2021-11-10
delete-service-accounts-and-roles-before-creation

TiansuYu:
Related with issue https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1359

The current version of helm chart does not handle hook tearing-down automatically. So after the uninstall, `spark-operator` will leave three artifacts `serviceAccount`, `ClusterRole` and `ClusterRoleBinding` inside the k8s config. Attempts to reinstall the same `spark-operator` will be blocked because `rbac.yaml` will try to install these as pre-install hook, despite the fact that they are already there.

Suggest to add the `before-hook-creation` inside `hook-delete-policy` as suggested by @swartz-k, to fix this issue.

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.12
+version: 1.1.13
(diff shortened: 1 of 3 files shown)


liyinan926:
Please bump up the chart version.

TiansuYu:
rebased with main to solve version conflicts issue

PR #1386, opened 2021-11-11
Add grafana dashboard to the helm chart

drazul:
Created Grafana dashboard to easy monitor SparkApps.

This dashboard get the ScheduledSparkApp name from the name of each SparkApp using a regex and group all metrics based on the ScheduledSparkApp name.

This dashboard can be simpler if we use kube_pod_labels metric but it doesn't work in the same way in kube-state-metrics > 2.0
https://github.com/kubernetes/kube-state-metrics/issues/1501

Screenshot 2021-11-11 at 10 42 40

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.9
+version: 1.1.10
 appVersion: v1beta2-1.2.3-3.1.1
 keywords:
   - spark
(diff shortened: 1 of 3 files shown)


liyinan926:
Looks like one of the integration tests failed with error: `Error:  templates/grafana-dashboards.yaml: object name must be between 0 and 253 characters: ""`. See https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/runs/4176372310?check_suite_focus=true.

drazul:
> Looks like one of the integration tests failed with error: `Error: templates/grafana-dashboards.yaml: object name must be between 0 and 253 characters: ""`. See https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/runs/4176372310?check_suite_focus=true.

CI Fixed.

I also updated the dashboard to show running spark jobs 

Screenshot 2021-11-15 at 17 12 37

liyinan926:
Could you please bump up the chart version?

drazul:
> Could you please bump up the chart version?

Sorry. I did a mistake when I was rebasing origin/master and the PR was closed.

I created another one with the same changes.
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1396

rubenssoto:
Is it possible to work when not use ScheduledSparkApp?

PR #1387, opened 2021-11-12
webhook timeout variable

sairamankumar2:
Webhook timeout by default is set to 30s (as we use AdmissionregistrationV1beta1) and sometimes that causes the request to fail when the number of pods in the cluster is huge.  We fix the webhook error by recreating the webhook manually with a lower timeout value (This does get modified when the operator pod restarts). 

We found kubernetes/kubernetes#71508, which talks about how the default client timeout is 30s, so if the webhooks have a timeout equal or greater, the client can give up before the apiserver has, which means the failurePolicy: ignore has no effect. 

This PR provides a way to set a default value for webhook timeout which allow the users to configure the timeout time based on their usage.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -86,7 +86,7 @@ type WebHook struct {
 	enableResourceQuotaEnforcement bool
 	resourceQuotaEnforcer          resourceusage.ResourceQuotaEnforcer
 	coreV1InformerFactory          informers.SharedInformerFactory
-	TimeoutSeconds                 *int32
+	timeoutSeconds                 *int32
 }
 
 // Configuration parsed from command-line flags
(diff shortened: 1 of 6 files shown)


sairamankumar2:
Also, can we set the default timeout value to `10s` as it is the default timeout value in webhook starting from `AdmissionregistrationV1`?

pkg/webhook/webhook.go
```diff
@@ -86,6 +86,7 @@ type WebHook struct {
 	enableResourceQuotaEnforcement bool
 	resourceQuotaEnforcer          resourceusage.ResourceQuotaEnforcer
 	coreV1InformerFactory          informers.SharedInformerFactory
+	TimeoutSeconds                 *int32
```
liyinan926, line 89: Nit: `timeoutSeconds` since it doesn't need to be exposed outside the package.

sairamankumar2: Renamed the variable to be timeoutSeconds and also bumped the chart version to have the intermediate fix

Issue #732, opened 2019-12-13
Reliably wait for spark application to start and then to finish


jkleckner:
See discussion in #719 with this last open issue.  Moving to a separate issue for clarity.

> > How can you launch a spark application and then reliably wait for it to finish?
> 
> I would explore extending `sparkctl` to support this kind of launch-and-wait use cases.

That could work, but it would be nice if the underlying api had this as a primitive so that there is no need to install another tool but the spark application could be waited on from `kubectl`.

_Originally posted by @jkleckner in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/719#issuecomment-565088850_

jkleckner:
Here is a workaround script for waiting for a test to complete and fetching the status.
Success is obtained when app_status is COMPLETED

```
spark_namespace=spark
APP_TEST_TIMEOUT=80
APP_TEST_NAME=app-test-${CI_JOB_ID}
app_status="UNDEFINED"

function check_app_test_status(){
  local counter
  counter=0
  app_status=$(kubectl get -n $spark_namespace sparkapplications ${APP_TEST_NAME} -o jsonpath='{.status.applicationState.state}')

  until [[ "${app_status}" = "COMPLETED" || "${app_status}" = "FAILED" ]]; do
    sleep 5

    if [ "${counter}" -ge "${APP_TEST_TIMEOUT}" ]; then
      echo "####### ERROR: reached timout - app test did not finish"
      kubectl -n $spark_namespace describe sparkapplications ${APP_TEST_NAME}
      app_status="TEST_TIMEOUT"
      break
    else
      counter=$((++counter))
    fi
    app_status=$(kubectl get -n $spark_namespace sparkapplications ${APP_TEST_NAME} -o jsonpath='{.status.applicationState.state}')
    echo '###' counter: ${counter} app_status:"${app_status}"
  done
}

...


kubectl apply --validate=true -f app-test.yaml
check_app_test_status

if [ "${app_status}" != "COMPLETED" ] ; then
  echo '################  ERROR RUNNING TEST CASE'
else
  echo '################  SUCCESS RUNNING TEST CASE'
fi

```

holdenk:
So I'm not seeing the appStatus field using the 2.4.5 version?

saimadineni:
@jkleckner  Has this issue been ever resolved natively or you've just been using this hack? 
Thanks

jkleckner:
[...]

PR #1411, opened 2021-11-30
Implement --wait flag on 'sparkctl create' in order to wait the completion

illomi7:
This PR implements the flag `--wait` or `-w` on `sparkctl create ... ` command.
You can use it when you need to know whetever your sparkapplication is finish in failed or in success state.
It will block until the job state is _COMPLETED_ or _FAILED_.

Related with issue #732


sparkctl/cmd/create.go
```diff
@@ -177,6 +181,29 @@ func createSparkApplication(app *v1beta2.SparkApplication, kubeClient clientset.
...
 	fmt.Printf("SparkApplication \"%s\" created\n", app.Name)
 
+	if WaitTillCompletion {
+		for {
+			appStatus, err := getSparkApplication(app.Name, crdClient)
+			if err != nil {
+				return fmt.Errorf("failed to get SparkApplication %s: %v", app.Name, err)
+			}
+
+			fmt.Printf("SparkApplication \"%s\" status: \"%s\"\n", app.Name,appStatus.Status.AppState.State)
```
liyinan926, line 191: You can use %q instead of %s, which adds double quotes for you I believe.

sparkctl/cmd/create.go
```diff
@@ -177,6 +181,29 @@ func createSparkApplication(app *v1beta2.SparkApplication, kubeClient clientset.
...
+
+			if appStatus.Status.AppState.State == v1beta2.FailedState ||
+				appStatus.Status.AppState.State == v1beta2.UnknownState {
```
liyinan926, line 194: Should this just be ignored and continue?

sparkctl/cmd/create.go
illomi7: No problem, I was only following the code-style that I already found in the file (ex: line 182)

sparkctl/cmd/create.go
illomi7: What are you suggesting to do?
Do you intend to remove the check on sparkapplication in failed state or peraphs you suggest to remove the return of the error and only print an error log?

liyinan926: I meant for the unknown state, we probably should ignore.

illomi7: Ok, we only add it to avoid that a unknown state of the job , probaby caused from a k8s pod error,  block the execution of the command.

alaurentinoofficial:
I recently created the PR #1506 to add logs to `sparkctl create` using the `--logs` flag and await the SparkApplication complete. 

I noticed that this PRs is very similar to mine, maybe it could be combined

I propose the command `sparkctl create` always waits the SparkApplication start and show the logs, if it exceeds the timeout returns a error. To avoid this behavior a `--daemon` or `-d` flag can be used to run in the background

What you guys @liyinan926 @illomi7  think?

Issue #1403, opened 2021-11-22
IfNotPresent with :latest


dcharbonnier:
Using IfNotPresent with `:latest` sounds like a very bad idea and this is the default.
I've seen some issue reported due to this problem.
The chart should contain the image version

autarchprinceps:
Yeah, I have to agree. That misses the entire point of helm chart versions & upgrades.

TomHellier:
I've just seen [my pr was merged](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1422), so the image should exist at [ghcr.io/googlecloudplatform/spark-operator:v1beta2-1.3.2-3.1.1](https://github.com/googlecloudplatform/spark-on-k8s-operator/pkgs/container/spark-operator)

Issue #1372, opened 2021-10-21
Failed to pull image "gcr.io/spark-operator/spark-operator:latest":


chlyzzo:
when i run ：
helm repo add spark-operator https://googlecloudplatform.github.io/spark-on-k8s-operator
and then:
helm install spark-operator-test spark-operator/spark-operator

get error:
 Failed to pull image "gcr.io/spark-operator/spark-operator:latest": rpc error: code = Unknown desc = Error response from daemon: Get https://gcr.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

so the image has access?

TomHellier:
I've just seen [my pr was merged](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1422), so the image should exist at [ghcr.io/googlecloudplatform/spark-operator:v1beta2-1.3.2-3.1.1](https://github.com/googlecloudplatform/spark-on-k8s-operator/pkgs/container/spark-operator)

PR #1415, opened 2021-12-01
Add docker build to github action

TomHellier:
https://travis-ci.org stopped building code in June, so this should be removed. Github actions does the same and more anyway.

Adding docker build into github actions, it would be good if this repository could move to use that as its docker container publishing method, as then all CI will be managed by one system.

The helm chart shouldn't default to use the "latest" docker container, as that hasn't been build since May 19th anyway, I've reverted this back to use the chart appVersion (it was using this previously)
Fixes:
#1403 
#1372


TomHellier:
It looks like the e2e integration tests are broken, they're expecting files to exist which are not there anymore.

I have raised https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1416 to address the problem at a later date

liyinan926 (approved):
Thanks for the PR!

PR #1420, opened 2021-12-03
[WIP] Add a hook to dowload the external jar if it is specified

bagnaram:
🚧 
I will squash and revert the CI changes when ready to merge. 

This introduces logic to the spark-submit to fetch external JAR files to local. This is useful to download jobs from an artifact repository without rebuilding container images for each job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 .PHONY: clean-sparkctl
 
 SPARK_OPERATOR_GOPATH=/go/src/github.com/bagnaram/spark-on-k8s-operator
-DEP_VERSION:=`grep DEP_VERSION= Dockerfile | awk -F\" '{print $$2}'`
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@
 .SILENT:
 .PHONY: clean-sparkctl
 
-SPARK_OPERATOR_GOPATH=/go/src/github.com/GoogleCloudPlatform/spark-on-k8s-operator
+SPARK_OPERATOR_GOPATH=/go/src/github.com/bagnaram/spark-on-k8s-operator
```
ssullivan, line 5: This should probably be GoogleCloudPlatform instead of bagnaram

bagnaram: Thank you, I will update this back when I am ready to review

PR #1427, opened 2021-12-09
Kubernetes v1.22 extensions/v1beta1 API removal

aneagoe:
Fixes https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1426.
This is a breaking change, meaning that on kubernetes versions older than 1.19 Ingress objects would fail to be created (everything else should still work as expected). It also defines the `pathType` statically as "ImplementationSpecific".

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ The following table lists the most recent few versions of the operator.
 | Operator Version | API Version | Kubernetes Version | Base Spark Version | Operator Image Tag |
 | ------------- | ------------- | ------------- | ------------- | ------------- |
 | `latest` (master HEAD) | `v1beta2` | 1.13+ | `3.0.0` | `latest` |
+| `v1beta2-1.3.3-3.1.1` | `v1beta2` | 1.16+ | `3.1.1` | `v1beta2-1.3.3-3.1.1` |
+| `v1beta2-1.3.2-3.1.1` | `v1beta2` | 1.16+ | `3.1.1` | `v1beta2-1.3.2-3.1.1` |
 | `v1beta2-1.3.0-3.1.1` | `v1beta2` | 1.16+ | `3.1.1` | `v1beta2-1.3.0-3.1.1` |
 | `v1beta2-1.2.3-3.1.1` | `v1beta2` | 1.13+ | `3.1.1` | `v1beta2-1.2.3-3.1.1` |
 | `v1beta2-1.2.0-3.0.0` | `v1beta2` | 1.13+ | `3.0.0` | `v1beta2-1.2.0-3.0.0` |
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
 version: 1.1.16
-appVersion: v1beta2-1.3.2-3.1.1
+appVersion: v1beta2-1.3.3-3.1.1
 keywords:
   - spark
 home: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.16
+version: 1.1.17
 appVersion: v1beta2-1.3.3-3.1.1
 keywords:
   - spark
--- a/main.go
+++ b/main.go
@@ -153,6 +153,10 @@ func main() {
 		glog.Fatal(err)
 	}
 
+	if err = util.InitializeIngressCapabilities(kubeClient); err != nil {
+		glog.Fatalf("Error retrieving Kubernetes cluster capabilities: %s", err.Error())
+	}
+
 	var batchSchedulerMgr *batchscheduler.SchedulerManager
 	if *enableBatchScheduler {
 		if !*enableWebhook {
--- a/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go
@@ -18,7 +18,7 @@ package v1beta2
 
 import (
 	apiv1 "k8s.io/api/core/v1"
-	extensions "k8s.io/api/extensions/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -322,7 +322,7 @@ type SparkUIConfiguration struct {
 	IngressAnnotations map[string]string `json:"ingressAnnotations,omitempty"`
 	// TlsHosts is useful If we need to declare SSL certificates to the ingress object
 	// +optional
-	IngressTLS []extensions.IngressTLS `json:"ingressTLS,omitempty"`
+	IngressTLS []networkingv1.IngressTLS `json:"ingressTLS,omitempty"`
 }
 
 // ApplicationStateType represents the type of the current state of an application.
--- a/pkg/apis/sparkoperator.k8s.io/v1beta2/zz_generated.deepcopy.go
+++ b/pkg/apis/sparkoperator.k8s.io/v1beta2/zz_generated.deepcopy.go
@@ -24,7 +24,7 @@ package v1beta2
 
 import (
 	v1 "k8s.io/api/core/v1"
-	v1beta1 "k8s.io/api/extensions/v1beta1"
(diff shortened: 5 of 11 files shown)


TomHellier:
@aneagoe please could you update this PR to support both options. It should be possible to write code which detects the k8s version and applies the appropriate ingress code.

I dont think it would be appropriate to merge this as it would move the minimum k8s version from 1.13~1.16 up quite a few versions

alexellis:
Glad this was useful to you. Changes like this kind of set the whole ecosystem back.

TomHellier:
LGTM :+1: thanks @aneagoe

Reamer:
Thank you very much for this contribution.
I think a PR like this should also change the corresponding [rbac.yml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2919f521ccde71fc54b8d72ba0474c5872c7d4aa/manifest/spark-operator-install/spark-operator-rbac.yaml#L42-L44)

aneagoe:
@Reamer I'm not quite sure how but it seems to be gracefully handled. Here's the clusterrole created by the helm chart on my 1.22 cluster:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    helm.sh/hook: pre-install
    helm.sh/hook-delete-policy: hook-failed, before-hook-creation
  creationTimestamp: "2021-12-03T09:00:55Z"
  labels:
    app.kubernetes.io/instance: spark-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: spark-operator
    app.kubernetes.io/version: v1beta2-1.3.1-3.1.1
    helm.sh/chart: spark-operator-1.1.15
  name: spark-operator
  resourceVersion: "9819502"
  uid: 16cc785d-04b4-43f9-81bc-520a53ad2485
rules:
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - '*'
- apiGroups:
  - ""
  resources:
  - services
  - configmaps
  - secrets
  verbs:
  - create
  - get
  - delete
  - update
- apiGroups:
  - extensions
  - networking.k8s.io
  resources:
  - ingresses
  verbs:
  - create
  - get
  - delete
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - update
  - patch
- apiGroups:
  - ""
  resources:
  - resourcequotas
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - create
  - get
  - update
  - delete
- apiGroups:
  - admissionregistration.k8s.io
  resources:
  - mutatingwebhookconfigurations
  - validatingwebhookconfigurations
  verbs:
  - create
  - get
  - update
  - delete
- apiGroups:
  - sparkoperator.k8s.io
  resources:
  - sparkapplications
  - sparkapplications/status
  - scheduledsparkapplications
  - scheduledsparkapplications/status
  verbs:
  - '*'
- apiGroups:
  - batch
  resources:
  - jobs
  verbs:
  - delete
```

As you can see, it contains the `networking.k8s.io` api group. Changing the rbac.yaml would require further modifications of the helm chart logic to support both, but does not seem necessary.

Reamer:
I do not use the helm charts, but it seems that helm uses this file [charts/spark-operator-chart/templates/rbac.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2e8b733f5ad029ceb8a949b153a53c434b8fcfe1/charts/spark-operator-chart/templates/rbac.yaml#L28). It would be good if the api group was also added in simple manifest file [manifest/spark-operator-install/spark-operator-rbac.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2919f521ccde71fc54b8d72ba0474c5872c7d4aa/manifest/spark-operator-install/spark-operator-rbac.yaml#L42).
Perhaps we should make this change in a separate PR?

aneagoe:
> I do not use the helm charts, but it seems that helm uses this file [charts/spark-operator-chart/templates/rbac.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2e8b733f5ad029ceb8a949b153a53c434b8fcfe1/charts/spark-operator-chart/templates/rbac.yaml#L28). It would be good if the api group was also added in simple manifest file [manifest/spark-operator-install/spark-operator-rbac.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2919f521ccde71fc54b8d72ba0474c5872c7d4aa/manifest/spark-operator-install/spark-operator-rbac.yaml#L42). Perhaps we should make this change in a separate PR?

It would make more sense, yes. At the same time, it might not be just a simple addition. I'm not sure how this behaves on older k8s that don't support `networking.k8s.io` ingresses.

Reamer:
> I'm not sure how this behaves on older k8s that don't support `networking.k8s.io` ingresses.
Neither do I, thank you for your PR.

aneagoe:
I've amended the `appVersion` to avoid the build failure. I'm not sure how to tackle the api change check though. @liyinan926 @TomHellier any advice?

TomHellier:
@aneagoe I guess technically the api has  changed, but as you say it should be backwards compatible as you've tested with v16 and v22. Perhaps that github workflow step could be amended to search for a string in the github commit like "(ignoreapi)" or something so we can just bypass it.

Unfortunately I think @liyinan926 is the only person with maintainer privileges on this repository, and I think he only checks this place once a month. I think ideally this repo would have more maintainers but as the repo states, this is not an officially supported project.

aneagoe:
The `build-api-docs` job fails at step `The API should not change once published`. This check would always fail if anything were to change:
```
0s
Run if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta1; then
  if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta1; then
    echo "sparkoperator.k8s.io/v1beta1 api has changed"
    false
  fi
  if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta2; then
    echo "sparkoperator.k8s.io/v1beta2 api has changed"
    false
  fi
  shell: /usr/bin/bash -e {0}
sparkoperator.k8s.io/v1beta2 api has changed
```

That file had to be changed.

aneagoe:
I've bumped the chart version as well, this should help pass the rest of the checks. Still not sure how I can get around the `build-api-docs` diff check as that's going to fail no matter what I do. @liyinan926 any advice on that?

aneagoe (commented):
Even though we're changing appVersion here, the `Check changes in resources used in docker file` workflow step in `build-spark-operator` is failing. I'm not really sure how I can address this, any input is appreciated.

aneagoe:
@TomHellier I believe https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1418/files introduces a bug in the workflow:
```
      - name: Check changes in resources used in docker file
        run: |
          DOCKERFILE_RESOURCES=$(cat Dockerfile | grep -P -o "COPY [a-zA-Z0-9].*? " | cut -c6-)
          for resource in $DOCKERFILE_RESOURCES; do
            # If the resource is different
            if ! git diff  --quiet origin/master -- $resource; then
              ## And the appVersion hasn't been updated
              if ! git diff --quiet charts/spark-operator-chart/Chart.yaml | grep +appVersion; then
                echo "resource used in gcr.io/spark-operator/spark-operator has changed in $resource, need to update the appVersion in charts/spark-operator-chart/Chart.yaml"
                git diff origin/master -- $resource;
                echo "failing the build... " && false
              fi
            fi
          done
```
Basically, `git diff --quiet charts/spark-operator-chart/Chart.yaml | grep +appVersion` would always fail since `--quiet` removes all output and can't be used together with `grep`. This seems to be the reason the check fails at that step for this PR.

liyinan926:
@aneagoe Looks like the operator code failed to build. Could you please fix that?

aneagoe:
> @aneagoe Looks like the operator code failed to build. Could you please fix that?

@liyinan926 It's not the operator build that's failing but a check in the github action (more details at https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1427#issuecomment-1010811978). I've created a separate PR to address that under https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1452.

Regarding the `build-api-docs` job failure at step `The API should not change once published`, I'm not sure what I can do as it would always fail in case of changes and the changes there were necessary for this PR.

aneagoe:
@liyinan926 thanks for merging https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1452, all pre-commit checks are now passing with the exception of `build-api-docs` which is expected to fail in this case. :tada:

liyinan926:
Thanks @aneagoe.

PR #1439, opened 2021-12-22
Modify istio enabled control flow in jobs to properly apply annotation when istio.enabled=false

seanturner026:
Currently setting `istio.enabled = false` has no effect on the jobs.

If `istio.enabled` => set the annotation to be false.

The logic should be

if `istio.enabled` => set the annotation to be true.
else => set the annotation to be false.

### DIFF
--- a/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml
+++ b/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml
@@ -13,6 +13,9 @@ spec:
       name: {{ include "spark-operator.fullname" . }}-webhook-cleanup
       {{- if .Values.istio.enabled }}
       annotations:
+        "sidecar.istio.io/inject": "true"
+      {{- else }}
+      annotations:
         "sidecar.istio.io/inject": "false"
       {{- end }}
     spec:
--- a/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml
+++ b/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml
@@ -11,13 +11,8 @@ spec:
   template:
     metadata:
       name: {{ include "spark-operator.fullname" . }}-webhook-cleanup
-      {{- if .Values.istio.enabled }}
       annotations:
-        "sidecar.istio.io/inject": {{ .Values.istio.enabled | quote }}
-      {{- else }}
-      annotations:
-        "sidecar.istio.io/inject": "false"
-      {{- end }}
+        "sidecar.istio.io/inject": {{ .Values.istio.enabled | default false | quote }}
     spec:
       serviceAccountName: {{ include "spark-operator.serviceAccountName" . }}
       restartPolicy: OnFailure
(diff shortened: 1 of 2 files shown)


charts/spark-operator-chart/templates/webhook-cleanup-job.yaml
```diff
@@ -12,6 +12,9 @@ spec:
     metadata:
       name: {{ include "spark-operator.fullname" . }}-webhook-cleanup
       {{- if .Values.istio.enabled }}
+      annotations:
+        "sidecar.istio.io/inject": "true"
```
dungdm93, line 16: Why not `"sidecar.istio.io/inject": {{ .Values.istio.enabled | quote }}`

seanturner026: I was applying the same style that already existed https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/2919f521ccde71fc54b8d72ba0474c5872c7d4aa/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml#L16

seanturner026: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1439/commits/8c0c27cca462a4b86a67c62e8e2c6bad745f2f04

charts/spark-operator-chart/templates/webhook-init-job.yaml
```diff
@@ -12,6 +12,9 @@ spec:
     metadata:
       name: {{ include "spark-operator.fullname" . }}-webhook-init
       {{- if .Values.istio.enabled }}
+      annotations:
+        "sidecar.istio.io/inject": {{ .Values.istio.enabled | quote }}
+      {{- else }}
       annotations:
         "sidecar.istio.io/inject": "false"
       {{- end }}
```
dungdm93, line 20: I mean:
```suggestion
      annotations:
        "sidecar.istio.io/inject": {{ .Values.istio.enabled | quote }}
```

dungdm93 (changes requested):

charts/spark-operator-chart/templates/webhook-init-job.yaml
seanturner026: I thought you might have meant this. I haven't done much helm in awhile. If `istio.enabled` isn't defined, would this cause an error?

Provided the answer is that there won't be an error, I'm totally happy with this :)

dungdm93: In this case, you can use `{{ .Values.istio.enabled | default false | quote }}`
So `null` or `""` (empty string) will be `"false"`

seanturner026: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1439/commits/6e507f651cbe66f688bf9df9d62fa7d674c6ae6f

seanturner026: So I wonder if the default here is supposed to be false.

Its come to my attention that istio prevents jobs from exiting because the sidecar container doesn't exit. Which means that the original intent of the code would be correct.

Thinking about it, I think the solution is actually leaving this code as is, and guiding the users to add pod labels that prevent istio from adding the sidecar.

I will admit that this solution would require some documentation as it's not immediately obvious how to proceed.

seanturner026:
Closing because I think that this is intended behavior https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1439#discussion_r780577356

PR #1451, opened 2022-01-12
Add automatic creation of RBAC for the driver

lamOrigin007:
Purpose:
Automatic creation of RBAC objects for the driver pod if ServiceAccount is not specified.


TomHellier:
Is there a related issue which would require this change?

This is a significant change to the rbac being given to the spark operator, and to my knowledge operators arnt typically given the ability to CRUD service accounts carte blanch across a cluster. Do you know of another kuberneres operator which does this?

PR #1454, opened 2022-01-14
Updated default registry to ghcr.io

aneagoe:
As the automated build system now pushes images to GitHub container registry, it makes sense to adjust the helm chart to use that as default instead of the old one.

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.17
+version: 1.1.18
 appVersion: v1beta2-1.3.3-3.1.1
 keywords:
   - spark
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -148,8 +148,8 @@ jobs:
 
       - name: Run chart-testing (install)
         run: |
-          docker build -t gcr.io/spark-operator/spark-operator:local .
-          minikube image load gcr.io/spark-operator/spark-operator:local
+          docker build -t ghcr.io/googlecloudplatform/spark-operator:local .
+          minikube image load ghcr.io/googlecloudplatform/spark-operator:local
           ct install
 
   integration-test:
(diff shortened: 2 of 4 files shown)


charts/spark-operator-chart/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.17
+version: 1.1.18
```
vara-bonthu, line 4: These chart versions are not available in the [repo](https://googlecloudplatform.github.io/spark-on-k8s-operator). The latest chart version available is `1.1.15`.

Chart version increment PR should also deploy the latest chart to the Helm Repo

```sh
▶ helm search repo spark-operator
NAME                         	CHART VERSION	APP VERSION        	DESCRIPTION
spark-operator/spark-operator	1.1.15       	v1beta2-1.3.1-3.1.1	A Helm chart for Spark on Kubernetes operator
```

TomHellier: I think it should work based on this line

https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/.github/workflows/release.yaml#L46

TomHellier: I believe the recent master build failed due to the api change @aneagoe made recently, so no docker container or chart was released

aneagoe: The spark operator image released just fine, but not the helm chart. I'm not sure why the chart did not release... I did bump the chart version in that PR.

aneagoe: Actually, it's really weird. The chart releaser actually ran, but did not detect any changes, so nothing was pushed. More info here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/runs/4806627798?check_suite_focus=true, check `Run chart releaser` step.

vara-bonthu: @aneagoe @TomHellier  I think i found the issue is due to the workflow itself. I have raised a new [PR](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1456) to correct this workflow. 
Please do let me know if you have any further suggestions

vara-bonthu:
👍

PR #1456, opened 2022-01-16
Github actions workflow fix for Helm chart deployment

vara-bonthu:
### Problem: 
Helm Chart release deployment workflow is broken since `spark-operator-chart-1.1.15` due to an issue within the release workflow. Release workflow `name: Release Spark-Operator Docker Image to github container registry` step is building the Docker image and pushing the image to Github container registry. This step also creates a git `tag` and pushes the tag to this repo. The subsequent git action step `name: Run chart-releaser` is pulling the latest tag from the repo to identify the changes in the chart folder. This Helm chart step could never find the changes in the repo due to tag commit from the previous step.

### Solution: 
This PR moved `git tag` and `git push` tags to  a new github action step `name: Release Spark-Operator Docker Image` at end of the workflow. This ensures Docker image is published and the Helm Chart is published before pushing the TAG to the repo.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -62,7 +62,7 @@ jobs:
       - name: Release Spark-Operator Docker Image
         run: |
           DOCKER_TAG=$(cat charts/spark-operator-chart/Chart.yaml | grep "appVersion: .*" | cut -c13-)
-          if git rev-parse -q --verify "refs/tags/$DOCKER_TAG" >/dev/null; then
+          if git rev-parse -q --verify "refs/tags/$DOCKER_TAG"; then
             echo "Spark-Operator Docker Image Tag $DOCKER_TAG already exists!"
           else
             git tag $DOCKER_TAG
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,7 +41,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Release Spark-Operator Docker Image to github container registry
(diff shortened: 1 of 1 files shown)


aneagoe:
@vara-bonthu although `chart-releaser-action` is quite new to me, the problem doesn't really seem to be related to the git tag steps/ordering. Those only extract the app version and not the chart version. From what I understand so far, the release process for the chart is only partly automated, requiring a manual git tag.
Further, docker image release and chart release should be detached; just because there's a newer image it doesn't necessarily mean we have to update and re-release the chart. As well, we might have chart changes that don't really require a new image. 
I'm wondering what would be the best way to test this...

aneagoe:
Ok, after a more careful review, it looks like changing the order would actually do the trick as it would force the `chart-releaser-action` (function `lookup_changed_charts`) to detect the changes.

vara-bonthu:
> lookup_changed_charts

Yes, absolutely. This [function](https://github.com/helm/chart-releaser-action/blob/565e7795cebb33bb2cf4f52ab8f779e21b9fa262/cr.sh#L59) is crucial which uses the `latest_tag -> git describe --tags --abbrev=0` to identify the Helm chart changes. 

- Now, the Helm chart deployment will be triggered since we moved the `git tag` and `git push tag` to the last step.
- Helm Chart release won't be triggered if there are no changes under `chart` folder however it can still deploy the new image with this workflow.
- Bumping the Helm Chart version is sufficient to detect the change and deploy the new version of Helm Chart with this change
- I have tested the workflow locally using `echo` statements without having to push the images to the repo to validate the functionality.

liyinan926:
Thanks for fixing the flow! Merging.

PR #1475, opened 2022-02-04
Add Operator volumes and volumeMounts in chart

ocworld:
It is needed to mount one or more volumes into spark operator pod  because of various reasons such as external ivy cache.

In this pr, it is supported to operator pod's volumes and volumeMounts in chart.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,7 +41,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Release Spark-Operator Docker Image to github container registry
+      - name: Build and Push Spark-Operator Docker Image to github container registry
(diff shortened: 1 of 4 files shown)


ocworld:
@ahnlab-oss

liyinan926:
Please update the chart version.

ocworld:
> Please update the chart version.

chart version is updated to 1.1.18

liyinan926:
Looks like you need to bump the Chart version again due to conflict.

liyinan926:
Looks like chart fails to build due to the following error. So you need to fix the linting.

```
Error: Error linting charts: Error processing charts
```

ocworld:
> Looks like chart fails to build due to the following error. So you need to fix the linting.
> 
> ```
> Error: Error linting charts: Error processing charts
> ```

It is fixed

PR #1482, opened 2022-02-16
Add ingress-class-name controller flag

### LINKED ISSUE #1445: add spec.ingressClassName api
when kubernetes version >= 1.19, `ingressClassName`  is used instead of `annotations: ingress.class`

I think we need to support this API

voyvodov:
Add a flag to set ingressClassName field for ingress
objects created for Spark UI.
This will make ingress compliant with Kubernetes >v1.19 and
better utilizing multiple ingress controllers

Should fix #1445


voyvodov:
@liyinan926 can I ask you for your check and approval for the workflow?

PR #1486, opened 2022-02-22
Webhook init and clean-up should use the same node selector the operator deployment

HawaiianSpork:
When deploying to a cluster with both windows and linux nodes, we need the webhook jobs to run on the linux nodes.


HawaiianSpork:
Will resubmit after CLA is signed.

PR #1494, opened 2022-03-17
Webhook init and clean-up should use the same node selector the operator deployment

HawaiianSpork:
When deploying to a cluster with both windows and linux nodes, we need the webhook jobs to run on the linux nodes.


liyinan926:
Sorry for taking so long to get to this. Could you please bump up the chart version? Thanks.

HawaiianSpork:
@liyinan926 version bumped

HawaiianSpork:
Handled by #1545

PR #1511, opened 2022-04-12
fix: add pre-upgrade hook to rbac resources

cwyl02:
In https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1379#issuecomment-966284850, a breaking change was introduced. 

The result is that when I did a helm upgrade from 1.1.6 chart to a 1.1.10+ chart, the spark-operator ClusterRole and ClusterRoleBinding resources are deleted. also see: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1407

In this PR, I added pre-upgrade value to all 3 resources. Although it won't create the RBAC resources upon the 1st `helm upgrade` from the old chart(I think this is helm behavior), a second `helm upgrade` will apply the previously non-hook RBAC resources.

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.1.22
+version: 1.1.23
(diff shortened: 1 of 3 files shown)


cwyl02:
hi @yuchaoran2011 could you approve me running the workflows?

liyinan926:
Could you please rebase this PR?

Issue #1526, opened 2022-05-12
Setup ingress support directly from Helm Chart


viktoriaas:
Hello, 
we set up `spark-on-k8s-operator` with UI enabled. We provide dynamic DNS and automatic `letsencrypt` certificates in our cluster so it is possible to create any url on HTTPS automatically just by setting a couple of annotations in `Ingress` + tls part.

We configured ingressURL to the specific pattern which includes namespace and application name. However, later we realized people who run `SparkApplication` tend to misconfigure `sparkUIOptions`, especially:
- ingress annotations required for generating a certificate
- `ingressTLS` part where they have to provide TLS host according to pattern configured in Operator which can be a rather long string (if the namespace is long).

On top of that, ingress configuration is not something directly connected to `SparkApplication` so we suggest possibility to set cluster-wide Ingress TLS support in Helm `values.yaml`. The modification requires the possibility to add custom annotations in values and a modification in code that detects if cluster-wide TLS support is desired. If yes, the host url in tls part of ingress can be set to the same url as in `rules[].host` because a certificate is going to be issued for that url (where app is served). 

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: spark-2-ui-ingress
  namespace: [namespace]
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-prod             <-- custom required annotation
    kubernetes.io/ingress.class: nginx                           <-- custom required annotation
    kubernetes.io/tls-acme: "true"                               <-- custom required annotation
    nginx.ingress.kubernetes.io/preserve-trailing-slash: "true"  <-- custom required annotation
    nginx.ingress.kubernetes.io/rewrite-target: /$2
  labels:
    sparkoperator.k8s.io/app-name: spark-2
rules:
  - host: [host_url]
    http:
      paths:
      - backend:
          service:
            name: spark-2-ui-svc
            port:
              number: 4040
        path: /[namespace]/[app_name](/|$)(.*)
        pathType: ImplementationSpecific
  tls:
  - hosts:
    - [host_url]
    secretName: [ingress_name]-tls
```

or0or1:
what is  spark-2-ui-svc?  how to get the service name ?  thanks.    @viktoriaas

PR #1527, opened 2022-05-12
Cluster-wide ingress TLS support

viktoriaas:
Hi, 
this is a pull request that implements cluster-wide ingress support directly from Helm chart. 

Changes:
- added ingressTLS section in `values.yaml` that allows enabling cluster-wide ingress and adding annotations to application's Ingress
- when ingressTLS support is enabled, tls host is set to the same url as configured `ingressUrlFormat`
- if user provides some annotations in `SparkApplication` spec
  - if such annotation was not configured by operator, it is added to all annotation
  - if such annotation was configured, value provided by user is used
- if user provides some items in `sparkUIOptions.ingressTLS` part of `SparkApplication` spec, this item is simpy added to list of `tls.hosts` that includes host configured by operator as well

For more information and motivation for this PR see #1526

The modification was tested with:
- Operator configured with cluster-wide TLS support
- example pi `SparkApplication` without `sparkUIOptions` section
Ingress was correctly created with `https` on right url

### DIFF
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -676,14 +676,16 @@ func (c *Controller) submitSparkApplication(app *v1beta2.SparkApplication) *v1be
 		}
 	}
 
+	driverInfo := v1beta2.DriverInfo{}
+
 	if c.enableUIService {
 		service, err := createSparkUIService(app, c.kubeClient)
 		if err != nil {
 			glog.Errorf("failed to create UI service for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 		} else {
-			app.Status.DriverInfo.WebUIServiceName = service.serviceName
-			app.Status.DriverInfo.WebUIPort = service.servicePort
-			app.Status.DriverInfo.WebUIAddress = fmt.Sprintf("%s:%d", service.serviceIP, app.Status.DriverInfo.WebUIPort)
+			driverInfo.WebUIServiceName = service.serviceName
+			driverInfo.WebUIPort = service.servicePort
+			driverInfo.WebUIAddress = fmt.Sprintf("%s:%d", service.serviceIP, app.Status.DriverInfo.WebUIPort)
 			// Create UI Ingress if ingress-format is set.
 			if c.ingressURLFormat != "" {
 				// We are going to want to use an ingress url.
@@ -703,8 +705,8 @@ func (c *Controller) submitSparkApplication(app *v1beta2.SparkApplication) *v1be
 					if err != nil {
 						glog.Errorf("failed to create UI Ingress for SparkApplication %s/%s: %v", app.Namespace, app.Name, err)
 					} else {
-						app.Status.DriverInfo.WebUIIngressAddress = ingress.ingressURL.String()
-						app.Status.DriverInfo.WebUIIngressName = ingress.ingressName
+						driverInfo.WebUIIngressAddress = ingress.ingressURL.String()
+						driverInfo.WebUIIngressName = ingress.ingressName
 					}
 				}
 			}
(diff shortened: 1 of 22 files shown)


liyinan926:
Please update the chart version.

viktoriaas:
> Please update the chart version.

could you please instruct me what you mean by that? it's my first time contributing both to code and chart

liyinan926:
Please bump up the chart version as defined in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/charts/spark-operator-chart/Chart.yaml, given that you modified the chart yamls.

viktoriaas:
I bumped the chart version to `1.1.22`, is it ok?

pkg/controller/sparkapplication/controller.go
```diff
@@ -79,6 +79,8 @@ type Controller struct {
 	ingressClassName  string
 	batchSchedulerMgr *batchscheduler.SchedulerManager
 	enableUIService   bool
+	enableIngressTLS bool
```
liyinan926, line 82: Nit: these two and the two fields above related to ingress can be grouped together.

pkg/controller/sparkapplication/controller.go
```diff
@@ -102,7 +106,7 @@ func NewController(
 	})
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, apiv1.EventSource{Component: "spark-operator"})
 
-	return newSparkApplicationController(crdClient, kubeClient, crdInformerFactory, podInformerFactory, recorder, metricsConfig, ingressURLFormat, ingressClassName, batchSchedulerMgr, enableUIService)
+	return newSparkApplicationController(crdClient, kubeClient, crdInformerFactory, podInformerFactory, recorder, metricsConfig, ingressURLFormat, ingressClassName, batchSchedulerMgr, enableUIService, enableIngressTLS, ingressTLSAnnotations)
```
liyinan926, line 109: Ingress related parameters can be grouped together.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -100,6 +101,16 @@ func getIngressResourceAnnotations(app *v1beta2.SparkApplication) map[string]str
 	return ingressAnnotations
 }
 
+func convertIngressStringToMap(ingressTLSAnnotations string) map[string]string {
+	annotations := strings.Split(ingressTLSAnnotations, ",")
+	ingressAnnotations := map[string]string{}
+	for _, a := range annotations {
+		parts := strings.Split(a, ":")
+		ingressAnnotations[parts[0]] = parts[1]
```
liyinan926, line 109: There should be a check to make sure parts has two elements.

pkg/controller/sparkapplication/controller.go
viktoriaas: I thought it makes sense to split the configuration because you might want to enableUI service but still let the user provide all other configurations. The `enableIngressTLS` is meant to be turned on only if you have a cluster wide provider and want simplify users' lives.

pkg/controller/sparkapplication/controller.go
viktoriaas: Oh I probably understand now - by grouping together you mean that if `ingressTLSAnnotations != []` then providin ingress is enabled, and  if `ingressTLSAnnotations == []` it is not?

viktoriaas:
@liyinan926 I wanted to move forward so I hope I understood grouping fields correctly and I just reordered the variables in `Controller` and `newSparkApplicationController` structs. In addition, I bumped Chart version again because in the meantime, it was bumped to `1.1.22` which I used. About checking 2 parts - see the comment as I don't think it's necessary.

liyinan926:
The changes look good to me. However, looks like a few integration tests failed. Can you fix them?

liyinan926 (approved):

liyinan926:
Looks like you also need to bump the chart version.

viktoriaas:
@liyinan926 I fixed the conflict with Chart version and bumped it again (because of new functionality). Please, can you approve the workflow? The implementation is the same and it would be nice to have the functionality integrated

PR #1535, opened 2022-05-23
fix executor event formatting

GaruGaru:
The formatter args for the executor events are passed as slice instead of vararg, this causes the events to be displayed as follows:
`Executor [exec-name %!s(int32=143) Error] failed with ExitCode: %!d(MISSING), Reason: %!s(MISSING)` 

Expected event format:
`Executor [exec-name Error] failed with ExitCode: 143, Reason: <reason>`


liyinan926:
The build failed. Could you please take a look and fix it?

GaruGaru:
@liyinan926 The checks are failing for the following reason ( since I've changed the code inside `pkg/` )
```Run DOCKERFILE_RESOURCES=$(cat Dockerfile | grep -P -o "COPY [a-zA-Z0-9].*? " | cut -c6-)
resource used in gcr.io/spark-operator/spark-operator has changed in pkg/, need to update the appVersion in charts/spark-operator-chart/Chart.yaml
```
Should I go ahead and update the `appVersion` in the chart in this PR ? 

eg:
from `v1beta2-1.3.5-3.1.1` to `v1beta2-1.3.6-3.1.1`

PR #1538, opened 2022-05-26
set driverState as running when driver container is running

dcoliversun:
This PR aims to set driverState as running when driver container is running. Previously, driverState was set to running only when the driver pod was running, regardless of whether the driver process (driver container) was started.


dcoliversun:
@liyinan926

dcoliversun:
I see e2e test failed. And log is as follow:
```plain
PASS
coverage: 11.6% of statements
ok  	github.com/GoogleCloudPlatform/spark-on-k8s-operator/sparkctl/cmd	0.029s	coverage: 11.6% of statements
2022/05/26 23:44:47 No kubeconfig found. Bypassing e2e tests
ok  	github.com/GoogleCloudPlatform/spark-on-k8s-operator/test/e2e	0.037s
?   	github.com/GoogleCloudPlatform/spark-on-k8s-operator/test/e2e/framework	[no test files]
FAIL
make: *** [Makefile:70: unit-test] Error 1
Error: Process completed with exit code 2.
```
Interesting. Exit code 2 signifies invalid usage of some shell built-in command. And, it looks normal from log. @liyinan926 Could u help me to find it out? Or, just re-run test?

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -169,7 +185,10 @@ func podStatusToDriverState(podStatus apiv1.PodStatus) v1beta2.DriverState {
 			}
 			return v1beta2.DriverFailedState
 		}
-		return v1beta2.DriverRunningState
+		if getDriverContainerRunningState(podStatus) != nil {
```
liyinan926, line 188: Looks like all what you need is to tell if all the driver containers are in running state, right? If so, the method can simply return a boolean instead of the state, and the condition becomes more natural.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -157,6 +161,18 @@ func getContainerTerminatedState(name string, podStatus apiv1.PodStatus) *apiv1.
 	return nil
 }
 
+func getContainerRunningState(name string, podStatus apiv1.PodStatus) *apiv1.ContainerStateRunning {
```
liyinan926, line 164: Looks like you can name this method `isDriverContainerRunning`.

pkg/controller/sparkapplication/sparkapp_util.go
```diff
@@ -145,6 +145,10 @@ func getExecutorContainerTerminatedState(podStatus apiv1.PodStatus) *apiv1.Conta
 	return state
 }
 
+func getDriverContainerRunningState(podStatus apiv1.PodStatus) *apiv1.ContainerStateRunning {
```
liyinan926, line 148: Is this method really needed given it's only wrapping a one-line call to another method?

pkg/controller/sparkapplication/sparkapp_util.go
dcoliversun: Yeah. `getDriverContainerRunningState` method has replaced by `isDriverContainerRunning` method.

pkg/controller/sparkapplication/sparkapp_util.go
dcoliversun: Thanks, renamed this method as `isDriverContainerRunning`

pkg/controller/sparkapplication/sparkapp_util.go
dcoliversun: Thanks, this method is lost.

dcoliversun:
`TestSyncSparkApplication_ExecutingState` unit test failed and I'm working on it.

PR #1545, opened 2022-06-06
Webhook init and cleanup should respect nodeSelector

erikcw:
Without `nodeSelector` being available, the init/cleanup pods will hang if k8s schedules them to run on ARM nodes (Graviton on EKS).


liyinan926:
Please bump up the chart version.

erikcw:
@liyinan926 Done.

PR #1564, opened 2022-07-03
Support using spark application metrics as operator metrics labels

amit-mazor:
### Goal:
This PR's goal is to allow adding additional labels to the operator metrics.

Currently, the only available label in the operator metrics is `app_type`, 
but I think it is very useful to add additional labels from our spark applications. 

For example internally, I added two labels in addition to 'app_type' that Im using in order to create alerts.


amit-mazor:
@liyinan926 Can you please review ?

PR #1565, opened 2022-07-03
Add PrometheusRule template for creating spark alerts

amit-mazor:
### Goal:
Enable creating prometheus alerts. 

I'v added here a new template for `PrometheusRule`, and disabled it by default in the values.


amit-mazor:
@liyinan926 Can you please review ?

Issue #216, opened 2018-07-16
Deal with internally created ConfigMap mounted onto SPARK_CONF_DIR


liyinan926:
PR apache/spark#20669 introduced a ConfigMap carrying Spark configuration properties in a file for the driver. The environment variable SPARK_CONF_DIR is set to point to the mount path /opt/spark/conf of the ConfigMap. This is in conflict with what spec.sparkConfigMap is designed to do. We need to find a solution to work with the internally created and mounted ConfigMap.

liyinan926:
@aditanase sorry for not responding sooner. I think that's a viable approach. And yes the admission webhook can be used to inject an init-container for doing that. Thanks for being willing to take on this. Looking forward to your implementation!

aditanase:
No worries. What are your thoughts on https://github.com/apache/spark/pull/22146?
Together with support for client mode (
https://github.com/apache/spark/commit/571a6f0574e50e53cea403624ec3795cd03aa204#diff-b5527f236b253e0d9f5db5164bdb43e9), it will be quite disruptive for this project. At best, it might negate the need for the webhook. At worst, it might make things like this issue vert hard to implement.
I'm planning to spend some time on that PR this week, you should probably check it out if you didn't already.

mrow4a:
@aditanase in the available timeline I don't see it disruptive - if webhook wont be needed for some users better it is. But to support earlier versions one would need webhook in the operator.

wweic:
Hey @aditanase @liyinan926 @mrow4a, wondering what's the plan to resolve this? One use case is to provide `hive-site.xml` in `/opt/spark/conf` so we can configure hive behavior, right now it seems I can not achieve that. Would love to hear your inputs.

liyinan926:
The approach proposed above of using an init-container to copy the files in `sparkConfigMap` to `$SPARK_CONF_DIR` of the main Spark container seems reasonable. This can be setup automatically by the operator.

wweic:
Thanks @liyinan926 ! I'll dig a bit and see if I can help.

wweic:
I setup an init-container and mount spark generated Volume(based on a configmap) and the operator generated Volume(based on a configmap) to it. Then I do a `cp` to copy from operator config to spark config. It throws error:
````
cp: cannot create regular file in  '/opt/spark/conf/test.xml': Read-only file system
````
[...]

PR #1568, opened 2022-07-06
Subpath support for config map volume mounts

bbenzikry:
This PR addresses several issues discussed at length in #216 
It allows for additional config files to be mounted into ``SPARK_CONF_DIR`` and adds compatibility to the upstream by changing ``DefaultSparkConfDir`` to ``/opt/spark/conf``
@tahesse worked quite a bit to handle tests, documentation and additional modifications to my original patch. 

It's important to note that this does not add any capability to [change the actual mounted spark.properties file](https://github.com/bbenzikry/spark-on-k8s-operator/blob/29bb31b25ca71152180b49fd688dfbc366ed681d/docs/troubleshooting.md), though we have discussed adding merge logic once this PR goes through.

### DIFF
--- a/pkg/webhook/patch.go
+++ b/pkg/webhook/patch.go
@@ -17,18 +17,19 @@ limitations under the License.
 package webhook
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
 	"github.com/golang/glog"
 
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta2"
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/config"
 	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/util"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 const (
@@ -42,7 +43,7 @@ type patchOperation struct {
 	Value interface{} `json:"value,omitempty"`
 }
 
-func patchSparkPod(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperation {
+func patchSparkPod(pod *corev1.Pod, app *v1beta2.SparkApplication, client kubernetes.Interface) []patchOperation {
 	var patchOps []patchOperation
 
 	if util.IsDriverPod(pod) {
@@ -51,7 +52,7 @@ func patchSparkPod(pod *corev1.Pod, app *v1beta2.SparkApplication) []patchOperat
 
 	patchOps = append(patchOps, addVolumes(pod, app)...)
 	patchOps = append(patchOps, addGeneralConfigMaps(pod, app)...)
-	patchOps = append(patchOps, addSparkConfigMap(pod, app)...)
+	patchOps = append(patchOps, addSparkConfigMap(pod, app, client)...)
 	patchOps = append(patchOps, addHadoopConfigMap(pod, app)...)
 	patchOps = append(patchOps, getPrometheusConfigPatches(pod, app)...)
 	patchOps = append(patchOps, addTolerations(pod, app)...)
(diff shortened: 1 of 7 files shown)


tahesse:
@liyinan926 may I kindly ask you to review this PR? You created the associated issue for years ago and this is the first iteration on dealing with the sparkConfigMap spec.
Thank you!

vanducng:
Hi team, wondering if everything going well. We are trying to integrate with Glue Catalog in which this PR would be helpful.

afterthought:
I'm able to test this now. Is there a prerelease available?

tahesse:
@afterthought Unfortunately not but if you check out this branch and build the docker image + subsequently the helm chart you can install it to a local minikube (`minikube image load <image name>`) which has access to your local docker images.

Looking forward to you testing!

pkg/webhook/patch.go
```diff
@@ -292,16 +293,78 @@ func addEnvironmentVariable(pod *corev1.Pod, envName, envValue string) *patchOpe
...
+			//       need to subPath mount in the default case.
+			vmPatchOp := addConfigMapVolumeMount(pod, config.SparkConfigMapVolumeName, config.DefaultSparkConfDir)
+			if vmPatchOp == nil {
+				return nil
+			}
+			patchOps = append(patchOps, *vmPatchOp)
 		}
-		patchOps = append(patchOps, *vmPatchOp)
+		// Patch custom sparkConfigMap from spec to have them subPath mounted
+		configMap, err := client.CoreV1().ConfigMaps(app.Namespace).Get(context.TODO(), *sparkConfigMapName, metav1.GetOptions{})
```
maknihamdi, line 348: `if err == nil ` I think

tahesse: Thanks for the review. The `if err == nil` is not required here because we the `else` clause here:

https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1568/files/7cf4c30947ecb2db0fdc732b42a8e10c929361a1#diff-0aa888a76ed441948985bc01c47945dc5572d3173b19e7beef96d7947920b763R364-R366

maknihamdi: I understood that we patch when we found the configmap with NO error, so I think the correct condition is`if err == nil {
//patch
} else {
			glog.Errorf("Could not get custom spark config map: %v", err)
		}` 
or `if configMap != nil {
//patch
} else {
			glog.Errorf("Could not get custom spark config map: %v", err)
		}`

bbenzikry: Hi, we're currently waiting for a response from maintainers on inclusion. I would appreciate waiting for their response before opening CRs, especially for cosmetics. I won't be replying to ANY non maintainer reviews until a response is given, so most items such as this one will be moot until that happens. Thanks.

tcassaert:
@bbenzikry @tahesse thanks a lot for this PR! 

I have built your branch and deployed the spark-operator image. However, I can't seem to find how to get it to work.

I'm using the spark-pi example to test:
```
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: spark-pi
  namespace: spark-jobs
spec:
  type: Scala
  mode: cluster
  image: "gcr.io/spark-operator/spark:v3.1.1"
  imagePullPolicy: Always
  mainClass: org.apache.spark.examples.SparkPi
  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.12-3.1.1.jar"
  sparkVersion: "3.1.1"
  sparkConfigMap: "spark-conf"
  restartPolicy:
    type: Never
  driver:
    cores: 1
    coreLimit: "1200m"
    memory: "512m"
    labels:
      version: 3.1.1
    serviceAccount: openeo
  executor:
    cores: 1
    instances: 1
    memory: "512m"
    labels:
      version: 3.1.1
```

With an example configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: spark-conf
  namespace: spark-jobs
data:
  spark-defaults.conf: |
    spark.eventLog.enabled=false
```

When deploying these, I get the following logs in the spark-operator:

```
1125 14:50:32.519600       9 event.go:282] Event(v1.ObjectReference{Kind:"SparkApplication", Namespace:"spark-jobs", Name:"spark-pi", UID:"93703710-ef06-40b8-857b-430d6af0ff99", APIVersion:"sparkoperator.k8s.io/v1beta2", ResourceVersion:
"2390167", FieldPath:""}): type: 'Normal' reason: 'SparkApplicationAdded' SparkApplication spark-pi was added, enqueuing it for submission
I1125 14:50:32.541951       9 sparkui.go:153] Creating an Ingress spark-pi-ui-ingress for the Spark UI for application spark-pi
I1125 14:50:32.550433       9 submission.go:65] spark-submit arguments: [/opt/spark/bin/spark-submit --class org.apache.spark.examples.SparkPi --master k8s://https://10.254.0.1:443 --deploy-mode cluster --conf spark.kubernetes.namespace=s
park-jobs --conf spark.app.name=spark-pi --conf spark.kubernetes.driver.pod.name=spark-pi-driver --conf spark.kubernetes.container.image=gcr.io/spark-operator/spark:v3.1.1 --conf spark.kubernetes.container.image.pullPolicy=Always --conf s
park.kubernetes.submission.waitAppCompletion=false --conf spark.kubernetes.driver.label.sparkoperator.k8s.io/app-name=spark-pi --conf spark.kubernetes.driver.label.sparkoperator.k8s.io/launched-by-spark-operator=true --conf spark.kubernet
es.driver.label.sparkoperator.k8s.io/submission-id=17178b93-ea44-4a66-80a9-d19df424a431 --conf spark.driver.cores=1 --conf spark.kubernetes.driver.limit.cores=1200m --conf spark.driver.memory=512m --conf spark.kubernetes.authenticate.driv
er.serviceAccountName=openeo --conf spark.kubernetes.driver.label.version=3.1.1 --conf spark.kubernetes.executor.label.sparkoperator.k8s.io/app-name=spark-pi --conf spark.kubernetes.executor.label.sparkoperator.k8s.io/launched-by-spark-op
erator=true --conf spark.kubernetes.executor.label.sparkoperator.k8s.io/submission-id=17178b93-ea44-4a66-80a9-d19df424a431 --conf spark.executor.instances=1 --conf spark.executor.cores=1 --conf spark.executor.memory=512m --conf spark.kube
rnetes.executor.label.version=3.1.1 local:///opt/spark/examples/jars/spark-examples_2.12-3.1.1.jar]
I1125 14:50:35.968539       9 webhook.go:247] Serving admission request
I1125 14:50:35.969805       9 patch.go:928] Processing mount {spark-local-dir-1 false /var/data/spark-df131b1c-e8e9-4d15-8b45-36c47d274c46  <nil> }(name spark-local-dir-1)
I1125 14:50:35.970055       9 patch.go:928] Processing mount {spark-conf-volume-driver false /opt/spark/conf  <nil> }(name spark-conf-volume-driver)
I1125 14:50:35.970123       9 patch.go:327] Replacing VolumeMount spark-conf-volume-driver at 1
I1125 14:50:35.970144       9 patch.go:332] New VolumeMount /spec/containers/0/volumeMounts/1
E1125 14:50:35.973582       9 patch.go:365] Could not get custom spark config map: <nil>
```

So I assume it's not using my configMap? Can'f find it in the spark-driver either when I get a shell in the pod.

Am I missing something to make this work?

tahesse:
@tcassaert thank you for testing this PR.

Regarding your test: the `spark-defaults.conf` is ignored by the upstream apache spark kubernetes resource handler because they generate a `spark.properties` which has higher precedence than the `spark-default.conf`.
Hence, your `ConfigMap` should look like

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: spark-conf
  namespace: spark-jobs
data:
  spark.properties: |
    spark.eventLog.enabled=false
```

However, we cannot just override the `spark.properties` as of now. The justification is given in the newly introduced `troubleshoot.md`:

https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/7cf4c30947ecb2db0fdc732b42a8e10c929361a1/docs/troubleshooting.md?plain=1#L5-L9

I did start to write some code to merge the upstream `spark.properties` with a `ConfigMap`-provided custom `spark.properties` here:
https://github.com/tahesse/spark-on-k8s-operator/blob/hotfix/webhook-patch-spark-properties/pkg/webhook/patch.go#L351-L376

and here:
https://github.com/tahesse/spark-on-k8s-operator/blob/hotfix/webhook-patch-spark-properties/pkg/webhook/patch.go#L969-L990

After some discussion with @bbenzikry, I did not include the properties patch code in this PR due to the following reasons:
- the upstream apache spark heavily relies on the auto-generated `spark.properties` which requires a discussion with the maintainers (optimally with the maintainers of this repo and the apache spark kubernetes resource manager maintainers) on how to continue
- various options exist to patch the `spark.properties` without breaking the kubernetes resource manager in the process
- danger of further diverging from the upstream apache spark kubernetes resource manager logic

So, in order to test this PR, you'd want to add any file used or required by the upstream apache spark kubernetes resource manager which is not the `spark.properties`, for example the `hive-site.xml` (see https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/786).

I hope this text sheds some light on the current process.

tcassaert:
@tahesse Thanks for your response!

If I understand it correctly, there's currently a way to set for example a `hive-site.xml` via the `sparkConfigMap` in the CRD, but there's not yet a way to set any `spark.*` configuration? That still has to go via `sparkConf`?

pkg/webhook/patch.go
mrendi29: @maknihamdi  is right ... we should patch the configmap only if there is **NO ERROR**, otherwise the current patch will always branch to else complaining that the configmap was not found... (tested this locally and it actually fails if err != nil)

bbenzikry: Hi @mrendi29, I'm sorry for not being clearer in my earlier response.
I wasn't trying to imply @maknihamdi is incorrect - I mentioned that due to a bit of history around this issue and to emphasize that investing any CR time here is irrelevant until the maintainers actually want to push this ahead. 
 
The original code ( and my own version ) has proper error handling https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/504c88e2003f32114981baea99cbbd1db1e82794/pkg/webhook/patch.go#L319C7-L319C7, so I'm guessing this regression happened during the last iteration.

I really don't mind updating the PR, but after two years I'm really unsure of the chances of it getting merged.

PR #1583, opened 2022-08-02
Suspend a running spark streaming application

maknihamdi:
suspend a running spark streaming application

We add a new State: suspended.

to switch to suspended state we only need to annotate SparkApplication object with a specific annotation.
to resume the job, we need to delete the annotation

we can switch to suspended from all non-final states


maknihamdi:
@liyinan926 It would be good if you could take a look when you have time, thanks!

PR #1586, opened 2022-08-03
Add option to specify hostNetwork in spark-operator deployment

tomgoren:
Solves https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1235

Because: https://medium.com/@denisstortisilva/kubernetes-eks-calico-and-custom-admission-webhooks-a2956b49bd0d


RustamGimadiev (approved):
I faced the same issue with Calico on EKS, hostNetwork is the solution

tomgoren:
Another way to work around this issue is via [Helm Post Rendering](https://helm.sh/docs/topics/advanced/#post-rendering) - that's what we ended up doing until this PR or something like it are merged.

PR #1590, opened 2022-08-07
Add new features: When a user submits a job to k8s using the native spark-submit, the spark-operator also senses the creation of the driver pod and creates the sparkApplication CR instance as well as the service ingress

xingyunyang01:
Add new features: When a user submits a job to k8s using the native spark-submit, the spark-operator also senses the creation of the driver pod and creates the sparkApplication CR instance as well as the service ingress


xingyunyang01:
@liyinan926 
It would be good if you could take a look when you have time, thanks!

PR #1610, opened 2022-09-05
Add `ingressClassName` chart value

vpbarb:
This PR adds the controller option `ingress-class-name` value in the Helm chart values.

The option was implemented here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1482


a-shkarupin:
@vpbarb could you please help me understand why this was closed?
I'm currently having issues setting ingress class name and this looked like a solution, so I'm confused why this didn't make it. Is there an alternative way out there?

vpbarb:
> @vpbarb could you please help me understand why this was closed? I'm currently having issues setting ingress class name and this looked like a solution, so I'm confused why this didn't make it. Is there an alternative way out there?

@a-shkarupin I made this PR before the spark-operator maintenance was moved to Kubeflow. I waited for 3 months for some feedback and realized that it would not be merged. So I closed this. I ended up having a fork used internally for our company needs.

PR #1617, opened 2022-10-02
update-api-docs: Added ApplicationStateTypes descriptions

namigaliyev:
Added to understand application status. The cause of the application state is not explained in any documents.


namigaliyev:
Hi @liyinan926 , have a look this pr please. I think It is useful for newcommers.

PR #1633, opened 2022-10-26
Fix #1393 : fix tolerations block in wrong segment for webhook jobs

### LINKED ISSUE #1393: [Chart] Webhook init and cleanup template files have wrong toleration block placement
Tolerations are placed outside the if block.

This is in both [webhook-cleanup-job.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/6939cbd887adda19984ed5f75fd49dcbfc7c09b4/charts/spark-operator-chart/templates/webhook-cleanup-job.yaml#L49) and [webhook-init-job.yaml](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/6939cbd887adda19984ed5f75fd49dcbfc7c09b4/charts/spark-operator-chart/templates/webhook-init-job.yaml#L39)

When webhooks are disabled, the file will only contain tolerations which causes this error from `helm-operator`:
```error
Warning  FailedReleaseSync  104s (x6 over 10m)  helm-operator  synchronization of release 'spark-operator' in namespace 'spark-operator' failed: installation failed: unable to build kubernetes objects from release manifest: error validating "": error validating data: [apiVersion not set, kind not set]
```

zhiminglim:
As per issue #1393 , this will ensure the tolerations block are within the if-block of the helm chart manifests for the Job resources.

Tested with `helm template` and tolerations set.


angopinath (commented):
looks good

Issue #1656, opened 2022-12-20
Honor SuccessfulRunHistoryLimit for manual executions


ArshiAAkhavan:
creating a `sparkapplication` from a `scheduledsparkapplication` using the following command
```bash
sparkctl create --from <schedule_spark_application_name> <spark_application_name>
```
would result in `sparkapplication` not being deleted after `SuccessfulRunHistoryLimit` successful apps being submitted.

it seems like `scheduledsparkapplication` only cares about those that are run automatically and not those that are run manualy

PR #1659, opened 2022-12-22
recognize manual submits in history of ScheduledSparkApplication

ArshiAAkhavan:
This Pull request changes how `sparkctl` creates new apps from an already declared `scheduledsparkapplication` by adding the required labels to the created object so that the operator can track the created `sparkapplication` as one of the `scheduledsparkapplicaion` runs.

This Pull request also changes how sorting for `sparkApps` works and sorts based on `CreatoionTimeStamp` first and `Name` second.
#1656


SmsS4:
Is there any update in this pull request?
We really need this feature in our company.

Issue #1546, opened 2022-06-08
The node was low on resource: ephemeral-storage. Container spark-kubernetes-driver was using 48136Ki, which exceeds its request of 0.


mostafaghadimi:
We are experiencing this `The node was low on resource: ephemeral-storage. Container spark-kubernetes-driver was using 48136Ki, which exceeds its request of 0.` error and found the pod evicted. I have two questions:
1. The pod is using a little amount of assigned memory. Is there any way to prevent the pod to write data into storage while it has free memory?
2. How to set `ephermal-storage` limit in order to get rid of the error?

mostafaghadimi:
@jkleckner 
Hi Jim 👋🏻 , 

You mean that there isn’t any free space left on the storage of the node which is running the pod?
Would you please give me more details about finding where this problem comes from and how can I resolve it?

Thanks.

jkleckner:
It can be either running out of space on the node or the request limit for your pod.

I have seen situations where we needed to limit a pod from using all local (ephemeral) storage and affecting other pods.

Check your monitoring of this resource.

See [1] for a definition of ephemeral storage as local node attached storage.

See [2] how you can limit (or not) the amount used by a pod.

[1] https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage
[2] https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

jkleckner:
Also note this cryptic comment that the `SPARK_WORKER_DIR` can be used to alter the volume where ephemeral data is stored.

[1] https://github.com/apache/spark/commit/71fc11357420b2207d1b29ab4502987293565a67

jkleckner:
I look at the local (ephemeral) disk usage on executor nodes and try to keep the peak of it during a day < 15% or so for headroom.

Also, on GKE, the IOPS are proportional to the volume size so over-provisioning the volume is also a throughput adjustment.

You can observe throttled I/O to see whether it is a bottleneck.

mostafaghadimi:
Thanks for your response, and I know how to limit the resources in **Pod/Deployment** and other objects in Kubernetes. I have tried the following:

```yaml
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: ScheduledSparkApplication
metadata:
  name: synonym-data-gathering-scheduled
  namespace: spark
spec:
  resources:
    limits:
      - default:
[...]

PR #1661, opened 2022-12-25
Add support for `ephemeral.volumeClaimTemplate` in helm chart CRDs

ArshiAAkhavan:
This pull request adds support for `ephemeral.volumeClaimTemplate` which is the Kubernetes way of handling temporary ephemeral disks.

This pull request can also be a solution to issue #1546


phyyou:
HI, @liyinan926. Can you help this PR to can be merged?

noahshpak:
FYI this works with `storage: '250G'` but not `storage: '250g'.

bscaleb:
Is there any documentation for using this field? I don't see any mention of `ephemeral` in the API spec for the CRDs. I'm also wondering what exactly it does. Thanks! @ArshiAAkhavan

fiedlerNr9:
> Is there any documentation for using this field? I don't see any mention of ephemeral in the API spec for the CRDs. I'm also wondering what exactly it does

+ 1 here. Anyone an example at hand how to use this?

ArshiAAkhavan:
> > Is there any documentation for using this field? I don't see any mention of ephemeral in the API spec for the CRDs. I'm also wondering what exactly it does
> 
> +1 here. Anyone an example at hand how to use this?

When I was using/developing Spark operator, I had a use case where I wanted drivers/executors to provision temporary storage on the fly and delete it when they were done. K8s allows this behavior via `EphemeralSotrage` for `pod`s and I used the same API here.

That being said, I haven't use the Spark Operator for quick long time and I'm not sure if it still works as expected.

PR #1674, opened 2023-01-27
add lifecycle to executor

tiechengsu:
lifecycle rule for driver was added in this commit
https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/commit/18572fa33bb0b99ed2f5ce24fa8be8f364c727f4

This PR add lifecycle to executor


liyinan926 (approved):
LGTM. Thanks.

PR #1684, opened 2023-02-27
feat: add helm flag for cert reload interval

DerekTBrown:
## What does this change do?
- Introduces a new Helm `values` flag to alter the webhook cert reload interval.
- Sets this value to 10 seconds by default.
   - If the certificate isn't loaded properly, the Spark Operator fails to do any work because it cannot communicate with Kubernetes. Reducing this to 10 seconds ensures the Spark Operator can stay available.


liyinan926:
Please bump the chart version.

PR #1685, opened 2023-02-28
Add webhook fail on error + fix docs generation

maxwell-gobel:
## Changes
- Fixed `helm-docs` make goal to correctly mount working directory
  - Bundled docs update that looked to be missing a few changes since original generation
- Added spot to configure `webhook-fail-on-error` flag via helm chart


Dlougach:
I think at least the `-webhook-fail-on-error` part is still worth adding.

PR #1719, opened 2023-04-07
Fix docs for Volcano integration

VVKot:
Docs had old chart flags for enabling batch scheduler/webhook, so it was not working.

@liyinan926 can you help merge this?


jiasheng55:
+1.
After some debugging, I finally found out that this helm install command was outdated :(
Hope this would get fixed soon.

Issue #1723, opened 2023-04-11
SparkApplication with volcano scheduler not working on OCP


disaster37:
Hi,

it seems SparkApplication not working with volcano sheduler on OCP.

I get the following error on SparkApplication Object

```
Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://10.54.0.1/api/v1/namespaces/spark-hpd1/pods. Message: admission webhook "validatepod.volcano.sh" denied the request: failed to get PodGroup for pod <spark-hpd1/spark-test-driver>: podgroups.scheduling.volcano.sh "spark-spark-test-pg" not found. Received status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=admission webhook "validatepod.volcano.sh" denied the request: failed to get PodGroup for pod <spark-hpd1/spark-test-driver>: podgroups.scheduling.volcano.sh "spark-spark-test-pg" not found, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=null, status=Failure, additionalProperties={}).
```

Same error on volcano-admission pod.

No logs on volcano-controller

I doesn't found whitch pod is supposed to create podGroup before create pod executor that run spark-submit.

It doesn't create any pod, no executor pod, no driver pod...

I follow the doc https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/volcano-integration.md

Spark-operator version: helm 1.1.27 It use image `ghcr.io/googlecloudplatform/spark-operator:v1beta2-1.3.8-3.1.1`
Volcano version: v1.7.0. It use image `volcanosh/vc-controller-manager:v1.7.0`

Add the spark applicatin:

```yaml
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: spark-test
  namespace: spark-hpd1
spec:
  type: Scala
  mode: cluster
  image: "gcr.io/spark-operator/spark:v3.1.1"
  imagePullPolicy: Always
  mainClass: org.apache.spark.examples.SparkPi
  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.12-v3.1.1.jar"
  sparkVersion: "3.1.1"
  batchScheduler: "volcano"   #Note: the batch scheduler name must be specified with `volcano`
  restartPolicy:
    type: Never
  driver:
[...]

PR #1724, opened 2023-04-12
fix: fix issue #1723 about spark-operator not working with volcano on OCP

disaster37:
Fix issue #1723 

We fix some parts of code:
- pkg/batchscheduler/volcano/volcano_scheduler.go: to return error when appear on podGroup create.
- pkg/batchscheduler/volcano/volcano_scheduler.go and manifest/spark-operator-install/spark-operator-rbac.yaml to fix right on OCP.

It fix error on OCP
```
podgroups.scheduling.volcano.sh "spark-spark-test-pg" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

### DIFF
--- a/pkg/batchscheduler/volcano/volcano_scheduler.go
+++ b/pkg/batchscheduler/volcano/volcano_scheduler.go
@@ -115,9 +115,12 @@ func (v *VolcanoBatchScheduler) getAppPodGroupName(app *v1beta2.SparkApplication
 }
 
 func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size int32, minResource corev1.ResourceList) error {
-	var err error
+	var (
+		err error
+		pg  *v1beta1.PodGroup
+	)
 	podGroupName := v.getAppPodGroupName(app)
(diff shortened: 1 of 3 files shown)


Aakcht:
Hi, @liyinan926  , any chance for reviewing this PR?

pkg/batchscheduler/volcano/volcano_scheduler.go
```diff
@@ -115,9 +115,12 @@ func (v *VolcanoBatchScheduler) getAppPodGroupName(app *v1beta2.SparkApplication
 }
 
 func (v *VolcanoBatchScheduler) syncPodGroup(app *v1beta2.SparkApplication, size int32, minResource corev1.ResourceList) error {
-	var err error
+	var (
+		err error
+		pg  *v1beta1.PodGroup
+	)
```
liyinan926, line 121: This change doesn't seem necessary given that `pg` is not used outside the if-else scope.

disaster37: You are right. I removed this change.

disaster37: what do you prefer ?

Aakcht:
@disaster37 @liyinan926 Thanks for taking a look!

PR #1795, opened 2023-07-14
CityMall is using k8s spark operator

murthy95:
* Use spark to run data migration jobs.


vara-bonthu:
@murthy95 Please resolve the conflicts so that we can merge the PR

PR #1805, opened 2023-07-27
Appropriate executor instance count defaults applied when dynamic allocation is enabled.

### LINKED ISSUE #1779: Dynamic executor allocation does not respect `minInstances = 0` and defaults to `instances = 1`
In a SparkApplication leveraging dynamic executor allocation with `minInstances = 0`, Spark Operator will set the instance count to 1, and will always create at least 1 executor.  This is due to the following function in `defaults.go`:

```
func setExecutorSpecDefaults(spec *ExecutorSpec, sparkConf map[string]string) {
        [...]
	if _, exists := sparkConf["spark.executor.instances"]; !exists && spec.Instances == nil {
		spec.Instances = new(int32)
		*spec.Instances = 1
	}
}
```

In the case of using dynamic allocation, `spark.executor.instances` is unset and thus the above logic is invoked.  

The desired behavior is that in the case where `minInstances = 0`, if no executors are required, then no executors will be created.

peter-mcclonski:
Resolves #1779


peter-mcclonski:
@ahartwellCpointe See above

PR #1827, opened 2023-08-30
Expose Prometheus port when custom config files are used

suryadutta:
> **Note**
>
> This is almost a duplicate of this PR: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/pull/1165
> I'm opening this because that PR was abandoned (was opened by a former employee at my company, SpotHero)

---

Right now, the patch that adds the Prometheus container port when Prometheus is enabled does not run if these two configurations are provided in the spec:

* `SparkApplication.spec.monitoring.metricsPropertiesFile` ([link to docs](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/api-docs.md#sparkoperator.k8s.io/v1beta2.MonitoringSpec:~:text=will%20be%20used.-,metricsPropertiesFile,-string))
* `SparkApplication.spec.monitoring.prometheus.configFile` ([link to docs](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/api-docs.md#sparkoperator.k8s.io/v1beta2.PrometheusSpec:~:text=as%20the%20default.-,configFile,-string))

This leads to an issue where the `jmx-exporter` port doesn't get exposed in either driver or executor containers if both of these configurations are set. More details here: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1053

---

The current behavior of the code is as follows:

If prometheus is enabled AND both `metricsPropertiesFile` and `prometheus.configFile` are not set:
   1. create a patch to create the Prometheus container port
   2. create a patch to create the VolumeMount at `config.PrometheusConfigMapMountPath`
   
However, only the second patch (VolumeMount) needs to be skipped if both `metricsPropertiesFile` and `prometheus.configFile` are set, since the default configs won't be mounted there.

This PR proposes modifying the logic to reflect this correct state.


suryadutta:
@liyinan926 - this PR is still waiting for a review.
 
If this project is no longer actively maintained, could you let the community know? I think this project is still being actively used by a large group of people/companies.

suryadutta:
> @suryadutta Thanks for the PR. I have not been able to do much for this repo in the past couple of years. We are trying to figure out what to do for this repo at this point.

@liyinan926 Totally understand, appreciate the response!

1350lumen:
The issue fixed in this PR still makes sense. Can it be reviewed and merged?

yuchaoran2011:
@1350lumen Thanks for the reminder and @suryadutta for the contribution. I somehow missed this PR earlier. This is a great fix

yuchaoran2011:
@suryadutta Build failed with
```
resource used in docker.io/kubeflow/spark-operator has changed in pkg/, need to update the appVersion in charts/spark-operator-chart/Chart.yaml
```
Could you make the change accordingly?

PR #1884, opened 2023-11-27
Chart: add patch permissions for spark operator SA to support spark 3.5.0

Aakcht:
While trying to run spark operator with Spark 3.5.0, I noticed the following errors:

```
Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. services "spark-pi-test-run-1-72fa8a8c0fd5639e-driver-svc" is forbidden: User "system:serviceaccount:spark:sparkoperator-spark-spark-operator" cannot patch resource "services" in API group "" in the namespace "spark-test-apps".
 at io.fabric8.kubernetes.client.KubernetesClientException.copyAsCause(KubernetesClientException.java:238)
 at io.fabric8.kubernetes.client.dsl.internal.OperationSupport.waitForResult(OperationSupport.java:518)
 at io.fabric8.kubernetes.client.dsl.internal.OperationSupport.handleResponse(OperationSupport.java:535)
 at io.fabric8.kubernetes.client.dsl.internal.OperationSupport.handlePatch(OperationSupport.java:430)
 at io.fabric8.kubernetes.client.dsl.internal.OperationSupport.handlePatch(OperationSupport.java:408)
 at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.handlePatch(BaseOperation.java:713)
 at io.fabric8.kubernetes.client.dsl.internal.HasMetadataOperation.lambda$patch$2(HasMetadataOperation.java:232)
 at io.fabric8.kubernetes.client.dsl.internal.HasMetadataOperation.patch(HasMetadataOperation.java:237)
 at io.fabric8.kubernetes.client.dsl.internal.HasMetadataOperation.patch(HasMetadataOperation.java:252)
 at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.serverSideApply(BaseOperation.java:1132)
 at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.serverSideApply(BaseOperation.java:92)
 at io.fabric8.kubernetes.client.extension.ResourceAdapter.serverSideApply(ResourceAdapter.java:325)
 at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
 at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
 at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
[...]

PR #1908, opened 2024-01-09
added patch permission for spark-operator for services

AndrewChubatiuk:
spark operator fails after some time to patch spark application service


zevisert:
This is something I've run into as well. This diff is also the exact same fix we've been using without error for quite a while now. It'd be great if this was merged.

zevisert:
This is already proposed in #1884, I think this PR should be closed since there's more information in #1884.

zevisert:
Coming back to life, it just got transferred to kubeflow since google couldn't provide maintainers anymore

Issue #1342, opened 2021-09-05
Support labels for SparkUIConfiguration


themoah:
Hi,

Is there an option to support additional labels on SparkUIConfiguration? Or an ability to propagate labels from SparkApplication to UI? As now service created by Spark Operator has only these key:
```
    sparkoperator.k8s.io/app-name: 
    sparkoperator.k8s.io/submission-id: 
```

But I'm looking for an option to add custom ones. 

Thanks!

Cian911:
Jumping in here to revive this... 

In my org, we're using [opa gatekeeper](https://github.com/open-policy-agent/gatekeeper) which has restrictions and best practices around labels for k8s manifests. Whilst we _can_ whitelist services from these restrictions, we'd rather not. So having the ability to set labels for services (or anything really) would be very handy.

Short of forking the repo and taking a stab at it myself - Is there any plans to accomodate this? @liyinan926

Cian911:
Ok, I made the change myself: https://github.com/kubeflow/spark-operator/pull/1952

PR #1952, opened 2024-04-06
feat: add support for custom service labels

Cian911:
There is no way currently to add custom labels to services created by the `spark-operator`. This can be very useful for anyone who uses [opa-gatekeeper](https://github.com/open-policy-agent/gatekeeper) for policy controls in their kubernetes clusters.

See related issues of users asking for this: https://github.com/kubeflow/spark-operator/issues/1342


yuchaoran2011 (changes requested):
Thanks for the contribution. Indentation has issues. Please make sure to run `go fmt` to reformat the code. `serviceLabels` sounds generic. Can we make it clear by renaming it to something like `sparkUiSvcLabels`?

Cian911:
> Thanks for the contribution. Indentation has issues. Please make sure to run `go fmt` to reformat the code. `serviceLabels` sounds generic. Can we make it clear by renaming it to something like `sparkUiSvcLabels`?

@yuchaoran2011 Thanks for taking a look, and the quick response. I have pushed a fix for the formatting :+1: 

However, with regards to renaming `serviceLabels`, I'm not sure I agree. The change I made, I feel, is adhering to the naming already in place for other attributes on the `SparkUIConfiguration` struct: [see here](https://github.com/Cian911/spark-operator/blob/9054d23ced1bbc171889f7b9ac069981a328ff37/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L304).

I _can_ make this change if insisted, but I personally don't see the benefit in the following configuration (as this suggestion might entail):

```yaml
sparkUIOptions:
    sparkUiSvcLabels:
      test-label/v1: 'true'
``` 

vs

```yaml
sparkUIOptions:
    serviceLabels:
      test-label/v1: 'true'
```

yuchaoran2011 (changes requested):
You are right. I missed that the newly added serviceLabels is under UI configurations. That's good then. One last change: another PR got merged before yours that took the 1.1.29 version number, now the chart number needs to be bump up to 1.1.30

Cian911:
@yuchaoran2011 Done :)

yuchaoran2011 (approved):
/lgtm

PR #1955, opened 2024-04-07
support multiple namespaces

AndrewChubatiuk:
replaced sparkJobNamespace with sparkJobNamespaces, added creation of spark role, serviceaccount and rolebinding for every spark job namespace

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -17,7 +17,18 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v4
         with:
-          fetch-depth: "0"
+          fetch-depth: '0'
+
+      - name: The API should not change once published
+        run: |
+          if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta1; then
+            echo "sparkoperator.k8s.io/v1beta1 api has changed"
+            false
+          fi
+          if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta2; then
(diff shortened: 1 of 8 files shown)


yuchaoran2011 (changes requested):
I'm not sure if changing Roles and RoleBindings to ClusterRoles and ClusterRoleBindings respectively is a good idea. RBAC permissions should be as tight as possible for security reasons. It's better to create a role and role binding for each application namespace

AndrewChubatiuk:
> I'm not sure if changing Roles and RoleBindings to ClusterRoles and ClusterRoleBindings respectively is a good idea. RBAC permissions should be as tight as possible for security reasons. It's better to create a role and role binding for each application namespace

Changed to Role and RoleBinding loop

yuchaoran2011 (commented):
Could you add more details to how this PR was tested? Specifically, it will be good to verify that webhook still works as expected. Some web hook resources are namespace specific. I don't remember which ones off the top of my head, but the webhook service and MutatingWebhookConfiguration objects are good to validate

AndrewChubatiuk:
@yuchaoran2011 I've beed running it in my project for several month in such setup. I see mutatingwebhooks for pods only and it monitors namespaces depending on operator's `-namespace`

AndrewChubatiuk:
it makes sense to migrate operator to kubebuilder v4, it simplifies setup and management a lot and also reduces amount of code lines

AndrewChubatiuk:
@yuchaoran2011 do you expect any other changes here?

yuchaoran2011 (approved):
/lgtm

PR #1956, opened 2024-04-07
Cleanup after golang upgrade

AndrewChubatiuk:
upgraded golang version in docker image
go.sum dependencies cleanup


.github/workflows/main.yaml
```diff
@@ -15,21 +15,10 @@ jobs:
...
-        run: |
-          if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta1; then
-            echo "sparkoperator.k8s.io/v1beta1 api has changed"
-            false
-          fi
-          if ! git diff --quiet origin/master -- pkg/apis/sparkoperator.k8s.io/v1beta2; then
-            echo "sparkoperator.k8s.io/v1beta2 api has changed"
-            false
-          fi
-
```
yuchaoran2011, line 32: what's the reason for removing this part?

Makefile
```diff
@@ -62,7 +62,7 @@ detect-crds-drift:
 
 clean:
 	@echo "cleaning up caches and output"
-	go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
+	go clean -r -x ./... 2>&1 >/dev/null
```
yuchaoran2011, line 65: We still want to clean the cache for the tests. Any reason those two flags got removed?

.github/workflows/main.yaml
AndrewChubatiuk: there are changes in formatting in golang 1.19 -> 1.22, which caused changes at these locations. cannot ignore formatting changes, cause format check is failing then

yuchaoran2011: But the removed code here is `git` command and shell script, not related to Go lang. Or am I missing something?

AndrewChubatiuk: this lines check if there were changes in `pkg/apis/sparkoperator.k8s.io/v1beta1` and as you can see from PR, there are formatting changes there

.github/workflows/release.yaml
```diff
@@ -4,73 +4,122 @@ on:
...
-          DOCKER_TAG=$(cat charts/spark-operator-chart/Chart.yaml | grep "appVersion: .*" | cut -c13-)
-          docker build -t gcr.io/spark-operator/spark-operator:${DOCKER_TAG} .
-          echo "Ideally, we'd release the docker container at this point, but the maintainer of this repo needs to approve..."
-          docker tag gcr.io/spark-operator/spark-operator:${DOCKER_TAG} ghcr.io/googlecloudplatform/spark-operator:${DOCKER_TAG}
-          if ! docker pull ghcr.io/googlecloudplatform/spark-operator:${DOCKER_TAG}; then
-            docker push ghcr.io/googlecloudplatform/spark-operator:${DOCKER_TAG}
-          else
-            echo "Spark-Operator Docker Image alredy exists"
-          fi
-
```
yuchaoran2011, line 55: Is the reason for removing these lines to get the CI to pass? If that's the case, I don't think it should be in this PR since it's not related to Go upgrade. As I was suggesting in another PR: https://github.com/kubeflow/spark-operator/pull/1884#issuecomment-2041159084, we should fix the pipelines first

AndrewChubatiuk: lines were not just simply removed, I've replaced it with multiarch image build + helm chart publishing

yuchaoran2011: You are right. I was looking for `docker push` but didn't realize that it was achieved via `docker/build-push-action`

yuchaoran2011 (approved):
/lgtm

.github/workflows/main.yaml
yuchaoran2011: @AndrewChubatiuk Now that the PR is merged, could you create another PR to add this part back?

PR #1964, opened 2024-04-12
publish chart independently, incremented both chart and image versions to trigger build of both

AndrewChubatiuk:
moved chart publishing to a separate job which runs after publish image is skipped or succeeded


vara-bonthu (approved):
/lgtm

PR #1954, opened 2024-04-06
upgraded golang and dependencies

AndrewChubatiuk:
Upgraded golang 1.19 -> 1.22 and all dependencies

PR #1630, opened 2022-10-24
update go to 1.19 + k8s.io libs to v0.25.3

ImpSy:
Hello,

This PR upgrade the go version to 1.19 and all k8s dependencies to v0.25.3
This upgrade would allow us to benefit from the fix regarding a memory leak in `k8s.io/client-go/util/workqueue`

This version of the operator has been running in production for the past week on our side (scheduling ~20K apps per day)

If I need to change anything do not hesitate to tell me

PR #1968, opened 2024-04-13
chore: remove k8s.io/kubernetes replaces and adapt to v1.29.3 apis

ajayk:


ajayk:
cc @AndrewChubatiuk

vara-bonthu:
@ajayk, it looks like the PR description is missing. Could you please provide more details to help us review your changes effectively? 

**Purpose of the Change:** Why is this change necessary?
**Problem Addressed:** What issue are you encountering that this change resolves?
**Solution:** How does this change solve the problem?
**Impact of Not Merging:** What would be the consequences if this PR is not merged?
**Existing Workarounds:** Is there a current workaround for this issue, and if so, what are its drawbacks?


Thank you for contributing, and we look forward to your detailed response!

debasishbsws:
@vara-bonthu It seems to me that while upgrading the k8s.io/kubernetes module version in PR: https://github.com/kubeflow/spark-operator/pull/1630 the person missed removing the replace statement present in go.mod file for Kubernetes https://github.com/kubeflow/spark-operator/blob/6ded3acd1c5058c1d737e40d6135b5606dc68e43/go.mod#L161

For this reason, even after Updating the module version in #1630 #1954, it is set to version 1.19.6, Which is 10 versions older than the current one 1.29.3 [link](https://github.com/kubeflow/spark-operator/pull/1954/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R29). This older version of Kubernetes has multiple CVE present in it.

just removing the replace line doesn't fix as there is a version mismatch with some other go modules this PR try to fix those with minimal changes.

yuchaoran2011 (approved):
Good catch! /lgtm

Issue #1574, opened 2022-07-08
Spark Driver Pod Getting Stuck in init state due to no driver configmap found


gangahiremath:
Hello,

 Spark Application driver pod get stuck in init state due to driver configmap not found issue; what could be possible reason for the same and this results in Spark Operator stopping as process id (mentioned below) responsible for SparkSubmit also gets stuck. This is not consistent, encountered by sample Pi Spark application that has run successfully on number of times in the same environment.

sparkxpix75ba3dfb-driver             0/1     Init:0/1          0          47m

Warning  FailedMount  19s (x20 over 31m)    kubelet            (combined from similar events): MountVolume.SetUp failed for volume "spark-conf-volume" : configmap "sparkxpix75ba3dfb-1657310554583-driver-conf-map" not found

Process ID invoked by SparkSubmit is also stuck

[...]

Issue #1489, opened 2022-02-25
Support setting driver/executor `memory` and `memoryLimit` separately


peay:
Spark's kubernetes executor [uses the same value for memory request and memory limit](https://github.com/apache/spark/blob/master/resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/features/BasicExecutorFeatureStep.scala#L212), and the current operator API matches that: although we have both `cores` and `coreLimit`, there is only `memory`.

However, in some cases, it can be useful to set a memory request that is lower than the memory limit, as a form of over-subscription, as Spark tasks will not always use all of the requested memory, and it can help increase overall cluster memory utilization.

In extreme cases, this could also be seen as a counter-part to how you can disable memory enforcement in YARN clusters, although kubernetes would give us here the opportunity to tune this much more finely than just enabling or disabling memory enforcement.

My assumption is that the operator could support different limit and request values even though Spark doesn't through the mutating webhook.

vpbarb:
Spark Operator doesn't make any magic here. When `coreLimit` is specified it's simply translated to `spark.kubernetes.driver.request.cores` or `spark.kubernetes.executo.request.cores` Spark property. It's Spark itself that sets pods requests and limits taking the values from Spark properties. As you can find [here](https://spark.apache.org/docs/latest/running-on-kubernetes.html), there are no similar properties for memory. The pod memory `requests` and `limits` are just calculated from `spark.{driver,executor}.memory` and `spark.{driver,executor}.memoryOverhead`. 

So regarding the topic, at first Spark should support this. But I doubt it will. Requesting less pod memory than it's declared for the Spark internals may lead to the pod dying because of OOM. Even if the user does it intentionally it will raise more questions.

pycgo:
+1

harryzhang2016:
any progress on this issue?

mk-raven:
Also interesting about progress

filippomascolo:
+1

TheDevilDan:
+1

catinwarmhands:
+1

missedone:
@jiakin888 , you may try this new feature that defines the pod template for driver and executor https://github.com/kubeflow/spark-operator/pull/2141

jiakin888:
> @jiakin888 , you may try this new feature that defines the pod template for driver and executor #2141

Unfortunately，it not work

bnetzi:
Fixed in

[...]

Issue #1138, opened 2021-01-21
Long delay between SparkApplicationAdded and SparkApplicationSubmitted


jystephan:
Hello all,

There's a small delay (usually 3-4 seconds) between the events SparkApplicationAdded and SparkApplicationSubmitted, as in the logs below:

```
16:41:08: [spark-operator] SparkApplicationAdded: SparkApplication word-count-20210121-164108-feuds was added, enqueuing it for submission 
16:41:11: [driver:default-scheduler] Scheduled: Successfully assigned spark-apps/word-count-20210121-164108-feuds-driver to <node-pool>
16:41:11: [driver:kubelet] FailedMount: MountVolume.SetUp failed for volume "spark-conf-volume" : configmap "word-count-20210121-164108-feuds-e635b37725d239f6-driver-conf-map" not found 
16:41:12: [spark-operator] SparkApplicationSubmitted: SparkApplication word-count-20210121-164108-feuds was submitted successfully 
```

I've observed that this delay increases when many applications are submitted roughly at the same time. 
In my test, I submitted 4 applications, and here are the times I got:

* App 1: delay of 10 seconds (SparkApplicationAdded 16:37:25, SparkApplicationSubmitted 16:37:35)
* App 2: delay of 13 seconds (SparkApplicationAdded 16:37:26, SparkApplicationSubmitted 16:37:39)
* App 3: delay of 14 seconds (SparkApplicationAdded 16:37:27, SparkApplicationSubmitted 16:37:41)
* App 4: delay of 16 seconds (SparkApplicationAdded 16:37:29, SparkApplicationSubmitted 16:37:45)

If I submit 10 applications, the delay grows to 35 seconds in my tests, and I think that by submitting even more applications I can make this delay arbitrarily large. 

My questions:
* What is the source of the delay? What is the main bottleneck?
* How can we troubleshoot this (which kind of logs should we look at, if any)? 
* How can we reduce this delay (e.g. spark-operator properties, or other changes)?

Advice/Feedback from the community much appreciated -- Thank you!

shinen:
Having similar issue. My case is even worst that usually 30~mins needed for an added sparkApplication to be submitted by the operator. 

Using latest helm chart to deploy to aks.

During the delay, no event is watched by the spark operator. What I can observe is that only the sparkApplication is created. No driver no executor and no events in spark operator log. 

The same also happens to delete event that terminated spark-driver does not get cleaned up after deleting the sparkApplication.

[...]

Issue #783, opened 2020-01-29
Is it possible to set memory limit for spark application POD through operator?


shensonj:
Is it possible to set memory limit for spark application POD through operator?  I would like to give higher memory for the container than JVM memory. Is there any way I can achieve this? I am having OOM error even though JAVA process looks fine.

@liyinan926 @mrow4a

shensonj:
@jamesthegiantpeach

liyinan926:
The memory limit of the driver/executor container is set to be the same as the memory request, taking in some overhead factor applied to the memory requested through `driver.memory` and `executor.memory`.

liyinan926:
Closing this and please use the other issue (#781) for further questions.

PR #1990, opened 2024-04-17
Draft: Performance mega boost - queue per app 

bnetzi:
## Issues related 
https://github.com/kubeflow/spark-operator/issues/1138
https://github.com/kubeflow/spark-operator/issues/1574
https://github.com/kubeflow/spark-operator/issues/783
https://github.com/kubeflow/spark-operator/issues/1489

## Purpose of this PR
First of all - this PR is mainly a Draft which we think should be discussed, and that's why we are submitting it even though we haven't added documentation not unit tests.
 
The original design of spark-operator posses one queue which is used for all the applications.
This design is causing a huge latency issue when trying to deal with hundreds / thousands of applications concurrently.

In benchmarks me and my fellows from Mobileye preformed, we showed clearly a linear latency increase depend on the amount of apps handled.
When getting to more than 500 applications, the avg time from creating a spark application object until pod creation is ~130 seconds for app to start. When going up to 1200 apps concurrently it can go Up to 20 minutes on average for each spark application to be created.
Scale up vertically would not be helpful as the CPU is doing nothing, most of the time is spent on the queue mutex.

The change we are presenting here is to  create a queue for each app.
It required a big change all around the code, but it is not changing the main flow in any way.

Our benchmarks showed that even with 1000 apps concurrently, Avg time for application creation is ~7 seconds

We also added a nice feature of using memoryLimit for driver / executor which is larger than the request by using the admission webhook.

**Proposed changes:**
-  create a queue for each app
- Add a parameter to controll qps and burst for k8s api
- adding memoryLimit option for executors and drivers

## Change Category

What are we still missing:

- fixing broken unit test
- documentation
- Peer review

I would point out that this code currently runs on our production environment with massive scale without any issues.

### DIFF
--- a/charts/spark-operator-chart/templates/deployment.yaml
+++ b/charts/spark-operator-chart/templates/deployment.yaml
@@ -74,6 +74,11 @@ spec:
         args:
         - -v={{ .Values.logLevel }}
         - -logtostderr
+        - -max-queue-time-without-update-in-minutes=30
+        - -queue-cleaner-interval-in-minutes=10
+        - -enable-profiling=false
+        - -api-qps={{ .Values.k8sBurstQps }}
+        - -api-burst={{ .Values.k8sQps }}
         {{- if eq (len $jobNamespaces) 1 }}
         - -namespace={{ index $jobNamespaces 0 }}
         {{- end }}
--- a/export-pprof.sh
+++ b/export-pprof.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# Generate a timestamp
+timestamp=$(date +%Y-%m-%d_%H-%M-%S)
+
+# Function to fetch goroutine profile
+fetch_goroutine_profile() {
+    local profile_type=$1  # 'start' or 'end'
+    local profile_path="/tmp/goroutine-profile-debug1-$profile_type-$timestamp.prof"
+    echo "Starting to fetch Goroutine profile ($profile_type) at $(date)..."
+    curl "http://localhost:6060/debug/pprof/goroutine?debug=1" -o "$profile_path"
+    echo "Completed fetching Goroutine profile ($profile_type) at $(date). File saved to $profile_path"
+}
+
+echo "Starting to fetch CPU profile at $(date)..."
+cpu_profile="/tmp/cpu-profile-30sec-$timestamp.prof"
+curl "http://localhost:6060/debug/pprof/profile?seconds=30" -o "$cpu_profile" &
+echo "CPU profile fetch initiated, running for 30 seconds..."
+
+echo "Starting to fetch Trace profile at $(date)..."
+trace_profile="/tmp/trace-profile-30sec-$timestamp.prof"
+curl "http://localhost:6060/debug/pprof/trace?seconds=30" -o "$trace_profile" &
+echo "Trace profile fetch initiated, running for 30 seconds..."
+
+echo "Fetching initial Goroutine profile..."
+fetch_goroutine_profile "start" &
+
+# Wait for CPU and trace profiling to complete
+wait
+
+echo "Starting to fetch final Goroutine profile after waiting for other profiles to complete..."
+fetch_goroutine_profile "end"
+
+echo "All profiling data collected"
+
+# Copying profiles to S3 bucket
+echo "CPU profile output - $cpu_profile"
+echo "Trace profile output - $trace_profile"
+echo "Initial Goroutine profile output - /tmp/goroutine-profile-debug1-start-$timestamp.prof"
+echo "Final Goroutine profile output - /tmp/goroutine-profile-debug1-end-$timestamp.prof"
+
--- a/pkg/controller/sparkapplication/controller.go
+++ b/pkg/controller/sparkapplication/controller.go
@@ -18,9 +18,12 @@ package sparkapplication
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
+	"math/big"
 	"os/exec"
-	"strings"
+
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
@@ -72,11 +75,18 @@ type AppQueueInfo struct {
 	LastUpdateTs time.Time
 }
 
+// A struck that will have map of queues and mutex for the map
+// We need many of those as one mutex will get us stuck in locks in high scale
+type AppQueuesMap struct {
+	appQueueMutex *sync.RWMutex
+	appQueueMap   map[string]AppQueueInfo
+}
+
 // Controller manages instances of SparkApplication.
 type Controller struct {
 	crdClient         crdclientset.Interface
 	kubeClient        clientset.Interface
-	appQueues         map[string]AppQueueInfo
+	appQueuesMaps     []AppQueuesMap
 	cacheSynced       cache.InformerSynced
 	recorder          record.EventRecorder
 	metrics           *sparkAppMetrics
@@ -88,21 +98,50 @@ type Controller struct {
 	enableUIService   bool
 }
 
-// This function checks if there is a queue for @appName, if not it creates a new AppQueueInfo struct for this @appName.
+func (c *Controller) GetRelevantMap(appName string) AppQueuesMap {
+	// Hash the string using SHA256
+	hash := sha256.Sum256([]byte(appName))
+
+	// Convert the hash to a big integer
+	hashInt := new(big.Int).SetBytes(hash[:])
+
+	// Calculate the modulo to get a value between 0 and queues len
+	modValue := new(big.Int).SetInt64(int64(len(c.appQueuesMaps)))
+	result64 := new(big.Int).Mod(hashInt, modValue)
+
+	result := int(result64.Int64())
+
+	// Print the resulting integer
+	glog.V(2).Infof("Map num for app %s: %d, mutex adress is %p", appName, result, c.appQueuesMaps[result].appQueueMutex)
+	return c.appQueuesMaps[result]
+}
+
 func (c *Controller) GetOrCreateRelevantQueue(appName string) AppQueueInfo {
-	queueInfo, exists := c.appQueues[appName]
+	m := c.GetRelevantMap(appName)
+	glog.V(2).Infof("for app %s, mutex adress is %p", appName, m.appQueueMutex)
+	m.appQueueMutex.RLock()
+	queueInfo, exists := m.appQueueMap[appName]
+	m.appQueueMutex.RUnlock()
 	if !exists {
+		m.appQueueMutex.Lock()
+		defer m.appQueueMutex.Unlock()
+		// check if the queue was created while locking the mutex
+		queueInfo, exists = m.appQueueMap[appName]
+		if exists {
+			return queueInfo
+		}
+
 		// Create a new queue for the app
-		controllerName := fmt.Sprintf("spark-application-controller-%v", appName)
+		queueName := fmt.Sprintf("spark-application-queue-%v", appName)
 		queue := workqueue.NewNamedRateLimitingQueue(&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(queueTokenRefillRate), queueTokenBucketSize)},
-			controllerName)
+			queueName)
 
 		queueInfo := AppQueueInfo{
 			Queue:        queue,
 			LastUpdateTs: time.Now(),
 		}
 
-		c.appQueues[appName] = queueInfo
+		m.appQueueMap[appName] = queueInfo
 
 		// create a worker for the app queue
 		go func(appName string) { c.runWorker(appName) }(appName)
@@ -112,7 +151,6 @@ func (c *Controller) GetOrCreateRelevantQueue(appName string) AppQueueInfo {
 	return queueInfo
 }
 
-// This function enqueues an event to the @appName queueInfo.
 func (c *Controller) AddToRelevantQueue(appName interface{}) {
 	// Use the interface as string
 	if appString, ok := appName.(string); ok {
@@ -135,6 +173,7 @@ func NewController(
 	ingressClassName string,
 	batchSchedulerMgr *batchscheduler.SchedulerManager,
 	enableUIService bool,
+	mapsNum int,
 ) *Controller {
 	crdscheme.AddToScheme(scheme.Scheme)
 
@@ -145,7 +184,7 @@ func NewController(
 	})
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, apiv1.EventSource{Component: "spark-operator"})
 
-	return newSparkApplicationController(crdClient, kubeClient, crdInformerFactory, podInformerFactory, recorder, metricsConfig, ingressURLFormat, ingressClassName, batchSchedulerMgr, enableUIService)
+	return newSparkApplicationController(crdClient, kubeClient, crdInformerFactory, podInformerFactory, recorder, metricsConfig, ingressURLFormat, ingressClassName, batchSchedulerMgr, enableUIService, mapsNum)
 }
 
 func newSparkApplicationController(
@@ -159,18 +198,24 @@ func newSparkApplicationController(
 	ingressClassName string,
 	batchSchedulerMgr *batchscheduler.SchedulerManager,
 	enableUIService bool,
+	mapsNum int,
 
 ) *Controller {
 
 	controller := &Controller{
 		crdClient:         crdClient,
 		kubeClient:        kubeClient,
-		appQueues:         make(map[string]AppQueueInfo),
+		appQueuesMaps:     make([]AppQueuesMap, mapsNum),
 		recorder:          eventRecorder,
 		ingressURLFormat:  ingressURLFormat,
 		ingressClassName:  ingressClassName,
 		batchSchedulerMgr: batchSchedulerMgr,
 		enableUIService:   enableUIService,
+	}
+	// Initiate the elements within appQueuesMaps
+	for i := range controller.appQueuesMaps {
+		controller.appQueuesMaps[i].appQueueMutex = &sync.RWMutex{}
+		controller.appQueuesMaps[i].appQueueMap = make(map[string]AppQueueInfo)
 	}
 
 	if metricsConfig != nil {
@@ -203,7 +248,7 @@ func newSparkApplicationController(
 	return controller
 }
 
-// StartQueueCleanupRoutine starts a go routine to clean up queues for SparkApplications that have not been updated for a certain period of time.
+// StartQueueCleanupRoutine starts a go routine to clean up queues for SparkApplications that have not been updated for a certain period.
 func (c *Controller) StartQueueCleanupRoutine(checkInterval time.Duration, maxAge time.Duration, stopCh <-chan struct{}) {
 	ticker := time.NewTicker(checkInterval)
 	defer ticker.Stop()
@@ -221,23 +266,37 @@ func (c *Controller) StartQueueCleanupRoutine(checkInterval time.Duration, maxAg
 
 // cleanupQueues deletes the queues for SparkApplications that have not been updated for a certain period.
 func (c *Controller) cleanupQueues(maxAge time.Duration) {
-	c.LogQueuesKeys()
-	if len(c.appQueues) == 0 {
-		glog.V(0).Info("No queues to clean up, skipping cleanup")
-		return
-	}
-	queuesBefore := len(c.appQueues)
-	now := time.Now()
-	glog.V(0).Infof("Cleaning up queues for SparkApplications at time: %v", now)
-	for appName, queueInfo := range c.appQueues {
-		glog.V(0).Infof("Checking queue for app %v", appName)
-		if now.Sub(queueInfo.LastUpdateTs) > maxAge {
-			glog.V(0).Infof("Sending %v to deletion due to inactivity", appName)
+	queuesBefore := 0
+	queuesAfter := 0
+	for i := range c.appQueuesMaps {
+		currM := c.appQueuesMaps[i].appQueueMap
+		currL := c.appQueuesMaps[i].appQueueMutex
+		if len(currM) == 0 {
+			glog.V(0).Info("No need to clean up Map %d as it is empty, skipping cleanup", i)
+			continue
+		}
+		queuesBefore += len(currM)
+		now := time.Now()
+		glog.V(0).Infof("Cleaning up queues for SparkApplications at time: %v", now)
+		// we prevent deadlock we mark the quesues for deletion and delete them outside the lock
+		var queuesToDelete []string
+		currL.RLock()
+		for appName, queueInfo := range currM {
+			glog.V(0).Infof("Checking queue for app %v", appName)
+			if now.Sub(queueInfo.LastUpdateTs) > maxAge {
+				glog.V(0).Infof("Sending %v to deletion due to inactivity", appName)
+				queuesToDelete = append(queuesToDelete, appName)
+			}
+		}
+		currL.RUnlock()
+		for _, appName := range queuesToDelete {
 			c.deleteImmediateQueue(appName)
 		}
+
+		queuesAfter += len(currM)
 	}
-	queuesAfter := len(c.appQueues)
 	glog.V(0).Infof("Cleaned up %d queues, %d remaining", queuesBefore-queuesAfter, queuesAfter)
+
 }
 
 // StopQueueCleanupRoutine stops the go routine that cleans up queues.
@@ -261,9 +320,12 @@ func (c *Controller) Start(maxAge time.Duration, checkInterval time.Duration, st
 func (c *Controller) Stop(stopCh chan struct{}) {
 	glog.Info("Stopping the SparkApplication controller")
 	c.StopQueueCleanupRoutine(stopCh)
-	for _, queueInfo := range c.appQueues {
-		queueInfo.Queue.ShutDown()
+	for i := range c.appQueuesMaps {
+		for _, queueInfo := range c.appQueuesMaps[i].appQueueMap {
+			queueInfo.Queue.ShutDown()
+		}
 	}
+
 }
 
 // Callback function called when a new SparkApplication object gets created.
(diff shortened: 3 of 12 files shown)


vara-bonthu:
/assign @ChenYi015 
/assign @yuchaoran2011 

Please review the changes when you get a chance

yuchaoran2011 (commented):
Impressive work! A few questions:
* What do we need K8s burst and QPS for?
* What's the impact on memory and CPU consumption of setting a higher or lower frequency of queue cleaning?
* Will there be edge cases where a queue's mutex is already deleted due to not receiving any updates for a while but the application still exists?

bnetzi:
> Impressive work! A few questions:
> 
> * What do we need K8s burst and QPS for?
> * What's the impact on memory and CPU consumption of setting a higher or lower frequency of queue cleaning?
> * Will there be edge cases where a queue's mutex is already deleted due to not receiving any updates for a while but the application still exists?

Hi @yuchaoran2011  thanks for reviewing!
- Qps:  we first did all the rest of the code and did not increase qps and Burst and the outcome was pretty good, but several issues still bothered us:
 1. CPU usage was very high at high scale
 2. Even though app submit was fast, application status updates were very slow and could take 20 minutes at scale, which was really bad for us
 3. At some point app were just stuck forever without status, even though the app was finished, and when spark operator restarted - it submitted the apps again as they had no status and no resources related
After profiling we discovered most of the time was spent on kiss API exponential back off that was caused from low qps and Burst
(We also tested just raising the qps and Burst without the rest of our changes and it was not preforming well)

- As for memory and CPU consumption -  we mostly tried to avoid running the queue clean often as it's locking a map when it cleans it. We haven't noticed high memory nor CPU but I can provide some CPU and memory consumption graphs to show the affects of amount of apps etc.

- about the edge case, yes of course it can happen, especially of you
-  set low values of queue cleanup intervals, but it does not cause any harm as every access to an app queue is always by the get or create method and the cleanup is not cleaning queues that aren't empty

luohao:
Thanks for this patch!

We also ran into similar issues at scale. What we observed was a race condition between the SparkOperator reconciliation and k8s Pod GC. We have autoscaler that terminates nodes when no Pods is running so completed Pods are deleted from API server when the node is terminated. If the SparkOperator could not update the status in time, it won't find the driver Pod and thus make the app as failed due to `driver pod not found`.

charts/spark-operator-chart/templates/deployment.yaml
```diff
@@ -74,6 +74,11 @@ spec:
         args:
         - -v={{ .Values.logLevel }}
         - -logtostderr
+        - -max-queue-time-without-update-in-minutes=30
+        - -queue-cleaner-interval-in-minutes=10
```
vara-bonthu, line 78: Can we make these two values configurable? Would there be a possibility for users to modify these values?

export-pprof.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Generate a timestamp
+timestamp=$(date +%Y-%m-%d_%H-%M-%S)
+
+# Function to fetch goroutine profile
+fetch_goroutine_profile() {
```
vara-bonthu, line 7: Is this script used anywhere? Couldn't find the references or docs.

vara-bonthu (commented):
@ben-tzionlmb Looking forward to see this feature to be merged. Would be nice to add some tests with this feature.

charts/spark-operator-chart/templates/deployment.yaml
yuchaoran2011: +1

yuchaoran2011:
@bnetzi Thanks for the clarification. That makes sense. The default is 50, if I remember correctly. Doubling the Spark operator QPS strikes a balance between allowing for more requests from operator to be processed and not overloading k8s API server. Overall the idea looks really solid. I second fleshing out this PR to include tests

pkg/controller/sparkapplication/controller.go
```diff
@@ -65,11 +69,24 @@ var (
...
+	LastUpdateTs time.Time
+}
+
+// A struck that will have map of queues and mutex for the map
+// We need many of those as one mutex will get us stuck in locks in high scale
+type AppQueuesMap struct {
+	appQueueMutex *sync.RWMutex
+	appQueueMap   map[string]AppQueueInfo
+}
+
```
yuchaoran2011, line 84: The naming of this struct is strange. The overall type name is `AppQueuesMap` but a field is named similarly `appQueueMap`. Can we give the struct type a more intuitive name? Something that indicates that it includes the mutex?

export-pprof.sh
bnetzi: This script is used for manual profiling which we've added as part of this feature (disabled by default). I probably need to add some documenting on how to use it

pkg/controller/sparkapplication/controller.go
bnetzi: Yes, it is not good naming, I will change it

bnetzi:
> @bnetzi Thanks for the clarification. That makes sense. The default is 50, if I remember correctly. Doubling the Spark operator QPS strikes a balance between allowing for more requests from operator to be processed and not overloading k8s API server. Overall the idea looks really solid. I second fleshing out this PR to include tests

The default in golang is actually 5 qps and 10 burst which is very low.

Maybe we should put default value which is lower than 100 and 200 by default for small k8s clusters or such with mix usage

vincentye38:
Please help me to understand your statement in the description more

> Scale up vertically would not be helpful as the CPU is doing nothing, most of the time is spent on the queue mutex.

For the original runWorker()->processNext(), `processNext()`'s `queue.Get()` get a spark-app to process. Multiple workers calling `runWorker()` each will pull spark-app items from `queue` sequentially because of queue mutex, but the items will be processed concurrently inside each worker's `processNext()`. Is `20 minutes on average` incurred by processing the item, but not by blocking in `queue.Get()`?

yuchaoran2011:
@bnetzi I think @vincentye38 brought up a valid point above. What he's proposing (i.e. having more workers concurrently process items) complements the solution proposed in your PR. Have you tried bumping up the number of workers to see what effect that has on performance?

bnetzi:
@vincentye38 @yuchaoran2011 Thanks for commenting

First I would mention that we obviously increased the amount of controller threads while raising the cpu, and it did not help at all. the cpu was 5% on a 32 vCPU instance. 

I think that in order to understand it we need to understand some of the mechanism and parameters involved here.

In my understanding (and this behavior is what we saw in the logs as well) each of the go routines is waiting in line to preform a q.Get() but the next item in queue be processed by another go routine only when q.Done() is called **which is after we are doing the sync**. It actually makes a lot of sense as if we have an event that triggers app creation, and then some app update - we want the creation to be done and the app state to be synced before we are preforming the next operation.
But this design makes **all other go routines to wait** for this task to be done. So the controller threads not making a significant impact on performance.

Also - the operator is using a a RateLimitingInterface workqueue, and this queue is also adding waits when the rate limit is reached, so it makes everything even worse. We tried to increase the limits and it helped a little but did not help enough, and that's why we decided on a queue per app approach.

Another point to mention here, (and in my opinion it is what  causing us to get to finally 20 minutes after running a long time in a large scale env) is that we have a vast amount of events per app. It is not just app creation, driver pod start and finish, but also **each executor pod state change is an event**. This is causing much longer waiting time as there are much more

vincentye38:
@bnetzi Thank you for explaining. It's interesting that the cpu utilization is so low. 

> each of the go routines is waiting in line to preform a q.Get() but the next item in queue be processed by another go routine only when q.Done() is called which is after we are doing the sync.

My understanding from the source code of q.Get() and q.Done() is that go routines can pickup other items from the queue. the queue deduplicates item which has identical one in the queue or in process when the item is inserted. The new identical item will be set aside into `dirty` set util the one in process is marked by q.done(item), then the new item will be moved from `dirty` to queue. WorkQueue is a dedup-queue. so q.done() doesn't block other go routine to q.Get() other item.

I guess the problem that cause low cpu utilization, low throughput and blocking items being dequeued is RateLimitingInterface. Could you mind to test out using WorkQueue directory without wrapping it with RateLimitingInterface? Thank you.

bnetzi:
> @bnetzi Thank you for explaining. It's interesting that the cpu utilization is so low.
> 
> > each of the go routines is waiting in line to preform a q.Get() but the next item in queue be processed by another go routine only when q.Done() is called which is after we are doing the sync.
> 
> My understanding from the source code of q.Get() and q.Done() is that go routines can pickup other items from the queue. the queue deduplicates item which has identical one in the queue or in process when the item is inserted. The new identical item will be set aside into `dirty` set util the one in process is marked by q.done(item), then the new item will be moved from `dirty` to queue. WorkQueue is a dedup-queue. so q.done() doesn't block other go routine to q.Get() other item.
> 
> I guess the problem that cause low cpu utilization, low throughput and blocking items being dequeued is RateLimitingInterface. Could you mind to test out using WorkQueue directory without wrapping it with RateLimitingInterface? Thank you.

Although from the source code it looks like it - I don't think the way you describe it is the way it actually works. We are missing something that would explain exactly why it doesn't work in parallel. 
The things I know for sure are - in reality it did not ran the important code segments in parallel, profiling with pprof showed most of time the code is stuck on mutex, and increasing the rate limit even to absurd numbers (100,000) did not change this behavior.

In any case - in my opinion the solution we came up with is a better design overall and it actually works for us in large scale production and after extensive tests. I don't have the capacity nor the resources to preform more tests at large scale just in order to understand the bug here, especially as we have a working version . However, If you can try to create a version without rate limit and test it yourself I think it would be beneficial

bnetzi:
Adding a google doc I've made for illustrating the issue and our solution 

https://docs.google.com/document/u/0/d/14x76jK7naRjxeFCkqUe7MgoDzZjYETixXXa828o0kSM/mobilebasic

luohao:
I wanted to share that we've been running a build where we replaced the RateLimitingInterface with the default workqueue. We've seen a significant reduction in reconciliation delay, even without increasing the client QPS as suggested in this PR. Occasionally, during peak hours, we still experience up to a 5-minute delay before the process starts, but this might not be entirely due to the operator reconciliation delay.

We haven't noticed any negative impacts from removing the rate limiter. This makes me wonder why the spark-operator uses it. What specific scenario is it designed to protect against?

robscc:
> I wanted to share that we've been running a build where we replaced the RateLimitingInterface with the default workqueue. We've seen a significant reduction in reconciliation delay, even without increasing the client QPS as suggested in this PR. Occasionally, during peak hours, we still experience up to a 5-minute delay before the process starts, but this might not be entirely due to the operator reconciliation delay.
> 
> We haven't noticed any negative impacts from removing the rate limiter. This makes me wonder why the spark-operator uses it. What specific scenario is it designed to protect against?

ratelimit is necessary to control the pressure of apiserver and limit the repeated updates of the spark application

pkg/controller/sparkapplication/controller_test.go
```diff
@@ -94,15 +95,41 @@ func TestOnAdd(t *testing.T) {
...
-	expectedKey, _ := cache.MetaNamespaceKeyFunc(app)
-	assert.Equal(t, expectedKey, key)
-	ctrl.queue.Forget(item)
+	ctrl.onAdd(app)
+	after := time.Now()
+	m := ctrl.GetRelevantMap(app.Name)
+	// Check that m is not nil.
+	assert.NotNil(t, m)
+	// Check that m.appQueueMap[appName].Queue is not nil.
+	assert.NotNil(t, m.appQueueMap[app.Name])
```
tiechengsu, line 107: should be 
key, _ := keyFunc(app)
m.appQueueMap[key]

pkg/controller/sparkapplication/controller.go
```diff
@@ -159,27 +248,84 @@ func newSparkApplicationController(
...
+	close(stopCh)
+}
+
 // Start starts the Controller by registering a watcher for SparkApplication objects.
-func (c *Controller) Start(workers int, stopCh <-chan struct{}) error {
+func (c *Controller) Start(maxAge time.Duration, checkInterval time.Duration, stopCh <-chan struct{}) error {
 	// Wait for all involved caches to be synced, before processing items from the queue is started.
 	if !cache.WaitForCacheSync(stopCh, c.cacheSynced) {
 		return fmt.Errorf("timed out waiting for cache to sync")
 	}
```
tiechengsu, line 312: With this change, we are breaking the leader election feature, as previously the worker thread creation is blocked by startCh so that only the leader operator has active worker thread. But we right now we are creating worker thread as new sparkapp comes in
https://www.kubeflow.org/docs/components/spark-operator/user-guide/leader-election/

xyzyng:
We are doing load test on spark operator and we also run into this issue. The test is run on an EKS cluster with 50 m5.4xlarge EC2 instance. We keep increasing the input traffic load by creating many spark application from client, we noticed spark job submitted from operator can keep up with the speed of the input rate at the begining, but once the input rate goes between 250 or 300 jobs per minute the spark-submit rate become stalled and cannot further increase. The start job latency once increased to more than 800 seconds. We are running operator with 30 worker and tried increase further but it does not really help. Looking at the metrics, we noticed the same behavior that CPU utilization of the operator is always below 10 vcores, but at the same time, the workqueue depth keep increasing to even more than 1000, where we have to stop the test. 

We suspect we are mostly hit by not able to increase qps and burst as we see lots of such errors in operator log 

```
I0926 11:50:49.403544      10 request.go:682] Waited for 39.99761059s due to client-side throttling, not priority and fairness, request: PUT:https://172.20.0.1:443/apis/sparkoperator.k8s.io/v1beta2/namespaces/spark-job4/sparkapplications/spark-pi-e6a3693e-6512-4e30-ad7e-69b4f33bdb66/status
I0926 11:50:59.403555      10 request.go:682] Waited for 39.997706246s due to client-side throttling, not priority and fairness, request: PUT:https://172.20.0.1:443/apis/sparkoperator.k8s.io/v1beta2/namespaces/spark-job3/sparkapplications/spark-pi-3fa65941-2ca1-44d4-8898-d1071ee0a2ee/status
```

But we may also be impacted by the single workqueue mutex given the high traffic. 

It will be really appreciated if this issue can be fixed in further spark operator versions. Thanks.

bnetzi:
For all the followers if this MR, per the massive change in spark operator design, I've waited with this MR to see if it is still relevant after the changes done.

I haven't had time to test V2 at scale yet, nor to accommodate my changes to V2. 
I would be happy to know if anyone already tested V2 at scale and have some conclusions

ImpSy:
My company actually went with a different implementation (customizing the workqueue ratelimiter) to handle high load clusters
The change from manually creating the controller (historical implemtation) to using controller-runtime (V2 changes) make it still relevant

I've actually open a PR to backport it here -> https://github.com/kubeflow/spark-operator/pull/2186 

To add context we implemented this to fix a customer issue because their airflow would submit 25K SparkApplication CR at the same time
Obviously this is not a silver bullet, you also need to increase cpu/memory and the number of controller threads

xyzyng:
> For all the followers if this MR, per the massive change in spark operator design, I've waited with this MR to see if it is still relevant after the changes done.
> 
> I haven't had time to test V2 at scale yet, nor to accommodate my changes to V2. I would be happy to know if anyone already tested V2 at scale and have some conclusions

I actually tested V2 operator on the cluster with same size (50 ec2 m5.4xlarge) today see similar behavior as V1. The job start latency is 2-4 seconds when the input rate is around 50-100 jobs per minutes. After increasing it to 150, start latency goes to 12s, which is stills table. Further increased to 200 job per minutes causes the start latency increased continuously, at one time it reached 150s. It is also interesting that after the operator run with 200 job per min for some time (maybe after 30 minutes), the start latency and running job number all dropped to 0 (and at same time SUBMITTED/COMPLETED/SUCCEEDING job number all keep increasing), if I randomly sample job from operator log, the latency still increases to more than 4 minutes.

bnetzi:
@xyzyng from what @ImpSy wrote it sounds like the main issue with V2 is just some tweaks with the controller runtime configurations (which make sense, since it used for many high volume operators)

@vara-bonthu - referring you to the discussion here, I think we just need to provide in general more options to configure the controller runtime, and that my PR is irrelevant

gangahiremath:
@bnetzi , @vara-bonthu , regarding the point '`referring you to the discussion here, I think we just need to provide in general more options to configure the controller runtime, and that my PR is irrelevant`', 

- Does it mean ' one queue per app and one goroutine per app' is not a solution for the performance issue faced? Is https://github.com/kubeflow/spark-operator/pull/2186 solution for the same
- Do we see performance opportunity improvement with the approach that we have tried? - https://github.com/kubeflow/spark-operator/issues/1574#issuecomment-1699668815

**Summary of changes** :

         - port spark-submit to golang
         - this removes JVM invocation, hence, performance-wise faster
         - no dependency on apache spark(as the frequency and quantity of changes to driver pod going to be minimal in future releases of apache spark)

vara-bonthu:
> @bnetzi , @vara-bonthu , regarding the point '`referring you to the discussion here, I think we just need to provide in general more options to configure the controller runtime, and that my PR is irrelevant`',
> 
> * Does it mean ' one queue per app and one goroutine per app' is not a solution for the performance issue faced? Is [FEATURE: add cli argument to modify controller workqueue ratelimiter #2186](https://github.com/kubeflow/spark-operator/pull/2186) solution for the same
> * Do we see performance opportunity improvement with the approach that we have tried? - [Spark Driver Pod Getting Stuck in init state due to no driver configmap found #1574 (comment)](https://github.com/kubeflow/spark-operator/issues/1574#issuecomment-1699668815)
> 
> **Summary of changes** :
> 
> ```
>      - port spark-submit to golang
>      - this removes JVM invocation, hence, performance-wise faster
>      - no dependency on apache spark(as the frequency and quantity of changes to driver pod going to be minimal in future releases of apache spark)
> ```

Thanks @gangahiremath ! Please add your feedback to this issue https://github.com/kubeflow/spark-operator/issues/2193

c-h-afzal:
Hey @bnetzi - 
 - we at Salesforce are running performance tests for Spark Operator using your PR branch.. From the recent conversation in this thread, it seems like V2 operator (without this PR) still suffers from performance issues as mentioned by @xyzyng . Can you please let us know the following:

1. Do you intend to pursue this MR and do you still see the one mutex per queue change as the major reason for low performance in V2?

2. Can you update your PR branch with the latest (or V2) changes and we can try to run performance tests on that branch. That’ll help us test your change+V2 for performance.

3. Do you or any of the other folks on this thread, suspect JVM spin-up times as a possible reason for spark job submission latency?   

Thank you.

bnetzi:
> @bnetzi , @vara-bonthu , regarding the point '`referring you to the discussion here, I think we just need to provide in general more options to configure the controller runtime, and that my PR is irrelevant`',
> 
> * Does it mean ' one queue per app and one goroutine per app' is not a solution for the performance issue faced? Is [FEATURE: add cli argument to modify controller workqueue ratelimiter #2186](https://github.com/kubeflow/spark-operator/pull/2186) solution for the same
> * Do we see performance opportunity improvement with the approach that we have tried? - [Spark Driver Pod Getting Stuck in init state due to no driver configmap found #1574 (comment)](https://github.com/kubeflow/spark-operator/issues/1574#issuecomment-1699668815)
> 
> **Summary of changes** :
> 
> ```
>      - port spark-submit to golang
>      - this removes JVM invocation, hence, performance-wise faster
>      - no dependency on apache spark(as the frequency and quantity of changes to driver pod going to be minimal in future releases of apache spark)
> ```

 I'm not sure that MR 2186 is the solution here, as there are also qps and burst configurations for other k8s API users (such as CRD Listers etc.)
Unfortunately I don't have the time nor resources to test V2 at scale and detect the performance issues there. I would probably be able to get to it only on mid November.
Would be great if someone else can test it.

As for the JVM overhead. Although it has some performance impact, moving to golang submit would require an operator version for every change in spark submit and would probably tend to break / have missing functions more often on new spark versions.
I believe we can get to a good enough solution without adding it.

bnetzi:
> Hey @bnetzi -
> 
> * we at Salesforce are running performance tests for Spark Operator using your PR branch.. From the recent conversation in this thread, it seems like V2 operator (without this PR) still suffers from performance issues as mentioned by @xyzyng . Can you please let us know the following:
> 
> 1. Do you intend to pursue this MR and do you still see the one mutex per queue change as the major reason for low performance in V2?
> 2. Can you update your PR branch with the latest (or V2) changes and we can try to run performance tests on that branch. That’ll help us test your change+V2 for performance.
> 3. Do you or any of the other folks on this thread, suspect JVM spin-up times as a possible reason for spark job submission latency?
> 
> Thank you.

I can't promise when, but if someone else wouldn't in the meantime - I will make sure V2 won't have scale performance issues. It might take a while but our platform is heavily based on spark operator and we do no intend to be stuck with V1.

The thing is in V2 the whole code was changed and many infra issues were fixed by moving to be based on controller runtime.
So the way I see it -  work queue per app might not longer be the solution. As I suggested in other comments in this MR, tweaking the k8s config (qps + burst) for the controller + the work queue might be a simpler more effective way here. But only testing would find out.

bnetzi:
> Another solution could be to move the submission phase to a dedicated pod That way app reporting will not have side effect on the most important part of the operator (launching spark applications)

I think in the first place the reason we have a queue is to make sure all actions on the pod are preformed by order. If you would separate it you might change this behavior. 
For example - what would happen if while the creation pod is trying to create the pod, there is an API request to delete the app and app is deleted before pod creation? Probably what would happen as the app is the owner ref, the pod would be created but the k8s garbage collector will terminate it eventually, but when? It might cause very weird bugs.

But in general I do think the spark operator pod is doing way too much work, especially executors related stuff and should be separated to different containers at least

ImpSy:
> I think in the first place the reason we have a queue is to make sure all actions on the pod are preformed by order. If you would separate it you might change this behavior.
I only made this proposition because that's the implementation that the `apache/spark-kubernetes-operator` decided to make
But I feel like 2 components running the reconciliation on the resources (Application CR in that case) would not break the order processing since the `Submitted` State would only be set after a successful submission by the `submission worker` and the controller would pick the lifecycle of the app afterwards

c-h-afzal:
@bnetzi  - Just wanted to share some feedback and our findings from testing on this PR. We (Salesforce) applied your PR and tested your changes at scale within our clusters. We compared your version, which  is essentially a "queueless" version (since there's only one job in each queue) against the version @gangahiremath had authored, which skips spark-submit thus avoiding all of the JVM latency. From our testing we found @gangahiremath version outperform significantly than the queueless version as well as the current master branch. 

Also, we found that this PR runs into OOM issues as soon as the cluster is stressed. Specifically, we tested for one job submission per second for 15 minutes, i.e. 900 jobs in total and the SO with this PR experienced OOM. I would guess that arises from the fact that all the controller threads are just spark-submitting the jobs as soon as they are enqueued in their respective queues and cause OOM.

Other than that, your approach is correct in the sense that having a single queue does affect SO performance whenever the rate of enqueueing becomes greater than the rate of dequeueing within the queue causing the queue length to grow. The performance is further affected because of the queue being guarded by a mutex. One approach could also be to use a concurrent lock-free single queue, which wouldn't require having a mutex. This PR does outperform the vanilla version of spark operator under light/normal load.

Hope this is helpful. Thanks.

bnetzi:
@c-h-afzal - thanks for the info!
This is really interesting, as we (Mobileye) are running this in production for months now in high load without any OOM issues, but it does require us to use 31 CPUs pod.
Also - we did notice that our k8s itself has issues of handling too many pods in a single namespace, so we stop submitting at ~3500 pods (drivers+ executors)

We also tried @gangahiremath version and under heavy load it didn't performed well for us.
So my Guess is there is a difference in the way we are testing that creates the difference is results.

The difference might be:
1. the qps config we used
2. the controller threads (which in our code defines the amount of queues maps) which is also very high in our case
3. The size of k8s cluster we are using
4. The amount of overall spark apps running concurrently 


I think it can be
P.s. I would not say though that this PR is queue less, as each app operations are in a queue, and it prevents concurrent actions on the same app, which is the point of the current queue.

Anyway - we started to work on a version that is compatible with spark operator V2, so more updates on that soon

c-h-afzal:
Gotcha! Thanks for response @bnetzi 

If it helps, we ran perf tests with Spark Operator running on a 4 vpcu and 16 GB memory node. We made all other variables e.g. controller threads, qps, the k8s cluster-size, number of spark jobs - constant across all the tests, so they shouldn’t have an impact on the results. Also, increasing the controller threads can only be as useful as the number of vCPUs/CPUs on the node. A number higher than available cores may not necessarily help.   

Given Ganga’s fix is working so well for us, we have gone ahead and [proposed](https://github.com/kubeflow/spark-operator/issues/2337) to make the spark app launch mechanism pluggable in spark operator. The default will always be spark-submit. The change would be a no-op for folks continuing to use spark-submit. The plugin mechanism would allow us to insert Ganga’s fix and circumvent spark-submit JVM penalty. If we find enough interest from other folks, we’ll also open-source Ganga’s fix in a separate repo.

And acknowledge your comment about queuless - the intent was to describe queless from the perspective of multiple spark apps but ops for a single app still need to queue.

Hoping to see your changes and future improvements in spark operator soon :)

bnetzi:
@c-h-afzal I think the specs are the main diff.
We ran our tests with 32 cores and ~100GB memory and got to 100% CPU usage at peaks (although memory was only 50% at peaks).
As for controller threads - in my code the controller threads are controlling the amount queue maps, so it is not coupled with amount of vCPUs, it only has more impact on memory. We have used 10,000 in our tests just to avoid map mutexes as much as possible.

I think it is great to have the submit with go as an opt-in option. I'll comment in the issue you opened.

bnetzi:
I'm closing this MR as it seems like the new operator release is working just fine with the right configuration.

We did the same stress test we did on the old operator.
our config:

spark operator controller pod:
 31 vCPUs

Worker queue config
  bucketQPS: '1000'
  bucketSize: '2000'

Controller config
  workers: 100
  maxTrackedExecutorPerApp: '1'

And we saw that all ran smooth without delays at all. **CPU was used 100% at max** (when submitted few hundred apps at once)

PR #1993, opened 2024-04-18
Add emptyDir sizeLimit support for local dirs

jacobsalway:
## Purpose of this PR

* Support the `sizeLimit` field for `emptyDir` volume types that are prefixed with `spark-local-dir-...`

## Change Category

- [ ] Documentation update

### Rationale

* If creating an `emptyDir` volume without a `spark-local-dir-...` prefix, the webhook will patch the appropriate volume onto the pod spec, but Spark will still create an `emptyDir` volume anyway https://github.com/apache/spark/blob/fd3f70c74ba63a0209a4a48a387048ef40380e0f/resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/features/LocalDirsFeatureStep.scala#L38-L40
* However if using a `spark-local-dir-...`, which the operator doesn't consider in the webhook and instead supplies at submission time, the sizeLimit is not carried through to the resulting volume in K8s
* It may be desirable to set a size limit if using an `emptyDir` volume e.g. if on an NVME node to prevent a node disk pressure condition

## Checklist

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.2.11
+version: 1.2.12
 appVersion: v1beta2-1.4.3-3.5.0
 keywords:
   - spark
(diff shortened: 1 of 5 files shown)


vara-bonthu:
@jacobsalway, thanks for the PR! It would be great if you could also update the documentation to reflect this new feature. Additionally, please address the failing pre-commit checks.

jacobsalway:
@vara-bonthu have bumped the app version and added a sizeLimit field to the volume example on the user guide. As an operator user I usually consult the that page and expected this to work by default given that it's supported in the SparkApplication CRD, so I think the example is good enough but happy to add additional docs if you prefer

vara-bonthu (approved):
/lgtm

PR #2003, opened 2024-04-24
Fix Docker Hub Credentials in CI

andreyvelich:
This should fix the `Login to Packages Container registry` GitHub action.

/assign @vara-bonthu @yuchaoran2011


vara-bonthu (approved):
/lgtm

Issue #1982, opened 2024-04-16
[FEATURE] Publish docker artifacts for multiple Spark versions with each release


peter-mcclonski:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?

Currently, each release of Spark Operator is associated with exactly one version of Spark-- Generally the latest minor/patch version.  Our organization is using Spark 3.4.2-- released Nov. '23.  We would like to use the most recent fixes and features present in Spark Operator without being forced to use a single, specific version of Spark.  

#### Describe the solution you would like

In particular, we would like to see releases of Spark Operator include support for multiple versions of Spark.  The length of support is likely a point for debate, but we would propose supporting the most recent patch version of the three most recent minor versions.

#### Describe alternatives you have considered

Alternatives include either forcing a project to align their spark version with the spark-operator spark version, or forcing the project to build and maintain their own version of the spark-operator docker image.

peter-mcclonski:
I would be happy to work on this, if the feature is desired.

PR #2024, opened 2024-05-10
(#1982) Publish Spark Operator image for multiple Spark versions on r…

peter-mcclonski:
…elease

### 🛑 Important:
Please open an issue to discuss significant work before you start. We appreciate your contributions and don't want your efforts to go to waste!

For guidelines on how to contribute, please review the [CONTRIBUTING.md](CONTRIBUTING.md) document.

## Purpose of this PR
Support multiple versions of Spark with our release process.  

**Proposed changes:**
- Builds and pushes Spark Operator built on Sparks 3.5.1, 3.4.3, and 3.3.4 as part of the release process.

## Change Category

- [ ] Documentation update

### Rationale

See #1982 

## Checklist

- [X] I have updated documentation accordingly.

### Additional Notes

Not sure of the best way to validate this action works prior to merging.  Would be eager for feedback on testing it.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,6 +56,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        spark_version:
+          - 3.5.1
+          - 3.4.3
+          - 3.3.4
         platform:
           - linux/amd64
           - linux/arm64
@@ -91,6 +95,8 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
+          build-args: |
+            SPARK_IMAGE=spark:${{ matrix.spark_version }}
           platforms: ${{ matrix.platform }}
           cache-to: type=gha,mode=max,scope=${{ env.SCOPE }}
           cache-from: type=gha,scope=${{ env.SCOPE }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -56,6 +56,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        spark_version:
+          - 3.5.1
```
peter-mcclonski, line 60: I don't like repeating this between the two jobs, but it was the _least complex_ solution.

charts/spark-operator-chart/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: spark-operator
 description: A Helm chart for Spark on Kubernetes operator
-version: 1.2.15
-appVersion: v1beta2-1.4.6-3.5.0
+version: 1.2.16
+appVersion: v1beta2-1.4.6
```
peter-mcclonski, line 5: I: The Spark version is separate from the Operator version.  New image tag format is consistent with the old one, but with parameterized Spark version rather than hardcoding it here.

vara-bonthu (approved):

vara-bonthu:
@peter-mcclonski Please resolve conflicts. Will review and merge the PR

yuchaoran2011 (commented):
The PR itself looks good. But I wonder if there really is a use case for building a separate image for each version. @peter-mcclonski Have you run into cases where, for example, operator built with Spark 3.4.3 and 3.5.1 result in difference behaviors?

peter-mcclonski:
> The PR itself looks good. But I wonder if there really is a use case for building a separate image for each version. @peter-mcclonski Have you run into cases where, for example, operator built with Spark 3.4.3 and 3.5.1 result in difference behaviors?

I actually have never tried it-- internally, we've always made a habit of aligning our Operator version to whatever version of spark we're using. If it doesn't actually need to be aligned (which I recall you mentioning in the first community call), then you're right, this may not be a value-add.

vara-bonthu:
Let's keep this open and see if users show interest in this feature. If we receive more thumbs up, we can maintain it; otherwise, we may consider de-scoping it.

josecsotomorales:
This is great! @peter-mcclonski I think we should update the version matrix here: https://github.com/kubeflow/spark-operator?tab=readme-ov-file#version-matrix

Issue #2098, opened 2024-07-25
[FEATURE] Support Yunikorn as a batch scheduler


jacobsalway:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?

[Apache YuniKorn](https://yunikorn.apache.org/) is an alternative Kubernetes scheduler that provides batch scheduling functionality, including gang scheduling, hierarchical queues and job ordering. The  Spark operator should support Yunikorn as a batch scheduler by supplying the necessary task group information on the driver pod annotations.

#### Describe the solution you would like

1. Scaffold a Yunikorn batch scheduler implementation under https://github.com/kubeflow/spark-operator/tree/master/pkg/batchscheduler
2. ...

sir3mat:
is it possible to use yunikorn with kubeflow trainer V2?

PR #2107, opened 2024-08-01
Support gang scheduling with Yunikorn

jacobsalway:
# Purpose

Add a scheduler implementation for [YuniKorn](https://yunikorn.apache.org) to support task group annotations on the driver pod for gang scheduling and queue labels on both driver and executor pods.

Completes the first dot point of https://github.com/kubeflow/spark-operator/issues/2098. Will add docs for the Spark operator in a separate PR, but will add docs to YuniKorn as well.

# Changes

* Add a new batch scheduler with the name `yunikorn`
* Calculate the resources required to schedule both driver and executor pods i.e. requests and executor counts, and use this to create task groups annotations on the driver pod to support [gang scheduling](https://yunikorn.apache.org/docs/next/user_guide/gang_scheduling/)
* Use the existing `batchSchedulerOptions.queue` to allow queue configuration. This is not a required field as Yunikorn supports [placement rules](https://yunikorn.apache.org/docs/0.8.0/configuration/placement_rules/) that allow an app to be placed without explicitly specifying a queue

# Testing

```bash
# clone my fork and checkout the branch
git clone git@github.com:jacobsalway/spark-operator.git
cd spark-operator && git checkout add-yunikorn-scheduler

# build the docker image
docker build . -t docker.io/kubeflow/spark-operator:add-yunikorn-scheduler

# create a kind cluster
kind create cluster --wait=1m

# load the operator image
kind load docker-image docker.io/kubeflow/spark-operator:add-yunikorn-scheduler

# install the operator with the forked build and the batch scheduler enabled
helm install spark-operator ./charts/spark-operator-chart \
  --create-namespace \
  --namespace spark-operator \
  --set image.tag=add-yunikorn-scheduler \
  --set controller.batchScheduler.enable=true

# install yunikorn
helm install yunikorn yunikorn \
  --repo https://apache.github.io/yunikorn-release \
  --create-namespace \
  --namespace yunikorn

# create the example application
kubectl apply -f examples/spark-pi-yunikorn.yaml

# observe that yunikorn will gang schedule the application
# task group pods prefixed with `tg-spark-...-spark-{driver/executor}-...` with requests
# that match the real pods will be created
```

https://github.com/user-attachments/assets/2e630b34-ef3c-457d-891f-5ad2bb438e12

## Change Category

- ~[ ] Documentation update~ will do in a separate PR to avoid this one getting any larger

## Checklist

- ~[ ] I have updated documentation accordingly.~ will do in a separate PR to avoid this one getting any larger


jacobsalway:
Still working on this, just sorting out some issues with converting the java byte string to Kubernetes resource values. Have re-read through the upstream code and have a better approach in mind

jacobsalway:
Stacked PR also available to allow the default batch scheduler to be set if not specified by the user https://github.com/jacobsalway/spark-operator/pull/1

internal/scheduler/yunikorn/scheduler.go
```diff
@@ -0,0 +1,148 @@
+package yunikorn
```
ChenYi015, line 1: We can add license comment in every new file, like https://github.com/kubeflow/spark-operator/blob/5972482ca4dbe059914a9b5b70edb98a82dfc0a6/internal/controller/sparkapplication/controller.go#L1-L15.

internal/scheduler/yunikorn/resourceusage/executor_count.go
```diff
@@ -0,0 +1,22 @@
...
+		if app.Spec.DynamicAllocation.MinExecutors != nil {
+			initialExecutors = max(initialExecutors, *app.Spec.DynamicAllocation.MinExecutors)
+		}
+		if app.Spec.DynamicAllocation.InitialExecutors != nil {
+			initialExecutors = max(initialExecutors, *app.Spec.DynamicAllocation.InitialExecutors)
+		}
+	}
+
+	return initialExecutors
+}
```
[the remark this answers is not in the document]
ChenYi015: Maybe we can move this func`NumInitialExecutors` to `pkg/util/sparkapplication.go` as this is an util function for SparkApplication.

internal/scheduler/yunikorn/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+const (
+	SchedulerName = "yunikorn"
+
+	// The names are set to match the Yunikorn gang scheduling example for Spark, but these can be any
+	// value as long as what's on the pod matches the task group definition
+	// https://yunikorn.apache.org/docs/next/user_guide/gang_scheduling/#enable-gang-scheduling-for-spark-jobs
+	driverTaskGroupName   = "spark-driver"
+	executorTaskGroupName = "spark-executor"
```
ChenYi015, line 21: I see that in Volcano scheduler, the PodGroup names are different between SparkApplications. So I am wondering whether we need to set distinct task group names for different SparkApplications?

ChenYi015 (commented):
Thanks for your contributions, I have left some comments.

internal/scheduler/yunikorn/scheduler.go
jacobsalway: For Yunikorn, a task group is a property of an application, so the task group names only need to be unique within the application they're a part of. 

Application IDs must be unique in the same way that a Volcano PodGroup name has to be, however there's a nice integration in Yunikorn to use the `spark-app-selector` label for the application ID.

internal/scheduler/yunikorn/scheduler.go
jacobsalway: I'll add the license comment to the top of all my new files, but what do you think if in a separate PR we add a license comment check as part of CI, and maybe a Makefile target for checking and adding locally?

internal/scheduler/yunikorn/resourceusage/executor_count.go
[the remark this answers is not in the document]
jacobsalway: Ah nice, good catch. I think I'd only been looking at https://github.com/apache/spark/blob/899fad4710bef174684deee64314ac483c16c494/core/src/main/scala/org/apache/spark/util/Utils.scala#L2534-L2557 when I wrote this implementation.

Will fix and move to utils tomorrow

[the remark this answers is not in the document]
jacobsalway: Fixed to match the core implementation and moved to `pkg/utils/sparkapplication.go`

internal/scheduler/yunikorn/scheduler.go
ChenYi015: I think it is a good point to add license check as part of CI. cc @andreyvelich @vara-bonthu @yuchaoran2011 .

ChenYi015 (approved):

internal/scheduler/yunikorn/scheduler.go
jacobsalway: Created an issue for it https://github.com/kubeflow/spark-operator/issues/2139

yuchaoran2011 (approved):
/lgtm
Great to see that all the edge cases including min overhead memory are covered.

PR #2121, opened 2024-08-13
Update release workflow and docs

ChenYi015:
## Purpose of this PR

Update release workflow, the release steps are as follows:

1. Update `VERSION` file
2. Update `version` and `appVersion` in `chart.yaml` and `make helm-docs`
3. Commit the changes and create a PR to the release branch
4. After the PR is raised, a workflow in `check-release.yaml` will be triggered to do some check e.g. whether the version matches semver pattern, whether the chart version/appVersion matches this version and whether the version tag already exists
5. Once the PR is merged, a workflow in `release.yaml` will be triggered to build and publish docker images to Docker hub and create a draft release
6. Once the draft release is published, a workflow in `release-helm-charts.yaml` will be triggered to update Helm repo index file

## Change Category

- [x] Documentation update

### Rationale

- Since a tag pushed by action workflow cannot trigger other workflows, we need to push the tag manually to trigger the release workflows.
- I have attempted to merge all the release-related workflows into one workflow file, but that requires more modifications to the workflow file, particularly because the current workflow for releasing Docker images functions well when it retrieves context from a tag push event.

## Checklist


.github/workflows/push-tag.yaml
```diff
@@ -1,44 +0,0 @@
...
-            echo "TAG_EXISTS=true" >> $GITHUB_ENV
-          else
-            echo "TAG_EXISTS=false" >> $GITHUB_ENV
-          fi
-
-      - name: Create and push tag
-        if: env.TAG_EXISTS == 'false'
-        run: |
-          git tag -a "$VERSION" -m "Release $VERSION"
-          git push origin "$VERSION"
```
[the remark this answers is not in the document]
andreyvelich: @vara-bonthu @yuchaoran2011 What do you think ?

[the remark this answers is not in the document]
ChenYi015: @andreyvelich Ok, I will do it as we discussed.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,64 @@
+name: Pre Release
```
andreyvelich, line 1: Why do you need to separate workflows between pre-release, release, and post-release ?

ChenYi015: @andreyvelich It is okay to place all the release workflows in one file, but I think doing so may make the release process more robust:

1. The pre-release workflow will do some checks when the release PR is raised e.g. whether the version matches semver pattern, whether the chart version/appVersion matches this version and whether the version tag already exists. If the pre-release check failes, then we should not merge the PR.
2. The release workflow will build and publish docker images to Docker hub and create a draft release.
3. Once the draft release is published, the post-release workflow will be triggered to update Helm repo index file to publish the Helm charts. Since the helm chart is packaged as an artifact of the release, so I think we should update the Helm repo index only after publishing the release.

What do you think? cc @vara-bonthu @yuchaoran2011

andreyvelich: Pre-release make sense to me! Maybe, we should rename this workflow to `check-release` to avoid confusion that this workflow for release candidates. WDYT @ChenYi015 ?

For the release and post-release, why we don't want to combine them into single workflow ? That should help us to avoid 2 commits:
- Spark Operator Official Release v2.0.0
- Add index for Spark operator chart v2.0.0.

Since it happens here: https://github.com/kubeflow/spark-operator/blob/75b4c6998516ab2dc14433c92e112f7991b1055f/.github/workflows/post-release.yaml#L64-L66.

I just think in the release workflow once step of image push is complete, we can update the helm charts with the new version.

The last steps of release workflow can be push to the GitHub.

ChenYi015: > Pre-release make sense to me! Maybe, we should rename this workflow to `check-release` to avoid confusion that this workflow is not to make release for RC.1/RC.0

Make sense, I will rename it.

> For the release and post-release, why we don't want to combine them into single workflow ? That should help us to avoid 2 commits:
> 
> * Spark Operator Official Release v2.0.0
> * Add index for Spark operator chart v2.0.0.

I think 2 commits cannot be avoided, for that the first commit is pushed to the release branch while the second one is pushed to the `gh-pages` branch, where [index.yaml](https://github.com/kubeflow/spark-operator/blob/gh-pages/index.yaml) file is hosted by GitHub pages.

> I just think in the release workflow once step of image push is complete, we can update the helm charts with the new version.

I think there might be a situation where if we publish the docker images and update the helm repo index file, but have not published the release yet, users will fail to `helm install` the latest version, for that the latest charts will not be available for download unless the release is published. WDYT @andreyvelich ?

andreyvelich: >I think 2 commits cannot be avoided, for that the first commit is pushed to the release branch while the second one is pushed to the gh-pages branch, where [index.yaml](https://github.com/kubeflow/spark-operator/blob/gh-pages/index.yaml) file is hosted by GitHub pages.

Oh, I missed it. What is the main motivation to publish the GitHub pages: https://kubeflow.github.io/spark-operator/ if we have Kubeflow docs: https://www.kubeflow.org/docs/components/spark-operator/overview/?

>I think there might be a situation where if we publish the docker images and update the helm repo index file, but have not published the release yet, 

Where we can get this case ? Do you mean that we publish the helm chart as part of GitHub release assets ?

ChenYi015: > Oh, I missed it. What is the main motivation to publish the GitHub pages: https://kubeflow.github.io/spark-operator/ if we have Kubeflow docs: https://www.kubeflow.org/docs/components/spark-operator/overview/?

A Helm chart repository actually is an HTTP server that hosts an `index.yaml` file, more information can be found [here](https://helm.sh/docs/topics/chart_repository/). And spark operator uses GitHub pages to serve helm chart repo, as this is a common choice.

When we add spark operator helm repo as follows:

```bash
helm repo add spark-operator https://kubeflow.github.io/spark-operator
```

The url actually is the GitHub pages url, when users want to `helm install` the spark operator chart, helm will retrieve the chart download url from [index.yaml](https://github.com/kubeflow/spark-operator/blob/gh-pages/index.yaml) file. For example, the index for spark operator chart `2.0.0-rc.0` is as follows: 
https://github.com/kubeflow/spark-operator/blob/0fcda776f6cc1b7cbb676705aede7c07ccb36828/index.yaml#L4-L23

> Do you mean that we publish the helm chart as part of GitHub release assets ?

Yes, the helm chart is published as release asserts. For example, the download url of Spark operator chart `v2.0.0-rc.0` is https://github.com/kubeflow/spark-operator/releases/download/v2.0.0-rc.0/spark-operator-2.0.0-rc.0.tgz.

andreyvelich: That makes sense, thanks for sharing!
Maybe we should rename this workflow to `helm-publish` to make it more declarative ?
Also, I can see that KServe using GitHub packages to publish the Helm release: https://github.com/kserve/kserve/blob/master/.github/workflows/helm-publish.yml.
That might help you to avoid separate branch: `gh-pages` for helm charts.

Do you prefer serving it via HTTP server or OCI storage ?

vara-bonthu (approved):
Thanks and the updated release workflow looks good to me. Approving this PR.
@andreyvelich add lgtm label once you are happy with your comments

docs/release.md
```diff
@@ -71,28 +71,42 @@ If you want to push changes to the `release-X.Y` release branch, you have to che
...
+4. Commit the changes:
 
     ```bash
     git add VERSION
     git add charts/spark-operator-chart/Chart.yaml
-    git commit -s -m "Release $VERSION"
-    git push
+    git add charts/spark-operator-chart/README.md
+    git commit -s -m "Spark Operator Official Release v${VERSION}"
+    git push origin release-X.Y
```
andreyvelich, line 98: Do we need this, since we make the commit to the release branch via `release` GitHub action ?

ChenYi015: I think this is needed, for that the `release` GitHub action is used to create a new draft release, not to make a commit to the release branch. So we need to update chart `version `/`appVersion` and README at the same time when we modify the `VERSION` file.

PR #2124, opened 2024-08-13
Increase frequency for Stale bot

andreyvelich:
I increased frequency for Stale bot (run every 2 hours) and number of operations per run.
This should help us to faster process 517 open issues and PRs.

/assign @vara-bonthu @yuchaoran2011 @ChenYi015

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -2,7 +2,7 @@ name: Mark stale issues and pull requests
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


ChenYi015 (approved):

ChenYi015:
@andreyvelich  There are many issues and PRs that need to be marked as stale, but actually, they are not due to the low frequency and low operations per run. Thanks for updating the stale workflow.
/lgtm

PR #2141, opened 2024-08-22
Support pod template for Spark 3.x applications

### LINKED ISSUE #1690: How can we pass the PSS/PSA required parameter as "seccompProfile" to sparkApplication driver and executor through template.
Hi, 
     using GKE(v1.25.6-gke) and EKS(v1.25.6-eks) cluster with PSS/PSA restricted. I tried to pass the PSS/PSA required parameters through the template to sparkApplication driver and executors but got the issue. below is the configuration which is I'm trying to use.
```
driver:
    template:
      spec:
        containers:
        - name: spark-driver-temp
          securityContext:
            readOnlyRootFilesystem: false
            capabilities:
              drop: [ ALL ]
            allowPrivilegeEscalation: false
            runAsNonRoot: true
	    seccompProfile:
              type: RuntimeDefault
           
```
getting the below issue while submitting sparkApplications to operator.
```
bash-4.4# kubectl apply -f pi.yaml
error: error validating "pi.yaml": error validating data: ValidationError(SparkApplication.spec.driver.template.spec.containers[0].securityContext): unknown field "seccompProfile" in io.k8s.sparkoperator.v1beta2.SparkApplication.spec.driver.template.spec.containers.securityContext; if you choose to ignore these errors, turn validation off with --validate=false
```
even tried without passing the **seccompProfile** parameter to securityContext so got the below error in the operator logs.
`pods \"spark-pi-driver\" is forbidden: violates PodSecurity \"restricted:latest\": seccompProfile (pod or container \"cspk-radish\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\").\n\tat io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:682)`

and I don't see any object defined for the **seccompProfile** in the [CRD](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/v1beta2-1.3.8-3.1.1/charts/spark-operator-chart/crds/sparkoperator.k8s.io_sparkapplications.yaml) to handled parameter. 
so wanted to know how can we pass the PSS/PSA parameters as **seccompProfile** through the template to sparkApplication driver and executors.

Thank you,
Mohan

### LINKED ISSUE #2101: [FEATURE] Use pod template feature in spark-submit for Spark 3.x applications
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?

As discussed in [#2049](https://github.com/kubeflow/spark-operator/pull/2049), it would be better to use Spark's native [pod template](https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template) feature in spark-submit. This will decrease the time to create spark pods, especially when deploying spark applications at a very large scale.

#### Describe the solution you would like

For Spark v3.x applications, use pod-template feature in spark-submit. 
For v2.x applications, still use webhook to patch spark pods.

ChenYi015:
## Purpose of this PR

Close #2101
Close #1690

**Proposed changes:**
- Add fields `spec.driver.template` and `spec.executor.template` to SparkApplication CRD
- Webhook will only mutate Spark pods with label `sparkoperator.k8s.io/mutated-by-spark-operator="true"`, and this label will be added by controller as needed.
- ~~For Spark 2.x applications, the webhook will be used to mutate Spark pods.~~
- ~~For Spark 3.x applications, the webhook will be used to mutate Spark driver/executor pods only if driver/executor pod template is not defined.~~
- For all spark applications, the webhook will still be used to mutate Spark pods no matter whether pod template is defined or not.
- Add example `spark-pi-pod-template.yaml`

## Change Category

- [x] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

## Checklist

### Additional Notes

- The resources section for cpu/memory requests and limits defined in the pod template will not work. Ref: https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template.


api/v1beta2/sparkapplication_types.go
```diff
@@ -407,6 +407,11 @@ type Dependencies struct {
 // SparkPodSpec defines common things that can be customized for a Spark driver or executor pod.
 // TODO: investigate if we should use v1.PodSpec and limit what can be set instead.
 type SparkPodSpec struct {
+	// Template is a pod template that can be used to define the driver or executor pod configurations that Spark configurations do not support.
+	// Spark version >= 3.0.0 is required.
+	// Ref: https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template.
+	// +optional
+	Template *corev1.PodTemplateSpec `json:"template,omitempty"`
```
jacobsalway, line 414: Will review the rest of the PR as well, but adding this as a struct rather than a binary/string field increases the number of properties in the resulting CRD by quite a bit. Is there any downside/considerations on the K8s side for very large CRDs?

ChenYi015: Good catch. I added a kubebuilder marker `+kubebuilder:validation:Schemaless` to mark the pod template field as schemaless to reduce the size of CRD.

internal/controller/sparkapplication/submission.go
```diff
@@ -302,6 +304,12 @@ func driverConfOption(app *v1beta2.SparkApplication) ([]string, error) {
 	property = fmt.Sprintf(common.SparkKubernetesDriverLabelTemplate, common.LabelLaunchedBySparkOperator)
 	args = append(args, "--conf", fmt.Sprintf("%s=%s", property, "true"))
 
+	// If Spark version is less than 3.0.0 or driver pod template is not defined, then the driver pod needs to be mutated by the webhook.
```
yuchaoran2011, line 307: As discussed during the community call, we can think about a future release (maybe 2.1) where we announce the deprecation of web hook and support for Spark 2.x

yuchaoran2011 (approved):

yuchaoran2011 (commented):
/lgtm

internal/controller/sparkapplication/submission.go
jacobsalway: Would we take the approach of moving to users only being able to specify a pod template, or map the existing struct fields and construct a pod template on the fly during submission?

vara-bonthu (approved):
@ChenYi015  Thanks for adding this much needed feature! PR looks good to me. You may need to resolve the conflicts to move forward.

ChenYi015:
@vara-bonthu Thanks for the review. I have resolved the merge conflicts.

internal/controller/sparkapplication/submission.go
ImpSy: I feel like mapping existing field to a pod template on the fly + new parameter `templateFile` go be the easiest migration path for existing user

Tom-Newton: For now I would personally find it useful to support webhook mutations and template on the same `SparkApplication`.

api/v1beta2/sparkapplication_types.go
Tom-Newton: I'm quite excited for this PR so I gave it a try. Maybe I did something wrong so feel free to ignore me but I got the following error:
```
Error from server (BadRequest): error when creating "<path>": SparkApplication in version "v1beta2" cannot be handled as a SparkApplication: strict decoding error: unknown field "spec.driver.template.metadata", unknown field "spec.driver.template.spec", unknown field "spec.executor.template.metadata", unknown field "spec.executor.template.spec"
```
when running my `kubectl` command. 

Removing `+kubebuilder:validation:Schemaless` and regenerating the CRD solved it for me.

josecsotomorales:
This is great stuff! Nicely done @ChenYi015 🚀

examples/spark-pi-pod-template.yaml
```diff
@@ -0,0 +1,169 @@
...
+  driver:
+    template:
+      metadata:
+        labels:
+          spark.apache.org/version: 3.5.3
+        annotations:
+          spark.apache.org/version: 3.5.3
+      spec:
+        containers:
+        - name: spark-kubernetes-driver
```
vara-bonthu, line 76: @ChenYi015 Shall we show nodeSelectors and Taints as well in the example? Mutating admission webhook is required for this but its not anymore once we merge this PodTemplates feature

vara-bonthu (approved):
/approve

ImpSy:
I know that the PR has been approved but I feel like introducing a new key is not the right implementation here

We should either build the template with the current CR if spark version >= 3 or have the possibility to specify a template file name to use

This implementation force a migration for our user if they want to get rid of webhook but it should be free for them (which is not the case here)

ChenYi015:
> I know that the PR has been approved but I feel like introducing a new key is not the right implementation here
> 
> We should either build the template with the current CR if spark version >= 3 or have the possibility to specify a template file name to use
> 
> This implementation force a migration for our user if they want to get rid of webhook but it should be free for them (which is not the case here)

@ImpSy 

1. It is somewhat difficult to build a template file form current CRD, it would be more easy and straight to use the template defined in `spec.[driver|executor].template`
2. If one want to specify a pod template file, he can use the spark conf `spark.kubernetes.[driver|executor].podTemplateFile`.
3. Maybe we should make webhook and pod template both work in this PR so that users can have time to migrate their applications to use pod templates.

Tom-Newton:
My user perspective is that I like introducing the new `template` fields now to enable new functionality but I'm not ready yet, to fully migrate away from the webhook, so I currently would like to use both. 

Before the webhook is deprecated though I think the template should be built from the current CR as ImpSy suggests.

examples/spark-pi-pod-template.yaml
ChenYi015: Done, I have added fields like nodeSelector, affinity and tolerations to pod template example.

ChenYi015:
I have updated this PR, the webhook will still be used to mutate Spark pods no matter pod template is defined or not. I will try to build the pod template file from the current CRD when we try to deprecate the webhook.

Tom-Newton (approved):
LGTM, though my review probably doesn't mean much :slightly_smiling_face:

vara-bonthu (approved):
/lgtm

PR #2146, opened 2024-08-27
feat: support driver and executor pod use different priority

Kevinz857:
### 🛑 Important:
Please open an issue to discuss significant work before you start. We appreciate your contributions and don't want your efforts to go to waste!

For guidelines on how to contribute, please review the [CONTRIBUTING.md](CONTRIBUTING.md) document.

## Purpose of this PR
This PR enhances the Spark Operator to support setting different priorities for driver and executor pods. This feature allows for more granular control over resource allocation and scheduling in Kubernetes clusters.

**Proposed changes:**
- Added support for specifying separate priorities for driver and executor pods in SparkApplication CRD
- Modified the controller logic to apply the specified priorities when creating driver and executor pods
- Updated relevant documentation to reflect the new feature

## Change Category

- [x] Documentation update

### Rationale

In complex Kubernetes environments, there's often a need to prioritize Spark driver pods differently from executor pods. This change allows users to fine-tune their Spark application deployments, potentially improving overall cluster resource utilization and application performance.

## Checklist

### Additional Notes

This feature introduces new fields in the SparkApplication CRD. Users will need to update their CRD definitions to take advantage of this new functionality. Backward compatibility is maintained for existing deployments.


Kevinz857:
@agemooij @mhausenblas @ChenYi015 Pls help make a review, thank you

internal/webhook/sparkpod_defaulter.go
```diff
@@ -490,9 +485,40 @@ func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) error
...
+
+	return nil
+}
+func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) error {
+	if *app.Spec.BatchSchedulerOptions.PriorityClassSplitting {
+		return addDriverExecutorSelfPriorityClassName(pod, app)
+	} else {
+		return addDriverExecutorUnifyPriorityClassName(pod, app)
+	}
+}
```
ChenYi015, line 520: From my understanding, the `priorityClassName` in `batchSchedulerOptions` is utilized by Volcano's PodGroup, and it is unrelated to the priorityClass of Spark pods. Given this, it seems unnecessary to have the `prioryClassSplitting` option or the `addDriverExecutorUnifyPriorityClassName`  function. WDYT?

Kevinz857: OK，prioryClassSplitting batchSchedulerOptions may be a bit inappropriate, let me clarify my usage scenario

The current business scenario has not yet used the ability of podGroup. The underlying K8s cluster is divided into multiple resource pools, and the SparkApplication task itself is defined with different priorities at the business level. Moreover, the resources of the underlying K8s cluster are limited. At this time, it is expected that high-priority Executors can run first when resources are insufficient, so the PriorityClass priority preemption scheduling mechanism of K8s itself is used

But because the default PriorityClass of SparkApplication is in the batchSchedulerOptions, it means that the driver and executor under the same SparkApplication will be set to the same priority; but in the actual business scenario, there may be problems, such as we know that the driver is the controller of the executor, once the drvier node is preempted and expelled, the entire SparkApplication will fail, and the expected effect of the business is that the Driver Pod cannot be preempted with high priority, and the Executor can be preempted with high priority, so that different priorities must be set for the Driver and Executor, such as setting the highest priority for the Driver, and the Executor can have a low priority, so that after the low-priority Executor is expelled, it will wait for high priority After the level Executor is executed and there are free resources, the Executors of these previously expelled low-priority tasks will be created by the corresponding drvier pod

If this is the case, how should this priority be defined?

ChenYi015: @Kevinz857 Perhaps we can do as follows: 
- If `spec.batchSchedulerOptions.priorityClassName` is defined, it will automatically be propagated to both driver and executor pods by default, as this is the current behavior so that we will not make a breaking change.
- Furthermore, if either `spec.driver.priorityClassName` or `spec.executor.priorityClassName` is specifically defined, it should take precedence over `spec.batchSchedulerOptions.priorityClassName`. In this way, we can eliminate the need to introduce `spec.batchSchedulerOptions.prioryClassSplitting` option as it is somewhat confusing. Also, users can define different priority classes for driver and executor pods.

Kevinz857: @ChenYi015 Very good suggestion, I can modify the code and resubmit MR according to this proposal.

Kevinz857: @ChenYi015 According to the latest proposal, if  if either spec.driver.priorityClassName or spec.executor.priorityClassName is specifically defined, it should take precedence over spec.batchSchedulerOptions.priorityClassName.

Modified the code according to this proposal, please help review it when you have time, thanks

internal/webhook/sparkpod_defaulter.go
```diff
@@ -475,21 +475,27 @@ func addSchedulerName(pod *corev1.Pod, app *v1beta2.SparkApplication) error {
...
 }
 
+func setPodPriorityClassName(pod *corev1.Pod, priorityClassName *string) {
+	if priorityClassName != nil && *priorityClassName != "" {
+		pod.Spec.PriorityClassName = *priorityClassName
+		pod.Spec.Priority = nil
+		pod.Spec.PreemptionPolicy = nil
+	}
+}
+
```
ChenYi015, line 485: Perhaps we could merge the logic of `setPodPriorityClassName` into `addPriorityClassName`.

Kevinz857: @ChenYi015 Already adapted merge the logic of setPodPriorityClassName into addPriorityClassName, thanks for ur suggestions

ChenYi015 (approved):

ChenYi015:
@Kevinz857  Thanks for the contribution. LGTM. Will wait for another approval.
The CI workflow needs approval from a maintainer to run @vara-bonthu @yuchaoran2011 
After the CI has run successfully, then this PR will be merged.
/hold

jacobsalway (commented):
I'll leave a better comment tomorrow on what the logic could look like, but just noting that unit tests aren't passing for me locally because of a nil pointer dereference on the `app.Spec.{Driver|Executor}.PriorityClassName`

Kevinz857:
> I'll leave a better comment tomorrow on what the logic could look like, but just noting that unit tests aren't passing for me locally because of a nil pointer dereference on the `app.Spec.{Driver|Executor}.PriorityClassName`

@jacobsalway  Thanks for the prompt, the judgment of whether the pointer is nil has been added. Looking forward to better revision suggestions tomorrow

ChenYi015:
> I'll leave a better comment tomorrow on what the logic could look like.

@jacobsalway, please retake a look when you have time.

ChenYi015:
@Kevinz857 The `build-api-docs` CI has failed, you need to `make build-api-docs` and commit the changes.

jacobsalway:
Reading into how the Volcano `PodGroup` priority class interacts with priority classes set on the pods within that group. Maybe it makes sense for us to remove this conflict altogether by removing the propagation of the batch scheduler options priority class to the pod priority class

ChenYi015:
> Reading into how the Volcano `PodGroup` priority class interacts with priority classes set on the pods within that group. Maybe it makes sense for us to remove this conflict altogether by removing the propagation of the batch scheduler options priority class to the pod priority class

@jacobsalway  Agree. The priority class of volcano pod group is unrelated to the ones of spark pods. But since this will cause a breaking change, we should denote it when we do the next release.

Kevinz857:
> > Reading into how the Volcano `PodGroup` priority class interacts with priority classes set on the pods within that group. Maybe it makes sense for us to remove this conflict altogether by removing the propagation of the batch scheduler options priority class to the pod priority class
> 
> @jacobsalway Agree. The priority class of volcano pod group is unrelated to the ones of spark pods. But since this will cause a breaking change, we should denote it when we do the next release.

@ChenYi015 Removed the definition and reference of spec.batchSchedulerOptions priorityClassName, and unified the definition and use of priority by driver pod and executor pod. Pls have a look when u are free. Thanks a lot

internal/scheduler/volcano/scheduler.go
```diff
@@ -207,10 +207,6 @@ func (s *Scheduler) syncPodGroup(app *v1beta2.SparkApplication, size int32, minR
 			if app.Spec.BatchSchedulerOptions.Queue != nil {
 				podGroup.Spec.Queue = *app.Spec.BatchSchedulerOptions.Queue
 			}
-			// Update pod group priorityClassName if it's specified in Spark Application
-			if app.Spec.BatchSchedulerOptions.PriorityClassName != nil {
-				podGroup.Spec.PriorityClassName = *app.Spec.BatchSchedulerOptions.PriorityClassName
-			}
```
ChenYi015: We need to add it back.

internal/webhook/sparkpod_defaulter.go
```diff
@@ -477,19 +477,19 @@ func addSchedulerName(pod *corev1.Pod, app *v1beta2.SparkApplication) error {
...
 func addPriorityClassName(pod *corev1.Pod, app *v1beta2.SparkApplication) error {
 	var priorityClassName *string
-	if app.Spec.BatchSchedulerOptions != nil {
-		priorityClassName = app.Spec.BatchSchedulerOptions.PriorityClassName
+
+	if util.IsDriverPod(pod) && *app.Spec.Driver.PriorityClassName != "" {
+		priorityClassName = app.Spec.Driver.PriorityClassName
+	} else if util.IsExecutorPod(pod) && *app.Spec.Executor.PriorityClassName != "" {
+		priorityClassName = app.Spec.Executor.PriorityClassName
 	}
```
ChenYi015, line 485:
```suggestion
	if util.IsDriverPod(pod) {
		priorityClassName = app.Spec.Driver.PriorityClassName
	} else if util.IsExecutorPod(pod) {
		priorityClassName = app.Spec.Executor.PriorityClassName
	}
```

vara-bonthu (approved):
Thanks for the PR @Kevinz857 🙌🏼 

/approve

@jacobsalway  I know you left some comments. Are you happy with the PR?

PR #2149, opened 2024-09-03
Fix: e2e test failes due to webhook not ready

ChenYi015:
## Purpose of this PR

**Proposed changes:**
- Wait for the webhooks to be ready in e2e test CI

## Change Category

- [ ] Documentation update

### Rationale

The [e2e CI](https://github.com/kubeflow/spark-operator/actions/runs/10678444394/job/29595500565) occasionally failed due to webhook not ready. 

## Checklist

### DIFF
--- a/test/e2e/suit_test.go
+++ b/test/e2e/suit_test.go
@@ -31,8 +31,11 @@ import (
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/cli"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


jacobsalway:
Acking the review assignment for tomorrow and thanks for taking a look at this before me

test/e2e/suit_test.go
```diff
@@ -53,6 +56,12 @@ import (
 const (
 	ReleaseName      = "spark-operator"
 	ReleaseNamespace = "spark-operator"
+
+	MutatingWebhookName   = "spark-operator-webhook"
+	ValidatingWebhookName = "spark-operator-webhook"
```
jacobsalway, line 61: It looks like these constants aren't used where I imagine you intended to use them on lines 148 and 149.

test/e2e/suit_test.go
jacobsalway, line 1: I like that you've pulled the `waitFor...` functions out into their own utilities. Should cut down on some boilerplate repetition if we add more e2e tests

test/e2e/suit_test.go
ChenYi015: You are right, I have fixed it.

jacobsalway:
Hmm I've run it a few times locally and was able to replicate the `connection refused` error once but not after that. I wonder if it's flakey depending on the exact poll timing vs when the bundle was updated and adding some buffer time might help?

> After you create the webhook configuration, the system will take a few seconds to honor the new configuration.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly

ChenYi015:
> I added another check to ensure that webhook service endpoints are not empty.

I did 10+ e2e tests locally, and all of them complete successfully.

jacobsalway:
Noting for posterity: we discussed this issue/change separately and think the flakiness is due to the `MutatingWebhookConfiguration` being created by Helm before the controller has generated the certs/the deployment or service objects are ready.

I'm happy to approve this PR as is/just add a time.Sleep buffer to fix the e2e flakiness if it's an issue, but we could also find a better fix

PR #2161, opened 2024-09-10
Update e2e tests

ChenYi015:
## Purpose of this PR

**Proposed changes:**
- Remove duplicate operator image build tasks in integration test
- Add time sleep buffer to ensure that webhooks are ready before running the e2e tests
- Check whether volumes are mounted in e2e tests
- Check whether resource requests and limits are set in e2e tests
- Remove the `coreLimit` field from all example spark applications except for `spark-pi-custom-resource.yaml`. For that, added `coreRequest` and `coreLimit` fields to it.

## Change Category

- [ ] Documentation update

## Checklist

### Additional Notes

The e2e tests of the recent PRs still failed due to the webhook not ready, even after checking the CA bundle and webhook service enpoints.

### DIFF
--- a/test/e2e/sparkapplication_test.go
+++ b/test/e2e/sparkapplication_test.go
@@ -26,11 +26,14 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"github.com/kubeflow/spark-operator/api/v1beta2"
+	"github.com/kubeflow/spark-operator/pkg/common"
 	"github.com/kubeflow/spark-operator/pkg/util"
 )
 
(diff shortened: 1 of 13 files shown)


test/e2e/suit_test.go
```diff
@@ -149,6 +149,8 @@ var _ = BeforeSuite(func() {
 	validatingWebhookKey := types.NamespacedName{Name: ValidatingWebhookName}
 	Expect(waitForMutatingWebhookReady(context.Background(), mutatingWebhookKey)).NotTo(HaveOccurred())
 	Expect(waitForValidatingWebhookReady(context.Background(), validatingWebhookKey)).NotTo(HaveOccurred())
+	// TODO: Remove this when there is a better way to ensure the webhooks are ready before running the e2e tests.
+	time.Sleep(10 * time.Second)
```
jacobsalway, line 153: I think this is fine as a temporary workaround. Two ways we can solve this in the future:

* Remove the dependency on the webhook altogether by constructing pod template files on the fly from the existing SparkApplication spec. Maybe via a ConfigMap into the driver pod?
* The controller should only start reconciling if the webhook is ready to serve.

test/e2e/sparkapplication_test.go
```diff
@@ -176,8 +209,26 @@ var _ = Describe("Example SparkApplication", func() {
...
+				}
+				if app.Spec.Driver.CoreRequest != nil {
+					Expect(container.Resources.Requests.Cpu().Equal(resource.MustParse(*app.Spec.Driver.CoreRequest))).To(BeTrue())
+				}
+				if app.Spec.Driver.CoreLimit != nil {
+					Expect(container.Resources.Limits.Cpu().Equal(resource.MustParse(*app.Spec.Driver.CoreLimit))).To(BeTrue())
+				}
+				Expect(container.Resources.Requests.Memory).NotTo(BeNil())
+				Expect(container.Resources.Limits.Memory).NotTo(BeNil())
+			}
```
jacobsalway, line 229: I'm curious: was there any specific reason to add this assertion? The request and limit fields on the pods are set via `spark-submit` args and aren't patched by the webhook issue that I believe you raised this PR as part of

ChenYi015: Yes, the cpu/memory requests/limits are not patched by the webhook. I added these assertions just to ensure that `spark-pi-custom-resouce.yaml` works as expected in setting resource requests/limits correctly.

jacobsalway: What do you think about removing the `coreLimit` field from the existing example applications and creating a new application that specifies both `coreRequest` and `coreLimit`, creating a new end-to-end test for that, then moving this assertion logic to that test? I feel like that captures the intention better but open to your opinion too obviously.

ChenYi015: @jacobsalway  Sounds good. I have removed the `coreLimit` field from all the example applications except for `spark-pi-custom-resource.yaml`. For that, I added the `coreRequest` and `coreLimit` fields to it.

Issue #2176, opened 2024-09-17
[BUG] Yunikorn taskgroups memory minResources is not including spark.executor.pyspark.memory


tcassaert:
## Description

When submitting an application with in `sparkConf` a value for `spark.executor.pyspark.memory`, this memory is added to the `request` of the executor, but it's not included in the `yunikorn.apache.org/task-groups` annotation. This makes the executor being stuck in `Pending` and Yunikorn reporting that the request for the pod is bigger than what the placeholder reserved.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Reproduction Code [Required]

Submit following `SparkApplication`:
```yaml
---
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: spark-pi-python
  namespace: default
spec:
  type: Python
  pythonVersion: "3"
  mode: cluster
  image: spark:3.5.2
  imagePullPolicy: IfNotPresent
  mainApplicationFile: local:///opt/spark/examples/src/main/python/pi.py
  sparkVersion: 3.5.2
  sparkConf:
    spark.executor.pyspark.memory: "2000"
  driver:
    annotations:
      yunikorn.apache.org/schedulingPolicyParameters: 'gangSchedulingStyle=Hard'
    cores: 1
    memory: 4096m
    memoryOverhead: 1200m
    serviceAccount: spark-operator-spark
  executor:
    instances: 2
    cores: 1
    memory: 4096m
    memoryOverhead: 1200m
  batchScheduler: yunikorn
  batchSchedulerOptions:
    queue: root.default
```

## Expected behavior

The task-group for the executors should have included the `spark.executor.pyspark.memory` in the `minResources.memory`.

## Actual behavior

The task-group looks like this:
```
yunikorn.apache.org/task-groups: '[{"name":"spark-driver","minMember":1,"minResource":{"cpu":"1","memory":"5296Mi"}},{"name":"spark-executor","minMember":2,"minResource":{"cpu":"1","memory":"5296Mi"}}]'
```

Where `minResource.memory` is a sum of `memory` and `memoryOverhead`. 
The request of the executor pods is however:
```yaml
    resources:
      limits:
        memory: 7296Mi
      requests:
        cpu: "1"
        memory: 7296Mi
```
Where the `resources.requests.memory` is a sum of `memory`, `memoryOverhead` and `spark.executor.pyspark.memory`. 

This results in pods that never get scheduled.

[...]

PR #2178, opened 2024-09-18
Account for spark.executor.pyspark.memory in Yunikorn gang scheduling

jacobsalway:
## Purpose

Account for `spark.executor.pyspark.memory` when calculating executor memory requests when gang scheduling with Yunikorn.

Fixes https://github.com/kubeflow/spark-operator/issues/2176

## Change Category

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/scheduler/yunikorn/resourceusage/memory.go
+++ b/internal/scheduler/yunikorn/resourceusage/memory.go
@@ -70,6 +70,26 @@ func memoryRequestBytes(podSpec *v1beta2.SparkPodSpec, memoryOverheadFactor floa
 	return memoryBytes + memoryOverheadBytes, nil
 }
 
+func executorPysparkMemoryBytes(app *v1beta2.SparkApplication) (int64, error) {
+	pysparkMemory, found := app.Spec.SparkConf["spark.executor.pyspark.memory"]
+	if app.Spec.Type != v1beta2.SparkApplicationTypePython || !found {
+		return 0, nil
+	}
+
+	// This fields defaults to mebibytes if no resource suffix is specified
+	// https://github.com/apache/spark/blob/7de71a2ec78d985c2a045f13c1275101b126cec4/docs/configuration.md?plain=1#L289-L305
+	if _, err := strconv.Atoi(pysparkMemory); err == nil {
+		pysparkMemory = pysparkMemory + "m"
+	}
+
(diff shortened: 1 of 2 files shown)


internal/scheduler/yunikorn/scheduler_test.go
```diff
@@ -223,6 +223,49 @@ func TestSchedule(t *testing.T) {
...
+						"memory": "896Mi", // 512Mi + 384Mi min overhead
+					},
+				},
+				{
+					Name:      "spark-executor",
+					MinMember: 2,
+					MinResource: map[string]string{
+						"cpu": "1",
+						// 512Mi + 384Mi min overhead + 500Mi spark.executor.pyspark.memory
+						"memory": "1396Mi",
```
jacobsalway, line 264: I wonder if it would be more robust to move these these to an end-to-end test suite and assert against the actual container request value? Thoughts?

internal/scheduler/yunikorn/resourceusage/memory.go
```diff
@@ -103,6 +122,11 @@ func executorMemoryRequest(app *v1beta2.SparkApplication) (string, error) {
...
 	}
 
+	pysparkMemoryBytes, err := executorPysparkMemoryBytes(app)
+	if err != nil {
+		return "", err
+	}
+
 	// See comment above in driver
-	return bytesToMi(requestBytes), nil
+	return bytesToMi(requestBytes + pysparkMemoryBytes), nil
```
jacobsalway, line 131: I think I'm also missing offheap bytes in this calculation https://github.com/apache/spark/blob/7de71a2ec78d985c2a045f13c1275101b126cec4/core/src/main/scala/org/apache/spark/resource/ResourceProfile.scala#L525-L532

internal/scheduler/yunikorn/scheduler_test.go
ChenYi015: Maybe we can add some e2e tests for yunikorn scheduler in the future.

ChenYi015 (approved):

ChenYi015:
@jacobsalway I think the same problem may happen to volcano scheduler and kube scheduler, perhaps we can abstract out a common util function to calculate the requested memory bytes.

jacobsalway:
Will create a PR in the near future to pull out the `resourceusage` package and also account for offheap memory usage.

Issue #2211, opened 2024-10-01
[FEATURE] Remove ned_bind_service requirement and option to disable ipv6 in the helm chart


cccsss01:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?
Ability to run the operator w/o bind_net_service

Currently it looks like the controller and webhook require net_bind_service because it has an ipv6 listener for *

#### Describe the solution you would like

The solution I would like is an option in the helm chart to disable ipv6, as well as not require net_bind_service 

#### Describe alternatives you have considered

Looking into setting the routes in the container?

jacobsalway:
Do you need anything specific for IPv6 or is removing the `cap_net_bind_service` flag on the binary enough for this issue?

cccsss01:
having a helm chart option to disable ipv6 would be great, right now 8081 and 8080 is configured to listen ipv6 *, which is what I think is causing the net_bind_service.  We have a developer working on adding
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
net.ipv6.conf.lo.disable_ipv6 = 1

to the pod to determine if that will get rid of the net_bind_service requirement.

https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#setting-sysctls-for-a-pod

jacobsalway:
@cccsss01 can you got into more detail on the request to disable IPv6 and what issues you're seeing? Admittedly I haven't worked in an IPv6 environment, but when I test my local build with `setcap 'cap_net_bind_service=+ep'` removed on a local IPv6 Kind cluster, the webhook is able to bind and successfully start.

cccsss01:
I think your assessment is likely perfect. The current hurdle we were looking at was the ipv6 attempting to listen on privileged ports.  With your current settings if that prevents it that's perfect.  The additional information I have, that may or may not be a problem with this application is w/ Cilium's Hubble UI, if you have ipv6 disabled on the host, and don't have ipv6 listeners disabled (provided in the helm chart), it will break the deployment.

PR #2216, opened 2024-10-03
Remove `cap_net_bind_service` from image

jacobsalway:
## Purpose of this PR

* Remove `setcap 'cap_net_bind_service=+ep'` from the build so the container can be run with all capabilities dropped on a non-privileged port
* Looking at other open source components, it seems more common to run on a non-privileged port and require either root or `NET_BIND_SERVICE` to run on a privileged port e.g. 443
  * e.g. https://open-policy-agent.github.io/gatekeeper/website/docs/vendor-specific/#running-on-private-gke-cluster-nodes
* The webhook runs on port 9443 by default which doesn't require any privilege escalation

Run the operator with the following values:

```yaml
controller:
  securityContext:
    capabilities:
      drop: ["ALL"]
webhook:
  securityContext:
    capabilities:
      drop: ["ALL"]
```

https://github.com/kubeflow/spark-operator/issues/2211

## Change Category

- [X] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,10 +20,6 @@ FROM golang:1.23.1 AS builder
 
 WORKDIR /workspace
 
-RUN apt-get update \
-    && apt-get install -y libcap2-bin \
-    && rm -rf /var/lib/apt/lists/*
-
 RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=bind,source=go.mod,target=go.mod \
     --mount=type=bind,source=go.sum,target=go.sum \
@@ -36,7 +32,6 @@ ARG TARGETARCH
 RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=cache,target="/root/.cache/go-build" \
     CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on make build-operator
-RUN setcap 'cap_net_bind_service=+ep' /workspace/bin/spark-operator
 
 FROM ${SPARK_IMAGE}
 


ImpSy:
Yeah I implemented a solution for allowing `runAsNonRoot: true` in the security context while selecting all possible ports

If we really want to `drop all capabilities` we can remove it (but we need to document that the container can't use ports <1024)

Dockerfile
```diff
@@ -36,7 +36,6 @@ ARG TARGETARCH
 RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=cache,target="/root/.cache/go-build" \
     CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on make build-operator
-RUN setcap 'cap_net_bind_service=+ep' /workspace/bin/spark-operator
```
ImpSy, line 39: you can also remove lines 23-25, since we only install `libcap2-bin` to be able to run `setcap`

jacobsalway: Thanks, good catch I missed that

jacobsalway:
Posting some examples I found of other projects: 

* [OPA Gatekeeper](https://open-policy-agent.github.io/gatekeeper/website/docs/vendor-specific/)
* [`metrics-server`](https://github.com/kubernetes-sigs/metrics-server/pull/884)
* `ingress-nginx` uses this field, however I think you'd usually expect NGINX to run on port 80 or 443 as it's usually used for web traffic and not an internal webhook. https://github.com/kubernetes/ingress-nginx/issues/10002

ChenYi015 (approved):

jacobsalway:
@ChenYi015 let's make sure we call this out specifically on whichever release changelog this ends up in

ChenYi015:
> let's make sure we call this out specifically on whichever release changelog this ends up in.

Sure, I will clarify this in the next release changelog (maybe `v2.0.2`).

/lgtm

Issue #2237, opened 2024-10-10
[BUG] Retries don't work


Tom-Newton:
## Description
Spark applications that fail submission with `onSubmissionFailureRetries` > 0 get stuck in `SUBMISSION_FAILED` state. Similarly applications that fail after submission with `onFailureRetries` > 0 get stuck in `FAILING` state. 

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Reproduction Code [Required]
Use the following config (the provided dummy values are enough to reproduce). This demonstrates the `onSubmissionFailureRetries`.
```
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: test
spec:
  image: "dummy"
  mainApplicationFile: "dummy"
  mode: cluster
  type: Python
  restartPolicy:
    type: OnFailure
    onSubmissionFailureRetries: 3
    onSubmissionFailureRetryInterval: 5
  sparkVersion: 3.5.0
```

Run `kubectl apply -f <yaml file containing the above content>`
Then watch the status of the `SparkApplication` custom resource. 

## Expected behavior
Given the dummy values we expect this to fail `spark-submit`, but it should retry 3 times then go into a `FAILED` status. 
```
  applicationState:
    errorMessage: "failed to run spark-submit: failed to run spark-submit: 24/10/10
      00:34:37 WARN NativeCodeLoader: Unable to load native-hadoop library for your
      platform... using builtin-java classes where applicable\nException in thread
      \"main\" org.apache.spark.SparkException: Failed to get main class in JAR with
      error 'File file:/opt/spark/work-dir/dummy does not exist'.  Please specify
      one with --class.\n\tat org.apache.spark.deploy.SparkSubmit.error(SparkSubmit.scala:1047)\n\tat
      org.apache.spark.deploy.SparkSubmit.prepareSubmitEnvironment(SparkSubmit.scala:528)\n\tat
      org.apache.spark.deploy.SparkSubmit.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:964)\n\tat
      org.apache.spark.deploy.SparkSubmit.doRunMain$1(SparkSubmit.scala:194)\n\tat
      org.apache.spark.deploy.SparkSubmit.submit(SparkSubmit.scala:217)\n\tat org.apache.spark.deploy.SparkSubmit.doSubmit(SparkSubmit.scala:91)\n\tat
[...]

Issue #2118, opened 2024-08-10
[BUG] SparkApplication in FAILING state has finish time


BalaMahesh:
## Description
We are using spark operator  v1beta2-1.6.2-3.5.0 in production. We have a spark application with the below policy. 

```
  restartPolicy:
    onFailureRetryInterval: 10
    onSubmissionFailureRetries: 5
    onSubmissionFailureRetryInterval: 20
    type: Always
```

When driver pod failed for some reason, operator has the below logs 

```
UID:"263d8e51-4a32-4832-94b5-f73043e4bd69", APIVersion:"sparkoperator.k8s.io/v1beta2", ResourceVersion:"1343465049", FieldPath:""}): type: 'Warning' reason: 'SparkDriverFailed' Driver **-regular-driver failed
I0810 13:25:06.814383      10 controller.go:860] Update the status of SparkApplication namespace/**-regular from:
{
  "sparkApplicationId": "spark-5a8712b706984da5bd90ac3fa7f35b86",
  "submissionID": "9b80c355-fc31-4680-b3ce-7ce34f7c31ec",
  "lastSubmissionAttemptTime": "2024-08-10T10:26:46Z",
  "terminationTime": null,
  "driverInfo": {
    "webUIServiceName": "**-regular-ui-svc",
    "webUIPort": 4040,
    "webUIAddress": "10.204.49.228:0",
    "podName": "**-regular-driver"
  },
  "applicationState": {
    "state": "RUNNING"
  },
  "executorState": {
    "**-8f53b6913bd57dcf-exec-7": "UNKNOWN",
    "**-8f53b6913bd57dcf-exec-8": "FAILED",
    "**-8f53b6913bd57dcf-exec-9": "COMPLETED"
  },
  "executionAttempts": 63,
  "submissionAttempts": 1
}
to:
{
  "sparkApplicationId": "spark-5a8712b706984da5bd90ac3fa7f35b86",
  "submissionID": "9b80c355-fc31-4680-b3ce-7ce34f7c31ec",
  "lastSubmissionAttemptTime": "2024-08-10T10:26:46Z",
  "terminationTime": "2024-08-10T13:25:06Z",
  "driverInfo": {
    "webUIServiceName": "**-regular-ui-svc",
    "webUIPort": 4040,
    "webUIAddress": "10.204.49.228:0",
    "podName": "**-regular-driver"
  },
  "applicationState": {
    "state": "FAILING",
[...]

PR #2241, opened 2024-10-13
Fix retries

Tom-Newton:
## Purpose of this PR
Fix retries for submission failure and application. Closes: https://github.com/kubeflow/spark-operator/issues/2237 and apparently maybe https://github.com/kubeflow/spark-operator/issues/2118 too.

**Proposed changes:**
- Convert `isNextRetryDue` into `timeUntilNextRetryDue` so we know how to set `RequeueAfter`. 
- After completing a reconcile where the outcome is to wait longer before retrying, make sure to set a requeue time. Without this the controller won't attempt to reconcile the `SparkApplication` again until 10 hours later after default Cache.SyncPeriod.
- Remove the `bool` return from `runSparkSubmit` and just return `error`, because the bool argument is equivalent to `err != nil`. Also remove the related code in `submitSparkApplication` since that now seems to be out of date. 
- Refactor `submitSparkApplication` and how its called a bit to ensure that submission and execution attempt counts, logging of errors and state are updated consistently. 
- Always return an error with meaningful name in `reconcileFailedSubmissionSparkApplication` if `validateSparkResourceDeletion` returns `false`. Previously if was returning `err` set from a prior and possibly unrelated function call. 
- Added new e2e tests for submission failure and application failure.

## Change Category

- [ ] Documentation update

## Checklist

- [x] I have updated documentation accordingly. - I don't think any change is required 
- [x] I have added tests that prove my changes are effective or that my feature works

### Additional Notes
I'm a golang noob


Tom-Newton:
I wasn't expecting to get reviewers assigned after opening as draft. This isn't ready for review yet.

Tom-Newton:
Just rebased to get the CI fix here too. I think this is ready for review now

josecsotomorales:
@ChanYiLin can we review this PR and create a new patch release if it looks good?

josecsotomorales:
@ChenYi015 can we review this PR and create a new release if it looks good?

jacobsalway:
Thanks for the tag. Sorry have been busy with work but will review this tonight.

josecsotomorales:
@ChenYi015 can we review this PR and create a new release if it looks good? @jacobsalway I'm happy to help with contributions

test/e2e/sparkapplication_test.go
```diff
@@ -236,6 +237,121 @@ var _ = Describe("Example SparkApplication", func() {
...
+				v1beta2.ApplicationStateFailedSubmission,
+			}
+			for _, app := range apps[:len(apps)-1] {
+				Expect(validStatuses).To(ContainElement(app.Status.AppState.State))
+			}
+
+			By("Checking driver does not exist")
+			driverPodName := util.GetDriverPodName(app)
+			_, get_driver_err := clientset.CoreV1().Pods(app.Namespace).Get(ctx, driverPodName, metav1.GetOptions{})
+			Expect(errors.IsNotFound(get_driver_err)).To(BeTrue())
```
ChenYi015, line 291:
```suggestion
            driverPodKey := types.NamespacedName{Namespace: app.Namespace, Name: driverPodName}
            err := k8sClient.Get(ctx, driverPodKey, &corev1.Pod{})
			Expect(errors.IsNotFound(err)).To(BeTrue())
```

We prefer using the k8s client provided by the controller-runtime if possible.

test/e2e/sparkapplication_test.go
```diff
@@ -236,6 +237,121 @@ var _ = Describe("Example SparkApplication", func() {
...
+			By("Waiting for SparkApplication to terminate")
+			key := types.NamespacedName{Namespace: app.Namespace, Name: app.Name}
+			apps, polling_err := collectSparkApplicationsUntilTermination(ctx, key)
+			Expect(polling_err).To(HaveOccurred())
+
+			By("Should eventually fail")
+			final_app := apps[len(apps)-1]
+			Expect(final_app.Status.AppState.State).To(Equal(v1beta2.ApplicationStateFailed))
+			Expect(final_app.Status.AppState.ErrorMessage).To(ContainSubstring("failed to run spark-submit"))
+			Expect(final_app.Status.SubmissionAttempts).To(Equal(*app.Spec.RestartPolicy.OnSubmissionFailureRetries + 1))
```
ChenYi015, line 277: In golang, we prefer using the `camelCase` naming convention instead of using an underscore. Ref: https://go.dev/doc/effective_go#mixed-caps.

internal/controller/sparkapplication/controller.go
```diff
@@ -709,50 +739,40 @@ func (r *Reconciler) submitSparkApplication(app *v1beta2.SparkApplication) error
...
-	app.Status.ExecutionAttempts = app.Status.ExecutionAttempts + 1
-	app.Status.LastSubmissionAttemptTime = metav1.Now()
-	r.recordSparkApplicationEvent(app)
 	return nil
 }
 
-// Helper func to determine if the next retry the SparkApplication is due now.
-func isNextRetryDue(app *v1beta2.SparkApplication) bool {
-	retryInterval := app.Spec.RestartPolicy.OnFailureRetryInterval
+func timeUntilNextRetryDue(app *v1beta2.SparkApplication) (time.Duration, error) {
```
ChenYi015, line 756: Maybe we can move the `timeUntilNextRetryDue` util func to file `pkg/util/sparkapplication.go`.

ChenYi015 (changes requested):

internal/controller/sparkapplication/controller.go
```diff
@@ -632,8 +641,29 @@ func (r *Reconciler) getSparkApplication(key types.NamespacedName) (*v1beta2.Spa
...
 // submitSparkApplication creates a new submission for the given SparkApplication and submits it using spark-submit.
-func (r *Reconciler) submitSparkApplication(app *v1beta2.SparkApplication) error {
+func (r *Reconciler) submitSparkApplication(app *v1beta2.SparkApplication) (returned_error error) {
 	logger.Info("Submitting SparkApplication", "name", app.Name, "namespace", app.Namespace, "state", app.Status.AppState.State)
+	// SubmissionID must be set before creating any resources to ensure all the resources are labeled.
+	app.Status.SubmissionID = uuid.New().String()
+	app.Status.LastSubmissionAttemptTime = metav1.Now()
+
+	defer func() {
+		app.Status.SubmissionAttempts = app.Status.SubmissionAttempts + 1
```
ChenYi015, line 651: We can put this line after line 648.

internal/controller/sparkapplication/controller.go
```diff
@@ -632,8 +641,29 @@ func (r *Reconciler) getSparkApplication(key types.NamespacedName) (*v1beta2.Spa
...
+
+		if returned_error == nil {
+			app.Status.AppState = v1beta2.ApplicationState{
+				State: v1beta2.ApplicationStateSubmitted,
+			}
+			app.Status.ExecutionAttempts = app.Status.ExecutionAttempts + 1
+		} else {
+			logger.Error(returned_error, "Failed to submit SparkApplication", "name", app.Name, "namespace", app.Namespace, "state", app.Status.AppState.State)
```
ChenYi015, line 659: Maybe we should use `logger.Info` instead of `logger.Error` which will print the stack trace, failed to submit spark application does not imply the failure of reconcile.

internal/controller/sparkapplication/controller.go
```diff
@@ -172,6 +172,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{Requeue: true}, err
 	}
 	logger.Info("Reconciling SparkApplication", "name", app.Name, "namespace", app.Namespace, "state", app.Status.AppState.State)
+	defer logger.Info("Finished reconciling SparkApplication", "name", app.Name, "namespace", app.Namespace)
```
ChenYi015, line 175:
```suggestion
	defer logger.Info("Finished reconciling SparkApplication", "name", app.Name, "namespace", app.Namespace, "state", app.Status.AppState.State)
```

We can add info about application state in the log statement.

Tom-Newton: I actually tried this but it doesn't work because the `app` object we use here is not the same one that we mutate during reconcile. I think we would need to call `getSparkApplication` again so I thought it probably wasn't worth it. Without status this log line's value is reduced, but I decided it was probably worth keeping.

josecsotomorales:
Hey @ChenYi015, it looks like change requests were addressed, can we get another review?

internal/controller/sparkapplication/controller.go
```diff
@@ -328,15 +322,22 @@ func (r *Reconciler) reconcileFailedSubmissionSparkApplication(ctx context.Conte
...
+				if err != nil {
+					return err
+				}
+				if timeUntilNextRetryDue <= 0 {
 					if r.validateSparkResourceDeletion(ctx, app) {
 						_ = r.submitSparkApplication(app)
 					} else {
 						if err := r.deleteSparkResources(ctx, app); err != nil {
 							logger.Error(err, "failed to delete resources associated with SparkApplication", "name", app.Name, "namespace", app.Namespace)
 						}
```
ChenYi015, line 335: The return error statement should be placed in the curly bracket.
```suggestion
						if err := r.deleteSparkResources(ctx, app); err != nil {
							logger.Error(err, "failed to delete resources associated with SparkApplication", "name", app.Name, "namespace", app.Namespace)
							return fmt.Errorf("resources associated with SparkApplication name: %s namespace: %s, needed to be deleted", app.Name, app.Namespace)
						}
```

internal/controller/sparkapplication/controller.go
```diff
@@ -328,15 +322,22 @@ func (r *Reconciler) reconcileFailedSubmissionSparkApplication(ctx context.Conte
...
-						return err
+						return fmt.Errorf("resources associated with SparkApplication name: %s namespace: %s, needed to be deleted", app.Name, app.Namespace)
```
ChenYi015, line 336:
```suggestion
```

internal/controller/sparkapplication/controller.go
Tom-Newton: The way I had it preserves the old behaviour and I think this suggestion would actaully introduce a bug. With this suggestion, if `deleteSparkResources` succeeds then reconcile will return `ctrl.Result{}, nil`, meaning it won't be requeued. But we need it to be requeued because after `r.validateSparkResourceDeletion` returned `false` its not going to try re-submitting the the application in this reconcile. 

We could make a slightly larger change so that `submitSparkApplication` is also called if `deleteSparkResources` is successful. That would allow the submission to happen in the same reconcile, which sounds like a good thing to me but I have little experience developing kubernetes controllers.

josecsotomorales: I recommend merging this PR in its current state. I tested it on my environment, and it fixes the `restartPolicy` issues we have been seeing.

ChenYi015: @Tom-Newton Yes, you are right. Will merge this PR first, we can optimize the reconcile logic later. For a failed-submission  app with max retries 3, the controller will need to reconcile 9 times now (NEW -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED_SUBMISSION -> FAILED).

ChenYi015 (approved):

internal/controller/sparkapplication/controller.go
Tom-Newton: I created an issue so this doesn't get forgot https://github.com/kubeflow/spark-operator/issues/2285

Issue #2244, opened 2024-10-14
[BUG] minResources of volcano podGroup didn't take into account dynamicAllocation and memoryOverheadFactor


kaka-zb:
## Description
We've been using spark operator and volcano for a long time in production env, however, there are some problems with the calculation of resource usage for volcano podGroup when the sparkapp is submitted.

The spark.dynamicAllocation.* & spark.kubernetes.memoryOverheadFactor params of spark are not taken into account  when calculating memory of minResources for volcano podGroup. As a result, the calculated minResources maybe smaller than real usage of sparkapp, and the gang scheduling maybe fail.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Environment & Versions

- Spark Operator App version: 2.0.1
- Helm Chart Version: 2.0.1
- Kubernetes Version: 1.25.7
- Apache Spark version: 3.4.3

kaka-zb:
BTW, i see that resourceusage directory implemented in yunikorn, but if you want, I can contribute our code for volcano, which has been verified by thousands times of spark task.

jacobsalway:
Hey, I wrote the `resourceusage` module for the Yunikorn batch scheduler. When I implemented this initially we discussed pulling these functions out into a more generic module for use across other batch schedulers. If you have code that also calculates the resulting pod resource fields I'd be happy to review and hopefully improve the existing solution.

kaka-zb:
@jacobsalway Thanks for reply, i will submit a draft PR and then you can reveiw and then see if that helps.

PR #2262, opened 2024-10-18
Draft: fix the minResources calculation logic of podgroup for volcano

kaka-zb:
## Purpose of this PR
As described in #2244, the spark.dynamicAllocation.* & spark.kubernetes.memoryOverheadFactor params of spark are not taken into account when calculating memory of minResources for volcano podGroup, so i created this PR and see if that helps. And then we can consider extracting `resourceusage` module to the `scheduler` directory.

Note that only `cluster deploy mode` is considered in this PR, after `resourceusage` module is extracted, i can import `resourceusage` module in volcano package and fix the minResources calculation logic in both `cluster deploy mode` and `client deploy mode`.

### DIFF
--- a/internal/scheduler/volcano/resource_usage.go
+++ b/internal/scheduler/volcano/resource_usage.go
@@ -1,0 +1,251 @@
+package volcano
+
+import (
+	"fmt"
+	"github.com/kubeflow/spark-operator/api/v1beta2"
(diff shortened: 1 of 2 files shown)


kaka-zb:
@jacobsalway, please take a look when you're free

PR #2264, opened 2024-10-18
Add workflow for releasing sparkctl binary

### LINKED ISSUE #1009: Provide release binary for sparkctl
Has anyone explored adding a `sparkctl` binary and/or Docker image to release process? Would be handy to share these with community to use. Just for sparkctl

ChenYi015:
## Purpose of this PR

Close #1009.

**Proposed changes:**
- Add workflow for releasing sparkctl binary

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -65,9 +65,51 @@ jobs:
             echo "Tag '${VERSION}' does not exist."
           fi
 
-  build_images:
+  release_sparkctl:
     needs:
       - check-release
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: true
+      matrix:
+        os:
+          - linux
+          - darwin
+        arch:
+          - amd64
+          - arm64
+
+    env:
+      GOOS: ${{ matrix.os }}
+      GOARCH: ${{ matrix.arch }}
+
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v4
+
+      - name: Read version from VERSION file
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -65,12 +65,49 @@ jobs:
...
 
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - linux
+          - darwin
+        arch:
+          - amd64
+          - arm64
```
ImpSy, line 82: The matrix here as no impact on the os / arch use since we rely on ubuntu-latest (linux/amd64)

But we could rely on go native capability to build cross-platform binaries by setting `GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }}` as env var when running `make build-sparkctl`

I would not recommend using matrix in that specific case because It create new runner for each variation and does not reuse the go build cache meaning that time we have to wait a full build time

ChenYi015: > But we could rely on go native capability to build cross-platform binaries by setting `GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }}` as env var when running `make build-sparkctl`

Make sense. Have added them as job envs.

> I would also not recommend using matrix in that specific case because It create new runner for each variation and does not reuse the go build cache meaning that time we have to wait a full build time

It will truly create 4 different runners, but they will build sparkctl binary concurrently, so actually the whole waiting time is shorter.

ImpSy (approved):

PR #2219, opened 2024-10-04
Set readOnlyRootFilesystem: true for the controller and webhook containers to improve default security posture

npgretz:
### 🛑 Important:
Please open an issue to discuss significant work before you start. We appreciate your contributions and don't want your efforts to go to waste!

For guidelines on how to contribute, please review the [CONTRIBUTING.md](CONTRIBUTING.md) document.

## Purpose of this PR
Provide a clear and concise description of the changes. Explain the motivation behind these changes and link to relevant issues or discussions.

By default, the controller and webhook containers write to their filesystems. DevOpSec at multiple organizations using the spark-operator chart require Kubernetes container security contexts to prohibit writing to the root filesystem ( [Slack Conversation](https://cloud-native.slack.com/archives/C074588U7EG/p1727852883726249), [Issue for Improved Security](https://github.com/kubeflow/spark-operator/issues/2218#issue-2565007558))

This PR will allow the spark-operator chart to be deployed with a read-only root filesystem to improve the default security posture and meet the requirements of users.

**Proposed changes:**
- Add an empty directory onto the webhook containers to create certs in
- Add an empty directory onto the controller containers to write Spark artifacts for Spark apps
- Set the default container security context for the webhook and controller to readOnlyRootFilesystem

## Change Category
Indicate the type of change by marking the applicable boxes:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

### Rationale

Multiple users of the Helm chart require an increased security posture in their Kubernetes deployments.

## Checklist
Before submitting your PR, please review the following:

- [ ] I have conducted a self-review of my own code.
- [ ] I have updated documentation accordingly.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [ ] Existing unit tests pass locally with my changes.

PR #2265, opened 2024-10-18
Update default container security context

ChenYi015:
## Purpose of this PR

**Proposed changes:**
- Update default container security context

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -158,10 +158,14 @@ controller:
     #   memory: 300Mi
 
   # -- Security context for controller containers.
-  securityContext: {}
-    # runAsUser: 1000
-    # runAsGroup: 2000
-    # fsGroup: 3000
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 185
+    runAsGroup: 185
+    fsGroup: 185
+    capabilities:
+      drop:
+      - ALL
 
   # -- Sidecar containers for controller pods.
   sidecars: []
@@ -301,10 +305,14 @@ webhook:
     #   memory: 300Mi
 
   # -- Security context for webhook containers.
-  securityContext: {}
-    # runAsUser: 1000
-    # runAsGroup: 2000
-    # fsGroup: 3000
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 185
+    runAsGroup: 185
+    fsGroup: 185
+    capabilities:
+      drop:
+      - ALL
 
   # Pod disruption budget for webhook to avoid service degradation.
   podDisruptionBudget:
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -161,6 +161,7 @@ controller:
   securityContext:
     runAsNonRoot: true
     fsGroup: 185
+    allowPrivilegeEscalation: false
     capabilities:
       drop:
       - ALL
@@ -306,6 +307,7 @@ webhook:
   securityContext:
     runAsNonRoot: true
     fsGroup: 185
+    allowPrivilegeEscalation: false
     capabilities:
       drop:
       - ALLL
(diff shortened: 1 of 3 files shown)


jacobsalway:
Might need to do some investigation to map out all the disk usage of `spark-submit` but what are your thoughts on also using an `emptyDir` volume in the controller and adding `readOnlyRootFilesystem: true` as well?

charts/spark-operator-chart/values.yaml
```diff
@@ -156,10 +156,14 @@ controller:
...
 
   # -- Security context for controller containers.
-  securityContext: {}
-    # runAsUser: 1000
-    # runAsGroup: 2000
-    # fsGroup: 3000
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 185
+    runAsGroup: 185
```
ImpSy, line 162: Hardcoding UserID and GroupID feels a bit to static
If we ever decide to change our base image those could change

But requiring `runAsNonRoot`, setting an FsGroup and dropping all capabilities is a great idea

ImpSy:
> Might need to do some investigation to map out all the disk usage of spark-submit but what are your thoughts on also using an emptyDir volume in the controller and adding readOnlyRootFilesystem: true as well?

Yeah I feel like `readOnlyRootFilesystem` is a bit too restrictive especially while we are actively modifying the usage of the controller with pod templates

But like the PR #2219 propose we could mount a  volume on /tmp and use `os.CreateTemp` everytime we need a local file

jacobsalway:
Could push the user and group directives down to the Dockerfile rather than the container security context. I'm not sold on read-only FS because any temporary disk usage could cause a panic but it's an enhancement we can make later down the line if requested/it's a common security requirement or implemented in the other PR.

charts/spark-operator-chart/values.yaml
```diff
@@ -156,10 +156,14 @@ controller:
     #   memory: 300Mi
 
   # -- Security context for controller containers.
-  securityContext: {}
-    # runAsUser: 1000
-    # runAsGroup: 2000
-    # fsGroup: 3000
+  securityContext:
+    runAsNonRoot: true
```
jacobsalway, line 160: Could add `allowPrivilegeEscalation: false` as well

ChenYi015: I did not add this as the default value of `allowPrivilegeEscalation` should be `false`.

ChenYi015: Done.

ChenYi015:
> Could push the user and group directives down to the Dockerfile rather than the container security context. I'm not sold on read-only FS because any temporary disk usage that isn't on a volume could cause a panic but it's an enhancement we can make later down the line if requested/it's a common security requirement or implemented in the other PR.

I have pushed the user and group directives down  to the Dockerfile. Agree on that we would better implement the readOnlyRootFileSystem in a dedicated PR.

charts/spark-operator-chart/values.yaml
ChenYi015: Have removed the `runAsUser` and `runAsGroup` and pushed them down to the Dockerfile.

jacobsalway (approved):
/lgtm

Works for me in local testing on Kind. As you said let's do `readOnlyRootFilesystem: true` in a separate PR.

cccsss01:
> > Might need to do some investigation to map out all the disk usage of spark-submit but what are your thoughts on also using an emptyDir volume in the controller and adding readOnlyRootFilesystem: true as well?
> 
> Yeah I feel like `readOnlyRootFilesystem` is a bit too restrictive especially while we are actively modifying the usage of the controller with pod templates
> 
> But like the PR #2219 propose we could mount a volume on /tmp and use `os.CreateTemp` everytime we need a local file

majority of these local files created don't need to be stored permanently correct?

PR #2266, opened 2024-10-19
Run e2e tests with Kubernetes version matrix

jacobsalway:
## Purpose of this PR

* Run e2e tests with Kubernetes version matrix containing n-3 of the current Kubernetes versions
## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -209,6 +209,9 @@ jobs:
 
   e2e-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        k8s_version: [v1.28.13, v1.29.8, v1.30.4, v1.31.1]
     steps:
       - name: Checkout source code
         uses: actions/checkout@v4
@@ -221,7 +224,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Create a Kind cluster
-        run: make kind-create-cluster
+        run: make kind-create-cluster KIND_K8S_VERSION=${{ matrix.k8s_version }}
 
       - name: Build and load image to Kind cluster
         run: |
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ LOCALBIN ?= $(shell pwd)/bin
 KUSTOMIZE_VERSION ?= v5.4.1
 CONTROLLER_TOOLS_VERSION ?= v0.15.0
 KIND_VERSION ?= v0.23.0
+KIND_K8S_VERSION ?= v1.29.3
 ENVTEST_VERSION ?= release-0.18
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION ?= 1.29.3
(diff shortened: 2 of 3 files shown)


.github/workflows/integration.yaml
```diff
@@ -209,6 +209,9 @@ jobs:
 
   e2e-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        k8s: [v1.28.13, v1.29.8, v1.30.4, v1.31.1]
```
jacobsalway, line 214: It doesn't look like Kind publishes image tags without the minor version e.g. `v1.31` so I had to grab the latest patch versions manually from https://hub.docker.com/r/kindest/node/tags. Is there a way to make this Dependabot compatible or should we simply bump all things that are related to Kubernetes versions manually on a schedule e.g. once a month?

Makefile
```diff
@@ -52,6 +52,7 @@ LOCALBIN ?= $(shell pwd)/bin
 KUSTOMIZE_VERSION ?= v5.4.1
 CONTROLLER_TOOLS_VERSION ?= v0.15.0
 KIND_VERSION ?= v0.23.0
+KIND_K8S_VERSION ?= v1.31.1
```
ChenYi015, line 55: Maybe we should change default k8s version to v1.29.3 as we use now.

.github/workflows/integration.yaml
ChenYi015, line 214: What about renaming `k8s` to `K8S_VERSION`?

ChenYi015: I think it's fine to specify the k8s version manually for now. I am not sure whether there is a a way to automatically update the k8s version matrix.

.pre-commit-config.yaml
```diff
@@ -1,10 +0,0 @@
-repos:
```
ChenYi015, line 1: I think we can keep this file, I often forget to `make helm-docs` manually, and it will help me do this.

.github/workflows/integration.yaml
jacobsalway: Yeah I couldn't find any examples of doing this automatically. Let's just do it manually.

.pre-commit-config.yaml
jacobsalway: Sure let's keep it

.github/workflows/integration.yaml
jacobsalway: Hmm I'm not sure if the naming of this matrix value matters too much here since it's going into `KIND_K8S_VERSION` anyway. I just took inspiration from https://github.com/apache/yunikorn-k8shim/blob/master/.github/workflows/pre-commit.yml#L50 when checking what other projects were doing

Makefile
jacobsalway: Do we need to keep `ENVTEST_K8S_VERSION`, `KIND_K8S_VERSION` and the Go client versions in sync? If there's no dependency I'd prefer to use the latest minor version i.e. 1.31

ChenYi015: Yes, I think we should keep the three k8s versions the same.

.github/workflows/integration.yaml
ChenYi015: Never mind, it's just a nit, I would prefer a more meaningful name.

jacobsalway: It's all good, meaningful variable names are a fair nit. Changing to `k8s_version`.

Makefile
jacobsalway: Changed to `1.29.3`. I reckon we could bump to the latest minor at some point soon though. I don't think there are any APIs that either Spark or the operator are using that would block us, unless there are any kubebuilder/controller-runtime dependency issues.

jacobsalway:
I'll document the official `n-3` (as governed by e2e test success) compatibility somewhere in a subsequent docs PR. I'd like to update the docs as mentioned during the community call.

ChenYi015 (approved):

PR #2312, opened 2024-11-05
Use NSS_WRAPPER_PASSWD instead of /etc/passwd as in spark-operator image entrypoint.sh

Aakcht:
## Purpose of this PR

Currently spark-operator entrypoint.sh has [this logic](https://github.com/kubeflow/spark-operator/blob/master/entrypoint.sh#L18). It is the same logic as in [entrypoint.sh of old spark images](https://github.com/apache/spark-docker/blob/master/3.3.0/scala2.12-java11-python3-ubuntu/entrypoint.sh#L30). It is intended for openshift and modifies /etc/passwd. It was working in the past, because in the older base spark images permissions for /etc/passwd [were modified](https://github.com/apache/spark-docker/blob/master/3.3.0/scala2.12-java11-python3-ubuntu/Dockerfile#L38).

New spark images [do not modify permissions of /etc/passwd](https://github.com/apache/spark-docker/blob/master/3.5.3/scala2.12-java11-ubuntu/Dockerfile). Therefore this entrypoint.sh logic does not work currently and when running spark-operator under random user spark-application submissions fails with the following error:
```
 org.apache.hadoop.security.KerberosAuthException: failure to login: javax.security.auth.login.LoginException: java.lang.NullPointerException: invalid null input: name
```

 This PR changes this logic to the [same logic as in newer spark images](https://github.com/apache/spark-docker/blob/master/3.5.3/scala2.12-java11-ubuntu/entrypoint.sh#L32).

**Proposed changes:**

- Modify entrypoint.sh logic of working with /etc/passwd and NSS_WRAPPER_PASSWD based on new spark image.

## Change Category

- [ ] Documentation update

## Checklist

### Additional Notes

I validated this PR by adding new entrypoint.sh to spark-operator docker image and checked that spark-operator can now submit applications when running under random user.


Aakcht:
Hi, @ChenYi015 @ImpSy @jacobsalway , any chance for reviewing this PR?

ChenYi015 (approved):

entrypoint.sh
```diff
@@ -4,21 +4,23 @@
...
+# It's to resolve OpenShift random UID case.
+# See also: https://github.com/docker-library/postgres/pull/448
+if ! getent passwd "$myuid" &> /dev/null; then
+    for wrapper in {/usr,}/lib{/*,}/libnss_wrapper.so; do
+      if [ -s "$wrapper" ]; then
+        NSS_WRAPPER_PASSWD="$(mktemp)"
+        NSS_WRAPPER_GROUP="$(mktemp)"
+        export LD_PRELOAD="$wrapper" NSS_WRAPPER_PASSWD NSS_WRAPPER_GROUP
+        mygid="$(id -g)"
+        printf 'spark:x:%s:%s:${SPARK_USER_NAME:-anonymous uid}:%s:/bin/false\n' "$myuid" "$mygid" "$SPARK_HOME" > "$NSS_WRAPPER_PASSWD"
```
ExNG, line 19: I think this line was supposed to look something like this:
```sh
printf 'spark:x:%s:%s:%s:%s:/bin/false\n' "$myuid" "$mygid" "${SPARK_USER_NAME:-anonymous uid}" "$SPARK_HOME" > "$NSS_WRAPPER_PASSWD"
```

Because currently the nss wrapper passwd looks something like:
```
spark:x:1000:1000:${SPARK_USER_NAME:-anonymous uid}:/foobar:/bin/false

```

the single quotes around the printf template prevent bash from interpolating $SPARK_USER_NAME

PR #2305, opened 2024-11-04
Fix: should not add emptyDir sizeLimit conf if it is nil

ChenYi015:
## Purpose of this PR

Close #2199 

**Proposed changes:**

- When running spark-submit, the emptyDir sizeLimit conf should not be added if sizeLimit is nil.

## Change Category

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

## Checklist

- [x] I have conducted a self-review of my own code.
- [ ] I have updated documentation accordingly.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [ ] Existing unit tests pass locally with my changes.

Issue #2199, opened 2024-09-27
[BUG] Pod creation creation fails on submission with invalid resource quantities


Cian911:
## Description

I've been scratching my head on this one for the past few days - without any resolution.

I am in the process of testing migrating the spark operator from `spark-operator-chart-1.4.6` to `v2.0.1` and have come across the following issues. It seems that submission fails at the point it tries to create a driver pod - with the following error around resource quantities:

```
Failure executing: POST at: https://127.0.01:443/api/v1/namespaces/spark-operator/pods.
      Message: Pod in version \"v1\" cannot be handled as a Pod: quantities must match
      the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'.
```

Below is the full error log.

```
status:
  applicationState:
    errorMessage: "failed to run spark-submit: failed to run spark-submit: 24/09/27
      14:55:49 WARN NativeCodeLoader: Unable to load native-hadoop library for your
      platform... using builtin-java classes where applicable\n24/09/27 14:55:49 INFO
      SparkKubernetesClientFactory: Auto-configuring K8S client using current context
      from users K8S config file\n24/09/27 14:55:50 INFO KerberosConfDriverFeatureStep:
      You have not specified a krb5.conf file locally or via a ConfigMap. Make sure
      that you have the krb5.conf locally on the driver image.\n24/09/27 14:55:50
      ERROR Client: Please check \"kubectl auth can-i create pod\" first. It should
      be yes.\nException in thread \"main\" io.fabric8.kubernetes.client.KubernetesClientException:
      Failure executing: POST at: https://127.0.01:443/api/v1/namespaces/spark-operator/pods.
      Message: Pod in version \"v1\" cannot be handled as a Pod: quantities must match
      the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'. Received
      status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=Pod
      in version \"v1\" cannot be handled as a Pod: quantities must match the regular
      expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$', metadata=ListMeta(_continue=null,
      remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}),
[...]

PR #2316, opened 2024-11-10
Fix: should not add emptyDir sizeLimit conf on executor pods if it is nil

Cian911:
## Purpose of this PR

**Proposed changes:**

This is a follow up fix from work already done in https://github.com/kubeflow/spark-operator/pull/2305 by @ChenYi015 

This PR adds a check at submission on executors pods if an emptyDir `sizeLimit` is `nil`.  

## Change Category

- [ ] Documentation update

### Rationale

Following some further testing of an older `SparkApplication` template wherein the volumeName (see issue https://github.com/kubeflow/spark-operator/issues/2199 for more details) had not been changed, I noticed behaviour wherein only the driver pod would get created and no executor pods would spawn. I believe this is due to the same issue outlined in https://github.com/kubeflow/spark-operator/pull/2305 which would make sense, as the check was only added on the driver pod submission, and not the executors.

## Checklist

### DIFF
--- a/internal/controller/sparkapplication/submission.go
+++ b/internal/controller/sparkapplication/submission.go
@@ -840,20 +840,22 @@ func executorVolumeMountsOption(app *v1beta2.SparkApplication) ([]string, error)
 		}
 		switch volumeType {
 		case common.VolumeTypeEmptyDir:
-			args = append(
-				args,
-				"--conf",
-				fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


ChenYi015 (approved):

ChenYi015:
@Cian911 Thanks for fixing this.
/lgtm

PR #2326, opened 2024-11-14
feat: manage `PodDisruptionBudget` for `SparkApplication` driver and executor

a7i:
## Purpose of this PR

Provide the ability to create PodDisruptionBudget per Spark Application

**Proposed changes:**

- Expose `PodDisruptionBudgetSpec` for driver definition
- - Expose `PodDisruptionBudgetSpec` for executor definition

## Change Category

- [ ] Documentation update

### Rationale

Our spark pipelines cannot be interrupted and during node drain, we want to prevent eviction of executor and driver pods. Once the pipeline is complete, then the node can be drained. This is natively supported via PodDisruptionBudget with `maxUnavailable: 0`

## Checklist


missedone:
@a7i , I'm wondering if we can use the new feature of the pod template to specify the PDB per PR https://github.com/kubeflow/spark-operator/pull/2141

missedone:
Ah right, it’s PDB which need a specific configuration item for it. My brain was stuck :(

jacobsalway:
Happy to take a look at the PR as well because this may be a useful feature, but if this is specifically for node draining, you could add annotations to prevent eviction: `karpenter.sh/do-not-disrupt: "true"` for Karpenter and `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"` for cluster-autoscaler.

a7i:
> "cluster-autoscaler.kubernetes.io/safe-to-evict"

Thanks! The context here is nodegroup upgrades, so we drain the nodes from the old nodegroup. so karpenter or cluster-autoscaler don't come into play.

Cian911:
This would be a great addition - thanks @a7i

jacobsalway:
> > "cluster-autoscaler.kubernetes.io/safe-to-evict"
> 
> Thanks! The context here is nodegroup upgrades, so we drain the nodes from the old nodegroup. so karpenter or cluster-autoscaler don't come into play.

Ah, so this is a user initiated drain and not one done automatically by either node provisioner e.g. Karpenter drift detection or node consolidation. In that case definitely understand your issue with needing to provision PDBs.

Will review this tonight.

a7i:
> Will review this tonight.

@jacobsalway let me know if there's interest and I can rebase

Issue #2268, opened 2024-10-20
[BUG] Generated clientsets/informers aren't updated


jacobsalway:
## Description

* The generated code under https://github.com/kubeflow/spark-operator/tree/master/pkg/client hasn't been updated in a while
* There is no CI check to ensure the generated code has been run if the API has been changed
* There doesn't seem to be a Makefile target to do this either. There are some scripts under `hack/` that look like they were used to run the codegen in the past that could be updated.

- [X] ✋ I have searched the open/closed issues and my issue is not listed

jacobsalway:
Closing because generated clientsets doesn't seem to be recommended or needed anymore. Will raise a PR to remove the `hack/` bash scripts from the repo at some point

If relevant in the future, to get the old setup working you need a `tools.go` style file to add the code-generator dependency to `go.mod`.

```go
//go:build tools

package tools

import (
    _ "k8s.io/code-generator"
)
```

References:
- https://github.com/kubernetes-sigs/kubebuilder/issues/1152#issuecomment-2466556889
- https://geek.jasonhancock.com/2023/07/28/k8s-crd-golang-kubebuilder/

Issue #2267, opened 2024-10-20
[BUG] Not possible to reference module by release tags


jacobsalway:
## Description

It's not possible to reference `github.com/kubeflow/spark-operator` by release tag in a Go module.

- [X] ✋ I have searched the open/closed issues and my issue is not listed.

## Reproduction Code [Required]

```
go mod init test
go get github.com/kubeflow/spark-operator
```

## Expected behavior

It should be possible reference this module via the release tags e.g. `require github.com/kubeflow/spark-operator/v2 v2.0.2`.

## Actual behavior

The module is downloaded with reference to a [pseudo-version](https://go.dev/ref/mod#pseudo-versions ) e.g. `require github.com/kubeflow/spark-operator v0.0.0-20241018113320-1509b341d4e6`. 

If you change the tag in `go.mod` and run `go mod tidy` you get this error:

```
go: test imports
	github.com/kubeflow/spark-operator/api/v1beta2: github.com/kubeflow/spark-operator/v2@v2.0.2: go.mod has non-.../v2 module path "github.com/kubeflow/spark-operator" (and .../v2/go.mod does not exist) at revision v2.0.2
```

ChenYi015:
Since `v2.2.0`, one can reference spark-operator as a go module by release tag:

```
go mod init test

# Use the latest version
go get github.com/kubeflow/spark-operator/v2@latest

# Use a specified version
go get github.com/kubeflow/spark-operator/v2@v2.2.1
```

PR #2368, opened 2024-12-23
Refactor sparkctl

ChenYi015:
## Purpose of this PR

**Proposed changes:**

Refactor sparkctl:

- Use controller-runtime client instead of generated crd client to interact with API server
- `sparkctl status` sub command is moved to `sparkctl get` so that it be consistent with `kubectl get`
- Adjust the sparkctl directory structure as follows:

```
$ tree -L 2 cmd/sparkctl
cmd/sparkctl
├── README.md
├── create
│   ├── create.go
│   ├── create_test.go
│   ├── gcs.go
│   ├── s3.go
│   └── testdata
├── delete
│   └── delete.go
├── event
│   └── event.go
├── forward
│   └── forward.go
├── get
│   └── get.go
├── list
│   └── list.go
├── log
│   └── log.go
└── main.go
```


jacobsalway:
Apologies for late review. What are your thoughts on removing `sparkctl` altogether? I think most of its functionality can already be achieved with `kubectl`, except for the job/dependency staging and creating a `SparkApplication` from an existing `ScheduledSparkApplication`. 

I personally haven't ever used `sparkctl` in nearly two years of running the operator in production and haven't used any custom CLIs for any other Kubernetes operators either. 

As for alternatives to `create` functionality: 

* Usually I've packaged my Spark jobs as pre-built container images, so I'm open to opinions on the job staging functionality. 
* For triggering a run from an existing scheduled application, one approach that comes to mind is adding an annotation [similarly to how Strimzi supports restarting Kafka connectors](https://strimzi.io/docs/operators/0.22.1/using#proc-manual-restart-connector-str).

ChenYi015:
Actually, I do not use `sparkctl` at all, it is totally fine with me to remove it altogether, but I am not sure whether or not anyone is using it. I refactored this binary using the controller-runtime client so that we can remove the generated clientsets, informers, and listers later.

vara-bonthu:
I agree with @jacobsalway  sparkctl might not be necessary, as users typically deploy jobs using kubectl or through schedulers. We can revisit this topic in our next meeting to decide whether to keep it or remove it. Removing it could reduce maintenance efforts and simplify the codebase.

jacobsalway:
Here are some actions I think we can take or discuss at the next meeting:
1. Consider removing `sparkctl` given its overlap with `kubectl` and to simplify the codebase.
2. Allow users to add a `k8s.sparkoperator.io/trigger=true` annotation to a `ScheduledSparkApplication` to replicate the functionality that would be lost from the removal of `sparkctl` that doesn't overlap with `kubectl`.
3. Fix the repo semantic versioning and fix the informer/lister/watcher code generation. I'd actually be against removing the generated code since I'm actually using it for some internal usecases, though with an older version of the CRs. https://github.com/kubeflow/spark-operator/issues/2267 and https://github.com/kubeflow/spark-operator/issues/2268.

jacobsalway:
On a side note the clientset code gen isn't relevant anymore https://github.com/kubeflow/spark-operator/issues/2268

vikas-saxena02:
I would say remove sparkctl alltogather. Happy to do that if you guys need help on that @ChenYi015  please let me know if you need someone to action it.

ChenYi015:
Close this for we have decided to deprecate `sparkctl` in future releases.

PR #2387, opened 2025-01-17
Suspend/Resume feature on SparkApplication

### LINKED ISSUE #2290: Add support for suspend

- Add a new field `.spec.suspend` to the SparkApplication CRD.
- The controller will not submit the SparkApplication except for the `.spec.suspend` is false.

### Why is this needed?

_No response_

_No response_

_No response_

### Additional context

_No response_

### Love this feature?

everpeace:
## Purpose of this PR

This PR implements Suspend/Resume operation on SparkApplication.

**Proposed changes:**

- Add `Suspend: true|false` field in `SparkApplicationSpec`
- State Transition:
  - Suspend: `**` --`Suspend: true`--> `SUSPENDING` --> `SUSPENDED` (`**`=_any non-Terminal Application State_)
  - Resume: `SUSPENDING | SUSPENDED` --`Suspend: false`--> `RESUMING` --> `SUBMITTED`

Fixes #2290 

## Change Category

- [ ] Documentation update

### Rationale

Supporting Suspend/Resume feature could make Kueue integration easier.

## Checklist


Kevinz857:
@everpeace Maybe u can resolve this conflict so that this PR can continue to move forward？

we also think about use kueue with sparkoperator , depend this suspend state feature . It would be better if you could push further down ~

everpeace:
@Kevinz857 Hi, thanks for your comment!! I rebaed my PR.

ChenYi015:
@Kevinz857 Would you mind help reviewing the PR?

Kevinz857:
> @Kevinz857 Would you mind help reviewing the PR?

@ChenYi015  Very happy, I will review this PR together in detail later.

everpeace:
@Kevinz857 Thanks for your review.

> Consider extracting the suspension check (if app.Spec.Suspend != nil && *app.Spec.Suspend) into a helper function like isSuspended(app) for clarity and potential reuse.

hmmm. `SparkApplication.Spec.Suspend` is not a pointer. So, I can't see any motivation for this.
https://github.com/kubeflow/spark-operator/blob/4d6dd1d2c6267e54e6518635b96e8919a49d813f/api/v1beta2/sparkapplication_types.go#L44

Perhaps, you saw `ScheduledSparkApplication`'s suspension check: 
https://github.com/kubeflow/spark-operator/blob/4d6dd1d2c6267e54e6518635b96e8919a49d813f/internal/controller/scheduledsparkapplication/controller.go#L109-L112

Could you elaborate on your suggestion if I missed something?

>  Also, using a named constant for the event string ("SparkApplicationSuspended") would improve consistency and readability.

hmmm. I think I already did. 
https://github.com/kubeflow/spark-operator/blob/4d6dd1d2c6267e54e6518635b96e8919a49d813f/pkg/common/event.go#L33-L38

Could you elaborate on your suggestion if I missed something?

> Field documentation enhancement
> In sparkapplication_types.go, the Suspend field comment could be more descriptive, such as:

Thanks. Fixed in [7926292](https://github.com/kubeflow/spark-operator/pull/2387/commits/7926292d9c5dcff8d0e93810079d2bc464920902)

> Test enhancement ideas
> Add unit tests for the suspension logic to ensure the controller skips job submission when .spec.suspend is true.

I think it's already covered:
https://github.com/kubeflow/spark-operator/blob/4d6dd1d2c6267e54e6518635b96e8919a49d813f/internal/controller/sparkapplication/controller_test.go#L579-L584

> Consider validating whether the correct Kubernetes events (Suspended/Resumed) are emitted using a fake event recorder.

IIUC, the current `SparkApplication` unit tests(`internal/controller/sparkapplication/controller_test.go`) do not assert emitted events at all. Although I fully agree your suggestion(we should assert emitted events in unit tests), I think the enhancement should be done in another PR so that all the unit tests assert emitted events. Or, should this PR cover event assertions only for suspend/resume in this PR?? If so, I would be happy to handle it.

andreyvelich:
/remove-lifecycle stale
@vara-bonthu @ChenYi015 @nabuskey I think, that would be a great feature in Spark Operator APIs to natively integrate it with Kueue!

cc @kannon92 @tenzen-y @mimowo

mimowo:
> @vara-bonthu @ChenYi015 @nabuskey I think, that would be a great feature in Spark Operator APIs to natively integrate it with Kueue!

Awesome effort! We had multiple people asking for the integration, so I'm happy to see it rolling.

> cc @kannon92 @tenzen-y @mimowo

Overall LGTM, let me comment here on the API itself, and I may also leave some small-ish comments to the code.

> hmmm. SparkApplicationSpec.Suspend is not a pointer. So, I can't see any motivation for it.

I don't remember / wasn't involved with the decision when introducing the field to core k8s as a ptr. 

I found this question from @soltysh in the k8s design, but there was no explicit answer: https://github.com/kubernetes/enhancements/pull/2234#discussion_r567014403, just the confirmation of the decision to use ptr in the [comment](https://github.com/kubernetes/enhancements/pull/2234#pullrequestreview-579533841).

Without knowing exact details, I think this is a general practice in k8s APIs to introduce new fields as pointers. I think it is because of the client-go compatibility. So, old API clients may fail to parse the API output after the field is introduced, even if it is not used in the cluster. Users may want to use the old clients because technically the API remains the v1beta2 version.

Maybe @deads2k could validate that statement to keep me honest.

mimowo:
@everpeace please rebase

test/e2e/sparkapplication_test.go
```diff
@@ -393,4 +395,86 @@ var _ = Describe("Example SparkApplication", func() {
...
+			Expect(k8sClient.Get(ctx, key, app)).To(Succeed())
+			app.Spec.Suspend = ptr.To(false)
+			Expect(k8sClient.Update(ctx, app)).To(Succeed())
+
+			By("Waiting for SparkApplication to Running")
+			Eventually(func(g Gomega) {
+				app := &v1beta2.SparkApplication{}
+				g.Expect(k8sClient.Get(ctx, key, app)).To(Succeed())
+				g.Expect(app.Status.AppState.State).To(Equal(v1beta2.ApplicationStateRunning))
+			}).WithTimeout(3 * time.Minute).Should(Succeed())
```
mimowo, line 469: How long it is typically in the Running state? Asking because this check is intrinsically flaky - if the app computes very quickly you may not notice the "Running" state, just "Completed" immediately. Note that "Eventually" performs periodic checks. Still, if the "running" lasts for >3s then it is very unlikely to flaky.

internal/controller/sparkapplication/controller.go
```diff
@@ -695,6 +724,146 @@ func (r *Reconciler) reconcileUnknownSparkApplication(ctx context.Context, req c
...
+				return err
+			}
+			if old.Status.AppState.State != v1beta2.ApplicationStateResuming {
+				return nil
+			}
+			app := old.DeepCopy()
+
+			r.recordSparkApplicationEvent(app)
+
+			_ = r.submitSparkApplication(ctx, app)
```
mimowo: Don't you need to fail on error here and retry? 

If the handling of error is not necessary, it is probably worth adding a comment why, because it is far from obvious.

test/e2e/sparkapplication_test.go
everpeace: > How long it is typically in the Running state?

Probably, a minute or so?? This test case uses this one: https://github.com/kubeflow/spark-operator/blob/43bcea28ad810c32379d772568ed9ef3d77b1d80/examples/spark-pi.yaml

> Asking because this check is intrinsically flaky - if the app computes very quickly you may not notice the "Running" state, just "Completed" immediately

So, I think this chance could be low.

internal/controller/sparkapplication/controller.go
everpeace: Good catch, I noticed this pattern is used in several places in the controller. Let me inspect this.

test/e2e/sparkapplication_test.go
mimowo: Sure, got it, then I'm fine keeping this check. Feel free to resolve the comment.

internal/controller/sparkapplication/controller.go
everpeace: The returned error of `r.submitSparkApplication` was not used anywhere in the controller. Instead, the submission result(success/failure) was recorded in `app.Status`. So I just removed the returned error from the method. fixed in https://github.com/kubeflow/spark-operator/pull/2387/commits/f4719c1c0f5a776c9317380b1f3f8a7eea317dc0

everpeace:
> Without knowing exact details, I think this is a general practice in k8s APIs to introduce new fields as pointers. I think it is because of the client-go compatibility. So, old API clients may fail to parse the API output after the field is introduced, even if it is not used in the cluster. Users may want to use the old clients because technically the API remains the v1beta2 version.

It makes sense. I changed suspend field as pointer in https://github.com/kubeflow/spark-operator/pull/2387/commits/43bcea28ad810c32379d772568ed9ef3d77b1d80.

everpeace:
@mimowo all your comments are addressed. PTAL 🙇

internal/controller/sparkapplication/controller.go
mimowo: I see, btw, my question was mostly exploratory, so if the repo maintainers prefer the old style, then I'm also fine with that.

mimowo:
LGTM, thank you 👍

soltysh:
> I found this question from @soltysh in the k8s design, but there was no explicit answer: [kubernetes/enhancements#2234 (comment)](https://github.com/kubernetes/enhancements/pull/2234#discussion_r567014403), just the confirmation of the decision to use ptr in the [comment](https://github.com/kubernetes/enhancements/pull/2234#pullrequestreview-579533841).
> 
> Without knowing exact details, I think this is a general practice in k8s APIs to introduce new fields as pointers. I think it is because of the client-go compatibility. So, old API clients may fail to parse the API output after the field is introduced, even if it is not used in the cluster. Users may want to use the old clients because technically the API remains the v1beta2 version.
> 
> Maybe @deads2k could validate that statement to keep me honest.

Just to clarify, since my name was called out the reason for going with `*bool` is two folded:
1. The ability to differentiate from the value being set and not.
2. Most importantly, the [k8s API guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) force all optional fields to be pointers.

Hope that helps.

mimowo:
LGTM, @andreyvelich what are the next steps for merging / releasing this enhancement?

andreyvelich:
/assign @nabuskey @vara-bonthu @ChenYi015 @yuchaoran2011 Please can you help with reviewing this PR?

We see a lot of interest of native integration between Kueue and SparkApplication CRD.

mimowo:
Hi folks @andreyvelich @nabuskey @vara-bonthu @ChenYi015 @yuchaoran2011, friendly ping.

tenzen-y:
@andreyvelich @everpeace @mimowo tnak you for driving this!

PR #2408, opened 2025-01-28
implement RuntimeClassName in SparkPodSpec

VaibhavBeriwala:
Added support to set RuntimeClassName in SparkPodSpec.

This is useful when we want to use a different Kubernetes RuntimeClassName depending on whether the Spark pods would be running on CPU vs GPU VMs.


ChenYi015:
@VaibhavBeriwala Thank you for your contribution! I think maybe the RuntimeClassName can be specified with the [pod template feature](https://github.com/kubeflow/spark-operator/pull/2141)? If that works, we would better avoid introducing a new field to the CRD.

VaibhavBeriwala:
@ChenYi015 Thanks for the review.
Yes, I think you are right. It's better to leverage the pod template feature rather than adding a new field to the CRD.

PR #2416, opened 2025-02-01
Support Kubernetes 1.32

jacobsalway:
## Purpose of this PR

**Proposed changes:**

- <Change 1>
- <Change 2>
- <Change 3>

## Change Category

- [ ] Documentation update

## Checklist


.golangci.yaml
```diff
@@ -1,7 +1,7 @@
 run:
   # Timeout for analysis, e.g. 30s, 5m.
   # Default: 1m
-  timeout: 1m
+  timeout: 2m
```
jacobsalway, line 4: Only takes 22s to execute on my local but was timing out in CI https://github.com/kubeflow/spark-operator/actions/runs/13088329194/job/36522115615?pr=2416

vara-bonthu (approved):
/approve

Does this mean users with older versions of K8s (1.28) needs to stick with old tags when this is merged? 

Have all tests been executed with these updates, and are there any known issues or regressions identified?

jacobsalway:
> Does this mean users with older versions of K8s (1.28) needs to stick with old tags when this is merged?

I don't think so. We might only officially support n-3 (which a lot of other open source projects in the Kubernetes ecosystem officially support) or whatever's in the end-to-end test matrix, but we're using stable APIs that `k8s.io/client-go` should support even with differing/older control plane versions (also see https://github.com/kubernetes/client-go?tab=readme-ov-file#compatibility-matrix)

I've expanded the end-to-end test matrix to include all the way down to 1.24 as is currently supported by Spark 3.5.4 https://spark.apache.org/docs/3.5.4/running-on-kubernetes.html#prerequisites.

> Have all tests been executed with these updates, and are there any known issues or regressions identified?

The end-to-end tests have run successfully, and I've manually tested all the examples and haven't found any issues so far. Before we tag a new release, we can come up with some other manual test scenarios to ensure no issues.

jacobsalway:
Bumping for review @ChenYi015

PR #2423, opened 2025-02-07
Add option for disabling leader election

### LINKED ISSUE #2384: Option to disable leader election in the controller

Hello,

Would it be possible to add a way to disable the leader election in the controller deployment? 

### Why is this needed?

It seems that leader election is set, even with a replica 1. It uses leases rights, and theses rights are not allowed in our Kubernetes for securities reasons. We'd like to be able to disable the Leader election

Add:
{{- if .Values.controller.leaderElection.enable }}
{{- end }}
In the Controller deployment to be able to disable the leader elections in the values.yaml

We could also use a value for "--leader-election=true"

### Additional context

_No response_

### Love this feature?

ChenYi015:
## Purpose of this PR

Close #2384

**Proposed changes:**

- Add option `controller.leaderElection.enable` and `webhook.leaderElection.enable` for disabling controller/webhook leader election
- When leader election was disabled, the related RBAC rules against leases will be removed.

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -84,6 +84,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | image.pullPolicy | string | `"IfNotPresent"` | Image pull policy. |
 | image.pullSecrets | list | `[]` | Image pull secrets for private image registry. |
 | controller.replicas | int | `1` | Number of replicas of controller. |
+| controller.leaderElection.enable | bool | `true` | Specifies whether to enable leader election for controller. |
 | controller.workers | int | `10` | Reconcile concurrency, higher values might increase memory usage. |
 | controller.logLevel | string | `"info"` | Configure the verbosity of logging, can be one of `debug`, `info`, `error`. |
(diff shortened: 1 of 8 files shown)


jacobsalway (commented):
Does it make sense to have leader election on the webhook at all? It's a mostly stateless service that can have multiple replicas. The only part that might be relevant to have leader election on is the TLS certificate generation/setting the secret, but this only needs to be done once and exponentially retries on conflict or error anyway so I think it's fine if multiple replicas attempt to do this.

The core of this PR looks good to me though. Happy to tick and add a separate PR for removing leader election.

ChenYi015:
For the TLS certificate generation process, I think it can work fine even without leader election. 

> Does it make sense to have leader election on the webhook at all? It's a mostly stateless service that can have multiple replicas. 

Multiple webhook server replicas can work at the same time with/without leader election. But for the controller part, 
two reconcilers are set respectively in the manager for patching caBundles of mutating/validating webhooks. I have not tested whether it can work as expected without leader election if there are multiple replicas. I think we would better keep it as it is.

jacobsalway:
> I have not tested whether it can work as expected without leader election if there are multiple replicas.

I'll have a look at this myself. In the mean time this PR doesn't introduce any breaking changes so happy to approve

PR #2441, opened 2025-02-14
Support non-standard Spark container names

jbhalodia-slack:
## Purpose of this PR

**Proposed changes:**
- If using a non-standard container name for Spark driver or executor container, assume that the first container in the list is the Spark container, which is what [Spark assumes](https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template). See **Rationale** section for further explanation.

## Change Category

- [ ] Documentation update

### Rationale

Current implementation only supports standard Spark container names. Driver container name must be `spark-kubernetes-driver` and executor container name must be `spark-kubernetes-executor` or `executor`. However, Pod template files can define multiple containers, and you can use the spark properties spark.kubernetes.driver.podTemplateContainerName and spark.kubernetes.executor.podTemplateContainerName to indicate which container should be used as a basis for the driver or executor. If not specified, or if the container name is not valid, Spark will assume that the first container in the list will be the driver or executor container [[Source](https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template)].

`podTemplateContainerName` config could be specified via `Spec.SparkConf` or in `spark-defaults.conf`, and in the webhook it's hard to determine if `podTemplateContainerName` is set via any one of the possible Spark configuration methods, so in the Webhook it is better to assume that if standard Spark container name is not found, the first container in the list is the driver or executor container.

## Checklist

### DIFF
--- a/internal/webhook/sparkpod_defaulter.go
+++ b/internal/webhook/sparkpod_defaulter.go
@@ -714,6 +714,7 @@ func findContainer(pod *corev1.Pod) int {
 		candidateContainerNames = append(candidateContainerNames, common.SparkExecutorContainerName, common.Spark3DefaultExecutorContainerName)
 	}
 
+	// return -1 if the pod is not a Driver or an Executor pod
 	if len(candidateContainerNames) == 0 {
 		return -1
 	}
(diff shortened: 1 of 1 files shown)


jbhalodia-slack:
@ChenYi015 @jacobsalway @vara-bonthu Can I get a PR review?

ChenYi015 (approved):

vara-bonthu (approved):
/lgtm

PR #2511, opened 2025-04-24
Add ShuffleTrackingEnabled to DynamicAllocation struct to allow disabling shuffle tracking

jbhalodia-slack:
## Purpose of this PR

Add ShuffleTrackingEnabled *bool to DynamicAllocation so that shuffle tracking can be disabled when using an external shuffle service. Set default for ShuffleTrackingEnabled to true if dynamicAllocation.enabled is true so that it's not a breaking change.

**Proposed changes:**

- Add ShuffleTrackingEnabled *bool to DynamicAllocation struct
- Set default for ShuffleTrackingEnabled to true if dynamicAllocation.enabled is true
- Set `spark.dynamicAllocation.shuffleTracking.enabled` spark conf based on ShuffleTrackingEnabled value

## Change Category

- [ ] Documentation update

### Rationale

Currently if dynamicAllocation is enabled, shuffle tracking is enabled by default. When using an external shuffle service,
we want dynamicAllocation to be enabled and want shuffle tracking to be disabled. So adding a variable to allow users to 
disable shuffle tracking is helpful when dynamicAllocation is enabled. By default shuffle tracking will be enabled so that 
it doesn't break existing use cases and its not a breaking change.

## Checklist

### DIFF
--- a/internal/controller/sparkapplication/submission.go
+++ b/internal/controller/sparkapplication/submission.go
@@ -1005,8 +1005,8 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
 	} else {
-	  	args = append(args, "--conf",
-		    fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
 	}
 	if dynamicAllocation.ShuffleTrackingTimeout != nil {
 		args = append(args, "--conf",
--- a/internal/controller/sparkapplication/submission.go
+++ b/internal/controller/sparkapplication/submission.go
@@ -1001,13 +1001,12 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationMaxExecutors, *dynamicAllocation.MaxExecutors))
 	}
+	shuffleTrackingEnabled := true
 	if dynamicAllocation.ShuffleTrackingEnabled != nil {
-		args = append(args, "--conf",
-			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
-	} else {
-		args = append(args, "--conf",
-			fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
+		shuffleTrackingEnabled = *dynamicAllocation.ShuffleTrackingEnabled
 	}
+	args = append(args, "--conf",
+		fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, shuffleTrackingEnabled))
 	if dynamicAllocation.ShuffleTrackingTimeout != nil {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationShuffleTrackingTimeout, *dynamicAllocation.ShuffleTrackingTimeout))
--- a/internal/controller/sparkapplication/submission.go
+++ b/internal/controller/sparkapplication/submission.go
@@ -989,10 +989,6 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 	args = append(args, "--conf",
 		fmt.Sprintf("%s=true", common.SparkDynamicAllocationEnabled))
 
-	// Turn on shuffle tracking if dynamic allocation is enabled.
-	args = append(args, "--conf",
-		fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
-
 	if dynamicAllocation.InitialExecutors != nil {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationInitialExecutors, *dynamicAllocation.InitialExecutors))
@@ -1004,6 +1000,10 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 	if dynamicAllocation.MaxExecutors != nil {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationMaxExecutors, *dynamicAllocation.MaxExecutors))
+	}
+	if dynamicAllocation.ShuffleTrackingEnabled != nil {
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
 	}
 	if dynamicAllocation.ShuffleTrackingTimeout != nil {
 		args = append(args, "--conf",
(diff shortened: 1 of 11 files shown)


internal/controller/sparkapplication/submission.go
```diff
@@ -1005,6 +1001,10 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationMaxExecutors, *dynamicAllocation.MaxExecutors))
 	}
+	if dynamicAllocation.ShuffleTrackingEnabled != nil {
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
+	}
```
ChenYi015, line 1007:
```suggestion
	if dynamicAllocation.ShuffleTrackingEnabled != nil {
		args = append(args, "--conf",
			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
	} else {
	  	args = append(args, "--conf",
		    fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
	}
```

Since we are going to deprecate the webhook in the future, it would be better if it can work without the webhook.

ChenYi015 (commented):
Thanks for your contribution! I have left some comments.

jbhalodia-slack:
@ChenYi015 Thank you, I've committed your edit. 🙇

ChenYi015:
@jbhalodia-slack The DCO check has failed, please sign off your commits with `git commit -s -m <message>`, then this PR will be merged.

jbhalodia-slack:
> @jbhalodia-slack The DCO check has failed, please sign off your commits with `git commit -s -m <message>`, then this PR will be merged.

@ChenYi015 Sorry about that, I've updated the commit message with the signoff string, thank you

internal/webhook/sparkapplication_defaulter.go
```diff
@@ -98,5 +98,9 @@ func defaultExecutorSpec(app *v1beta2.SparkApplication) {
 				app.Spec.Executor.Instances = util.Int32Ptr(1)
 			}
 		}
+		// Set default for ShuffleTrackingEnabled to true if dynamicAllocation.enabled is true.
+		if enableDynamicAllocation && app.Spec.DynamicAllocation.ShuffleTrackingEnabled == nil {
+			app.Spec.DynamicAllocation.ShuffleTrackingEnabled = util.BoolPtr(true)
+		}
```
ChenYi015, line 104: We should move this out of the `if app.Spec.Executor.Instances == nil` statement. Could you refactor the `defaultExecutorSpec` func as follows:

1. Check whether dynamic allocation is enabled.
2. If executor instances is not specified and dynamic allocation is disabled, then default executor instances to `1`.
3. If `ShuffleTrackingEnabled` is nil and dynamic allocation is enabled, then default it to `true`.

jbhalodia-slack: That makes sense, tho I have a question about setting executor.instances to 1. spark-submit sets the default for `--num-executors` to 2, so if `spark.executor.instance` is not set then I would expect `spark.executor.instance` to default to 2. So do we still want to set a default for `app.Spec.Executor.Instances`?

ChenYi015: > spark-submit sets the default for --num-executors to 2, so if spark.executor.instance is not set then I would expect spark.executor.instance to default to 2. 

Make sense. We would better to keep the default behavior to be consistent with spark-submit. Let us remove the default logic for `app.Spec.Executor.Instances` in another PR.

jbhalodia-slack: Ok sounds good, I've refactored the function, let me know if it needs further refactoring 🙇

internal/controller/sparkapplication/submission.go
```diff
@@ -1005,6 +1001,13 @@ func dynamicAllocationOption(app *v1beta2.SparkApplication) ([]string, error) {
 		args = append(args, "--conf",
 			fmt.Sprintf("%s=%d", common.SparkDynamicAllocationMaxExecutors, *dynamicAllocation.MaxExecutors))
 	}
+	if dynamicAllocation.ShuffleTrackingEnabled != nil {
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=%t", common.SparkDynamicAllocationShuffleTrackingEnabled, *dynamicAllocation.ShuffleTrackingEnabled))
+	} else {
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=true", common.SparkDynamicAllocationShuffleTrackingEnabled))
+	}
```
nabuskey, line 1010: nit: we could avoid this else by setting the default string value to true. It's just a styling preference imo. The function is nice right now with no else clause.

nabuskey (approved):
Great addition for sure. I love seeing support for RSS here.

internal/controller/sparkapplication/submission.go
jbhalodia-slack: @nabuskey thanks for the review, I've made the update

internal/webhook/sparkapplication_defaulter.go
```diff
@@ -86,17 +86,26 @@ func defaultDriverSpec(app *v1beta2.SparkApplication) {
...
-		if !enableDynamicAllocation && app.Spec.SparkConf != nil {
-			if dynamicConf, _ := strconv.ParseBool(app.Spec.SparkConf[common.SparkDynamicAllocationEnabled]); dynamicConf {
-				enableDynamicAllocation = true
-			}
-			if !enableDynamicAllocation && app.Spec.SparkConf[common.SparkExecutorInstances] == "" {
-				app.Spec.Executor.Instances = util.Int32Ptr(1)
-			}
+
+	isDynamicAllocationSpecEnabled := app.Spec.DynamicAllocation != nil && app.Spec.DynamicAllocation.Enabled
+
```
ChenYi015, line 91: I think we should check whether dynamic allocation is enabled from both application spec and Spark conf here. It would be better if we can define an util function like `IsDynamicAllocationEnabled` within file `pkg/util/sparkapplication.go`. 

Then we can use just one variable `isDynamicAllocationEnabled` to represent it, like `isDynamicAllocationEnabled := util.IsDynamicAllocationEnabled(app)`.

internal/webhook/sparkapplication_defaulter.go
```diff
@@ -86,17 +86,26 @@ func defaultDriverSpec(app *v1beta2.SparkApplication) {
...
+		executorInstancesFromConf := ""
+		if app.Spec.SparkConf != nil {
+			isDynamicAllocationConfEnabled, _ = strconv.ParseBool(app.Spec.SparkConf[common.SparkDynamicAllocationEnabled])
+			executorInstancesFromConf = app.Spec.SparkConf[common.SparkExecutorInstances]
+		}
+
+		if executorInstancesFromConf == "" && !isDynamicAllocationConfEnabled {
+			app.Spec.Executor.Instances = util.Int32Ptr(1)
 		}
 	}
```
ChenYi015, line 104:
```suggestion
	if app.Spec.Executor.Instances == nil &&
		app.Spec.SparkConf[common.SparkExecutorInstances] == "" &&
		!isDynamicAllocationEnabled {
		app.Spec.Executor.Instances = util.Int32Ptr(1)
	}
```

internal/webhook/sparkapplication_defaulter.go
jbhalodia-slack: @ChenYi015 Thanks for the suggestion, I've made the update

ChenYi015 (approved):

ChenYi015:
@jbhalodia-slack Thanks for this great contribution.
/lgtm
/approve

Issue #2460, opened 2025-03-05
Make UI ingress's `tls` and `annotations` configurable in the helm config


Tom-Newton:
### What feature you would like to be added?

The `SparkApplication` resource has `ingressAnnotations` and `ingressTLS` parameters for controlling the spark UI ingress that is created https://github.com/kubeflow/spark-operator/blob/master/docs/api-docs.md#sparkuiconfiguration. Meanwhile the ingress URL for the spark UI is configured by `controller.uiIngress.urlFormat` in the helm config https://github.com/kubeflow/spark-operator/tree/master/charts/spark-operator-chart#values.

It would be useful if we could also configure default  `ingressAnnotations` and `ingressTLS` in the helm config. 

### Why is this needed?

`controller.uiIngress.urlFormat` is constant across all `SparkApplication`s and its likely that ingress `tls` and `annotations` will be too. So it could be a nice convenience thing to have it configurable in the helm config. 

I also have a specific use-case where I'm using an integration for the `spark-operator` and I don't have much control over the creation of the `SparkApplication`, so I can't use the `ingressAnnotations` and `ingressTLS` that exist on this. For context the integration I'm using is https://docs.flyte.org/en/latest/flytesnacks/examples/k8s_spark_plugin/index.html

### Describe the solution you would like

New helm config options for default `annotations` and `tls`. 

### Describe alternatives you have considered

My specific problem with Flyte could be solved with a new feature in Flyte. I'm not really confident what the best solution is but I imagine this would be easier to implement and I think it could have some broader value. 

### Additional context
If there is some consensus that this is a reasonable thing to do then I might be able to contribute it. 

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

Tom-Newton:
I really want this feature, so I'm going to start working on it.

Tom-Newton:
I think the PR is ready for review: https://github.com/kubeflow/spark-operator/pull/2513

PR #2513, opened 2025-04-26
Make default ingress tls and annotations congurable in the helm config

Tom-Newton:
## Purpose of this PR
Support helm chart level defaults for ingress TLS and annotations, so they don't need to be configured on every SparkApplication. 
Closes: https://github.com/kubeflow/spark-operator/issues/2460

**Proposed changes:**

- Add CLI arguments `--ingress-tls` and `--ingress-annotations` and use them when TLS/annotations are not provided in the SparkApplication's spec. These are provided as JSON strings in the CLI. 
- Expose these via the helm config. The user configures them as normal in the helm yaml and they get serialised to JSON, so they can be read by the controller. 
- Wrote unit tests. 

## Change Category

- [ ] Documentation update

## Checklist

### Additional Notes
I considered adding an e2e test to cover the helm and CLI configuration. However, I couldn't see a way to run the e2e tests with multiple helm configs. I could have changed all the e2e tests to use these new configs, but then we loose test coverage for when the new configs are not used. 

I'm a golang noob.

### DIFF
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -73,10 +73,10 @@ spec:
         - --ingress-class-name={{ . }}
         {{- end }}
         {{- with .Values.controller.uiIngress.tls }}
-        - --ingress-tls={{ . | toJson | quote }}
+        - --ingress-tls={{ . | toJson }}
         {{- end }}
         {{- with .Values.controller.uiIngress.annotations }}
-        - --ingress-annotations={{ . | toJson | quote }}
+        - --ingress-annotations={{ . | toJson }}
         {{- end }}
         {{- end }}
         {{- if .Values.controller.batchScheduler.enable }}
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -72,6 +72,9 @@ spec:
         {{- with .Values.controller.uiIngress.ingressClassName }}
         - --ingress-class-name={{ . }}
         {{- end }}
+        {{- with .Values.controller.uiIngress.ingressTLS }}
+        - --ingress-tls={{ . | toJson }}
+        {{- end }}
         {{- end }}
         {{- if .Values.controller.batchScheduler.enable }}
         - --enable-batch-scheduler=true
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -72,8 +72,11 @@ spec:
         {{- with .Values.controller.uiIngress.ingressClassName }}
         - --ingress-class-name={{ . }}
         {{- end }}
-        {{- with .Values.controller.uiIngress.ingressTLS }}
-        - --ingress-tls={{ . | toJson }}
+        {{- with .Values.controller.uiIngress.defaultIngressTLS }}
+        - --default-ingress-tls={{ . | toJson }}
+        {{- end }}
+        {{- with .Values.controller.uiIngress.defaultIngressAnnotations }}
+        - --default-ingress-annotations={{ . | toJson }}
         {{- end }}
         {{- end }}
         {{- if .Values.controller.batchScheduler.enable }}
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -72,11 +72,11 @@ spec:
         {{- with .Values.controller.uiIngress.ingressClassName }}
         - --ingress-class-name={{ . }}
         {{- end }}
-        {{- with .Values.controller.uiIngress.defaultIngressTLS }}
-        - --default-ingress-tls={{ . | toJson }}
+        {{- with .Values.controller.uiIngress.tls }}
+        - --ingress-tls={{ . | toJson | quote }}
         {{- end }}
-        {{- with .Values.controller.uiIngress.defaultIngressAnnotations }}
-        - --default-ingress-annotations={{ . | toJson }}
+        {{- with .Values.controller.uiIngress.annotations }}
+        - --ingress-annotations={{ . | toJson | quote }}
         {{- end }}
         {{- end }}
         {{- if .Values.controller.batchScheduler.enable }}
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -74,10 +74,12 @@ controller:
     urlFormat: ""
     # -- Optionally set the ingressClassName.
     ingressClassName: ""
+    # -- Optionally set default TLS configuration for the Spark UI's ingress. `ingressTLS` in the SparkApplication spec overrides this.
     defaultIngressTLS: []
     # - hosts:
     #   - "*.example.com"
     #   secretName: "example-secret"
+    # -- Optionally set default ingress annotations for the Spark UI's ingress. `ingressAnnotations` in the SparkApplication spec overrides this.
     defaultIngressAnnotations: {}
       # key1: value1
       # key2: value2
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -75,12 +75,12 @@ controller:
     # -- Optionally set the ingressClassName.
     ingressClassName: ""
     # -- Optionally set default TLS configuration for the Spark UI's ingress. `ingressTLS` in the SparkApplication spec overrides this.
-    defaultIngressTLS: []
+    tls: []
     # - hosts:
     #   - "*.example.com"
     #   secretName: "example-secret"
     # -- Optionally set default ingress annotations for the Spark UI's ingress. `ingressAnnotations` in the SparkApplication spec overrides this.
-    defaultIngressAnnotations: {}
+    annotations: {}
       # key1: value1
       # key2: value2
 
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -74,6 +74,7 @@ controller:
     urlFormat: ""
     # -- Optionally set the ingressClassName.
     ingressClassName: ""
+    ingressTLS: []
 
   batchScheduler:
     # -- Specifies whether to enable batch scheduler for spark jobs scheduling.
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -74,7 +74,13 @@ controller:
     urlFormat: ""
     # -- Optionally set the ingressClassName.
     ingressClassName: ""
-    ingressTLS: []
+    defaultIngressTLS: []
+    # - hosts:
+    #   - "*.example.com"
+    #   secretName: "example-secret"
+    defaultIngressAnnotations: {}
+      # key1: value1
+      # key2: value2
 
   batchScheduler:
     # -- Specifies whether to enable batch scheduler for spark jobs scheduling.
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -138,7 +138,10 @@ func NewStartCommand() *cobra.Command {
 		},
 		PreRunE: func(_ *cobra.Command, args []string) error {
 			if ingressTLSstring != "" {
-				return json.Unmarshal([]byte(ingressTLSstring), &ingressTLS)
+				err := json.Unmarshal([]byte(ingressTLSstring), &ingressTLS)
+				if err != nil {
+					return err
+				}
 			}
 			return nil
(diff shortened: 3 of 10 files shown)


thetaiter-swf (approved):

Tom-Newton:
@ChenYi015 sorry for the direct ping, but you helped me get a couple of other things merged. I was hoping you could review this, when you get a chance.

nabuskey (commented):
I wonder if defining default TLS settings with JSON is the way to go.  Probably not a bad experience if using Helm but I don't really like to deal with escaping JSON strings in general.

cmd/operator/controller/start.go
```diff
@@ -128,12 +133,26 @@ func init() {
...
+				err := json.Unmarshal([]byte(defaultIngressTLSstring), &defaultIngressTLS)
+				if err != nil {
+					return err
+				}
+			}
+			if defaultIngressAnnotationsString != "" {
+				return json.Unmarshal([]byte(defaultIngressAnnotationsString), &defaultIngressAnnotations)
+			}
+			return nil
+		},
```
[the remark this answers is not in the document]
Tom-Newton: Good catch! I think your memory is correct 
```
	if c.PreRunE != nil {
		if err := c.PreRunE(c, argWoFlags); err != nil {
			return err
		}
	} else if c.PreRun != nil {
		c.PreRun(c, argWoFlags)
	}
```
I've consolidated into PreRunE

cmd/operator/controller/start.go
```diff
@@ -128,12 +133,26 @@ func init() {
...
 		},
+		PreRunE: func(_ *cobra.Command, args []string) error {
+			if defaultIngressTLSstring != "" {
+				err := json.Unmarshal([]byte(defaultIngressTLSstring), &defaultIngressTLS)
+				if err != nil {
+					return err
+				}
+			}
+			if defaultIngressAnnotationsString != "" {
+				return json.Unmarshal([]byte(defaultIngressAnnotationsString), &defaultIngressAnnotations)
```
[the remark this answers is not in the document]
Tom-Newton: Returned formatted error

Tom-Newton:
> I wonder if defining default TLS settings with JSON is the way to go. Probably not a bad experience if using Helm but I don't really like to deal with escaping JSON strings in general.

Thanks for reviewing. I think that is a fair assesment. The problem is I can't think of any better options... Do you have any suggestions?

Tom-Newton:
> I wonder if defining default TLS settings with JSON is the way to go. Probably not a bad experience if using Helm but I don't really like to deal with escaping JSON strings in general.

Thanks for reviewing. I think that is a fair assesment, but I can't think of any better options... Do you have any suggestions?

nabuskey:
> > I wonder if defining default TLS settings with JSON is the way to go. Probably not a bad experience if using Helm but I don't really like to deal with escaping JSON strings in general.
> 
> Thanks for reviewing. I think that is a fair assesment, but I can't think of any better options... Do you have any suggestions?

Hmm maybe read from a file instead, likely from CMs. I'd say adding support YAML and JSON is a good idea too. 


I'd like to hear others' opinions on this though. I do love this feature proposed here and would love to see it implemented.

Tom-Newton:
> Hmm maybe read from a file instead, likely from CMs. I'd say adding support YAML and JSON is a good idea too.

I've been thinking a bit about how to implement this. It does add a bit of complexity and I'm not decided how to organise things. It feels a bit strange to mix with some config provided via config maps and some via CLI arguments. 

I can implement it with config maps if anyone has strong opinions, but I think my preference would be to leave it as is for now. If we want to change to using config maps, I would propose that as a seperate PR to do a full migration to config maps.

nabuskey:
> It feels a bit strange to mix with some config provided via config maps and some via CLI arguments.

Agreed on this but I honestly I am not sure if escaped json string is good either. I'd love to hear others' opinions and get this implemented.

@ChenYi015 @jacobsalway @vara-bonthu Opinions on this?

Tom-Newton:
@ChenYi015, sorry to be impatient but do you have any time to look at this or could you nominate another reviewer? Based on the number of :+1:s on [the issue](https://github.com/kubeflow/spark-operator/issues/2460) I think it is quite a popular feature.

internal/controller/sparkapplication/controller_test.go
```diff
@@ -46,31 +47,199 @@ var _ = Describe("SparkApplication Controller", func() {
 			Name:      appName,
 			Namespace: appNamespace,
 		}
+		ingressKey := types.NamespacedName{
+			Name:      appName + "-ui-ingress",
```
ChenYi015: We can use `util.GetDefaultUIIngressName()` instead of construct ingress name manually.

charts/spark-operator-chart/values.yaml
```diff
@@ -74,6 +74,15 @@ controller:
     urlFormat: ""
     # -- Optionally set the ingressClassName.
     ingressClassName: ""
+    # -- Optionally set default TLS configuration for the Spark UI's ingress. `ingressTLS` in the SparkApplication spec overrides this.
+    defaultIngressTLS: []
```
ChenYi015: What about simplifying `controller.uiIngress.defaultIngressTLS` to `controller.uiIngress.tls`?

charts/spark-operator-chart/values.yaml
```diff
@@ -74,6 +74,15 @@ controller:
...
+    # - hosts:
+    #   - "*.example.com"
+    #   secretName: "example-secret"
+    # -- Optionally set default ingress annotations for the Spark UI's ingress. `ingressAnnotations` in the SparkApplication spec overrides this.
+    defaultIngressAnnotations: {}
```
ChenYi015: What about simplifying `controller.uiIngress.defaultIngressAnnotations` to `controller.uiIngress.annotations`?

internal/controller/sparkapplication/controller.go
```diff
@@ -57,11 +57,13 @@ var (
...
-	EnableUIService       bool
-	IngressClassName      string
-	IngressURLFormat      string
-	DefaultBatchScheduler string
+	Namespaces                []string
+	EnableUIService           bool
+	IngressClassName          string
+	IngressURLFormat          string
+	DefaultIngressTLS         []networkingv1.IngressTLS
+	DefaultIngressAnnotations map[string]string
```
ChenYi015:
```suggestion
	IngressClassName          string
	IngressURLFormat          string
	IngressTLS                []networkingv1.IngressTLS
	IngressAnnotations        map[string]string
```

Let us simplify the variable naming to be consistent with `IngressClassName` and `IngressURLFormat`.

charts/spark-operator-chart/templates/controller/deployment.yaml
```diff
@@ -72,6 +72,12 @@ spec:
         {{- with .Values.controller.uiIngress.ingressClassName }}
         - --ingress-class-name={{ . }}
         {{- end }}
+        {{- with .Values.controller.uiIngress.defaultIngressTLS }}
+        - --default-ingress-tls={{ . | toJson }}
```
ChenYi015:
```suggestion
        - --default-ingress-tls={{ . | toJson | quote }}
```

Maybe we need to add a `quote` func to avoid unexpected flag parsing.

charts/spark-operator-chart/templates/controller/deployment.yaml
```diff
@@ -72,6 +72,12 @@ spec:
...
+        {{- end }}
+        {{- with .Values.controller.uiIngress.defaultIngressAnnotations }}
+        - --default-ingress-annotations={{ . | toJson }}
```
ChenYi015:
```suggestion
        - --default-ingress-annotations={{ . | toJson | quote }}
```

internal/controller/sparkapplication/web_ui.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
 	"github.com/kubeflow/spark-operator/v2/pkg/common"
 	"github.com/kubeflow/spark-operator/v2/pkg/util"
+	networkingv1 "k8s.io/api/networking/v1"
```
ChenYi015: Let us group the import statements into three groups, one group for std lib, one for third-party libs and the last one for spark-operator.

cmd/operator/controller/start.go
```diff
@@ -18,11 +18,15 @@ package controller
...
 import (
 	"crypto/tls"
+	"encoding/json"
 	"flag"
+	"fmt"
 	"os"
 	"slices"
 	"time"
 
+	networkingv1 "k8s.io/api/networking/v1"
```
ChenYi015: Let us move import statements of third-party libraries into one group.

internal/controller/sparkapplication/driveringress.go
```diff
@@ -80,14 +80,20 @@ func (r *Reconciler) createDriverIngress(app *v1beta2.SparkApplication, driverIn
...
 	ingressName := fmt.Sprintf("%s-ing-%d", app.Name, *driverIngressConfiguration.ServicePort)
 	if util.IngressCapabilities.Has("networking.k8s.io/v1") {
-		return r.createDriverIngressV1(app, service, ingressName, ingressURL, ingressClassName)
+		return r.createDriverIngressV1(app, service, ingressName, ingressURL, ingressClassName, nil, nil)
 	}
 	return r.createDriverIngressLegacy(app, service, ingressName, ingressURL)
 }
 
-func (r *Reconciler) createDriverIngressV1(app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string) (*SparkIngress, error) {
+func (r *Reconciler) createDriverIngressV1(app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string, defaultIngressTLS *[]networkingv1.IngressTLS, defaultIngressAnnotations *map[string]string) (*SparkIngress, error) {
```
ChenYi015:
```suggestion
func (r *Reconciler) createDriverIngressV1(app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string, defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
```

Let us use `[]networkingv1.IngressTLS` instead of `*[]networkingv1.IngressTLS` and use `map[string]string` instead of `*map[string][string]`.

ChenYi015 (commented):
@Tom-Newton Thanks for your contribution! This PR overall LGTM, have left some comments.

internal/controller/sparkapplication/controller_test.go
Tom-Newton: Done. I had to move the `appConfig` though so that I could reference it in the calls to `util.GetDefaultUIIngressName()`.

charts/spark-operator-chart/templates/controller/deployment.yaml
Tom-Newton: I tried this but it actually caused it to break. It interprets the entire JSON as a string 
```
Error: failed parsing ingress-annotations JSON string from CLI: json: cannot unmarshal string into Go value of type map[string]string
```

I think `toJson` probably does sufficient quoting on its own.

charts/spark-operator-chart/templates/controller/deployment.yaml
Tom-Newton: I had the same issue as https://github.com/kubeflow/spark-operator/pull/2513#discussion_r2137273880

internal/controller/sparkapplication/controller.go
Tom-Newton: Ok, I made the change here and in a few related places that I guess you would also want to remove "default".

I think my preference would have been to leave it as it was because "default" provides clarity on how this interacts when the same config is set in the `SparkApplication` spec. I'm not going to argue about it though.

Tom-Newton (commented):
Thanks for reviewing @ChenYi015. There is [one comment](https://github.com/kubeflow/spark-operator/pull/2513#discussion_r2137273880) where I couldn't make the change you suggested, but all the others I've made changes that hopefully align with what you were suggesting.

ChenYi015:
> > It feels a bit strange to mix with some config provided via config maps and some via CLI arguments.
> 
> Agreed on this but I honestly I am not sure if escaped json string is good either. I'd love to hear others' opinions and get this implemented.
> 
> @ChenYi015 @jacobsalway @vara-bonthu Opinions on this?

I think, in the future, we can introduce a config API just like [jobset](https://github.com/kubernetes-sigs/jobset/blob/main/api/config/v1alpha1/configuration_types.go), [kuberay-operator](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/config/v1alpha1/configuration_types.go). Then we can create a ConfigMap for holding all configurations and deprecate flag-based configurations.

Issue #2525, opened 2025-05-07
Spark metrics depends on Prometheus


blcksrx:
### What happened?

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

I have noticed that `MetricsProperties` and `MetricsPropertiesFile` depends on `PrometheusMonitoring` has been enabled or not.

I would like to mention that although majority of k8s users use Prometheus as observation service but spark metrics does not limits to Prometheus beside of this fact that it gives users more flexibility.

Please let me know your opinion on the issue, I'll be glad to raise a PR.

https://github.com/kubeflow/spark-operator/blob/master/pkg/util/sparkapplication.go#L241-L264

### Environment & Versions

- Kubernetes Version:
- Spark Operator Version:
- Apache Spark Version:

### Impacted by this bug?

Give it a 👍 We prioritize the issues with most 👍

PR #2529, opened 2025-05-08
#2525 spark metrics in depends on prometheus

blcksrx:
## Purpose of this PR

**Proposed changes:**

- <Change 1>
- <Change 2>
- <Change 3>

## Change Category

- [ ] Documentation update

## Checklist

### Additional Notes

#2525

### DIFF
--- a/internal/controller/sparkapplication/monitoring_config.go
+++ b/internal/controller/sparkapplication/monitoring_config.go
@@ -35,7 +35,7 @@ import (
 
 func configPrometheusMonitoring(app *v1beta2.SparkApplication, client client.Client) error {
 	port := common.DefaultPrometheusJavaAgentPort
-	if app.Spec.Monitoring.Prometheus.Port != nil {
+	if app.Spec.Monitoring.Prometheus != nil && app.Spec.Monitoring.Prometheus.Port != nil {
 		port = *app.Spec.Monitoring.Prometheus.Port
 	}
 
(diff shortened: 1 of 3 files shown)


nabuskey:
A bit confused here. This PR doesn't seem to address the issue linked. The functions modified are called when building prometheus configurations only. This means it doesn't allow for other monitoring solutions to work like mentioned in the issue.

blcksrx:
I understand the confusion, it just naming and I have avoid to touch/rename more than that to make it simple
Although the function name is `configPrometheusMonitoring` but it builds the configMap file for metrics.properties.

https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/monitoring_config.go#L125

nabuskey:
Hmm as far as I can tell, `configPrometheusMonitoring` is called if the prometheus field is not nil. So this means you have to put something in the prometheus field. I think writings tests here to show your intentions and results help here.

blcksrx:
yes Exactly, my intention was to write the test but I saw that all the monitoring_test has been commented out

blcksrx:
@nabuskey I've rebased my PR with your recent fix and added my test case there

api/v1beta2/sparkapplication_types.go
```diff
@@ -640,8 +640,10 @@ type Port struct {
 // MonitoringSpec defines the monitoring specification.
 type MonitoringSpec struct {
 	// ExposeDriverMetrics specifies whether to expose metrics on the driver.
+	//	+optional
```
nabuskey, line 643: Do they need to be optional? let's fix indent if so.

blcksrx: My bad, it was my first intention but i changed my mind, apparently forgot to remove the comment

blcksrx: fixed

vara-bonthu (approved):
/approve

PR #2560, opened 2025-06-12
Update golangci lint

joshuacuellar1:
## Purpose of this PR
Golangci-lint has already released a v2 version of the linter, but the project is still running v1. The configuration file syntax has also changed, so I updated it. When I rerun the linters, they catch several errors (uncaught errors, etc.), so I fixed those as well.

**Proposed changes:**
Update Golangci-lint version with config file
Fix errors detected by linters

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/controller/sparkapplication/driveringress.go
+++ b/internal/controller/sparkapplication/driveringress.go
@@ -145,15 +145,15 @@ func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.Spa
 	}
 
 	if len(ingressResourceAnnotations) != 0 {
-		ingress.ObjectMeta.Annotations = ingressResourceAnnotations
+		ingress.ObjectMeta.Annotations = ingressResourceAnnotations //nolint:staticcheck
 	}
 
 	// If we're serving on a subpath, we need to ensure we use the capture groups
 	if ingressURL.Path != "" && ingressURL.Path != "/" {
-		if ingress.ObjectMeta.Annotations == nil {
-			ingress.ObjectMeta.Annotations = make(map[string]string)
+		if ingress.ObjectMeta.Annotations == nil { //nolint:staticcheck
+			ingress.ObjectMeta.Annotations = make(map[string]string) //nolint:staticcheck
 		}
-		ingress.ObjectMeta.Annotations["nginx.ingress.kubernetes.io/rewrite-target"] = "/$2"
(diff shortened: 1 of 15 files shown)


internal/controller/sparkapplication/driveringress.go
```diff
@@ -145,15 +145,15 @@ func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.Spa
 	}
 
 	if len(ingressResourceAnnotations) != 0 {
-		ingress.ObjectMeta.Annotations = ingressResourceAnnotations
+		ingress.ObjectMeta.Annotations = ingressResourceAnnotations //nolint:staticcheck
```
ChenYi015: Do we have a way to fix this lint check instead of skipping lint?

joshuacuellar1: When I run `make go-lint-fix`, the linter changes the line to `ingress.Annotations`, removing the `ObjectMeta` part. This also happens to all the other lines I added the `\\nolint:staticcheck` comment to. I wasn't sure if removing `ObjectMeta` changes how the code runs, so I left them like this.

Is the `ObjectMeta` part necessary?

ChenYi015: For struct embedding, I think it is fine to leave out the `ObjectMeta` part and access the `Annotations` field directly, see https://gobyexample.com/struct-embedding.

pkg/util/util_test.go
```diff
@@ -30,13 +30,13 @@ import (
 
 var _ = Describe("GetMasterURL", func() {
 	BeforeEach(func() {
-		os.Setenv(common.EnvKubernetesServiceHost, "127.0.0.1")
-		os.Setenv(common.EnvKubernetesServicePort, "443")
+		os.Setenv(common.EnvKubernetesServiceHost, "127.0.0.1") //nolint:errcheck
```
ChenYi015:
```suggestion
		Expect(os.Setenv(common.EnvKubernetesServiceHost, "127.0.0.1")).NotTo(HaveOccurred())
```

We can use `Expect(err).NotTo(HaveOccurred())` to avoid skipping the lint check.

joshuacuellar1: Done.

ChenYi015 (approved):

Issue #2538, opened 2025-05-28
Spark 4.0 support?


sunchao:
### What feature you would like to be added?

[Spark 4.0](https://spark.apache.org/releases/spark-release-4-0-0.html) just released and curious if there is ongoing effort to support it. Thanks.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

PR #2564, opened 2025-06-19
upgrade to Spark 4.0.0

nabuskey:
## Purpose of this PR

This PR upgrades Spark to 4.0.0. 

There are some changes to core from 3.5 to 4.0 but as far as I can tell, they don't affect us: https://spark.apache.org/docs/latest/core-migration-guide.html#upgrading-from-core-35-to-40

There hasn't been many changes to spark-submit: https://github.com/apache/spark/commits/master/docs/submitting-applications.md

I do have some concerns:
- are we going to maintain a separate branch for spark 3 for those who haven't or cannot upgrade to 4 yet?
- if yes, do we need a separate release process / automation?

fixes: #2538 

**Proposed changes:**

- Update Dockerfile and examples to reference 4.0.0.
- Update Docker image references to use FQDN.

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=index.docker.io/library/spark:4.0.0
+ARG SPARK_IMAGE=docker.io/library/spark:4.0.0
 
 FROM golang:1.24.1 AS builder
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=spark:3.5.5
+ARG SPARK_IMAGE=index.docker.io/library/spark:4.0.0
 
 FROM golang:1.24.1 AS builder
 
--- a/config/samples/v1beta2_scheduledsparkapplication.yaml
+++ b/config/samples/v1beta2_scheduledsparkapplication.yaml
@@ -11,7 +11,7 @@ spec:
   template:
     type: Scala
     mode: cluster
-    image: index.docker.io/library/spark:4.0.0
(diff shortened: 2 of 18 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=spark:3.5.5
+ARG SPARK_IMAGE=index.docker.io/library/spark:4.0.0
```
ChenYi015: What is the benefits of using FQDN? As it will increase the length of the image name.

nabuskey: It's less ambiguous. The short name relies on the engine's registry config where short names are resolved using its default configuration. In case of Docker, it pulls from docker.io.

Although unlikely, you may end up pulling from the wrong registry. 

https://www.redhat.com/en/blog/container-image-short-names

Using FQDN protects by eliminating ambiguity.

ChenYi015: Sounds good. Should we use `docker.io` rather than `index.docker.io`, as it is the default registry name of Docker hub?

nabuskey: Ah good point. Lemme update.

config/samples/v1beta2_scheduledsparkapplication.yaml
```diff
@@ -11,23 +11,23 @@ spec:
   template:
     type: Scala
     mode: cluster
-    image: spark:3.5.5
+    image: index.docker.io/library/spark:4.0.0
```
ChenYi015: This and many other places should also be updated.

josecsotomorales:
Hey folks! I'm happy to take this PR to get a new version released ASAP, just let me know if you would like me to take it 🚀

ChenYi015 (approved):

Issue #1408, opened 2021-11-25
FileNotFoundException prometheus.yaml


yevsh:
gcr.io/spark-operator/spark-operator:v1beta2-1.2.0-3.0.0
Spark version: 2.4.5
kubernetes: 1.18.10

in sparkapplication yaml:

```
 monitoring:
    exposeDriverMetrics: true
    exposeExecutorMetrics: true
    prometheus:
      jmxExporterJar: "/opt/spark/jars/jmx_prometheus_javaagent.jar"
      port: 8090
```

from log:

```
 + CMD=("$SPARK_HOME/bin/spark-submit" --conf "spark.driver.bindAddress=$SPARK_DRIVER_BIND_ADDRESS" --deploy-mode client "$@")                 
 + exec /usr/bin/tini -s -- /opt/spark/bin/spark-submit --conf spark.driver.bindAddress=10.244.1.108 --deploy-mode client --properties-file /o 
 Exception in thread "main" java.lang.reflect.InvocationTargetException                                                                        
     at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)                                                                            
     at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)                                                          
     at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                                  
     at java.lang.reflect.Method.invoke(Method.java:498)                                                                                       
     at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)                                                
     at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)                                               
 Caused by: java.io.FileNotFoundException: /etc/metrics/conf/prometheus.yaml (No such file or directory)                                       
     at java.io.FileInputStream.open0(Native Method)    
```
as i understand,  this file created/mounted by spark as **i don’t  have it in image** .
and I know that this file is not required inside the image.

Also it was working...and now not any more, can it be related to cluster roles/serviceaccount/webhook?

please advise.

yevsh:
i am still getting this with spark operator v1beta2-1.3.3-3.1.1

Do I also need to create webhook?

PR #2576, opened 2025-06-29
Add enterprise-gateway as sub chart to spark-operator

fresende:
## Purpose of this PR

Adds an optional dependency for the Enterprise Gateway as a subchart to the Spark Operator Helm chart to enable Jupyter notebooks to launch kernels in Spark Operator. The chart will only install the gateway if `jupyter-gateway.enable` is set to true.

Still waiting on the dependency [#1408](https://github.com/jupyter-server/enterprise_gateway/pull/1408)  before we can continue.

## Change Category

- [ X] Feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [X ] I have conducted a self-review of my own code.

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -30,6 +30,13 @@ keywords:
 
 home: https://github.com/kubeflow/spark-operator
 
+dependencies:
+  - name: enterprise-gateway
+    version: 3.3.0-dev.0
+    repository: https://lresende.github.io/helm-chart/
+    alias: jupyter-gateway
+    condition: jupyter-gateway.enable
+
 maintainers:
 - name: yuchaoran2011
   email: yuchaoran2011@gmail.com
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -435,3 +435,10 @@ certManager:
   # See [cert-manager.io/v1.Certificate](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate).
   # @default -- 1/3 of issued certificate’s lifetime.
   renewBefore: ""
+
+  jupyter-gateway:
+    # -- Specifies whether to enable Jupyter Gateway.
+    enable: true
+    namespace: enterprise-gateway
+    kip:
+      enabled: false


charts/spark-operator-chart/Chart.yaml
```diff
@@ -30,6 +30,13 @@ keywords:
 
 home: https://github.com/kubeflow/spark-operator
 
+dependencies:
+  - name: enterprise-gateway
+    version: 3.3.0-dev.0
+    repository: https://lresende.github.io/helm-chart/
```
ImpSy, line 36: Please back port your chart code into this repo
Not allowing external reference is the best way to protect the project against supply chain attack

fresende: Just to clarify, the enterprise-gateway chart I'm referencing is from the official Jupyter Enterprise Gateway Helm chart ([source](https://github.com/jupyter-server/enterprise_gateway/tree/main/etc/kubernetes/helm/enterprise-gateway)), and I’ll be pinning it to a specific version once my changes are [released](https://github.com/jupyter-server/enterprise_gateway/releases/tag/v3.2.3) — currently at 3.3.0-dev.0, to ensure security, stability, and reproducibility.
Given that, would it still be necessary to vendor the entire chart into this repo? Maintaining a full copy locally could become a maintenance burden as the upstream continues to evolve.

vikas-saxena02: @fresende the enterprise gateway is not actively maintained and the last commit was about a year ago. I dont think its a good idea to use something which is not likely to be maintained actively cc: @jacobsalway @vara-bonthu  @yuchaoran2011

ChenYi015: Do we have a reason that we must include enterprise-gateway as a sub chart? We can install enterprise-gateway separately, right?

fresende: This is part of a current GSoC project sponsored by the Kubeflow community, focused on integrating Spark with Kubeflow notebooks:
https://docs.google.com/document/d/1ZrWCB8p_ZR5oyH6sClBT43QNNK-QrnUXOzQVu39ivLw/edit?usp=sharing

Based on our earlier Slack discussion:
https://cloud-native.slack.com/archives/C074588U7EG/p1750351418736009
It was suggested that using a subchart would be the best approach. I’ve been collaborating with the Jupyter community to update Enterprise Gateway and have created a PR that adds it as a subchart accordingly.

If this direction doesn’t seem ideal, I can shift focus to documenting how to deploy it manually, but that would lead to a more complex and less user-friendly experience.

yuchaoran2011: My opinion is the same as @ChenYi015 mentioned above. I don't think another component should be included in the Spark operator chart. That will open up a precedent for other components to be added later, causing the deployment to become bloated. 
As a start, documenting how Spark operator and EG can work together will be good.

Shekharrajak: @fresende can we have documentation in spark operator page / GitHub readme somewhere to how to use it with Jupyter Gateway and working example ?

PR #2663, opened 2025-10-07
bugfix: hook.upgradeCrd use cases

pvbouwel:
When using the new `hook.upgradeCrd` there are 2 cases that currently fail.

1) When a CRD is missing  it cannot be created due to missing role permissions
2) If the CRD was previously managed differently force conflicts should be used in order to make sure Helm can start managing it.

## Purpose of this PR

Fix the use cases mentioned above

**Proposed changes:**

- Add create permissions for CRDs (not only get and patch) The documenttation says that `hook.upgradeCrd`
- Allow (opt-in) force-conflicts for updating the CRDs

## Change Category

- [ ] Documentation update

### Rationale

If users want to auto-manage there CRD upgrades the initial install is also something they like to delegate to the helm chart.

1) It can be that CRDs have been added over time (e.g. sparkconnects.sparkoperator.k8s.io) but then they won't exist when running the helm hook to upgrade CRDs and it will fail since it does not have permissions to create a CRD
2) If users had custom versions of the CRDs they could get conflicts and the recommendation to use --force-conflicts to manage all the fields by the HELM chard but this was not manageable via the helm chart.

## Checklist

### Additional Notes

I used these changes myself succesfully to upgrade a deployment of the spark-operator.


pvbouwel:
I can fix the helm tests and extend them to cover case with the extra value but  I'll await feedback whether there is interest to get this in or if other naming is preferred.

ChenYi015:
@pvbouwel Thanks for fixing the role permission issue. I prefer not to introduce `--force-conflicts` for that we will avoid introducing breaking changes to CRDs. And if users have customized CRDs with breaking changes, they should be responsible for managing the CRDs.

pvbouwel:
I don't fully get `for that we will avoid introducing breaking changes to CRDs.`  I'd say it would enable it.  It was anyway optionally and defaulted to not have it.

But I can follow the reasoning that if anybody has done something to the CRDs they hopefully know how to get to a helm-managed state if they'd want to (which might more often not be the case)

So I updated the initial comment and the scope to just fix the permission issue (+corresponding test).

ChenYi015:
> I don't fully get `for that we will avoid introducing breaking changes to CRDs.` I'd say it would enable it. It was anyway optionally and defaulted to not have it.

The community will not make any breaking changes to the CRDs in the future (e.g. renaming or removing fields). So `--force-conflicts` option is not needed in most cases. If the community has a real need for this option, we will consider adding it back.

PR #2668, opened 2025-10-15
Support controlling `SparkApplication` labels mutation to spark pods

### LINKED ISSUE #2669: FeatureRequest: Control which SparkApplication labels  are mutated spark pods

User often would like to control which `SparkApplication` labels are mutated/not mutated to the Spark pods created by the `SparkApplication`.

### Why is this needed?

https://github.com/kubernetes-sigs/kueue/pull/7268 requires this feature.

In `SparkApplication` integration in Kueue, Kueue expects `SparkApplication` to have `kueue.x-k8s.io/queue-name` label:

```yaml
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: spark-pi
  labels:
    kueue.x-k8s.io/queue-name: queue-a
```

Currently, Spark Operator copies all the `SparkApplication` labels to driver/executor pods. Thus, resultant driver/executor pods like this:

```yaml
kind: Pod
metadata:
  labels:
    kueue.x-k8s.io/queue-name: queue-a

    # managed labels should not be controlled by users
    spark-role: driver
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app
---
kind: Pod
metadata:
  labels:
    kueue.x-k8s.io/queue-name: queue-a

    # managed labels should not be controlled by users
    spark-role: executor
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app
```

In this situation, Kueue activates [Plain `Pod` integration](https://kueue.sigs.k8s.io/docs/tasks/run/plain_pods/) for such pods(opt-in feature) instead of `SparkApplication` integration. This means each pod are treated as queue-ed item instaed of `SparkApplication`. This is a problem.

Why this happens:
- Kueue basically activates an integration in the mutating webhook:
  - If a received `Pod` has no owner reference and `queue-name` label, then activate `Pod` integration
  - If a received `Pod` has `SparkApplication` owner reference, then skip activating plain pod integration, and already activated `SparkApplication` integration will take effect
- But, 
  - spark-operator _always_ copies _all_ the labels down to the spark pods, and
  - spark-operator also puts `OwnerReference` on the Spark driver pod in the [mutating webhook](https://github.com/kubeflow/spark-operator/blob/v2.3.0/internal/webhook/sparkpod_defaulter.go#L178-L185).
- Thus, the `OwnerReference` race can happen because Kubernetes does not have an explicit specification on the order of the webhook called. 
- Consequently, Kueue's webhook might see a spark driver pod without an owner reference, but just with `queue-name` label. This could activate plain pod integration mistakenly. 

I could see two problems:
- Setting `OwnerReference` in webhook is problem-prone
- _Always_ copying _all_ the labels to spark pods could be a bit non-flexible
  - User CANNOT define parent-only labels

Probably, I could see two approaches:
- Put `OwnerReference` at submission time in the controller instead of webhook
  - _added: #2670 can fix this (also fix [#2502](https://github.com/kubeflow/spark-operator/issues/2502) partially)_
- Introduce a field that enables users to control which labels are mutated to the Spark driver/executor pod
  - _added: #2668 can fix this_

Both approaches could solve the issue described above. However, I think I'd like to focus the second one here because in the Kubernetes ecosystem, parent resource labels are normally NOT mutated to the child pods. For example, `Deployment` labels will not be copied to the child pods. Instead, user can define labels for child pods in `Deployment.Spec.Template.Metadata.Labels` field. And, `SparkApplication` already has `spec.{driver|executor}.labels` for the case.

However, we cannot stop the label mutation because it's a breaking change. So, I'd like to propose introducing a field like `sparkApplicationLabelsMutation` which can filter/filter-out target labels to be mutated/not mutated in `SparkApplication` effectively:

```yaml
kind: SparkApplication
metadata:
  name: spark-app
  labels:
    keyA: valueA
    keyB: valueB
spec:
  driver:
    # when not specified, all the labels are copied down for backward compat
    sparkApplicationLabelsMutation:
      # multiple match condition can be defined. OR-ed.
      labelKeysMatches:
      # this condition matches any labels except for "keyB"
      - fixed: keyB
        # regex: "..."
        invert: true 
  executor:
    sparkApplicationLabelMutation:
      # this condition matches any labels
      labelKeysMatches:
      - regex: '^.*'
```

This `SparkApplication` would generate a driver/executor pod with the following labels:

```yaml
kind: Pod
metadata:
  labels:
    # SparkApplication's user-defined labels
    # any labels except for "keyB" because "fixed: keyB, invert: true"
    keyA: valueA

    # managed labels should not be controlled by users
    spark-role: driver
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app
---
kind: Pod
metadata:
  labels:
    # SparkApplication's user-defined labels
    # all the labels are matched because "regex: '^.*'"
    keyA: valueA
    keyB: valueB

    # managed labels should not be controlled by users
    spark-role: executor
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app
```

In the Kueue integration, I plan to introduce the `SparkApplication` webhook in Kueue, which injects the match rules like this:

```yaml
kind: SparkApplication
metadata:
  name: spark-app
  labels:
    kueue.x-k8s.io/queue-name: queue-a
spec:
  driver:
    sparkApplicationLabelsMutation:
      labelKeysMatches:
      # Kueue's mutating webhook injects this match rule
      - fixed: "kueue.x-k8s.io/queue-name"
        invert: true
  executor:
    sparkApplicationLabelMutation:
      labelKeysMatches:
      # Kueue's mutating webhook injects this match rule
      - fixed: "kueue.x-k8s.io/queue-name"
        invert: true
```

_No response_

### Additional context

_No response_

### Love this feature?

everpeace:
## Purpose of this PR

**Proposed changes:**

- Introduce `SparkApplication.Spec.{Driver,Executor}.SparkApplicationLabelMutation` field  which provides users to control which labels in SparkApplication are mutated to the Spark pods

```yaml
kind: SparkApplication
metadata:
  name: spark-app
  labels:
    keyA: valueA
    keyB: valueB
spec:
  driver:
    sparkApplicationLabelsMutation:
      labelKeysMatches:
      - fixed: keyA
  executor:
    sparkApplicationLabelsMutation:
      labelKeysMatches:
      - regex: '^.*'
----
#
# Above SparkApplication generates Spark driver/executor pod like this:
#
kind: Pod
metadata:
  labels:
    # SparkApplication's label
    # keyB is excluded because "keyB" doesn't match "fixed: keyA"
    keyA: valueA

    # managed labels should not be controlled by users
    spark-role: driver
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app

---
kind: Pod
metadata:
  labels:
    # SparkApplication's label
    # all the labels are mutated because any label key matches "regex: '^.*'"
    keyA: valueA
    keyB: valueB

    # managed labels should not be controlled by users
    spark-role: executor
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/app-name: spark-app
```

## Change Category

- [ ] Documentation update

### Rationale

Fixes #2669 

## Checklist


ChenYi015:
It is a bit compliated for users to use a field like `sparkApplicationLabelsMutation` when integrating with Kueue. Let us use the controller-reference way first and see wether the community has a request for this feature.
/hold for review

everpeace:
ref: https://github.com/kubeflow/spark-operator/issues/2669#issuecomment-3435883335

As discussed in the community meeting on 2025-10-17, we plan to introduce a simpler solution.

Issue #2502, opened 2025-04-09
Deprecation of webhook by moving all functionality into the pod template


ChenYi015:
The community has decided to deprecate the webhook in the future. Create this umbrella issue to track which issues we need to solve.

huskysun:
Hi @ChenYi015, is there a design doc or something similar that captures this idea? Like what's the problem with the current webhook, what's the planned path for the deprecation and such? Thanks.

PR #2670, opened 2025-10-16
Set ControllerReference on driver pod and non-Controller OwnerReference on executor pod at submission time

everpeace:
This feature is required in https://github.com/kubernetes-sigs/kueue/pull/7268 

## Purpose of this PR

**Proposed changes:**

- This PR puts an owner reference at submission time instead of webhook
  - `OwnerReference(controller=true)` on driver pod
  - `OwnerReference(controller=false)` on executor pod
    - because it already has the controller reference to the driver pod.
    - a non-controller owner reference just makes other controllers find the executor pod is managed(but not contrlled) by `SparkApplication` easily
- for backward compatibility, I made a mutation webhook idempotent on `OwnerReference` (i.e. just skip if it already has it)

## Change Category

- [ ] Documentation update

### Rationale

Typical kubernetes controllers expect `OwnerReference` is set on resources when the webhook is called. Thus, if multiple webhooks are deployed, there could be a race on them. Particularly, this can be a problem in Kueue integration(_also described in https://github.com/kubeflow/spark-operator/issues/2669_).

- Kueue basically activates integration in the mutating webhook:
  - If `Pod` has no owner reference and `queue-name` label, then activate `Pod` integration
  - If `Pod` has `SparkApplication` owner reference and `queue-name` label, then activate `SparkApplication` integration
- But, spark-operator also puts `OwnerReference` on the Spark driver pod in the [mutating webhook](https://github.com/kubeflow/spark-operator/blob/v2.3.0/internal/webhook/sparkpod_defaulter.go#L178-L185).
- Thus, the `OwnerReference` race happens because Kubernetes does not have an explicit spec on the order of the webhook called. 
- Consequently, Kueue's webhook might see a spark driver pod without an owner reference. This could activate plain pod integration mistakenly. 

Moreover, executor pods do not have `OwnerReference` to `SparkApplication` but just have it to their driver pod. This makes other controllers difficult to find that the executor pods are managed under `SparkApplicaton`. This PR also addresses this case by setting `OwnerReference(controller=false)` at submission time.

_It can be a part of https://github.com/kubeflow/spark-operator/issues/2502_

## Checklist

### DIFF
--- a/internal/controller/sparkapplication/submission.go
+++ b/internal/controller/sparkapplication/submission.go
@@ -22,7 +22,12 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
+	"slices"
 	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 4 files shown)


nabuskey:
Do we want to add controller ownership on the executor pods? 
I assume we'd have two owner references as a result? One from the driver and the other from the controller. Having multiple is not a problem but I feel like having a clear lineage (operator -> driver -> executor) is a good idea.

everpeace:
> I feel like having a clear lineage (operator -> driver -> executor) is a good idea.

This is clean. But, in this case, other controllers or users have to traverse `OwnerReference(controller=true)` up to the root. Although `OwnerReference(controller=false)` is just informational, I believe it would be useful on many occasions.

PR #2697, opened 2025-10-31
Added SparkApp name validator to accept valid DNS-1035 format label names

aryankumar04:
## Purpose of this PR

This PR adds early validation of SparkApplication `metadata.name` to ensure it conforms to DNS-1035 label requirements (same as Kubernetes Service names). This prevents applications with invalid names from being accepted and failing later when the controller attempts to create the web UI Service.

**Problem:**
Currently, SparkApplications with invalid names (e.g., starting with numbers like `28478459-kafka-to-gcs-streaming-6j783x`) are accepted by the API server but fail during reconciliation with errors like:
```
Failed to reconcile SparkApplication: failed to create web UI service: Service "..." is invalid: 
metadata.name: Invalid value: "...": a DNS-1035 label must consist of lower case alphanumeric 
characters or '-', start with an alphabetic character, and end with an alphanumeric character
```

**Solution:**
Add webhook validation to reject invalid names at admission time with concise, user-friendly error messages.

**Proposed changes:**

- Added `validateName()` function to check DNS-1035 label compliance
- Integrated name validation into `ValidateCreate()` and `ValidateUpdate()` webhook handlers
- Improved error messages to be concise and actionable (e.g., "name must start with a letter (not a number)")
- Prevents late-stage failures by catching invalid names early in the admission webhook

## Change Category

- [ ] Documentation update

### Rationale

**Why this change is needed:**
1. **Early Failure Detection**: Currently, invalid names are only detected when the controller tries to create Services, resulting in confusing errors long after submission.
2. **Better User Experience**: Users get immediate, clear feedback about naming requirements instead of delayed reconciliation errors.
3. **Consistency**: Ensures SparkApplication names follow Kubernetes naming conventions from the start, preventing downstream resource creation failures.
4. **Reduced Confusion**: The previous error messages were verbose and technical. The new messages are concise and explain the specific issue (e.g., "must start with a letter").

**Impact:**
- ✅ **Non-breaking**: Only rejects names that would have failed anyway during reconciliation
- ✅ **Backward Compatible**: Does not affect existing valid SparkApplications
- ✅ **Clear Errors**: Provides actionable error messages for common naming mistakes

## Checklist

### Additional Notes

**Error Message Examples:**
- Invalid name starting with number: `"invalid SparkApplication name \"28478459-kafka-to-gcs-streaming-6j783x\": name must start with a letter (not a number)"`
- Invalid name too long: `"invalid SparkApplication name \"...\": name exceeds 63 characters"`
- Invalid name starts with uppercase: `"invalid SparkApplication name \"...\": name must start with a lowercase letter"`

**Valid Name Requirements:**
- Must start with a lowercase letter [a-z]
- Can contain lowercase letters, numbers [0-9], and hyphens [-]
- Must end with a letter or number (not a hyphen)
- Maximum 63 characters
- Examples: `my-job`, `job1`, `data-pipeline-01`, `a1`

**Testing:**
- Tested with names starting with numbers (should be rejected)
- Tested with valid names (should be accepted)
- Tested with names containing uppercase letters (should be rejected)
- Tested with names starting/ending with hyphens (should be rejected)
- Verified error messages are concise and actionable

**Related:**
This validation prevents the same issue that would occur when creating the web UI Service during reconciliation, but catches it earlier with better error messages.

### DIFF
--- a/internal/webhook/scheduledsparkapplication_validator_test.go
+++ b/internal/webhook/scheduledsparkapplication_validator_test.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2025 The Kubeflow authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
+)
+
+func TestScheduledSparkApplicationValidatorValidateCreate(t *testing.T) {
+	validator := NewScheduledSparkApplicationValidator()
+
+	t.Run("returns nil for unrelated object types", func(t *testing.T) {
+		warnings, err := validator.ValidateCreate(context.Background(), &v1beta2.SparkApplication{})
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		if len(warnings) != 0 {
+			t.Fatalf("expected no warnings, got %v", warnings)
+		}
+	})
+
+	t.Run("accepts ScheduledSparkApplication instances", func(t *testing.T) {
+		app := &v1beta2.ScheduledSparkApplication{}
+		warnings, err := validator.ValidateCreate(context.Background(), app)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		if len(warnings) != 0 {
+			t.Fatalf("expected no warnings, got %v", warnings)
+		}
+	})
+}
+
+func TestScheduledSparkApplicationValidatorValidateUpdate(t *testing.T) {
+	validator := NewScheduledSparkApplicationValidator()
+
+	t.Run("returns nil for unrelated object types", func(t *testing.T) {
+		warnings, err := validator.ValidateUpdate(
+			context.Background(),
+			&v1beta2.ScheduledSparkApplication{},
+			&v1beta2.SparkApplication{},
+		)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		if len(warnings) != 0 {
+			t.Fatalf("expected no warnings, got %v", warnings)
+		}
+	})
+
+	t.Run("accepts ScheduledSparkApplication instances", func(t *testing.T) {
+		oldApp := &v1beta2.ScheduledSparkApplication{}
+		newApp := &v1beta2.ScheduledSparkApplication{}
+		warnings, err := validator.ValidateUpdate(context.Background(), oldApp, newApp)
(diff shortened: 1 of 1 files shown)


internal/webhook/sparkapplication_validator.go
```diff
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
+	validation "k8s.io/apimachinery/pkg/util/validation"
```
ChenYi015:
```suggestion
	"k8s.io/apimachinery/pkg/util/validation"
```

internal/webhook/sparkapplication_validator.go
```diff
@@ -148,6 +159,29 @@ func (v *SparkApplicationValidator) validateSpec(_ context.Context, app *v1beta2
...
+		// Create a concise, user-friendly error message
+		var issue string
+		if len(name) == 0 {
+			issue = "name cannot be empty"
+		} else if len(name) > 63 {
+			issue = "name exceeds 63 characters"
+		} else if name[0] >= '0' && name[0] <= '9' {
+			issue = "name must start with a letter (not a number)"
+		} else if !(name[0] >= 'a' && name[0] <= 'z') {
+			issue = "name must start with a lowercase letter"
```
ChenYi015:
```suggestion
		} else if unicode.IsDigit(name[0]) {
			issue = "name must start with a letter (not a number)"
		} else if !unicode.IsLower(name[0]) {
			issue = "name must start with a lowercase letter"
```

internal/webhook/sparkapplication_validator.go
aryankumar04: Thanks for the suggestion

internal/webhook/sparkapplication_validator.go
aryankumar04: Thanks for the suggestion

internal/webhook/sparkapplication_validator.go
```diff
@@ -148,6 +159,29 @@ func (v *SparkApplicationValidator) validateSpec(_ context.Context, app *v1beta2
...
+			issue = "name cannot be empty"
+		} else if len(name) > 63 {
+			issue = "name exceeds 63 characters"
+		} else if unicode.IsDigit(name[0]) {
+			issue = "name must start with a letter (not a number)"
+		} else if !unicode.IsLower(name[0]) {
+			issue = "name must start with a lowercase letter"
+		} else {
+			issue = "name must contain only lowercase letters, numbers, and hyphens, and end with a letter or number"
+		}
```
ChenYi015: Sorry for the wrong feedback. Since `name[0]` is a byte, we need to change it back to the original.

aryankumar04: Changed back to original @ChenYi015

internal/webhook/sparkapplication_validator.go
vjanelle: I don't think it's necessary for this to only be lowercase letters?

Also you could use a regex for most of this instead

vjanelle:
```
`^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$`
```

aryankumar04: This is updated with regex: `^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`

Issue #2679, opened 2025-10-27
[Spark Connect] AssertionError: Expected hostname or IPv6 IP enclosed in []


tiagotxm:
### What happened?

I'm encountering an `AssertionError` when using Spark Connect with the Spark Operator in an IPv6 Kubernetes cluster.

`Exception in thread "main" java.lang.AssertionError: assertion failed: Expected hostname or IPv6 IP enclosed in [] but got <IPV6_ADDRESS_HERE>`

I have tried forcing IPv6 configurations in my SparkConnect deployment manifest, but the error persists. E.g

```
sparkConf:
  spark.kubernetes.driver.service.ipFamilies: "IPv6"
  spark.driver.extraJavaOptions: "-Djava.net.preferIPv6Addresses=true"
  spark.executor.extraJavaOptions: "-Djava.net.preferIPv6Addresses=true"
  spark.driver.bindAddress: "[::]"
  spark.executor.bindAddress: "[::]"
  spark.driver.host: "[::]"
```

I also modified the Spark Operator's deployment container env vars following [docs](https://github.com/awslabs/data-on-eks/blob/main/analytics/terraform/spark-eks-ipv6/helm-values/spark-operator-values.yaml#L12):
```
env:
- name: _JAVA_OPTIONS
  value: "-Djava.net.preferIPv6Addresses=true"
- name: KUBERNETES_DISABLE_HOSTNAME_VERIFICATION
  value: "true"
```

Has anyone successfully configured Spark Connect via the Operator in an IPv6 environment? Is there a different sparkConf required to fix it? 

### Reproduction Code

Deployment Spark Connect on IPv6 k8s cluster

### Expected behavior

The IP address should be surrounded by `[]`

### Impacted by this bug?

Give it a 👍 We prioritize the issues with most 👍

tiagotxm:
I will try to work on it

ChenYi015:
Closed by #2703.

PR #2703, opened 2025-11-03
Fix driver host configuration to handle IPv6 addresses

tiagotxm:
When the driver pod IP is an IPv6 address, Spark expects the address to be enclosed in square brackets (e.g., `[2001:db8::1]`). The previous approach set `spark.driver.host` unconditionally via:

```go
args = append(args, "--conf", "spark.driver.host=${POD_IP}")
```

This fails Spark's strict hostname checks for IPv6 addresses.

## Purpose of this PR

**Proposed changes:**

Fix `spark.driver.host` to Enclose IPv6 Addresses in Brackets to Prevent AssertionError ([#2679](https://github.com/kubeflow/spark-operator/issues/2679))

## Change Category

- [ ] Documentation update

## Checklist

### Additional Notes

- Fixes the `AssertionError` on IPv6 clusters by ensuring Spark receives a valid host identifier.
- Compatible with both IPv4 and IPv6 environments.

image

### DIFF
--- a/internal/controller/sparkconnect/options.go
+++ b/internal/controller/sparkconnect/options.go
@@ -170,9 +170,9 @@ func driverConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 	}
 
 	args = append(args, "--conf", "spark.driver.bindAddress=0.0.0.0")
-	
+
 	driverHost := "$(host=${POD_IP}; if [[ $host == *:* ]] && [[ $host != \\[* ]]; then echo \"[$host]\"; else echo \"$host\"; fi)"
-    args = append(args, "--conf", fmt.Sprintf("spark.driver.host=%s", driverHost))
+	args = append(args, "--conf", fmt.Sprintf("spark.driver.host=%s", driverHost))
 	args = append(args, "--conf", "spark.driver.port=7078")
 
 	// Driver pod name


tiagotxm:
Hi maintainers,

I'm working on a fix for [#2679](https://github.com/kubeflow/spark-operator/issues/2679) to properly enclose IPv6 addresses in brackets for `spark.driver.host`
I noticed that in [pkg/util/util.go](https://github.com/tiagotxm/spark-operator/blob/f0e45055f3c72020ac4341779b821284dd4f14d9/pkg/util/util.go), the `GetMasterURL()` function already handles the IPv6 bracket requirement with logic like:


```
if strings.Contains(kubernetesServiceHost, ":") && !strings.HasPrefix(kubernetesServiceHost, "[") {
    return fmt.Sprintf("k8s://https://[%s]:%s", kubernetesServiceHost, kubernetesServicePort), nil
}
```
Would it make sense to generalize or reuse this condition for setting values such as spark.driver.host, to ensure consistent IPv6 handling across the operator codebase? 

Any feedback or suggestions would be appreciated!

Thanks!

dbbvitor (approved):

ChenYi015:
> Would it make sense to generalize or reuse this condition for setting values such as spark.driver.host, to ensure consistent IPv6 handling across the operator codebase?

I am afraid not. Since we cannot determine the server pod IP before creating it, the operator cannot construct the final paramater `--conf spark.driver.host=${POD_IP}` in advance.

dbbvitor:
> > Would it make sense to generalize or reuse this condition for setting values such as spark.driver.host, to ensure consistent IPv6 handling across the operator codebase?
> 
> I am afraid not. Since we cannot determine the server pod IP before creating it, the operator cannot construct the final paramater `--conf spark.driver.host=${POD_IP}` in advance.

I think he is suggesting something like having the dynamic bracket wrapping logic in go instead of bash.

Maybe the snippet he quoted could be its own function so it could be used standalone or inside the `GetMasterURL()` function, avoiding duplication.

Maybe something like:

```go
func driverConfOption(...) {
//...

    podIp, err := util.dynamicIPv6Brackets(POD_IP)
    if err != nil {
        return nil, fmt.Errorf("failed to proccess IPv6 address: %v", err)
    }
    args = append(args, "--conf",
        fmt.Sprintf("spark.driver.host=%s", podIp)
    )

}
```

ChenYi015:
It is good to create a util function to bracket IP address as needed. But this function cannot be reused in the Spark connect case because the code must be executed in server pod, not in operator itself.

ChenYi015 (approved):
@tiagotxm Thanks for fixing the IPv6 issue.
/approve

PR #2742, opened 2025-11-19
feat(scheduledsparkapplication): configurable timestampPrecision (nan…

### LINKED ISSUE #2602: Ability to configure the precision of the timestamp attached to the application name when running scheduled spark applications

Currently, when using the scheduled spark application functionality, each run is automatically appended with the current time in nanoseconds. See [here](https://github.com/kubeflow/spark-operator/blob/8ca0176d6c26002920a5f8d049f7272e2f6eaaaf/internal/controller/scheduledsparkapplication/controller.go#L241). This is a 19 digit number that causes the name of the job run to become very long. 

Unfortunately, due to naming schemes within my organization, I would like to make my application name 45 characters long. When this is combined with the additional dash and 19 digit number, my resulting job name exceeds the 63 character limit imposed by Kubernetes. 

As an hourly running job, nanoseconds is a far greater level of precision than I would ever need in my job naming. I would like to be able to configure my scheduled spark application spec to choose from the nanos/micros/millis/seconds options that are available in the go time package for the purposes of appending a timestamp to my scheduled job run: https://pkg.go.dev/time 

### Why is this needed?

Long automatic appending causes name to exceed Kubernetes length limits

I would like to be able to configure my scheduled spark application spec to choose from the nanos/micros/millis/seconds options that are available in the go time package for the purposes of appending a timestamp to my scheduled job run: https://pkg.go.dev/time 

_No response_

### Additional context

_No response_

### Love this feature?

rahul810050:
feat(scheduledsparkapplication): add configurable timestampPrecision for run name generation

This PR introduces a new field `.spec.timestampPrecision` that allows users to control the precision of the timestamp suffix added to `SparkApplication` names generated by `ScheduledSparkApplication`. Supported values are:

`nanos` | `micros` | `millis` | `seconds` | `minutes`

The default remains `nanos` for full backward compatibility.

## Summary

Previously, scheduled runs always used `time.UnixNano()` to generate the timestamp suffix, producing a 19-digit value. When combined with long application names, this frequently caused the run name to exceed Kubernetes’ 63-character limit.

This PR makes the timestamp precision configurable so users can choose a shorter suffix if needed. A new `minutes` option is also added to match Kubernetes CronJob controller behavior, which only schedules in minute granularity.

## Key Changes

- **API / CRD**
  - Added new optional field `.spec.timestampPrecision` to `ScheduledSparkApplicationSpec`
  - Enum validation: `nanos`, `micros`, `millis`, `seconds`, `minutes`
  - Default: `"nanos"` (preserves current behavior)
  - Regenerated CRDs using `make generate` + `make manifests`

- **Controller**
  - Added `formatTimestamp()` helper to format timestamps according to the selected precision
  - Updated run-name generation to use this helper
  - `minutes` mode computes `Unix()/60` to stay consistent with CronJob naming semantics

- **Tests**
  - Added `format_timestamp_test.go` to validate timestamp length for all supported precisions
  - Updated `envtest` setup helper script for contributors

- **Helm Chart**
  - Added optional value `controller.scheduledSparkApplication.timestampPrecision`
  - Defaults to `"nanos"`

## Why This Is Needed (Fixes #2602)

Users with long application names often hit Kubernetes’ 63-character name limit because the operator always appends a 19-digit nanosecond timestamp.

Allowing precision selection reduces the suffix size:

| Precision | Digits | Example Use Case |
|----------|--------|------------------|
| minutes  | ~8     | Matches CronJob granularity; shortest valid suffix |
| seconds  | 10     | Common for hourly/daily jobs |
| millis   | 13     | High-rate jobs |
| micros   | 16     | Advanced workloads |
| nanos    | 19     | Current behavior |

This gives users flexibility while keeping backward compatibility.

## How I Tested

```bash
gofmt -s -w .
make generate
make manifests

# setup envtest
bash scripts/setup-envtest-binaries.sh
export KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/v1.32.0-linux-amd64"

# run unit tests
go test ./internal/controller/scheduledsparkapplication -v
```
- Verified CRD generated correctly (enum + default)
- Ensured all timestamp precisions produce expected digit lengths
- Confirmed controller creates SparkApplication names with correct suffixes

### Checklist:

- [x] Self-review completed
- [x] CRD + controller changes implemented
- [x] Unit tests added and passing
- [x] Helm values updated
- [x] Backward compatibility preserved

### DIFF
--- a/internal/controller/scheduledsparkapplication/controller_test.go
+++ b/internal/controller/scheduledsparkapplication/controller_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2024 The Kubeflow authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package scheduledsparkapplication
 
 import (


ChenYi015:
Hi @rahul810050 , you do not have to create a new PR for every update. You can run git push command by adding an extra `--force` flag.

rahul810050:
> Hi @rahul810050 , you do not have to create a new PR for every update. You can run git push command by adding an extra `--force` flag.

sure @ChenYi015 i will keep it in mind...thanks!

api/v1beta2/scheduledsparkapplication_types.go
```diff
@@ -23,6 +23,7 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// ScheduledSparkApplicationSpec defines the desired state of ScheduledSparkApplication.
```
ChenYi015: Duplicated.

api/v1beta2/scheduledsparkapplication_types.go
```diff
@@ -38,6 +39,16 @@ type ScheduledSparkApplicationSpec struct {
...
+	// TimestampPrecision controls the precision of the timestamp appended to generated
+	// SparkApplication names for scheduled runs.
+	//
+	// Allowed values: "nanos", "micros", "millis", "seconds", "minutes"
+	// +kubebuilder:validation:Enum=nanos;micros;millis;seconds;minutes
+	// +optional
+	// +kubebuilder:default:=nanos
+	// Defaults to "nanos" to preserve current behavior.
+	TimestampPrecision string `json:"timestampPrecision,omitempty"`
+
```
ChenYi015: This field should be removed as we have dicussed.

internal/controller/scheduledsparkapplication/controller.go
```diff
@@ -42,6 +42,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"os"
+	"strconv"
+
```
ChenYi015: Would be better to group the imports with the specified order, i.e. stdlib, third-party and self imports.

scripts/setup-envtest-binaries.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
ChenYi015: Is this file needed? We can run unit tests and e2e tests directly without setting up envtest binaries.

charts/spark-operator-chart/values.yaml
```diff
@@ -57,6 +57,10 @@ hook:
     tag: ""
 
 controller:
+  # -- default precision for ScheduledSparkApplication timestamp suffix
+  scheduledSparkApplication:
+    timestampPrecision: "nanos"
+
```
ChenYi015:
```suggestion
  # ScheduledSparkApplication controller configurations.
  scheduledSparkApplication:
    # -- Default timestamp precision for the scheduled SparkApplication name suffix.
    # Can be one of `nanos`, `micros`, `millis`, `seconds` and `minutes`.
    timestampPrecision: nanos
```

Would be better to move this value after `controller.batchScheduler` as it is a less common used configuration. Then run `make helm-docs` to update the Helm chart README.

internal/controller/scheduledsparkapplication/controller.go
```diff
@@ -236,9 +239,24 @@ func (r *Reconciler) createSparkApplication(
...
+	precision := "nanos" // fallback default
+
+	if envPrecision, ok := os.LookupEnv("SCHEDULED_SA_TIMESTAMP_PRECISION"); ok && strings.TrimSpace(envPrecision) != "" {
+		precision = strings.TrimSpace(envPrecision)
+	} else if strings.TrimSpace(scheduledApp.Spec.TimestampPrecision) != "" {
+		precision = strings.TrimSpace(scheduledApp.Spec.TimestampPrecision)
+	}
+
+	suffix := formatTimestamp(precision, t)
+
```
ChenYi015, line 222: We can add a new controller flag to be consistent with the current implementation, like `--timestamp-precision`, instead of adding a new env. 
See: https://github.com/kubeflow/spark-operator/blob/f53373e7e94ec6eb69d2f6829e61e9c9497e71d6/cmd/operator/controller/start.go

scripts/setup-envtest-binaries.sh
rahul810050: Actually I am on Arch Linux so I needed this for my machine

ChenYi015: Is there any error when running `make unit-test`?

api/v1beta2/scheduledsparkapplication_types.go
```diff
@@ -20,9 +20,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
-
```
ChenYi015, line 25: Changes to this file should be removed.

charts/spark-operator-chart/templates/controller/deployment.yaml
```diff
@@ -135,10 +136,13 @@ spec:
...
         {{- end }}
-        {{- with .Values.controller.env }}
+
+        # --- Controller environment variables (preserve user configured controller.env if any).
+        {{- if .Values.controller.env }}
         env:
-        {{- toYaml . | nindent 8 }}
+{{ toYaml .Values.controller.env | nindent 8 }}
         {{- end }}
+
```
ChenYi015: Let us revert it back to avoid unnecessary changes.

ChenYi015:
Some changes still need to be made:

- We do not have to remove`controller.scheduledSparkApplication` from `values.yaml`.
- Add a new flag parameter named `ScheduledSATimestampPrecision` , like: https://github.com/kubeflow/spark-operator/blob/f53373e7e94ec6eb69d2f6829e61e9c9497e71d6/cmd/operator/controller/start.go#L86-L96
- Then pass the value to `scheduledsparkapplication.Options`, like: https://github.com/kubeflow/spark-operator/blob/f53373e7e94ec6eb69d2f6829e61e9c9497e71d6/cmd/operator/controller/start.go#L443-L448

internal/controller/scheduledsparkapplication/format_timestamp_test.go
```diff
@@ -0,0 +1,26 @@
+package scheduledsparkapplication
+
```
ChenYi015: Since the `formatTimestampLengths` util function is defined in `controller.go`, it is suggested to put the related unit tests in file `controller_test.go` .

scripts/setup-envtest-binaries.sh
rahul810050: @ChenYi015   thanks for asking!

Yes — when I run make unit-test on my Arch Linux setup, it fails because envtest cannot find the Kubernetes API server binaries (kube-apiserver / etcd). Arch Linux does not ship compatible versions by default, and setup-envtest also fails due to version mismatch.

That’s why I added the helper script — it downloads the exact binaries that controller-runtime expects and places them in the correct directory, allowing the unit tests and e2e tests to run successfully on my machine.

If the project prefers not to include this script, I’m totally fine removing it and relying only on the documented workflow. I mainly added it to help cross-platform contributors like myself.

Please let me know what you prefer and I’ll update the PR accordingly!

ChenYi015: I have raised a PR #2751 to fix this issue. Could you help verifying whether it can work as expected?

ChenYi015:
Some changes still need to be made:

1. Changes to file `api/v1beta2/scheduledsparkapplication_types.go` and `config/crd/bases/sparkoperator.k8s.io_scheduledsparkapplications.yaml` should be discarded since we decide not to modify `ScheduledSparkApplication` CRD.
2. File `scripts/setup-envtest-binaries.sh` should be removed and I have created a PR #2751 to fix `make unit-test` does not work on Arch linux.
3. File `internal/controller/scheduledsparkapplication/format_timestamp_test.go` should be removed because the added unit tests have been moved to `internal/controller/scheduledsparkapplication/controller_test.go`.
4. Comments for `controller.scheduledSparkApplication.timestampPrecision` in `values.yaml` need to be updated as suggested by the previsous review comments.
5. Run `make generate`, `make manifests` and `make helm-docs` and commit the changes.

rahul810050:
> Some changes still need to be made:
> 
> 1. Changes to file `api/v1beta2/scheduledsparkapplication_types.go` and `config/crd/bases/sparkoperator.k8s.io_scheduledsparkapplications.yaml` should be discarded since we decide not to modify `ScheduledSparkApplication` CRD.
> 2. File `scripts/setup-envtest-binaries.sh` should be removed and I have created a PR [Update ENVTEST_K8S_VERSION for e2e and unit tests #2751](https://github.com/kubeflow/spark-operator/pull/2751) to fix `make unit-test` does not work on Arch linux.
> 3. File `internal/controller/scheduledsparkapplication/format_timestamp_test.go` should be removed because the added unit tests have been moved to `internal/controller/scheduledsparkapplication/controller_test.go`.
> 4. Comments for `controller.scheduledSparkApplication.timestampPrecision` in `values.yaml` need to be updated as suggested by the previsous review comments.
> 5. Run `make generate`, `make manifests` and `make helm-docs` and commit the changes.

Hii @ChenYi015 !! thanks for the clarification!! I will update the PR soon

rahul810050:
@ChenYi015 could you please review it ??

charts/spark-operator-chart/values.yaml
```diff
@@ -82,35 +82,6 @@ controller:
...
-    # -- Optionally set default TLS configuration for the Spark UI's ingress. `ingressTLS` in the SparkApplication spec overrides this.
-    tls: []
-    # - hosts:
-    #   - "*.example.com"
-    #   secretName: "example-secret"
-    # -- Optionally set default ingress annotations for the Spark UI's ingress. `ingressAnnotations` in the SparkApplication spec overrides this.
-    annotations: {}
-      # key1: value1
-      # key2: value2
-
```
ChenYi015, line 113: Why are you removing these lines?

config/rbac/role.yaml
```diff
@@ -70,7 +70,6 @@ rules:
 - apiGroups:
   - sparkoperator.k8s.io
   resources:
-  - scheduledsparkapplications
```
ChenYi015, line 73: Changes to the `role.yaml` file should be discarded, we need to keep these RBAC rules.

internal/controller/scheduledsparkapplication/controller.go
```diff
@@ -81,185 +88,182 @@ func NewReconciler(
...
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the ScheduledSparkApplication object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.2/pkg/reconcile
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
ChenYi015, line 91: Why are you removing these lines since they are unrelated to the timestamp precision feature? Removing these will also lead to unexpected changes to `config/rbac/role.yaml` file.

ChenYi015 (commented):
Please just make changes that are related to the timestamp precision feature and avoid unnecessary changes like comments removing, values removing and variables renaming.

PR #2768, opened 2025-12-10
Feature: Support hot update executor priorityClassName

houyuting:
## Purpose of this PR

 Update executor priorityClassName not restart sparkapp

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/controller/sparkapplication/event_filter.go
+++ b/internal/controller/sparkapplication/event_filter.go
@@ -173,7 +173,7 @@ func (f *EventFilter) Update(e event.UpdateEvent) bool {
 
 	// The spec has changed except for Spec.Suspend.
 	// This is currently best effort as we can potentially miss updates and end up in an inconsistent state.
-	if !equality.Semantic.DeepEqual(oldApp.Spec, newApp.Spec) {
+	if !equality.Semantic.DeepEqual(oldApp.Spec, newApp.Spec) && !f.canHotUpdate(oldApp, newApp) {
 
 		// Only Spec.Suspend can be updated
 		oldApp.Spec.Suspend = newApp.Spec.Suspend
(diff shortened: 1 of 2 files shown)


ChenYi015:
This will not work because the lifecycle of executor pods are managed by Spark driver, the driver process will not try to reload configurations when running.

houyuting:
> This will not work because the lifecycle of executor pods are managed by Spark driver, the driver process will not try to reload configurations when running.

@ChenYi015 hi , The driver does not need to reload the configuration because priorityClassName takes effect at the Pod level. The purpose of this feature is that when the executor priorityClassName of a Spark application changes, the spark job itself does not need to be restarted; the new priorityClassName only needs to take effect on newly created Pods, With this feature, we can support the following scenarios:
1. If the job is in the submitted state, all Pods can use the new priorityClassName.
2. If the job is in the running state and has already obtained all its Pods, an external controller can choose to apply the new priorityClassName by deleting Pods so that the newly created Pods pick up the new value. If the job has not yet obtained all Pods, you can also choose to let only newly created Pods use the new priorityClassName. The exact behavior can be adjusted based on the business scenario. like 

-  low -> high, we can delete pending pod manual let high priorityClassName take  effect on new pod, this is a faster way to obtain resources .
- high- > low, we can delete all running pod manual ，release high priorityClassName pod and apply low priorityClassName pod requeue.

This capability is already being used in our production environment and works correctly.

ChenYi015:
> The driver does not need to reload the configuration because priorityClassName takes effect at the Pod level. 

Yes, your are right. I just relized that the priorityClassName of executor pods are mutated by the webhook:

https://github.com/kubeflow/spark-operator/blob/a531c93bdf9b449a55e46748a9106f6f2dc3fc1b/internal/webhook/sparkpod_defaulter.go#L534-L550

houyuting:
> > The driver does not need to reload the configuration because priorityClassName takes effect at the Pod level.
> 
> Yes, your are right. I just relized that the priorityClassName of executor pods are mutated by the webhook:
> 
> https://github.com/kubeflow/spark-operator/blob/a531c93bdf9b449a55e46748a9106f6f2dc3fc1b/internal/webhook/sparkpod_defaulter.go#L534-L550
hi @ChenYi015  The priorityClassName passed in through the parameter app *v1beta2.SparkApplication in this piece of code is already the updated value.

ChenYi015:
@houyuting Another PR #2786 have already been merged, in which operator will be able to hot update webhook patched fields (including `.spec.executor.priorityClassName`) without restarting the app. Does this feature statisfy your needs?

houyuting:
> @houyuting Another PR #2786 have already been merged, in which operator will be able to hot update webhook patched fields (including `.spec.executor.priorityClassName`) without restarting the app. Does this feature statisfy your needs?

Yes, Since we're from the same company working on the same use case, I'll go ahead and close this PR.

Issue #2766, opened 2025-12-09
Feature Request: Selective Restart Strategy for SparkApplication Updates


Kevinz857:
### What feature you would like to be added?

A **selective restart strategy** for SparkApplication updates that avoids unnecessary full application restarts when only specific components (executors) or configurations (services/ingress) are modified.

Currently, any change to the SparkApplication spec triggers a complete application restart by setting the status to `ApplicationStateInvalidating`. This proposal introduces intelligent change detection and selective update mechanisms that respect Spark's Driver-Executor architecture constraints.

### Current Pain Points
- **Inefficient restarts**: Changing executor `priorityClassName` restarts the entire application including the driver
- **Production downtime**: Long-running Spark Streaming applications lose all state during unnecessary restarts  
- **Resource waste**: Healthy driver pods are terminated when only executor changes are needed
- **Operational friction**: Simple scaling or configuration updates require full application cycles

### Business Impact
- **Streaming applications**: Critical for 24/7 data pipelines that cannot afford frequent restarts
- **Cost optimization**: Avoid terminating expensive, long-running computations
- **DevOps efficiency**: Enable faster configuration updates and scaling operations
- **Resource utilization**: Keep healthy components running during partial updates

### Technical Context
From [`internal/controller/sparkapplication/event_filter.go`](https://github.com/kubeflow/spark-operator/blob/main/internal/controller/sparkapplication/event_filter.go#L137-L148):

```go
if !equality.Semantic.DeepEqual(oldApp.Spec, newApp.Spec) {
    // Force-set the application status to Invalidating which handles clean-up and application re-run.
    newApp.Status.AppState.State = v1beta2.ApplicationStateInvalidating
    // ... complete restart for ANY spec change
}
```

### 🚨 **Critical Architectural Constraint**
**Spark Driver cannot be restarted independently** because:
- Driver maintains application state (SparkContext, task scheduling, RDD lineage)
- Driver-Executor connections are stateful and cannot be reestablished 
- Executors detect Driver disconnection and terminate automatically
- New Driver instance loses all knowledge of existing Executors

#### 1. **Full Restart Required** (Driver-Related Changes)
- **Driver Configuration**: `driver.priorityClassName`, `driver.cores`, `driver.memory`, `driver.javaOptions`
- **Application Logic**: `type`, `mainClass`, `mainApplicationFile`, `arguments`
- **Global Configuration**: `sparkConf`, `hadoopConf` (most keys)
[...]

PR #2786, opened 2025-12-21
feat: skip reconcile for webhook-patched executor fields

Kevinz857:
## Summary

  When only webhook-patched executor fields change (PriorityClassName, NodeSelector, Tolerations, Affinity, SchedulerName), skip triggering a reconcile since these fields are automatically applied by the mutating webhook when new pods are created.

  This optimization reduces unnecessary application restarts when users update executor scheduling configurations.

  ## Changes

  - Add `isWebhookPatchedFieldsOnlyChange()` to detect webhook-only field changes
  - Skip reconcile and emit event when only webhook fields changed
  - Add comprehensive unit tests for change detection logic

  ## Supported Fields

  The following executor fields can now be updated without triggering a reconcile:
  - `Executor.PriorityClassName`
  - `Executor.NodeSelector`
  - `Executor.Tolerations`
  - `Executor.Affinity`
  - `Executor.SchedulerName`

  ## How It Works

  These fields are patched by the mutating webhook when executor pods are created. When they change:
  1. Already running pods are not affected
  2. New executor pods will automatically use the updated values
  3. No application restart is needed

  ## Test Plan

  - [x] Unit tests for `isWebhookPatchedFieldsOnlyChange()` function
  - [x] Unit tests for `hasExecutorWebhookFieldChanges()` function
  - [x] All existing tests pass

  Partial fix for #2766

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -91,6 +91,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | hook.affinity | object | `{}` | Affinity for the Helm hook Job. |
 | hook.tolerations | list | `[]` | List of node taints to tolerate for the Helm hook Job. |
 | controller.replicas | int | `1` | Number of replicas of controller. |
+| controller.featureGates | list | `[]` | Feature gates to enable or disable specific features. Example: featureGates: - name: PartialRestart   enabled: true |
 | controller.revisionHistoryLimit | int | `10` | The number of old history to retain to allow rollback. |
 | controller.leaderElection.enable | bool | `true` | Specifies whether to enable leader election for controller. |
 | controller.leaderElection.leaseDuration | string | `"15s"` | Leader election lease duration. |
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -124,6 +124,9 @@ spec:
         {{- if .Values.controller.maxTrackedExecutorPerApp }}
         - --max-tracked-executor-per-app={{ .Values.controller.maxTrackedExecutorPerApp }}
         {{- end }}
+        {{- if .Values.controller.featureGates }}
+        - --feature-gates={{ range $index, $gate := .Values.controller.featureGates }}{{ if $index }},{{ end }}{{ $gate.name }}={{ $gate.enabled }}{{ end }}
+        {{- end }}
         {{- if or .Values.prometheus.metrics.enable .Values.controller.pprof.enable }}
         ports:
         {{- if .Values.controller.pprof.enable }}
--- a/charts/spark-operator-chart/tests/controller/deployment_test.yaml
+++ b/charts/spark-operator-chart/tests/controller/deployment_test.yaml
@@ -828,3 +828,33 @@ tests:
           path: spec.template.spec.containers[?(@.name=="spark-operator-controller")].args
           content: --leader-election-retry-period=15s
 
+  - it: Should not contain `--feature-gates` arg if `controller.featureGates` is not set
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[?(@.name=="spark-operator-controller")].args
+          pattern: --feature-gates=.*
+
+  - it: Should contain `--feature-gates` arg with single feature gate if `controller.featureGates` is set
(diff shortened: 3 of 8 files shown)


ChenYi015:
@Kevinz857 Thanks for your contribution! Since this feature will cause a breaking change, I would like to introduce a feature gate machanism first and add an alpha feature like `PartialRestart` for this feature request and it will default to `false`. Users can choose whether to enable this feature or not. Once this feature turns into GA state, we will enable it by default. WDYT?

Kevinz857:
@ChenYi015  Thanks for your feedback! The feature gate approach sounds reasonable to me.

  I agree that introducing a feature gate mechanism for this change is a good idea, as it:
  1. Allows users to opt-in and test the feature in their environments
  2. Provides a safe rollout path from alpha → beta → GA
  3. Avoids unexpected behavior changes for existing users

  I'll work on implementing a feature gate (e.g., `PartialRestart`) that:
  - Defaults to `false` in alpha stage
  - Guards the service/ingress hot-update logic
  - Can be enabled via controller configuration

  Could you point me to any existing feature gate implementation in the codebase I should follow, or would you prefer I propose a new mechanism?

  Thanks!

ChenYi015:
@Kevinz857 I have created a issue https://github.com/kubeflow/spark-operator/issues/2792 for this feature. A reuseable codebase can be found [here](https://github.com/kubernetes/component-base/blob/b8fb4e3d1155bc23613fb7ecc08c8784e2bdba88/featuregate/feature_gate.go). Example repo that supports feature gate https://github.com/kubernetes-sigs/jobset/blob/main/pkg/features/features.go.

Kevinz857:
@ChenYi015   Thanks for creating the issue and providing the reference!

  I've looked at the jobset implementation - it's clean and straightforward.

  Would you like me to:
  1. **Wait** for the feature gate mechanism to be implemented first, then rebase my PR on top of it?
  2. **Implement** the feature gate mechanism as part of this PR  (or a separate PR)?

  I'm happy to help implement the feature gate if needed. Let me know  your preference!

Kevinz857:
@ChenYi015  Thanks for the feedback! I've updated this PR to use the feature gate mechanism from #2794.

  Changes made:

  1. Added PartialRestart feature gate in pkg/features/features.go:
    - Default: false
    - PreRelease: Alpha
   
  2. Updated event_filter.go to check the feature gate before skipping reconcile:
  ```go
  if features.Enabled(features.PartialRestart) && f.isWebhookPatchedFieldsOnlyChange(oldApp, newApp) {
      // skip reconcile logic
  }
```

  3. Rebased on latest master (includes #2794)

  Users can enable this feature by setting `--feature-gates=PartialRestart=true` on the controller.

  Please take another look when you have a chance!

**Comment by ChenYi015** [comment]
@Kevinz857  To make it work, we also need to:

1. Add a new CLI flag like `--feature-gates`, an example usage can be `--feature-gates=feature1=true,feature2=false`.
2. Add new field `featureGates` to Helm chart `values.yaml`, for example:

    ```yaml
    featureGates:
    - name: feature1
      enable: true
    - name: feature2
      enable: false
    ```
3. Update controller deployment Helm template to include `--feature-gates` flag when `featureGates` are specified.
4. Add associated Helm chart unit tests.

**Comment by Kevinz857** [comment]
@ChenYi015  Thanks for the detailed feedback! I've implemented all the requested changes:

  Changes made:

  1. CLI flag ```--feature-gates``` - Added support in ```cmd/operator/controller/start.go``` using 
 ```go
utilfeature.DefaultMutableFeatureGate.AddFlag()
```
  2. Helm chart values.yaml - Added featureGates field:
```go
  controller:
    featureGates:
    - name: PartialRestart
      enabled: true
```
  3. Helm deployment template - Updated to include ```--feature-gates``` flag when featureGates are specified
  4. Helm chart unit tests - Added tests for:
    - Default case (no feature gates)
    - Single feature gate
    - Multiple feature gates
  5. Helm documentation - Updated README.md via helm-docs

  Usage example:
```bash
  helm install spark-operator spark-operator/spark-operator \
      --set 'controller.featureGates[0].name=PartialRestart,controller.featureGates[0].enabled=true'
```
  Please take another look when you have a chance. Thanks!

ChenYi015 (approved):

ChenYi015:
@Kevinz857 Thanks for this great contribution!
/approve

PR #2787, opened 2025-12-21
feat: hot update Service/Ingress in Running state

Kevinz857:
## Summary

  Allow updating Service and Ingress resources without restarting the application when only service/ingress related fields change.

  When the application is in Running state and only service/ingress fields change, the controller will update the Service/Ingress resources directly without transitioning to Invalidating state.

  ## Changes

  - Add `isServiceIngressFieldsOnlyChange()` to detect service/ingress-only changes
  - Skip setting Invalidating state for service/ingress-only changes in Running state
  - Add `updateServiceIngressResources()` to update resources in Running state
  - Add comprehensive unit tests for change detection logic

  ## Supported Fields

  The following fields can now be updated without restarting the application:
  - `SparkUIOptions` (ServiceAnnotations, ServiceLabels, IngressAnnotations, IngressTLS, etc.)
  - `DriverIngressOptions` (ServiceAnnotations, ServiceLabels, IngressAnnotations, IngressTLS, etc.)
  - `Driver.ServiceAnnotations`
  - `Driver.ServiceLabels`

  ## How It Works

  1. When spec changes are detected in event filter, check if only service/ingress fields changed
  2. If application is in Running state and only these fields changed, don't set state to Invalidating
  3. Trigger reconcile which will update Service/Ingress resources using existing create-or-update logic
  4. Application continues running without interruption

  ## Test Plan

  - [x] Unit tests for `isServiceIngressFieldsOnlyChange()` function
  - [x] Unit tests for `hasServiceIngressFieldChanges()` function
  - [x] All existing tests pass
  - [ ] Manual test: Update ServiceAnnotations while app is running

  Partial fix for #2766

### DIFF
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -469,6 +469,15 @@ func (r *Reconciler) reconcileRunningSparkApplication(ctx context.Context, req c
 				return err
 			}
 
+			// Update Service/Ingress resources if needed.
+			// These functions use create-or-update pattern, so they will update
+			// existing resources if configuration has changed.
+			if err := r.updateServiceIngressResources(ctx, app); err != nil {
+				logger.Error(err, "Failed to update Service/Ingress resources")
+				// Don't return error to avoid blocking the reconcile loop.
+				// Service/Ingress updates are best-effort during running state.
+			}
+
 			if err := r.updateSparkApplicationStatus(ctx, app); err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


internal/controller/sparkapplication/controller.go
```diff
@@ -469,6 +469,15 @@ func (r *Reconciler) reconcileRunningSparkApplication(ctx context.Context, req c
 				return err
 			}
 
+			// Update Service/Ingress resources if needed.
+			// These functions use create-or-update pattern, so they will update
+			// existing resources if configuration has changed.
+			if err := r.updateServiceIngressResources(ctx, app); err != nil {
+				logger.Error(err, "Failed to update Service/Ingress resources")
```
nabuskey, line 476: can we put this in events only or events and controller log? Makes debugging issues related to services easier.

nabuskey (commented):
This supports creating and updating only correct? Are there reasons for not supporting deletions?

Kevinz857:
@nabuskey   Thanks for the review!

  **Regarding deletions:**
  You're right, the current implementation only handles create/update.I intentionally kept the scope limited for the first iteration because:
1. Deletion logic is more complex - we need to track which resources were created by previous configurations
2. The current create-or-update pattern follows existing codebase patterns
3. Wanted to get feedback before expanding scope

  I can add deletion support if you think it's necessary for this PR. Or we could track it as a follow-up enhancement?

  **Regarding events:**
  Good suggestion! I'll add Kubernetes events when Service/Ingress resources are updated. Something like:
  - `ServiceUpdated` / `IngressUpdated` for successful updates
  - `ServiceUpdateFailed` / `IngressUpdateFailed` for errors

  Will push an update soon. Thanks!

nabuskey:
Please let us know when it's ready for reviews.

Kevinz857:
@nabuskey  Ready for review. I've added Kubernetes events for Service/Ingress updates:
  - SparkUIServiceUpdated / SparkUIServiceUpdateFailed
  - SparkUIIngressUpdated / SparkUIIngressUpdateFailed
  - SparkDriverIngressServiceUpdated / SparkDriverIngressServiceUpdateFailed
  - SparkDriverIngressUpdated / SparkDriverIngressUpdateFailed

internal/controller/sparkapplication/event_filter.go
```diff
@@ -184,6 +184,24 @@ func (f *EventFilter) Update(e event.UpdateEvent) bool {
...
+			isServiceIngressFieldsOnlyChange(oldApp, newApp) {
+			f.logger.Info("Only Service/Ingress fields changed, updating resources without restart",
+				"name", newApp.Name, "namespace", newApp.Namespace)
+			f.recorder.Eventf(
+				newApp,
+				corev1.EventTypeNormal,
+				"SparkApplicationServiceUpdate",
+				"SparkApplication %s Service/Ingress configuration updated",
+				newApp.Name,
+			)
```
nabuskey: I don't think events here  is necessary. This is just a predicate.

internal/controller/sparkapplication/event_filter.go
```diff
@@ -184,6 +184,24 @@ func (f *EventFilter) Update(e event.UpdateEvent) bool {
 			return true
 		}
 
+		// Check if only Service/Ingress related fields changed while application is running.
+		// These resources can be updated without restarting the application.
+		if newApp.Status.AppState.State == v1beta2.ApplicationStateRunning &&
+			isServiceIngressFieldsOnlyChange(oldApp, newApp) {
+			f.logger.Info("Only Service/Ingress fields changed, updating resources without restart",
```
nabuskey: Info level log is likely not needed here. Debug logging instead?

internal/controller/sparkapplication/event_filter.go
```diff
@@ -245,6 +263,76 @@ func (f *EventFilter) filter(app *v1beta2.SparkApplication) bool {
...
+	clearServiceIngressFields(oldCopy)
+	clearServiceIngressFields(newCopy)
+
+	// Also zero out Suspend field as it's handled separately
+	oldCopy.Spec.Suspend = nil
+	newCopy.Spec.Suspend = nil
+
+	// If specs are equal after clearing service/ingress fields,
+	// then only service/ingress fields changed
+	return equality.Semantic.DeepEqual(oldCopy.Spec, newCopy.Spec)
```
nabuskey: These bits are confusing imo. If we know which fields corresponds to ingress and service object changes, do we need to copy the objects? 
I am concerned about DeepCopy()'s performance implications too. If you have a large deployment, it could added up quick with it being called

nabuskey (commented):
Thanks for the update! Please see my comments.

Kevinz857:
@nabuskey  Thanks for the review! I've addressed all your comments:

  1. **Removed event from predicate** - You're right, the predicate is not the right place for events. Removed the `f.recorder.Eventf()` call from `event_filter.go`.

  2. **Changed log level to Debug** - Changed `f.logger.Info()` to `f.logger.V(1).Info()` for debug level logging.

  3. **Optimized DeepCopy for better performance** - Instead of copying the entire `SparkApplication`, now only copying the `Spec` portion:
     ```go
     // Before: copied entire SparkApplication
     oldCopy := oldApp.DeepCopy()

     // After: only copy Spec for better performance
     oldSpec := oldApp.Spec.DeepCopy()
     This reduces memory allocation and improves performance in large deployments.

  Ready for another round of review. Thanks!
  ```

nabuskey:
> Optimized DeepCopy for better performance - Instead of copying the entire SparkApplication, now only copying the Spec portion:

My question was around avoiding deep copy entirely. Do we need to deep copy or can we avoid it for performance gains?

PR #2810, opened 2026-01-15
Fix SparkConnect nil executor template panic

### LINKED ISSUE #2807: [SparkConnect] Nil pointer dereference in imageOption when Executor.Template is not specified
### What happened?

The SparkConnect controller panics with a nil pointer dereference when reconciling a SparkConnect resource that does not specify `spec.executor.template`.

In `internal/controller/sparkconnect/options.go`, the `imageOption` function accesses `template.Spec.Containers` without checking if `template` is nil:

### Reproduction Code

1. Deploy Spark Operator with SparkConnect controller enabled:
```bash
helm install spark-operator spark-operator/spark-operator \
  --set controller.namespaces="{default}" \
  --set controller.image.tag=local \
  --set controller.image.pullPolicy=Never
```

2. Create a minimal SparkConnect resource without executor template:
```yaml
apiVersion: sparkoperator.k8s.io/v1alpha1
kind: SparkConnect
metadata:
  name: test-spark
  namespace: default
spec:
  sparkVersion: "3.5.0"
  image: apache/spark:3.5.0
  server: {}
  executor:
    instances: 2
```

3. Apply the resource:
```bash
kubectl apply -f sparkconnect.yaml
```

4. Check controller logs:
```bash
kubectl logs -l app.kubernetes.io/component=controller -n default
```

## Expected Behavior
The controller should handle missing `executor.template` gracefully and use default values.

## Actual Behavior
Controller panics with:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Environment & Versions

- Kubernetes Version:
- Spark Operator Version:
- Apache Spark Version:

### Additional context

_No response_

### Impacted by this bug?

sameerdattav:
## What this PR does

Fixes #2807
This PR makes SparkConnect image resolution nil-safe and adds regression coverage.

## Controller fix

- Adds a nil check for `conn.Spec.Executor.Template` before accessing `template.Spec.Containers`
- Uses a local `template` variable consistently when iterating over containers

## Preserved behavior

- If an executor container image is specified in the template, it overrides `spec.image`
- If no executor template is provided, the image falls back to `spec.image` (or empty if unset)

## Tests

Adds a regression/unit test in `options_test.go` that:
- Creates a `SparkConnect` with `spec.image` set and no `spec.executor.template`
- Calls `imageOption`

### Assertions

- No error is returned
- Both of the following are set to `spec.image`:
  - `spark.kubernetes.container.image`
  - `spark.kubernetes.executor.container.image`

## Verification

```bash
go test ./internal/controller/sparkconnect

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=docker.io/library/spark:4.0.0
+ARG SPARK_IMAGE=docker.io/library/spark:4.0.1
 
 FROM golang:1.24.10 AS builder
 
--- a/config/samples/v1beta2_scheduledsparkapplication.yaml
+++ b/config/samples/v1beta2_scheduledsparkapplication.yaml
@@ -11,11 +11,11 @@ spec:
   template:
     type: Scala
     mode: cluster
-    image: docker.io/library/spark:4.0.0
+    image: docker.io/library/spark:4.0.1
     imagePullPolicy: IfNotPresent
     mainClass: org.apache.spark.examples.SparkPi
     mainApplicationFile: local:///opt/spark/examples/jars/spark-examples.jar
-    sparkVersion: 4.0.0
+    sparkVersion: 4.0.1
     restartPolicy:
       type: Never
     driver:
--- a/config/samples/v1beta2_sparkapplication.yaml
+++ b/config/samples/v1beta2_sparkapplication.yaml
@@ -8,11 +8,11 @@ metadata:
 spec:
   type: Scala
   mode: cluster
-  image: docker.io/library/spark:4.0.0
+  image: docker.io/library/spark:4.0.1
   imagePullPolicy: IfNotPresent
   mainClass: org.apache.spark.examples.SparkPi
   mainApplicationFile: local:///opt/spark/examples/jars/spark-examples.jar
-  sparkVersion: 4.0.0
+  sparkVersion: 4.0.1
   driver:
     labels:
       version: 4.0.0
--- a/examples/spark-pi-configmap.yaml
+++ b/examples/spark-pi-configmap.yaml
@@ -21,11 +21,11 @@ metadata:
 spec:
   type: Scala
   mode: cluster
-  image: docker.io/library/spark:4.0.0
+  image: docker.io/library/spark:4.0.1
   imagePullPolicy: IfNotPresent
   mainClass: org.apache.spark.examples.SparkPi
   mainApplicationFile: local:///opt/spark/examples/jars/spark-examples.jar
-  sparkVersion: 4.0.0
+  sparkVersion: 4.0.1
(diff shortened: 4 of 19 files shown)

### CODE REVIEW

File: internal/controller/sparkconnect/options.go [hunk 1]
```diff
@@ -83,9 +83,9 @@ func imageOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 	}
 
 	template := conn.Spec.Executor.Template
-	if len(template.Spec.Containers) != 0 {
+	if template != nil && len(template.Spec.Containers) != 0 {
 		index := 0
```
Shekharrajak, line 87:
```suggestion
		index := -1
```
If we did not find `if container.Name == common.Spark3DefaultExecutorContainerName` condition then this check will help to know there was no index found in containers list.

Shekharrajak: In line 99 I think we need to throw exception 

```
if executorImage == "" {
    return nil, fmt.Errorf("executor image is not specified")
}
```

Shekharrajak:
Can you also please compare how such cases are handled in spark application yaml ?

sameerdattav:
@Shekharrajak  I’ve pushed the suggested changes.

I checked SparkApplication (`submission.go`) as well — if no image is set, it doesn’t error and doesn’t set any image config, so failures happen later (more permissive).
In SparkConnect, we already error immediately when the image is not set (stricter), so I kept the executor logic consistent with that.

internal/controller/sparkconnect/options_test.go
```diff
@@ -38,23 +38,34 @@ var _ = Describe("Options functions", func() {
                 Spec: v1alpha1.SparkConnectSpec{
                     SparkVersion: "3.5.0",
                     Image:        &image,
-                    Server: v1alpha1.ServerSpec{},
-                    Executor: v1alpha1.ExecutorSpec{},
+                    Server:       v1alpha1.ServerSpec{},
```
Shekharrajak: let's avoid these kind of unnecessary changes

sameerdattav: This is a new test file added in this PR. I didn’t intentionally change anything here — not sure why Git is showing this as a change from the last commit.

ChenYi015:
@sameerdattav Need to run `make go-fmt` to format the code and commit the changes to pass the CI.

sameerdattav:
@ChenYi015 , Thanks for pointing that out!
I ran the command and comitted, kindly check again.
Thank you!

sameerdattav:
@ChenYi015 , I've noticed that multiple ci checks were failing and made some changes,
Kindly check now.
Thank you.

ChenYi015:
@sameerdattav Could you check what happened to your code base? I can see that there are lots of unnessary file changes.
image

PR #2816, opened 2026-01-22
feat: add ingress-agnostic mode for improved compatibility with various ingress controllers

### LINKED ISSUE #2781: Decouple Spark Operator from ingress-nginx

I would like the spark operator's Ingress management to be decoupled from ingress-nginx

### Why is this needed?

Ingress-nginx is deprecated and will be EOL in March 2026. After that, usage of it will disappear

Currently, the Ingress objects are created by the Operator in runtime using some parameters inputted to the helm chart. For instance inputting this:

```
  uiIngress:
    enable: true
    ingressClassName: traefik

    urlFormat: /spark

    annotations:
      kubernetes.io/tls-acme: "true"
      traefik.ingress.kubernetes.io/router.middlewares: "oauth2-proxy-oauth-errors@kubernetescrd,traefik-oauth2-proxy-auth@kubernetescrd"

```

Will yield an Ingress object like this:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/tls-acme: "true"
    nginx.ingress.kubernetes.io/rewrite-target: /$2
    traefik.ingress.kubernetes.io/router.middlewares: oauth2-proxy-oauth-errors@kubernetescrd,traefik-oauth2-proxy-auth@kubernetescrd
  name: spark-pi2-ui-ingress
spec:
  ingressClassName: traefik
  rules:
  - http:
      paths:
      - backend:
          service:
            name: spark-pi-ui-svc
            port:
              number: 4040
        path: /spark(/|$)(.*)
        pathType: ImplementationSpecific
```

Note that the field "path:" has been altered from the helm value input, and the "rewrite-target" annotation have been added. For my specific setup, the annotation doesn't do anthing and the path attribute is broken (it's very Implementation Specific to ingress-nginx).

For my specific case I'm using traefik, which has two requirements: 1) path is not a regex, and 2) i need to be able to create a "Middleware" object for path rewriting.

I'm not suggesting that spark-operator should accomodate traefik specifically, just that it should get out of the way, and let the cluster operator do the setup that's needed for their cluster.

**Suggestion:**

1. Have the "Ingress" object be created in the helm template rendering, *not* at runtime by the operator
2. Don't alter the path input
3. Make sure all the requests go into the "controller", and make sure that it in turn proxies the response from the active Application Pods
   
   Jupyterhub operator does something similar; the ingress goes into a "proxy" Pod like so:
   
```
  rules:
  - host: <host>
    http:
      paths:
      - backend:
          service:
            name: proxy-public
            port:
              name: http
        path: /
        pathType: Prefix
```

_No response_

### Additional context

I'm not sure how important or useful the dashboard is in the spark-operator, since it's ephemeral and short-lived. This is just a suggestion from a k8s cluster operator on how to solve an existing issue.

### Love this feature?

Denyme24:
## Purpose of this PR
fixes #2781 

**Proposed changes:**

- added an opt-in ingress-agnostic mode (Helm value/CLI flag + per-SparkApplication override) to disable nginx-specific path regex mangling and `nginx.ingress.kubernetes.io/rewrite-target` injection.
- In agnostic mode: keep the user-supplied path as-is, use `PathType: Prefix`, and do not add nginx rewrite annotations; legacy mode remains the default for backward compatibility.
- Wire the option through controller (web UI + driver ingress), Helm values/deployment, and added tests covering both legacy and agnostic modes; regenerate API types/CRDs as needed.

## Change Category

- [ ] Documentation update

## Checklist


Denyme24:
hey team @ChenYi015 @nabuskey, please review the changes once.
thank you.

PR #2817, opened 2026-01-24
feat: Expose metrics-labels flag via Spark Operator Helm chart

IgorSteps:
## Purpose of this PR

Expose the Spark Operator's `metrics-labels` flag as a Helm chart value so users can add custom labels (for example using standard K8s labels) to the Spark Operator standard metrics. Right now, the chart hard codes that flag to `app_type` so the [documented](https://www.kubeflow.org/docs/components/spark-operator/getting-started/#:~:text=The%20following%20is,of%20these%20metrics.) ability to add your own label keys is not actually possible by developers using your Helm chart. 

This makes it much easier to build practical monitoring on whole platforms that use Spark Operator. Developers would be able to filter your standard metrics by e.g., namespace, app, component or team. They would have the ability to build common dashboards and setup targeted alerts without coming with elaborate ways to include enough information in the only available `app_type` label to achieve that instead. 

**Proposed changes:**

- Adds `prometheus.metrics.labels` value to the values.yaml;
- Uses it in the deployment if specified, otherwise defaults to `app_type` for backwards compatibility;
- Helm chart automatic doc updates and unit test.

## Change Category

- [ ] Documentation update

### Rationale

This keeps the current `app_type` value as default but lets users define their own custom labels.

## Checklist

### Additional Notes

None.

### DIFF
--- a/charts/spark-operator-chart/Chart.yaml
+++ b/charts/spark-operator-chart/Chart.yaml
@@ -20,7 +20,7 @@ name: spark-operator
 
 description: A Helm chart for Spark on Kubernetes operator.
 
-version: 2.4.0
+version: 2.3.0
 
 appVersion: 2.3.0
 
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -1,6 +1,6 @@
 # spark-operator
 
-![Version: 2.4.0](https://img.shields.io/badge/Version-2.4.0-informational?style=flat-square) ![AppVersion: 2.3.0](https://img.shields.io/badge/AppVersion-2.3.0-informational?style=flat-square)
+![Version: 2.3.0](https://img.shields.io/badge/Version-2.3.0-informational?style=flat-square) ![AppVersion: 2.3.0](https://img.shields.io/badge/AppVersion-2.3.0-informational?style=flat-square)
 
 A Helm chart for Spark on Kubernetes operator.
 


charts/spark-operator-chart/Chart.yaml
```diff
@@ -20,7 +20,7 @@ name: spark-operator
 
 description: A Helm chart for Spark on Kubernetes operator.
 
-version: 2.3.0
```
ChenYi015, line 23: Just leave `version` unmodified, we will bump it in a dedicated PR when needed.

charts/spark-operator-chart/README.md
```diff
@@ -1,6 +1,6 @@
 # spark-operator
 
-![Version: 2.3.0](https://img.shields.io/badge/Version-2.3.0-informational?style=flat-square) ![AppVersion: 2.3.0](https://img.shields.io/badge/AppVersion-2.3.0-informational?style=flat-square)
```
ChenYi015, line 3: Helm docs also need to be updated when chart version is changed back to `2.3.0`.

ChenYi015 (approved):
/approve

ChenYi015:
@IgorSteps You need to sign off your commits to pass the DCO checks, see https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits.

ChenYi015 (approved):
/approve

PR #2835, opened 2026-02-03
fix: allow same ServicePort with different hosts in driverIngressOptions

### LINKED ISSUE #2662: [BUG] Unable to set the same port for different hosts in driverIngressOptions
### What happened?

There is the rule to deny such behavior but how to route traffic from different hosts to the same port?
I also couldn't find any reason for this requirement. Does anyone know the reason for the restriction?

```go
if servicePorts[*item.ServicePort] {
    return fmt.Errorf("DriverIngressOptions has duplicate ServicePort: %d", *item.ServicePort)
}
```
It seems like there is no way to solve this with the current functionality.

### Reproduction Code

This is how this feature was planned to be used:
```yaml
driverIngressOptions:
  - servicePort: 4040
    servicePortName: "spark-ui-1"
    ingressURLFormat: "private.domain.com"
  - servicePort: 4040
    servicePortName: "spark-ui-2"
    ingressURLFormat: "public.domain.com"
```
Port 4040 is selected as the default port opened for web UI
I also set `- --ingress-class-name=nginx`

### Expected behavior

Spark job web UI can be reached from both private.domain.com and public.domain.com

### Actual behavior

Raises an error: DriverIngressOptions has duplicate ServicePort

### Environment & Versions

- Kubernetes Version: 1.25.11
- Spark Operator Version: 2.3.0
- Apache Spark Version: 3.5.0

### Additional context

_No response_

### Impacted by this bug?

puwun:
Fixes #2662

## Purpose of this PR
Users want to expose the same Spark driver port (e.g 4040 for Spark UI) through multiple ingresses with different hostnames. The current validation incorrectly rejects this configuration as "duplicate ServicePort".

**Proposed changes:**

- Modified validation to check `(ServicePort, IngressURLFormat)` pairs instead of just `ServicePort`, allowing same port with different hosts
- Updated service and ingress naming to include index for multiple entries with the same port to ensure unique resource names
- Added tests for the new behavior

## Change Category

## Checklist

### Additional Notes

With this fix, the following configuration is now valid:

```yaml
driverIngressOptions:
  - servicePort: 4040
    ingressURLFormat: "private.domain.com"
  - servicePort: 4040
    ingressURLFormat: "public.domain.com"
```

### DIFF
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -377,8 +377,8 @@ func (r *Reconciler) reconcileSubmittedSparkApplication(ctx context.Context, req
 				}
 			}
 
-			for _, driverIngressConfiguration := range app.Spec.DriverIngressOptions {
-				service, err := r.createDriverIngressServiceFromConfiguration(ctx, app, &driverIngressConfiguration)
+			for i, driverIngressConfiguration := range app.Spec.DriverIngressOptions {
+				service, err := r.createDriverIngressServiceFromConfiguration(ctx, app, &driverIngressConfiguration, i)
 				if err != nil {
 					return fmt.Errorf("failed to create driver ingress service for SparkApplication: %v", err)
 				}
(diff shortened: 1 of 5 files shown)


nabuskey (commented):
Why do you want to use index for ingress name uniqueness?

PR #2866, opened 2026-03-05
fix: Patch memory resources being duplicated

CarterFendley:
## Purpose of this PR

This PR is to fix an error where memory requests are doubled for spark applications when quota enforcement is enabled. The changes added are in two main categories listed below.

**IMPORTANT NOTE:** Before merge, I will add some additional tests to validate this logic. Putting up the PR now for some early feedback.

**Proposed changes:**

1. Fix the memory resource validation
2. Add more information when resources are rejected
3. Other misc things
    - Add a default for when no memory is set [here](https://github.com/CarterFendley/spark-operator/blob/8755dc43b01052ed56e93928e8125b85e382c04a/internal/webhook/resourcequota.go#L217)

## Change Category

- [ ] Documentation update

### Rationale

For the changes respectively:
1. Resource quotas for memory do not seem to be evaluated correctly.
2. Add more informative information so a user can know which resource they are exceeding.

## Checklist

#### Change 2 - Add more information when resources are rejected.

First I will explain the new error message because that is simple. With a change to [validateResourceQuota](https://github.com/CarterFendley/spark-operator/blob/8755dc43b01052ed56e93928e8125b85e382c04a/internal/webhook/resourcequota.go#L283) and [validateResourceUsage](https://github.com/CarterFendley/spark-operator/blob/8755dc43b01052ed56e93928e8125b85e382c04a/internal/webhook/sparkapplication_validator.go#L186) the new error message looks like:
```
Error from server (Forbidden): error when creating "examples/spark-pi.yaml": admission webhook "validate-sparkoperator-k8s-io-v1beta2-sparkapplication.sparkoperator.k8s.io" denied the request: failed to validate resource quota "default/compute-quota" with error "Exceeded 'memory' quota: requested=3584Mi, used=0, limit=2Gi (new total would be: 3584Mi)
```
This matches the native Spark K8 integration more closely and removes guess work.

#### Change 1 - Fix the memory resource validation.

At a high level, because
- `getSparkPodMemoryRequests` was setting `memory` and `requests.memory` [here](https://github.com/CarterFendley/spark-operator/blob/610ec74d9df46b0aeee71bec53100768c8c06cb7/internal/webhook/resourcequota.go#L210-L213)
- `getMemoryLimits` reused this function [here](https://github.com/CarterFendley/spark-operator/blob/610ec74d9df46b0aeee71bec53100768c8c06cb7/internal/webhook/resourcequota.go#L219)

Then [this SumResourceList](https://github.com/CarterFendley/spark-operator/blob/610ec74d9df46b0aeee71bec53100768c8c06cb7/internal/webhook/resourcequota.go#L73-L78) will double `memory` and `requests.memory`.

This can be seen while running the `examples/spark-pi.yaml` after enabling `--enable-resource-quota-enforcement=true` and with the following resource quota:
```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: compute-quota
spec:
  hard:
    pods: "10"
    cpu: "2"
    memory: 2Gi
```
This is what produced the error message shown above.
```
Error from server (Forbidden): error when creating "examples/spark-pi.yaml": admission webhook "validate-sparkoperator-k8s-io-v1beta2-sparkapplication.sparkoperator.k8s.io" denied the request: failed to validate resource quota "default/compute-quota" with error "Exceeded 'memory' quota: requested=3584Mi, used=0, limit=2Gi (new total would be: 3584Mi)
```
As this error message states, the calculated requested memory is `3584Mi`. However this spark job only requests `512m` for the driver and a single executor (see [here](https://github.com/CarterFendley/spark-operator/blob/610ec74d9df46b0aeee71bec53100768c8c06cb7/examples/spark-pi.yaml#L35)). This plus the default overhead should give us a total application request of:
```
(512 + 384) * 2 = 1792
```
However due to the reasons shown above we are getting:
```
((512 + 384) * 2) * 2 = 3584
```

This PR changes `getSparkPodMemoryRequests` (renamed `getSparkPodMemory`) and `getMemoryRequests` (renamed `getMemory`) to return a simple `int64`. Then `getMemoryRequests` and `getMemoryLimits` set the specific resource name [here](https://github.com/CarterFendley/spark-operator/blob/8755dc43b01052ed56e93928e8125b85e382c04a/internal/webhook/resourcequota.go#L236-L255).

Because this will now produce `memory`, `requests.memory` *and* `limits.memory` without any key duplicated, then the doubling is removed.

*Another note:* This is my first commit to the spark-operator side of kubeflow so if there is any convention which I should be following and I have neglected, please let me know and I will make changes :+1:

### DIFF
--- a/internal/webhook/resourcequota.go
+++ b/internal/webhook/resourcequota.go
@@ -153,13 +153,16 @@ func getSparkPodCoresLimits(podSpec *v1beta2.SparkPodSpec, replicas int64) (core
 	return resourceList, nil
 }
 
-func getMemoryRequests(app *v1beta2.SparkApplication) (corev1.ResourceList, error) {
+func getMemory(
+	app *v1beta2.SparkApplication,
+	parseLimit bool,
+) (int64, error) {
 	// If memory overhead factor is set, use it. Otherwise, use the default value.
 	var memoryOverheadFactor float64
 	if app.Spec.MemoryOverheadFactor != nil {
 		parsed, err := strconv.ParseFloat(*app.Spec.MemoryOverheadFactor, 64)
 		if err != nil {
-			return nil, err
+			return -1, err
 		}
 		memoryOverheadFactor = parsed
 	} else if app.Spec.Type == v1beta2.SparkApplicationTypeJava {
@@ -169,9 +172,14 @@ func getMemoryRequests(app *v1beta2.SparkApplication) (corev1.ResourceList, erro
 	}
 
 	// Calculate driver pod memory requests.
-	driverResourceList, err := getSparkPodMemoryRequests(&app.Spec.Driver.SparkPodSpec, memoryOverheadFactor, 1)
+	driverMemory, err := getSparkPodMemory(
+		&app.Spec.Driver.SparkPodSpec,
+		memoryOverheadFactor,
+		1,
+		parseLimit,
+	)
 	if err != nil {
-		return nil, err
+		return -1, err
 	}
 
 	// Calculate executor pod memory requests.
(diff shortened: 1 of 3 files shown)


internal/webhook/sparkapplication_validator.go
```diff
@@ -112,6 +112,7 @@ func (v *SparkApplicationValidator) ValidateUpdate(ctx context.Context, oldObj r
 	// Validate SparkApplication resource usage when resource quota enforcement is enabled.
 	if v.enableResourceQuotaEnforcement {
 		if err := v.validateResourceUsage(ctx, newApp); err != nil {
+			logger.Info("Resource request validation failed: %s", err)
```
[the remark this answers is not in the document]
CarterFendley: I can switch to `.Error` if the team wants here, not really a fatal but could be considered and error depending on what people think.

internal/webhook/resourcequota.go
```diff
@@ -169,54 +172,86 @@ func getMemoryRequests(app *v1beta2.SparkApplication) (corev1.ResourceList, erro
...
-	executorResourceList, err := getSparkPodMemoryRequests(&app.Spec.Executor.SparkPodSpec, memoryOverheadFactor, replicas)
+	executorMemory, err := getSparkPodMemory(
+		&app.Spec.Executor.SparkPodSpec,
+		memoryOverheadFactor,
+		replicas,
+		parseLimit,
+	)
 	if err != nil {
-		return nil, err
+		return -1, err
```
vjanelle, line 197: why return -1?

CarterFendley: Could not be `nil` due to `int64` type, figured it was better than `0`.

internal/webhook/sparkapplication_validator.go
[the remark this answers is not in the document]
vjanelle: .Info doesn't do sprintf, IIRC, so this won't work the way you're expecting

ChenYi015:
Overall LGTM. The `logger.info` comment needs to be solved. It would be better if you can add corresponding unit tests to make sure it works.

CarterFendley:
@ChenYi015 Thanks for taking a look! Now that you have given the LGTM of the overall design, will definitely create tests :)

CarterFendley:
Okay!

The error message is now the one suggested by copilot, and I have [created a parameterized test](https://github.com/CarterFendley/spark-operator/blob/62eb7c256298467a52a388e46e2a482b47649c8b/internal/webhook/sparkapplication_validator_test.go#L217-L267) which tests under different memory quotas!

CarterFendley:
@ChenYi015 Can I get a re-review?

internal/webhook/resourcequota.go
nabuskey: I think it's a good idea to return a zero value to avoid having magic numbers

internal/webhook/resourcequota.go
```diff
@@ -169,54 +172,86 @@ func getMemoryRequests(app *v1beta2.SparkApplication) (corev1.ResourceList, erro
...
-		}
-		memoryBytes = parsed
+	var err error
+	if parseLimit && podSpec.MemoryLimit != nil {
+		memoryBytes, err = parseJavaMemoryString(*podSpec.MemoryLimit)
+	} else if podSpec.Memory != nil {
+		// Either in the case of parsing requests or if parsing limits but no limit is supplied
+		memoryBytes, err = parseJavaMemoryString(*podSpec.Memory)
+	} else {
+		memoryBytes = common.DefaultMemoryBytes
```
nabuskey, line 217: Do we need this? Right now, it uses the zero value of int64. We set the field in the defaulter so we shouldn't hit this. But if we hit this for some reason, the calculated values are different compared to what we get right now (`0 + default overhead` vs `defaultMemoryBytes + default overhead`).

PR #2868, opened 2026-03-06
fix: avoid duplicate web UI service/ingress creation during reconciliation

### LINKED ISSUE #2701: Double request to create ingress and service
### What happened?

While upgrading to v2.3.0 we have faced issue on Kyverno side due to attempts to create same ingress url multiple times.

Investigating more, we found that this issue happens intermittently while application is on submitted state, so wanted to ask about that as we have not faced that on the older version from Google or maybe we are missing some config on the new deployment

Sample effect of this, is about lifecycle transitions because this conflict is preventing SparkApplication from transitioning from Submitted state to Running state ending up being zombie and not being Garbage collected after `timeToLiveSeconds`

Below are screenshots of where we have faced that

Image
Image

### Reproduction Code

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Environment & Versions

- Kubernetes Version:
- Spark Operator Version:
- Apache Spark Version:

### Additional context

_No response_

### Impacted by this bug?

YassinNouh21:
## Purpose of this PR

Fixes #2701

While in `Submitted` state, reconciliation is triggered multiple times (driver pod created, pod phase transitions, status updates). Each reconciliation re-attempted `Create()` on the same web UI service and ingress, causing:
- Wasteful API calls (failed Create + Update) on every cycle
- Kyverno policy violations from the failed Create attempts before `AlreadyExists` is returned
- Potential zombie SparkApplications stuck in Submitted state

**Proposed changes:**

- Add a status guard (`WebUIServiceName == ""`) in `reconcileSubmittedSparkApplication` to skip web UI resource creation when resources were already created in a previous reconciliation
- Switch from `Create → catch AlreadyExists → Update` to `Get → if NotFound → Create, else → Update` pattern in `createDriverIngressService`, `createDriverIngressV1`, and `createDriverIngressLegacy` to avoid unnecessary failed API calls that trigger Kyverno policy violations

## Change Category

- [ ] Documentation update

### Rationale

The root cause is that `reconcileSubmittedSparkApplication` has no guard before creating web UI resources. Every reconciliation while in Submitted state re-runs the creation logic. The old `Create → catch AlreadyExists` pattern generates a failed Create API call before falling through to Update. With Kyverno policies, this failed Create triggers a policy violation before the `AlreadyExists` error is returned — causing the conflict reported in #2701.

The fix addresses both layers:
1. **Status guard**: Once `WebUIServiceName` is set in status, subsequent reconciliations skip the creation block entirely — zero unnecessary API calls
2. **Get-then-Create**: Even for `DriverIngressOptions` (which don't have a status guard), the Get-first pattern avoids the failed Create that triggers Kyverno violations. The Update path now properly merges onto the existing object (preserving `ResourceVersion`)

## Checklist

### Additional Notes

- All existing unit tests pass
- All 13 e2e tests pass on Kind cluster (k8s v1.32.0)
- All CI checks pass locally: `go mod tidy`, `go-fmt`, `go-vet`, `build-operator`

### DIFF
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -351,8 +351,7 @@ func (r *Reconciler) reconcileSubmittedSparkApplication(ctx context.Context, req
 			}
 
 			// Create web UI service for spark applications if enabled.
-			// Guard: skip if resources were already created in a previous reconciliation.
-			if r.options.EnableUIService && app.Status.DriverInfo.WebUIServiceName == "" {
+			if r.options.EnableUIService {
 				service, err := r.createWebUIService(ctx, app)
 				if err != nil {
 					return fmt.Errorf("failed to create web UI service: %v", err)
--- a/internal/controller/sparkapplication/driveringress.go
+++ b/internal/controller/sparkapplication/driveringress.go
@@ -171,9 +171,13 @@ func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.Spa
 		}
 		// Ingress does not exist, create it.
 		if err := r.client.Create(ctx, ingress); err != nil {
-			return nil, fmt.Errorf("failed to create ingress %s/%s: %v", ingress.Namespace, ingress.Name, err)
+			if !errors.IsAlreadyExists(err) {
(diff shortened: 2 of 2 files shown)


ChenYi015 (approved):

YassinNouh21:
@ChenYi015 will you merge it ?

PR #2878, opened 2026-03-14
fix: Add label selector to ConfigMap cache to prevent OOM via informer flooding

RobuRishabh:
## Purpose of this PR

Currently the operator is vulnerable to OOMKill via controller-runtime informer cache flooding.

The `ByObject` cache config in `cmd/operator/controller/start.go` lists `&corev1.ConfigMap{}: {}` with an explicitly **empty** config — no label selector, no field selector, no namespace restriction. This creates an unfiltered cluster-wide ConfigMap informer that caches ALL ConfigMaps in the cluster.

In contrast, Pods in the same `ByObject` map already correctly use a `LaunchedBySparkOperator` label selector. ConfigMaps had no such protection.

A user with permissions to create ConfigMaps can OOMKill the operator (512Mi limit) by pre-loading large ConfigMaps (~900KB each) across multiple namespaces. The unfiltered informer caches all of them, exceeding the memory limit.

**Proposed changes:**

- Add `LabelCreatedBySparkOperator` label selector to the existing `ByObject[ConfigMap]` cache entry, restricting the informer to only cache operator-managed ConfigMaps. This follows the same pattern already used for Pods.
- Add `LabelCreatedBySparkOperator` label to `buildPrometheusConfigMap()` so Prometheus monitoring ConfigMaps pass through the filtered cache.
- Handle the upgrade path: existing Prometheus ConfigMaps created before this fix (without the label) are gracefully updated with the label instead of causing `AlreadyExists` errors when the filtered cache returns `NotFound` for unlabeled ConfigMaps.
- Propagate the label during ConfigMap updates, not just creation.
- Add test coverage for the label assertion on all existing test cases and a new upgrade scenario test.

## Change Category

- [ ] Documentation update

### Rationale

Without this fix, any user with permissions to create ConfigMaps can OOMKill the operator by pre-loading large ConfigMaps (~900KB each) across multiple namespaces. The unfiltered cluster-wide ConfigMap informer caches all of them, exceeding the 512Mi memory limit (Exit Code 137). Adding a label selector restricts the cache to only operator-managed ConfigMaps, matching the existing pattern used for Pods.

The upgrade path must also be handled because Prometheus ConfigMaps created by previous operator versions lack the new label. After the upgrade, these ConfigMaps become invisible to the filtered cache (`client.Get` returns `NotFound`), but still exist in the cluster (`client.Create` returns `AlreadyExists`). The fix catches this case and updates the existing ConfigMap with the label.

## Checklist

### Additional Notes

### DIFF
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -415,7 +415,11 @@ func newCacheOptions() cache.Options {
 					common.LabelLaunchedBySparkOperator: "true",
 				}),
 			},
-			&corev1.ConfigMap{}:                  {},
+			&corev1.ConfigMap{}: {
+				Label: labels.SelectorFromSet(labels.Set{
+					common.LabelCreatedBySparkOperator: "true",
+				}),
+			},
 			&corev1.PersistentVolumeClaim{}:      {},
 			&corev1.Service{}:                    {},
 			&v1beta2.SparkApplication{}:          {},
--- a/internal/controller/sparkapplication/monitoring_config.go
+++ b/internal/controller/sparkapplication/monitoring_config.go
@@ -33,7 +33,7 @@ import (
 	"github.com/kubeflow/spark-operator/v2/pkg/util"
 )
 
-func configPrometheusMonitoring(ctx context.Context, app *v1beta2.SparkApplication, client client.Client) error {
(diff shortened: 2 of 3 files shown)


nabuskey (approved):
LGTM

andreyvelich (commented):
/approve

PR #2898, opened 2026-03-30
fix(rbac): Replace hardcoded namespace: default with namespace: system

RobuRishabh:
## Purpose of this PR

The `config/rbac/spark-application-rbac.yaml` file hardcodes `namespace: default` in its ServiceAccount, Role, RoleBinding, and RoleBinding subjects. This locks deployments to the `default` namespace and prevents deployers from controlling the target namespace via Kustomize, Helm, or `kubectl apply -n`.

**Proposed changes:**

- Remove `namespace: default` from ServiceAccount metadata
- Remove `namespace: default` from Role metadata
- Remove `namespace: default` from RoleBinding metadata
- Remove `namespace: default` from RoleBinding subjects

## Change Category

- [ ] Documentation update

### Rationale

Hardcoding `namespace: default` breaks multi-tenant and custom-namespace installations. Base manifests should omit explicit namespace fields and let the deployer decide, which is the standard Kubernetes/kubebuilder convention.

## Checklist

### DIFF
--- a/config/rbac/spark-application-rbac.yaml
+++ b/config/rbac/spark-application-rbac.yaml
@@ -17,12 +17,10 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: spark-operator-spark
-  namespace: default
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  namespace: default
   name: spark-role
 rules:
 - apiGroups:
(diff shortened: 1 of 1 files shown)


nabuskey (approved):
LGTM

nabuskey (changes requested):
Actually we should just remove explicit namespace references. Namespaces are typically specified in kustomization files, correct?

RobuRishabh:
> Actually we should just remove explicit namespace references. Namespaces are typically specified in kustomization files, correct?

@nabuskey Thanks for the review, I've updated the PR to remove the explicit `namespace` fields entirely instead of replacing them with `system`.

nabuskey (approved):
LGTM

nabuskey (approved):
LGTM

vara-bonthu (approved):
/approve

Issue #2905, opened 2026-04-08
SparkApplication can get stuck in PENDING_RERUN because retry reconciliation does not requeue when cleanup is incomplete


skaushik15:
### What happened?

A SparkApplication failed and the operator moved it to PENDING_RERUN. The operator deleted the driver pod and web UI Service, but one reconcile pass logged 

> Resources associated with SparkApplication still exist 

and then stopped. The SparkApplication remained stuck in PENDING_RERUN and was never retried.

Logs for refernece:
`
63:2026-04-08T00:30:32.023Z	INFO	sparkapplication/controller.go:194	Reconciling SparkApplication	{"controller": "spark-application-controller", "namespace": "harness-helm-new", "name": "hlu-ledger-1775608200001434796", "reconcileID": "4907d4d1-bbaa-4c2d-842c-e4304f5d28a4", "state": "RUNNING"}
64:2026-04-08T00:30:32.042Z	INFO	sparkapplication/controller.go:218	Finished reconciling SparkApplication	{"controller": "spark-application-controller", "namespace": "harness-helm-new", "name": "hlu-ledger-1775608200001434796", "reconcileID": "4907d4d1-bbaa-4c2d-842c-e4304f5d28a4"}
66:2026-04-08T00:30:32.474Z	INFO	sparkapplication/controller.go:194	Reconciling SparkApplication	{"controller": "spark-application-controller", "namespace": "harness-helm-new", "name": "hlu-ledger-1775608200001434796", "reconcileID": "dd9aa0b0-6632-4676-b85a-0a2a42318588", "state": "RUNNING"}
67:2026-04-08T00:30:32.521Z	INFO	sparkapplication/controller.go:218	Finished reconciling SparkApplication	{"controller": "spark-application-controller", "namespace": "harness-helm-new", "name": "hlu-ledger-1775608200001434796", "reconcileID": "dd9aa0b0-6632-4676-b85a-0a2a42318588"}
68:2026-04-08T00:30:33.956Z	INFO	sparkapplication/event_handler.go:87	Spark pod updated	{"name": "hlu-ledger-1775608200001434796-driver", "namespace": "harness-helm-new", "oldPhase": "Running", "newPhase": "Failed"}
[...]

PR #2906, opened 2026-04-08
fix: requeue pending rerun while resources exist

skaushik15:
Please refer https://github.com/kubeflow/spark-operator/issues/2905 for more details
## Summary
- avoid prematurely finalizing `PENDING_RERUN` when the previous application resources still exist
- keep reconciliation in `PENDING_RERUN` by returning a requeue result until old resources are cleaned up
- add controller tests to verify the requeue behavior and guard against regressions

## Test plan
- [x] unit tests for pending rerun path added/updated
- [x] validate controller behavior in `PENDING_RERUN` when resources exist

### DIFF
--- a/api/v1beta2/sparkapplication_types.go
+++ b/api/v1beta2/sparkapplication_types.go
@@ -107,7 +107,8 @@ type SparkApplicationSpec struct {
 	// This is best effort and actual retry attempts can be >= the value specified.
 	// +optional
 	FailureRetries *int32 `json:"failureRetries,omitempty"`
-	// RetryInterval is the unit of intervals in seconds between submission retries.
+	// RetryInterval is the unit of intervals in seconds between retry/poll attempts
+	// in controller-managed loops (for example submission retries and pending rerun cleanup polling).
 	// +optional
 	RetryInterval *int64 `json:"retryInterval,omitempty"`
 	// This sets the major Python version of the docker
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -52,6 +52,8 @@ import (
 	"github.com/kubeflow/spark-operator/v2/pkg/util"
 )
 
+const pendingRerunRequeueDelay = 5 * time.Second
+
 // Options defines the options of the controller.
 type Options struct {
 	Namespaces            []string
@@ -500,6 +502,7 @@ func (r *Reconciler) reconcileRunningSparkApplication(ctx context.Context, req c
 func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 	key := req.NamespacedName
+	var result ctrl.Result
 	retryErr := retry.RetryOnConflict(
 		retry.DefaultRetry,
 		func() error {
@@ -518,7 +521,13 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context, 
 				r.recordSparkApplicationEvent(app)
 				r.submitSparkApplication(ctx, app)
 			} else {
-				logger.Info("Resources associated with SparkApplication still exist")
+				// Retry cleanup in case deletion is only partially complete, for example if the driver pod is gone
+				// but the UI Service is still terminating.
+				logger.Info("Resources associated with SparkApplication still exist, retrying deletion")
+				if err := r.deleteSparkResources(ctx, app); err != nil {
+					logger.Error(err, "failed to delete spark resources")
+				}
+				result.RequeueAfter = pendingRerunRequeueDelay
 			}
 			if err := r.updateSparkApplicationStatus(ctx, app); err != nil {
 				return err
@@ -528,9 +537,9 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context, 
 	)
 	if retryErr != nil {
 		logger.Error(retryErr, "Failed to reconcile SparkApplication")
-		return ctrl.Result{}, retryErr
+		return result, retryErr
 	}
-	return ctrl.Result{}, nil
+	return result, nil
 }
 
 func (r *Reconciler) reconcileInvalidatingSparkApplication(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -523,31 +523,22 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context, 
 				if app.Status.TerminationTime.IsZero() {
 					app.Status.TerminationTime = metav1.Now()
 				}
-				if util.PendingRerunCleanupRetryBudgetExceeded(app) {
-					app.Status.TerminationTime = metav1.Now()
-					app.Status.AppState.State = v1beta2.ApplicationStateFailed
-					app.Status.AppState.ErrorMessage = "failed to delete resources before rerun after exhausting cleanup retries"
-					r.recordSparkApplicationEvent(app)
-				} else {
-					timeUntilNextRetryDue, err := util.TimeUntilNextRetryDue(app)
-					if err != nil {
-						app.Status.TerminationTime = metav1.Now()
-						app.Status.AppState.State = v1beta2.ApplicationStateFailed
-						app.Status.AppState.ErrorMessage = fmt.Sprintf("failed to determine pending rerun cleanup retry: %v", err)
-						r.recordSparkApplicationEvent(app)
-					} else if timeUntilNextRetryDue <= 0 {
-						// Retry cleanup in case deletion is only partially complete, for example if the driver pod is gone
-						// but the UI Service/executors are still terminating.
-						logger.Info("Resources associated with SparkApplication still exist, retrying deletion")
-						if err := r.deleteSparkResources(ctx, app); err != nil {
-							logger.Error(err, "failed to delete spark resources")
-						}
-						result.Requeue = true
-					} else {
-						// Keep polling until Kubernetes GC has finished removing all resources associated with the app.
-						logger.Info("Resources associated with SparkApplication still exist, waiting for GC to complete")
-						result.RequeueAfter = timeUntilNextRetryDue
+				timeUntilNextRetryDue, err := util.TimeUntilNextRetryDue(app)
+				if err != nil {
+					// If no retry interval can be computed for this state, requeue with default rate limiter backoff.
+					result.Requeue = true
+				} else if timeUntilNextRetryDue <= 0 {
+					// Retry cleanup in case deletion is only partially complete, for example if the driver pod is gone
+					// but the UI Service/executors are still terminating.
+					logger.Info("Resources associated with SparkApplication still exist, retrying deletion")
+					if err := r.deleteSparkResources(ctx, app); err != nil {
+						logger.Error(err, "failed to delete spark resources")
 					}
+					result.Requeue = true
+				} else {
+					// Keep polling until Kubernetes GC has finished removing all resources associated with the app.
+					logger.Info("Resources associated with SparkApplication still exist, waiting for GC to complete")
+					result.RequeueAfter = timeUntilNextRetryDue
 				}
 			}
 			if err := r.updateSparkApplicationStatus(ctx, app); err != nil {
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -52,7 +52,11 @@ import (
 	"github.com/kubeflow/spark-operator/v2/pkg/util"
 )
 
-const pendingRerunRequeueDelay = 5 * time.Second
+const (
+	pendingRerunRequeueDelay = 5 * time.Second
+	// Keep PendingRerun cleanup retry budget aligned with the default Pod termination grace period (30s).
+	pendingRerunCleanupRetryWindow = 30 * time.Second
+)
 
 // Options defines the options of the controller.
 type Options struct {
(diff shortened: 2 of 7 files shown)


vizsmurali (approved):

internal/controller/sparkapplication/controller.go
```diff
@@ -518,7 +521,13 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context,
...
 				r.submitSparkApplication(ctx, app)
 			} else {
-				logger.Info("Resources associated with SparkApplication still exist")
+				// Retry cleanup in case deletion is only partially complete, for example if the driver pod is gone
+				// but the UI Service is still terminating.
+				logger.Info("Resources associated with SparkApplication still exist, retrying deletion")
+				if err := r.deleteSparkResources(ctx, app); err != nil {
+					logger.Error(err, "failed to delete spark resources")
+				}
+				result.RequeueAfter = pendingRerunRequeueDelay
```
dineshkumar181094: This  cause infinite loop, you are Requeueing the same spark app, If it is failed to delete  may be because of permission issues the event will be in queue for ever.

The backoff function is already retrying multiple time. 

If you want to transition to Failed that will make more sense. 

I don't know How you landed here, the sparkapp will come into this state only after it has completed the deleteSparkResources. Until there is some race condition or someone manually create spark app in between.

skaushik15: Thanks, good catch on the unbounded requeue risk in PENDING_RERUN.
I've fixed it.
It can happen when driver pod and UI were deleted(as we can see in the logs ) but some other resources like executor was not cleaned up yet.
Another user has seen exact same behaviour.
https://github.com/kubeflow/spark-operator/issues/2905#issuecomment-4209528311

internal/controller/sparkapplication/controller.go
```diff
@@ -515,10 +522,31 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context,
 			logger.Info("Pending rerun SparkApplication", "state", app.Status.AppState.State)
 			if r.validateSparkResourceDeletion(ctx, app) {
 				logger.Info("Successfully deleted resources associated with SparkApplication", "state", app.Status.AppState.State)
+				// This timestamp is used only for bounding PendingRerun cleanup retries.
+				app.Status.TerminationTime = metav1.Time{}
 				r.recordSparkApplicationEvent(app)
```
dineshkumar181094, line 520: please use updateSparkApplicationStatus method to update the state.

internal/controller/sparkapplication/controller.go
```diff
@@ -515,10 +522,31 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context,
...
+					app.Status.TerminationTime = metav1.Now()
+				}
+
+				if time.Since(app.Status.TerminationTime.Time) >= pendingRerunCleanupRetryWindow {
+					app.Status.AppState.State = v1beta2.ApplicationStateFailed
+					app.Status.AppState.ErrorMessage = fmt.Sprintf(
+						"failed to delete resources before rerun within %s",
+						pendingRerunCleanupRetryWindow,
+					)
+					logger.Info("Pending rerun cleanup timed out, transitioning to failed state")
```
dineshkumar181094: please use updateSparkApplicationStatus method to update the state.

dineshkumar181094 (changes requested):

internal/controller/sparkapplication/controller.go
```diff
@@ -52,6 +52,12 @@ import (
 	"github.com/kubeflow/spark-operator/v2/pkg/util"
 )
 
+const (
+	pendingRerunRequeueDelay = 5 * time.Second
+	// Keep PendingRerun cleanup retry budget aligned with the default Pod termination grace period (30s).
+	pendingRerunCleanupRetryWindow = 30 * time.Second
+)
+
```
nabuskey: Let's make this more flexible. Hardcoded values like these are very difficult to maintain.

nabuskey (changes requested):
The change for this problem should be very similar to what we do in `reconcileFailedSubmissionSparkApplication`. 

If GC hasn't completed yet, we should update the result object with TimeUntilNextRetryDue, then re-queue.

internal/controller/sparkapplication/controller.go
skaushik15: Done

internal/controller/sparkapplication/controller.go
skaushik15: Done

skaushik15:
> The change for this problem should be very similar to what we do in `reconcileFailedSubmissionSparkApplication`.
> 
> If GC hasn't completed yet, we should update the result object with TimeUntilNextRetryDue, then re-queue.

Made the change in line with reconcileFailedSubmissionSparkApplication

tariq-hasan:
> > The change for this problem should be very similar to what we do in `reconcileFailedSubmissionSparkApplication`.
> > If GC hasn't completed yet, we should update the result object with TimeUntilNextRetryDue, then re-queue.
> 
> Made the change in line with reconcileFailedSubmissionSparkApplication

I am wondering if we might be conflating the semantics of `PENDING_RERUN` somewhat with these changes. As per this [description](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L140-L182), if `PENDING_RERUN` is just a placeholder for garbage collection and if its parent states - `FAILING` and `SUCCEEDING` - already encapsulate the semantics for restart policy, linear backoff and retry logic then would we want to reuse that logic again in `PENDING_RERUN`?

If `PENDING_RERUN` is just a polling step to validate garbage collection I am wondering if we want to have a clean reusable abstraction for just that logic - which is basically just [this portion](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L436-L442). This would still leave open the question of the polling interval but that would be a different consideration than the retry semantics that was already run through in the prior state.

Also, the `SUSPENDED` state suffers from the same issue as `PENDING_RERUN`: https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L815. I am wondering if we want to make the same changes there as well.

skaushik15:
> > > The change for this problem should be very similar to what we do in `reconcileFailedSubmissionSparkApplication`.
> > > If GC hasn't completed yet, we should update the result object with TimeUntilNextRetryDue, then re-queue.
> > 
> > 
> > Made the change in line with reconcileFailedSubmissionSparkApplication
> 
> I am wondering if we might be conflating the semantics of `PENDING_RERUN` somewhat with these changes. As per this [description](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L140-L182), if `PENDING_RERUN` is just a placeholder for garbage collection and if its parent states - `FAILING` and `SUCCEEDING` - already encapsulate the semantics for restart policy, linear backoff and retry logic then would we want to reuse that logic again in `PENDING_RERUN`?
> 
> If `PENDING_RERUN` is just a polling step to validate garbage collection I am wondering if we want to have a clean reusable abstraction for just that logic - which is basically just [this portion](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L436-L442). This would still leave open the question of the polling interval but that would be a different consideration than the retry semantics that was already run through in the prior state.
> 
> Also, the `SUSPENDED` state suffers from the same issue as `PENDING_RERUN`: https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L815. I am wondering if we want to make the same changes there as well.

Agree @tariq-hasan other states have similar issues and we should fix them.
Spark application shouldn't be stuck in a non terminal state forever, it should graceful fail so that failure strategies can be applied

internal/controller/sparkapplication/controller.go
```diff
@@ -515,10 +516,39 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context,
...
+						if err := r.deleteSparkResources(ctx, app); err != nil {
+							logger.Error(err, "failed to delete spark resources")
+						}
+						result.Requeue = true
+					} else {
+						// Keep polling until Kubernetes GC has finished removing all resources associated with the app.
+						logger.Info("Resources associated with SparkApplication still exist, waiting for GC to complete")
+						result.RequeueAfter = timeUntilNextRetryDue
+					}
+				}
```
tariq-hasan, line 551:
```suggestion
				logger.Info("Resources associated with SparkApplication still exist, retrying deletion")
				if err := r.deleteSparkResources(ctx, app); err != nil {
					logger.Error(err, "Failed to delete resources associated with SparkApplication")
				}
				result.RequeueAfter = r.options.ResourceCleanupPollInterval
```

Add to that I believe we ought to have something like this for the resource cleanup where we set up the poll interval as a configurable option since the retry was already taken care of in the prior step.

nabuskey: This is in line with what I had in mind. Make use of `sec.RetryInterval` with a default value and backoff just to avoid adding yet another option to the spec / flags. Currently the field is not used at all. Perhaps it makes sense to rename or update description of the field to avoid confusion. What do you think?

nabuskey (commented):
@tariq-hasan' s suggestion is great. The core of the issue here is that we return an empty result object instead of setting the requeue field. By not setting the field, reconciliation only happens when global reconcile timer is hit.

For this to happen, we need to set retry interval and backoff logic similar to how it's handled in other states.

internal/controller/sparkapplication/controller.go
tariq-hasan: That makes sense to me. Probably ideal to use this field than to add a new one in the spec or manifests.

nabuskey: @skaushik15 Does this make sense?

We should be able to set the default value here: api/v1beta2/defaults.go. 5 seconds default is good since it's the default value for other similar intervals.
Feel free to ping on Slack too.

skaushik15: Hey ,
it does make sense, but is there a plan to move it to terminal state instead of just retrying with a backoff interval? would moving it to a failed state after certain retries make more sense?

skaushik15: Made the changes as per suggestion

tariq-hasan: My understanding is that we can rephrase the question in this way: "_Does K8s guarantee garbage collection (in this case for the Spark driver pod, Spark Web UI Ingress and Spark Web UI Service)? If so should we rely on that guarantee and avoid any custom logic for app-level failures? On the other hand if there is no guarantee then should we do app-level failure to compensate for that lack of guarantee?_"

The original issue for the stuck state happened not because we did not fail the app. Rather it happened because we did not requeue in the first place. So the two aspects are fundamentally different and we are already taking care of requeueing.

In so far as garbage collection is concerned K8s does not guarantee deletion in a bounded time since finalizers, webhooks, offline nodes, etc. can block it, but this sounds more like a rare edge case than what this current issue is dealing with.

api/v1beta2/sparkapplication_types.go
```diff
@@ -107,7 +107,8 @@ type SparkApplicationSpec struct {
 	// This is best effort and actual retry attempts can be >= the value specified.
 	// +optional
 	FailureRetries *int32 `json:"failureRetries,omitempty"`
-	// RetryInterval is the unit of intervals in seconds between submission retries.
+	// RetryInterval is the unit of intervals in seconds between retry/poll attempts
+	// in controller-managed loops (for example submission retries and pending rerun cleanup polling).
 	// +optional
```
tariq-hasan, line 112: I think the `submission retries` phrasing should be removed since this is not a legitimate use case for `RetryInterval`.

internal/controller/sparkapplication/controller.go
```diff
@@ -515,10 +516,30 @@ func (r *Reconciler) reconcilePendingRerunSparkApplication(ctx context.Context,
...
 				logger.Info("Successfully deleted resources associated with SparkApplication", "state", app.Status.AppState.State)
+				app.Status.TerminationTime = metav1.Time{}
 				r.recordSparkApplicationEvent(app)
 				r.submitSparkApplication(ctx, app)
 			} else {
-				logger.Info("Resources associated with SparkApplication still exist")
+				if app.Status.TerminationTime.IsZero() {
+					app.Status.TerminationTime = metav1.Now()
+				}
+				timeUntilNextRetryDue, err := util.TimeUntilNextRetryDue(app)
```
tariq-hasan, line 526: The termination time is used for app failure tracking and therefore should not to be used to calculate backoff for cleanup polling. Otherwise this will conflate the use case for this field.

As a practical example, if the app transitions from `PENDING_RERUN` to `FAILED`, the [TTL calculation](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L701-L716) in `reconcileTerminatedSparkApplication` would use the cleanup start timestamp rather than the [actual failure timestamp](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L449-L451).

pkg/util/sparkapplication.go
```diff
@@ -111,15 +111,29 @@ func ShouldRetry(app *v1beta2.SparkApplication) bool {
...
+			baseInterval := time.Duration(*retryInterval) * time.Second
+			if baseInterval > 0 {
+				elapsed := time.Since(lastAttemptTime.Time)
+				if elapsed < 0 {
+					elapsed = 0
+				}
+				attemptsDone = int32(elapsed/baseInterval) + 1
+			}
+		}
 	}
```
tariq-hasan, line 135: Looking at the original code for `TimeUntilNextRetryDue` there is an implicit assumption that this is a helper function that enables restarts to occur after app failures.

This change therefore intermingles two different concerns - app failure retries and cleanup polling - within the same block of code. It would be better to move the logic for cleanup polling to a different helper function to avoid confusion and maintain a clean separation of concerns.

PR #2917, opened 2026-04-20
fix(rbac): Fix Helm RBAC bugs, add trailing newline, and add unit test

RobuRishabh:
## Purpose of this PR

The Spark Operator's Helm chart RBAC was scaffolded with overly broad defaults and never pruned to match actual controller/webhook behavior. This PR brings the Helm chart `policyRules` into perfect alignment with the Go `+kubebuilder:rbac` markers and Kustomize `config/rbac/role.yaml` (tightened in companion PR #2914) by auditing every permission against actual Go code (`r.client.*` calls, `Watches`, `Owns`, `controllerutil.*` usage, and `spark-submit` subprocess behavior).

Additionally, this PR fixes a bug where the webhook ClusterRole was missing cluster-wide permissions when `jobNamespaceSelector` is configured, and fixes a race condition in the namespace filtering e2e tests that caused deterministic CI failures.

**Proposed changes:**

- Tighten controller `policyRules` in `controller/_helpers.tpl`: removed excess verbs and the entire `persistentvolumeclaims` resource block
- Remove `nodes: get` from the controller ClusterRole in `controller/rbac.yaml`
- Tighten webhook `policyRules` in `webhook/_helpers.tpl` : remove `pods` block entirely, narrow `resourcequotas` to `list` only, narrow `sparkapplications` to `get, list, watch`, and remove all `status`/`finalizer`/`scheduledsparkapplication` resources
- Fix webhook ClusterRole in `webhook/rbac.yaml` to include `policyRules` when `jobNamespaceSelector` is set, matching the controller's existing behavior
- Fix e2e namespace filtering test race condition: add `Eventually` wait in `BeforeEach` to ensure the controller's informer cache has indexed new namespaces before creating SparkApplications
- Added rule-content test cases (`contains`/`notContains` assertions) to `controller/rbac_test.yaml` to prevent RBAC regressions, including `jobNamespaceSelector` coverage
- Added rule-content test cases (`contains`/`notContains` assertions) to `webhook/rbac_test.yaml` to prevent RBAC regressions, including `jobNamespaceSelector` coverage

## Change Category

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

### Rationale

Every removed permission was verified against the Go source:

| Removed Permission | Evidence |
|---|---|
[...]

PR #2919, opened 2026-04-21
feat(kustomize): Add core deployment infrastructure and E2E test suite

RobuRishabh:
## Purpose of this PR

This PR adds complete Kustomize deployment infrastructure for the Spark Operator, enabling users to deploy without Helm using `kubectl apply -k config/default/`. It also promotes Kustomize to a **first-class deployment strategy** alongside Helm by integrating it into the existing Go e2e test suite and the main `integration.yaml` CI workflow.

Currently, the `config/` directory contains auto-generated scaffolding from `controller-gen` and Kubebuilder, but lacks the manifests needed for a working standalone deployment (controller/webhook Deployments, ServiceAccounts, RBAC bindings, etc.). This PR fills that gap, bringing Kustomize deployments to parity with Helm.

**Proposed changes:**

- Add Kustomize deployment manifests: `config/default/` (entry point with namespace and image parameterization via `images` directive derived from `VERSION` file), `config/manager/` (controller Deployment + SA), `config/rbac/` (ClusterRole/Binding, leader election Role), `config/webhook/` (webhook Deployment, SA, RBAC, service wiring)
- Separate Spark job RBAC into `config/spark-rbac/` — a standalone kustomization users apply to each application namespace independently from the operator infrastructure
- Add Go-based static lint test (`test/kustomize/kustomize_build_test.go`) validating manifest structure, RBAC rules, namespace consistency, and security settings — analogous to `helm-unittest` for the Helm chart
- Refactor `test/e2e/suite_test.go` to support both Helm and Kustomize deployment via `DEPLOY_METHOD` env var (defaults to `helm` for backward compatibility); image override only applied when `SPARK_OPERATOR_IMAGE` is explicitly set
- Add `deploy_method: [helm, kustomize]` matrix dimension to the `e2e-test` job in `integration.yaml` (4 k8s versions × 2 deploy methods = 8 jobs)
- Add `kustomize-lint.yaml` GitHub Actions workflow for Kustomize-specific lint validation, and corresponding Makefile targets
- Update Makefile `e2e-test` target to accept `DEPLOY_METHOD`

## Change Category

- [ ] Documentation update

### Rationale

The Spark Operator supports Helm-based deployment but has no working Kustomize path despite having a `config/` directory. Several pre-existing files in `config/` had latent bugs (missing `apiGroups`, wrong Service selector, no webhook `objectSelector`) that were never caught because they were never applied to a live cluster — upstream E2E tests use Helm exclusively.

This PR enables `kubectl apply -k config/default/` as a first-class deployment method, runs the **same Go e2e test suite** against both Helm and Kustomize deploys in CI, and adds a Go-based static lint test to prevent structural regressions. The Kustomize structure also lays the groundwork for downstream consumers who overlay on top of `config/default/`.

## Checklist

### Additional Notes

**Pre-existing bugs fixed in this PR:**

| Bug | Root Cause | Fix |
|-----|-----------|-----|
| `ClusterRole "spark-operator-controller" is invalid: rules[0].apiGroups: Required value` | `+kubebuilder:rbac` markers used `groups=,` instead of `groups="",` causing `controller-gen` to omit `apiGroups: [""]` for core resources | Fixed Go markers in `controller.go` and `reconciler.go`; regenerated `config/rbac/role.yaml` |
| Webhook Service routed to controller pod instead of webhook pod | `config/webhook/service.yaml` used Kubebuilder default selector `control-plane: controller-manager` | Updated selector to `app.kubernetes.io/component: webhook` |
| Operator pods fail to start (webhook deadlock) | `mutate-pod` webhook in `manifests.yaml` intercepted ALL pod creation including operator's own pods | Added strategic merge patch with `objectSelector: sparkoperator.k8s.io/launched-by-spark-operator: "true"` (matching Helm chart behavior) |

**Architecture:**

| Directory | Purpose | Namespace |
|-----------|---------|-----------|
| `config/default/` | Operator infrastructure (controller, webhook, CRDs, RBAC) | `spark-operator` (set by transformer) |
| `config/spark-rbac/` | Spark job RBAC (SA, Role, RoleBinding for driver pods) | User-specified via `-n <app-namespace>` |

**Dual deployment strategy in e2e:**

| Deploy Method | Controlled By | Webhook Names | Namespace Filtering Tests |
|---------------|--------------|---------------|--------------------------|
| `helm` (default) | Helm Go SDK + `ci-values.yaml` | `spark-operator-webhook` | Run (uses `jobNamespaceSelector`) |
| `kustomize` | `kubectl apply -k config/default/` | `mutating/validating-webhook-configuration` | Skipped (no selector configured) |

The Go e2e suite (`test/e2e/`) now runs against both deployment methods in CI via the `integration.yaml` matrix. All 9 functional tests pass with Kustomize; 4 namespace filtering tests are skipped (they require Helm-specific `jobNamespaceSelector` configuration).

**How to test locally:**

```bash
make kind-create-cluster
make docker-build IMAGE_TAG=local
make kind-load-image IMAGE_TAG=local

# Format Go code
make go-fmt

# Static lint (no cluster needed, <2 seconds)
make kustomize-lint

# Go e2e with Kustomize deploy
SPARK_OPERATOR_IMAGE=ghcr.io/kubeflow/spark-operator/controller:local make e2e-test DEPLOY_METHOD=kustomize

# Go e2e with Helm deploy (existing behavior)
make e2e-test DEPLOY_METHOD=helm
```


test/kustomize/test-operator-install.sh
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env bash
```
alinaryan: I noticed all the tests are in bash, not Go, but the existing e2e tests use Ginkgo/Go. Are we ok with mainaining two different test frameworks here? This could potentially make maintenance harder over time.

config/webhook/clusterrole.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - scheduledsparkapplications/status
+      - scheduledsparkapplications/finalizers
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
alinaryan: Does the webhook actually need all these verbs or would get/list/watch suffice?

config/default/params.env
```diff
@@ -0,0 +1,3 @@
+# Spark Operator image parameters
+SPARK_OPERATOR_CONTROLLER_IMAGE=ghcr.io/kubeflow/spark-operator/controller:v2.5.0-rc.0
+SPARK_OPERATOR_WEBHOOK_IMAGE=ghcr.io/kubeflow/spark-operator/controller:v2.5.0-rc.0
```
alinaryan: Do we have a plan to keep these version strings in sync with releases?

alinaryan (commented):
Thanks for opening this. Please see my comments. I understand you addressed a few bug fixes while working on this. It might be better to open separate PRs in the future. If one of the RBAC or webhook fixes needs to be backported or causes a regression, it's tied up with this 1300+ line kustomize PR.

config/webhook/clusterrole.yaml
RobuRishabh: No, the webhook doesn't need all those verbs, Great catch, and I'll tighten it to match the Helm chart.

I already tightened the webhook RBAC in the Helm chart in PR #2917 .  I should have applied the same tightening here.
I'll tighten the Kustomize webhook ClusterRole to match what I already have in the Helm chart (PR #2917):
```
- apiGroups: [""]
  resources: [resourcequotas]
  verbs: [list]
- apiGroups: [sparkoperator.k8s.io]
  resources: [sparkapplications]
  verbs: [get, list, watch]
```

config/default/params.env
RobuRishabh: Not sure, but my thought was I will raise this issue in the community meeting and ask maintainers if we can follow the same pattern as Helm's Chart.yaml/values.yaml: update params.env as part of the release process.
1. During a release, the same CI step that bumps appVersion in Chart.yaml would also update SPARK_OPERATOR_CONTROLLER_IMAGE and SPARK_OPERATOR_WEBHOOK_IMAGE in params.env.
2. For local development / CI, the image is always overridden via the SPARK_OPERATOR_IMAGE env var (as shown in the Makefile targets), so the hardcoded default is only used by users running a plain kubectl apply -k config/default/.

I can add a comment in params.env documenting this, or we could add a CI check that verifies params.env stays in sync with the Helm chart's appVersion. What would you prefer?

RobuRishabh:
> Thanks for opening this. Please see my comments. I understand you addressed a few bug fixes while working on this. It might be better to open separate PRs in the future. If one of the RBAC or webhook fixes needs to be backported or causes a regression, it's tied up with this 1300+ line kustomize PR.

Thanks for the review @alinaryan  You're right, in the future I'll split bug fixes into their own PRs for easier backporting and bisection. The three bug fixes here (empty apiGroups in kubebuilder markers, webhook Service selector, webhook objectSelector deadlock) were discovered because they were latent issues in the config/ scaffolding that only surfaced when I actually applied the manifests to a live cluster.

test/kustomize/test-operator-install.sh
nabuskey: I'd rather use Go.  As I understand it, the only thing that should change is the deployment method. The same tests should run against both.

Makefile
```diff
@@ -170,10 +170,41 @@ unit-test: setup-envtest ## Run unit tests.
...
+kustomize-verify-rbac: ## Verify RBAC permissions for controller and webhook (requires cluster).
+	@echo "Running RBAC verification..."
+	bash test/kustomize/verify-rbac.sh
+
+.PHONY: kustomize-test-spark-pi
+kustomize-test-spark-pi: ## Run Spark Pi functional test (requires cluster with operator installed).
+	@echo "Running Spark Pi functional test..."
+	CLEANUP=$(CLEANUP) bash test/kustomize/test-spark-pi.sh
+
+.PHONY: kustomize-e2e
```
nabuskey: are these necessary? seems like very specific tasks that should be part of the install task. spark pi stuff can be in e2e

Makefile
```diff
@@ -170,10 +170,41 @@ unit-test: setup-envtest ## Run unit tests.
 	go tool cover -html=cover.out -o cover.html
 	@echo "Coverage report available at cover.html"
 
+DEPLOY_METHOD ?= helm
```
nabuskey: Let's move this up with others.

.github/workflows/kustomize-lint.yaml
```diff
@@ -0,0 +1,39 @@
...
+  kustomize-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Validate Kustomize build
+        run: make kustomize-lint
```
nabuskey, line 46: lint only for now? Add actual tests later? is that the plan?

config/default/params.env
nabuskey: I have to review the release process but we should be able to grab this from a marker

test/kustomize/test-operator-install.sh
RobuRishabh: ok. I'll remove the bash test scripts and rely on the Go e2e suite with DEPLOY_METHOD=kustomize for all functional/integration testing. 
I will also add go-based static manifest linter for the Kustomize deployment to keep it in parity with helm-unittest.

Makefile
RobuRishabh: Done

.github/workflows/kustomize-lint.yaml
RobuRishabh: Correct, this workflow is lint only. I will rename it from `kustomize-e2e.yaml` to `kustomize-lint.yaml` to make that explicit. 
It runs `make kustomize-lint`, which is now a Go test (`test/kustomize/kustomize_build_test.go`) that validates the `kubectl kustomize config/default/` output -- resource counts, RBAC rules, namespace consistency, security settings. No cluster needed.

Functional e2e tests run via the existing `integration.yaml` workflow, which has the `deploy_method: [helm, kustomize]` matrix dimension (4 k8s versions x 2 deploy methods = 8 CI jobs). So the CI strategy is:
- kustomize-lint.yaml: static manifest validation on any change to `config/**` or `test/kustomize/**`
- integration.yaml: full e2e suite (`make e2e-test DEPLOY_METHOD=kustomize) against real Kind clusters

Makefile
RobuRishabh: No, I've removed all the Kustomize Makefile targets (`kustomize-operator-install`, `kustomize-verify-rbac`, `kustomize-test-spark-pi`, `kustomize-e2e`) along with their backing bash scripts.

The only Kustomize-specific target remaining is `kustomize-lint`, which I've converted from a bash script to a Go test (`test/kustomize/kustomize_build_test.go`). It runs `kubectl kustomize config/default/` and validates the output structure (resource counts, RBAC rules, namespace consistency, security settings), no cluster needed. This is the Kustomize equivalent of helm-unittest for the Helm chart.

config/default/params.env
nabuskey: Can we grab from the version file? 

```
~/spark-operator$ cat VERSION
v2.5.0-rc.0
```
I gotta update the file btw

RobuRishabh: Done

alimaredia:
@RobuRishabh I'm wondering if the `kustomize_build_test.go` is even needed. As long as the E2E tests are being run against the operator deployed via the Kustomize manifests **and** there is some kind of Helm - Kustomize drift detection test running in CI (which I know you're working on) then that should cover testing the functionality and the contents of the Kustomize manifests.

Am I missing some value add that the go test already has that wouldn't be covered in those two cases?

test/e2e/suite_test.go
```diff
@@ -170,11 +215,113 @@ var _ = AfterSuite(func() {
...
+	kustomizeDir := filepath.Join(repoRoot, "config", "default")
+
+	By("Uninstalling the Spark operator via Kustomize")
+	deleteCmd := exec.Command("kubectl", "delete", "-k", kustomizeDir, "--ignore-not-found", "--timeout=120s")
+	deleteCmd.Stdout = GinkgoWriter
+	deleteCmd.Stderr = GinkgoWriter
+	_ = deleteCmd.Run()
+}
+
+func createSparkJobRBAC(ctx context.Context, namespace string) {
```
alimaredia: @RobuRishabh This function looked a little fishy, and digging in deeper the existence of this function ties into a bigger issue with the way the kustomize manifests are laid out right now.

The createSparkJobRBAC function exists because the kustomize `namespace: spark-operator` transformer puts everything into the operator's namespace, but e2e tests run Spark apps in `default`.

Rhe function works around this by recreating the same ServiceAccount, Role, and RoleBinding with hardcoded values — essentially duplicating what's already in config/rbac/spark-application-rbac.yaml.

If we moved those resources into their own `config/spark-rbac/` directory with a `kustomization.yaml` that doesn't set a namespace or sets the namespace to default, the test could just run `kubectl apply -k config/spark-rbac -n default` and we wouldn't need `createSparkJobRBAC` at all.

This is also just a nicer separation — operator infrastructure (controller, webhook, CRDs) stays in `config/default`, and the resources users actually need in their application namespaces live in config/spark-rbac/. Since the ServiceAccount, Role (spark-role), and RoleBinding (spark-role-binding) are all manually written rather than generated by `make manifests`, pulling them out into a separate directory doesn't interfere with anything.

test/e2e/suite_test.go
```diff
@@ -170,11 +215,113 @@ var _ = AfterSuite(func() {
...
 
-	By("Tearing down the test environment")
-	err = testEnv.Stop()
-	Expect(err).ToNot(HaveOccurred())
-})
+func installViaKustomize() {
+	repoRoot := filepath.Join("..", "..")
+	kustomizeDir := filepath.Join(repoRoot, "config", "default")
+
+	image := os.Getenv("SPARK_OPERATOR_IMAGE")
```
alimaredia: This function should just read the operator image from a `config/defualt/params.env` or from the images in`kustomization.yaml` and only override it if the environment variable SPARK_OPERATOR_IMAGE is already set. That would remove the a hardcoded fallback in the tests and the rewriting of kustomization.yaml.

config/webhook/role.yaml
```diff
@@ -0,0 +1,47 @@
...
+  namespace: system
+  labels:
+    app.kubernetes.io/name: spark-operator
+    app.kubernetes.io/instance: spark-operator
+    app.kubernetes.io/component: webhook
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - events
```
alimaredia: We had talked about removing `events` here since it's not necessary.

RobuRishabh: ya, will remove it from here and place it in clusterrole.yaml

test/e2e/suite_test.go
RobuRishabh: So as per maintainers prev comment https://github.com/kubeflow/spark-operator/pull/2919#discussion_r3155996310 I actually removed `params.env` entirely and replaced it with  the images directive in kustomization.yaml using newTag: v2.5.0-rc.0 (from the VERSION file).

You're right, I need to get rid of this hardcoded fallbacks or unnecessary file rewriting, so I have done these changes now:
- If SPARK_OPERATOR_IMAGE is set → override the newTag in kustomization.yaml, restore on cleanup
- If SPARK_OPERATOR_IMAGE is NOT set → use whatever image tag is already in kustomization.yaml as-is (no rewriting, no fallback)
- The kustomization.yaml images.newTag (derived from VERSION file via make kustomize-set-image) is the single source of truth

test/e2e/suite_test.go
RobuRishabh: Its an excellent suggestion, and very architecturally correct. Thanks for catching this. I will do this separation.

RobuRishabh:
> @RobuRishabh I'm wondering if the `kustomize_build_test.go` is even needed. As long as the E2E tests are being run against the operator deployed via the Kustomize manifests **and** there is some kind of Helm - Kustomize drift detection test running in CI (which I know you're working on) then that should cover testing the functionality and the contents of the Kustomize manifests.
> 
> Am I missing some value add that the go test already has that wouldn't be covered in those two cases?

@alimaredia, I think we should have this file. so our whole purpose is to make kustomize in parity with helm. The lint test (`kustomize_build_test.go`) serves the same role as `helm-unittest` does for the Helm chart, fast structural validation without a cluster. It catches things that neither e2e nor drift detection covers:
- Kustomize-specific features: namespace transformer correctness, replacements propagation, objectSelector patch applied
- Structural regressions: missing resources, broken security context (even if both Helm and Kustomize are equally broken, drift passes but lint would catch it)

Once drift detection lands in a follow-up PR, we can re-evaluate and potentially slim down the lint test by removing checks that drift covers (like RBAC rule comparison), but the Kustomize-specific structural checks will remain valuable.

test/e2e/suite_test.go
```diff
@@ -170,11 +214,80 @@ var _ = AfterSuite(func() {
...
-	By("Tearing down the test environment")
-	err = testEnv.Stop()
-	Expect(err).ToNot(HaveOccurred())
-})
+func installViaKustomize() {
+	repoRoot := filepath.Join("..", "..")
+	kustomizeDir := filepath.Join(repoRoot, "config", "default")
+	kustomizationPath := filepath.Join(kustomizeDir, "kustomization.yaml")
+
+	image := os.Getenv("SPARK_OPERATOR_IMAGE")
```
nabuskey: we have `IMAGE_TAG` in Makefile. I think it makes sense to use that instead. I was trying to run this locally and didn't work at first. I would avoid adding a very specific env var like this here.

nabuskey (commented):
What is the kustomization deployment model we are testing? kubectl embedded version or standalone? The tests here seem to use embedded but in makefile we download standalone binary.

RobuRishabh:
> What is the kustomization deployment model we are testing? kubectl embedded version or standalone? The tests here seem to use embedded but in makefile we download standalone binary.

@nabuskey The tests intentionally use `kubectl apply -k` / `kubectl kustomize` (embedded), that's the deployment path we document for users (`kubectl apply -k config/default/`).

The standalone `kustomize` binary in the Makefile is pre-existing Kubebuilder scaffolding from `kubebuilder init` it was there before this PR and is only used by the `install-crd` / `uninstall-crd` targets. Nothing in this PR uses or depends on it. Those CRD targets could equally use `kubectl kustomize config/crd | kubectl apply -f -`, but I didn't want to change existing targets outside the scope of this PR.

If you'd like, I can clean up `install-crd`/`uninstall-crd` to use embedded `kubectl kustomize` as well and remove the standalone binary dependency entirely. I am happy to do that in this PR or a follow-up. whatever you recommend, I will do that.

test/e2e/suite_test.go
RobuRishabh: Good point, agreed. I'll update installViaKustomize() to read IMAGE_TAG instead of SPARK_OPERATOR_IMAGE and have the Makefile pass it through

nabuskey (approved):
During today's call we decided to use embedded version of kustomize for all testing. To do that, we need to remove the standalone kustomize stuff (mostly in Make). We will work on them in followup PRs.

nabuskey (approved):
LGTM

vara-bonthu (approved):
/approve

PR #2933, opened 2026-05-07
feat(ci): Implement drift detection as Go test in test/drift/

RobuRishabh:
## Purpose of this PR

This PR implements an automated **Helm-Kustomize Drift Detection Pipeline** that uses the Helm chart as the "source of truth" and compares its rendered output against Kustomize build output, reporting any unauthorized divergence.

**Proposed changes:**

- Add `test/drift/drift_test.go` — a Go test package that renders both Helm and Kustomize outputs into typed Kubernetes objects and performs semantic comparison of RBAC rules, webhook configurations, and deployment specs
- Add `test/drift/drift-check-values.yaml` — Helm values that align template output with Kustomize defaults for fair comparison
- Add `.github/workflows/kustomize-drift-check.yaml` — CI workflow triggered on PRs/pushes affecting `config/`, `charts/`, `internal/`, or `test/drift/`
- Update `Makefile`:
  - Add `drift-check` target for local developer experience (`make drift-check`)
  - Exclude `test/drift/` from `unit-test` target (it has its own dedicated workflow)

## Change Category

- [ ] Documentation update

### Rationale

The Kustomize manifests were abandoned and broke because there was no automated gate to ensure they stayed current with Helm chart updates. A byte-for-byte comparison isn't feasible due to inherent differences in naming, labeling, and parameterization schemes between Helm and Kustomize. Instead, this test performs **semantic comparison** by:

1. Parsing rendered YAML into typed Go structs (`rbacv1.ClusterRole`, `appsv1.Deployment`, `admissionregistrationv1.MutatingWebhookConfiguration`, etc.)
2. Normalizing expected differences (naming conventions, namespace variable substitution, rule merging)
3. Comparing the fields that matter: RBAC verbs/resources, webhook paths/operations/failurePolicy, and deployment args/ports/probes/securityContext

This approach uses only libraries already in `go.mod` (`k8s.io/api`, `k8s.io/apimachinery`, `testify`) and follows the same pattern as the existing `test/kustomize/` package — no new external dependencies.

**Note:** The drift-check workflow uses `continue-on-error: true` because 7 genuine drift items currently exist between Helm and Kustomize. These will be fixed in follow-up PRs, after which the `continue-on-error` will be removed to make the check blocking.

## Checklist

### Additional Notes

The values file is placed under `test/drift/` (not `charts/spark-operator-chart/ci/`) to avoid chart-testing automatically using it as an install test configuration.

The test currently detects 7 genuine drift items:

| Area | Drift |
|------|-------|
| WebhookClusterRole | Kustomize has extra permissions (pods, scheduledsparkapplications, finalizers/status) |
| ControllerRole | Kustomize missing events permission |
| MutatingWebhookPaths | Kustomize missing scheduledsparkapplication webhook |
| MutatingWebhookRules | Pod webhook operations differ (CREATE vs CREATE+UPDATE) |
| ValidatingWebhookPaths | Kustomize has duplicate sparkapplication entry |
| ControllerArgs | Kustomize missing --feature-gates, --kube-api-burst/qps, --scheduled-spark-application-timestamp-precision |
| WebhookArgs | Kustomize missing --kube-api-burst/qps |

These represent real accumulated drift that will be fixed in follow-up PRs. Once fixed, the CI gate will prevent future drift from being introduced silently.

---

## Commands to test

```bash
# Run via Makefile (same as CI)
make drift-check

# Run Go test directly
go test ./test/drift/ -v -count=1

# Run with explicit binary paths
HELM=/opt/homebrew/bin/helm go test ./test/drift/ -v -count=1

# Run with both explicit paths
HELM=/opt/homebrew/bin/helm KUSTOMIZE=./bin/kustomize-v5.4.1 go test ./test/drift/ -v -count=1

# Run specific subtest groups
HELM=/opt/homebrew/bin/helm go test ./test/drift/ -v -count=1 -run "TestDrift/RBAC"
HELM=/opt/homebrew/bin/helm go test ./test/drift/ -v -count=1 -run "TestDrift/WebhookConfiguration"
HELM=/opt/homebrew/bin/helm go test ./test/drift/ -v -count=1 -run "TestDrift/DeploymentSpecs"

# Verify existing unit tests still pass (drift is excluded)
make unit-test

# Verify compilation
go vet ./test/drift/
```

### DIFF
--- a/test/drift/drift_test.go
+++ b/test/drift/drift_test.go
@@ -1,0 +1,909 @@
+/*
+Copyright 2026 The Kubeflow authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package drift_test detects semantic drift between the Helm chart (source of
+// truth) and the Kustomize manifests. It renders both sides, parses them into
+// typed Kubernetes objects, and compares the fields that matter: RBAC rules,
+// webhook configurations, and deployment specs.
+package drift_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// ---------------------------------------------------------------------------
+// Name mapping between Helm and Kustomize
+// ---------------------------------------------------------------------------
+
+// Helm uses "spark-operator-controller" / "spark-operator-webhook" for all
+// resource names (derived from fullname helper). Kustomize uses short names
+// for namespace-scoped resources and the same names for cluster-scoped ones.
+
+const (
+	helmReleaseName = "spark-operator"
+	helmNamespace   = "spark-operator"
+
+	// ClusterRole names (same on both sides)
+	controllerClusterRole = "spark-operator-controller"
+	webhookClusterRole    = "spark-operator-webhook"
+
+	// Role names differ
+	helmControllerRole = "spark-operator-controller"
+	kustControllerRole = "spark-operator-controller"
+	helmWebhookRole    = "spark-operator-webhook"
+	kustWebhookRole    = "spark-operator-webhook"
+
+	// Deployment names (same on both sides after kustomize namePrefix)
+	helmControllerDeploy = "spark-operator-controller"
+	kustControllerDeploy = "spark-operator-controller"
+	helmWebhookDeploy    = "spark-operator-webhook"
+	kustWebhookDeploy    = "spark-operator-webhook"
+
+	// Webhook configuration names differ
+	helmMutatingWH = "spark-operator-webhook"
+	kustMutatingWH = "mutating-webhook-configuration"
+	helmValidWH    = "spark-operator-webhook"
+	kustValidWH    = "validating-webhook-configuration"
+
+	// Container names differ (Kustomize uses short names)
+	helmControllerContainer = "spark-operator-controller"
+	kustControllerContainer = "controller"
+	helmWebhookContainer    = "spark-operator-webhook"
+	kustWebhookContainer    = "webhook"
+)
+
+// Resource kinds that legitimately exist on only one side.
+var ignoredKinds = map[string]bool{
+	"Job":                      true, // Helm CRD hook
+	"PodMonitor":               true, // Helm prometheus
+	"Certificate":              true, // Helm cert-manager
+	"Issuer":                   true, // Helm cert-manager
+	"ConfigMap":                true, // Kustomize params
+	"Namespace":                true, // Kustomize creates NS
+	"CustomResourceDefinition": true, // Covered by detect-crds-drift
+	"PodDisruptionBudget":      true, // Neither side produces PDB by default (Helm: gated behind podDisruptionBudget.enable + replicas>1; Kustomize: config/pdb/ not included in config/default/kustomization.yaml)
+}
+
+// ---------------------------------------------------------------------------
+// Rendering helpers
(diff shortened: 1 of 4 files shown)


test/drift/drift_test.go
```diff
@@ -0,0 +1,909 @@
...
+				Resources:       append([]string{}, r.Resources...),
+				ResourceNames:   append([]string{}, resourceNames...),
+				NonResourceURLs: append([]string{}, r.NonResourceURLs...),
+			}
+		}
+	}
+
+	var out []rbacv1.PolicyRule
+	for _, r := range merged {
+		r.Resources = sortedCopy(r.Resources)
```
alinaryan: This should be dedupSorted(sortedCopy(r.Resources)) to match what you do for ResourceNames on the next line

test/drift/drift_test.go
```diff
@@ -0,0 +1,909 @@
...
+				ke, ok := kustByPath[he.Path]
+				if !ok {
+					continue
+				}
+				assert.Equal(t, he.ObjectSelector, ke.ObjectSelector,
+					"MutatingWebhook %s: objectSelector differs", he.Path)
+				// Helm in cluster-wide mode (jobNamespaces: [""]) omits namespaceSelector,
+				// while Kustomize hardcodes it via strategic-merge patches (controller-gen
+				// cannot emit it). Only compare when Helm actually renders one.
+				if he.NamespaceSelector != nil {
```
alinaryan, line 668: When Helm's namespaceSelector is nil (cluster-wide mode), the comparison is skipped entirely. This means Kustomize could have a stale or wrong selector and the test would silently pass. Consider adding an else branch that asserts ke.NamespaceSelector is also nil, so you catch the case where Kustomize has a selector it shouldn't.

test/drift/drift_test.go
```diff
@@ -0,0 +1,909 @@
...
+			for _, he := range helmEntries {
+				ke, ok := kustByPath[he.Path]
+				if !ok {
+					continue
+				}
+				assert.Equal(t, he.ObjectSelector, ke.ObjectSelector,
+					"ValidatingWebhook %s: objectSelector differs", he.Path)
+				// See MutatingWebhookSelectors comment: skip namespaceSelector when
+				// Helm renders nil (cluster-wide mode).
+				if he.NamespaceSelector != nil {
```
alinaryan, line 769: same comment here

test/drift/drift_test.go
```diff
@@ -0,0 +1,909 @@
...
+// Resource kinds that legitimately exist on only one side.
+var ignoredKinds = map[string]bool{
+	"Job":                      true, // Helm CRD hook
+	"PodMonitor":               true, // Helm prometheus
+	"Certificate":              true, // Helm cert-manager
+	"Issuer":                   true, // Helm cert-manager
+	"ConfigMap":                true, // Kustomize params
+	"Namespace":                true, // Kustomize creates NS
+	"CustomResourceDefinition": true, // Covered by detect-crds-drift
+	"PodDisruptionBudget":      true, // Optionally present
```
alinaryan: why is PDB ignored? If both Helm and Kustomize produce one, differences in minAvailable and MaxUnavailable would be skipped

test/drift/drift_test.go
RobuRishabh: Good catch, thank you. I'll fix this

test/drift/drift_test.go
RobuRishabh: actually this concern is dependent on [PR #2941](https://github.com/kubeflow/spark-operator/pull/2941)

So, Kustomize is expected to have a non-nil `namespaceSelector`, it's hardcoded to `default` via strategic-merge patches in [PR #2941](https://github.com/kubeflow/spark-operator/pull/2941). The patches exist because `controller-gen` can't emit namespaceSelector from kubebuilder markers, so Kustomize adds it via patches that survive make manifests regeneration.

The scenario after this PRs  merged:
- Helm (cluster-wide mode, `jobNamespaces: [""]`): `namespaceSelector = nil` (intentionally omitted)
- Kustomize (after PR #2941): `namespaceSelector` = scoped to default (hardcoded via patch)
So `he.NamespaceSelector == nil` and `ke.NamespaceSelector != nil` is the correct, expected state. An `else` asserting `ke.NamespaceSelector == nil` would fail immediately.

The conditional check is intentional: the drift test uses `jobNamespaces: [""]` (cluster-wide) in `drift-check-values.yaml` to make RBAC rules land in the ClusterRole (matching Kustomize's structure). This forces Helm to omit `namespaceSelector`. If someone later changes the values to a non-cluster-wide mode (e.g., `["default"]`), Helm will render a `namespaceSelector` and the comparison will automatically activate catching any real drift.

The objectSelector comparison on the line above remains unconditional that's the security-critical one.

test/drift/drift_test.go
RobuRishabh: Ditto

test/drift/drift_test.go
RobuRishabh: PDB is ignored because neither side produces one with default settings:

- Helm gates PDB behind `controller.podDisruptionBudget.enable` (defaults `false`) and requires `replicas > 1`
- Kustomize has a PDB file at `config/pdb/` but it's not included in `config/default/kustomization.yaml`

I'll improve the comment to explain this better. If both deployment methods start producing PDB by default in the future, we can add a comparison subtest.

nabuskey:
Awaiting PRs #2936 and #2941 to merge. Then rebase this PR. The detection CI should run ok afterwards.

nabuskey (approved):
/approve

Issue #2788, opened 2025-12-22
Restart of Spark Operator may Result in Duplicated Submission


kohn:
### What happened?

- [x] ✋ I have searched the open/closed issues and my issue is not listed.
When I restart spark-operator, sometimes job fails with the error of "driver pod already exist". I checked the kubernetes event, and find the following event:
- T0: got the event of SparkApplicationSubmitted
- T1: old spark-operator exit
- T2: new spark-operator started
- T3: got the event of SparkApplicationAdded
- T4: got the event of SparkApplicationSubmissionFailed
- T5: got the event of SparkApplicationFailed

I guess this bug happens when spark-operator abruptly exist while spark-submit command is done but updateSparkApplicationStatus is not executed so that SparkApplication status is still "new"(""). Then the new spark-operator is up and try to re-submit again as the status is "new".

### Reproduction Code

Keep submitting lots of jobs and restart the spark-operator

### Expected behavior

jobs can be started

### Actual behavior

some jobs fail with the error of "driver pod already exist"

### Environment & Versions

- Kubernetes Version: 1.33
- Spark Operator Version: 2.3.0
- Apache Spark Version:

### Impacted by this bug?

Give it a 👍 We prioritize the issues with most 👍

tasier:
I’ve encountered this issue as well.

Denyme24:
hey @kohn @ChenYi015, i would like to work on this issue .

make submission idempotent by checking for an existing driver pod before calling spark-submit, reuse it if it matches the SparkApplication, and only submit when no valid driver pod exists.

can i get assigned?

rahul810050:
@Denyme24 I am currently working on this...

a7i:
Hey folks, this has gone stale.

Let me know what you think about this PR: https://github.com/kubeflow/spark-operator/pull/2932

a7i:
## Production context

We saw this failure mode in production on **2026-06-03** with two `SparkApplication` resources created at the same time, while the operator controller was rolling to a new version. This matches the restart/duplicate-submission class of problems described in this issue.

**Trigger:** A controller Deployment rollout started shortly before two new `SparkApplication` objects were created. The old and new controller pods were from different ReplicaSets.

**Sequence (both applications, identical pattern):**

1. `SparkApplication` created/updated successfully (201).
2. Operator service account `create` `{name}-driver` pod (201) ~2s later.
[...]

PR #2934, opened 2026-05-08
fix(controller): cancel in-flight spark-submit on controller shutdown

a7i:
## Purpose of this PR

Reduce duplicate driver submissions when the controller restarts mid-submit (refs #2788). Complementary to adoption in #2932: adoption recovers races this cannot avoid; this PR narrows the window by making in-flight `spark-submit` cancellable on shutdown.

On SIGTERM, controller-runtime cancels the reconcile context immediately. The post-submit status write uses that same context, so the real lever is interrupting the `spark-submit` JVM before it creates a driver pod the controller will never record. Previously `runSparkSubmit` used `exec.Command` without context propagation.

**Proposed changes:**

- `SparkSubmitter.runSparkSubmit`: use `exec.CommandContext`, send SIGTERM on cancellation (not SIGKILL), and wait up to `ShutdownGracePeriod` before force-killing the child.
- Reuse `--driver-pod-creation-grace-period` for `SparkSubmitter.ShutdownGracePeriod` (WaitDelay after SIGTERM). Leave `Manager.GracefulShutdownTimeout` at controller-runtime's 30s default (not overridden). No separate `--graceful-shutdown-timeout` flag.
- Helm: `controller.terminationGracePeriodSeconds` (default `60`), must exceed the 30s manager shutdown window.
- Startup validation: reject `driver-pod-creation-grace-period` >= 30s so WaitDelay stays below GracefulShutdownTimeout.
- Unit tests for `runSparkSubmit` (SIGTERM on cancel, WaitDelay before force-kill) and chart deployment rendering.

### How the defaults were chosen

| Setting | Default | Role |
|---------|---------|------|
| `driverPodCreationGracePeriod` | `10s` | Retry "driver pod not found" after submit; `cmd.WaitDelay` after SIGTERM to spark-submit |
| `GracefulShutdownTimeout` | `30s` (controller-runtime default) | Manager cleanup window after SIGTERM; unchanged from today |
| `terminationGracePeriodSeconds` | `60` | Kubelet ceiling; must stay above GracefulShutdownTimeout so the manager can reap in-flight JVMs before SIGKILL |

Ordering: `driverPodCreationGracePeriod` < `GracefulShutdownTimeout` (30s) < `terminationGracePeriodSeconds`.

Typical `spark-submit` in kubernetes mode with `submission-wait-app-completion=false` takes ~5-10s; worst case on a busy cluster ~15-20s. Tune `driverPodCreationGracePeriod` up if your cluster needs more headroom (must stay below 30s).

## Change Category

- [ ] Documentation update

### Rationale

Refs https://github.com/kubeflow/spark-operator/issues/2788. Complementary to #2932.

## Checklist

### Additional Notes

Locally validated:

- `go test ./internal/controller/sparkapplication/ -run TestRunSparkSubmit -timeout 120s`
- `helm unittest charts/spark-operator-chart -f 'tests/controller/deployment_test.yaml'`


internal/controller/sparkapplication/submission.go
```diff
@@ -59,22 +59,25 @@ func (*SparkSubmitter) Submit(ctx context.Context, app *v1beta2.SparkApplication
...
 
-	// Try submitting the application by running spark-submit.
+	// Try submitting the application by running spark-submit. The reconcile context is
+	// threaded into spark-submit so that controller shutdown (or any other context
+	// cancellation) terminates the child process instead of letting it run to completion
+	// and create a driver pod whose status update will never be persisted.
 	logger.Info("Running spark-submit", "arguments", args)
-	if err := runSparkSubmit(args); err != nil {
+	if err := runSparkSubmit(ctx, args); err != nil {
 		return fmt.Errorf("failed to run spark-submit: %v", err)
```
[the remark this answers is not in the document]
a7i: ignoring, this wasn't part of my changes

internal/controller/sparkapplication/submission.go
```diff
@@ -59,22 +59,25 @@ func (*SparkSubmitter) Submit(ctx context.Context, app *v1beta2.SparkApplication
...
-func runSparkSubmit(args []string) error {
+func runSparkSubmit(ctx context.Context, args []string) error {
 	sparkHome, present := os.LookupEnv(common.EnvSparkHome)
 	if !present {
 		return fmt.Errorf("env %s is not specified", common.EnvSparkHome)
 	}
 	command := filepath.Join(sparkHome, "bin", "spark-submit")
-	cmd := exec.Command(command, args...)
+	cmd := exec.CommandContext(ctx, command, args...)
 	_, err := cmd.Output()
```
tariq-hasan, line 96: I am wondering if we want to have something like the following instead to allow the JVM process to have a graceful shutdown. I say this because, in the case of context cancellation, the default [`CommandContext`](https://github.com/golang/go/blob/master/src/os/exec/exec.go#L501-L503) happens to do a direct SIGKILL without leaving scope for the process to exit gracefully.
```suggestion
    cmd := exec.CommandContext(ctx, command, args...)
    cmd.Cancel = func() error {
        return cmd.Process.Signal(syscall.SIGTERM)
    }
    cmd.WaitDelay = 10 * time.Second
    _, err := cmd.Output()
```

internal/controller/sparkapplication/submission.go
```diff
@@ -59,22 +59,25 @@ func (*SparkSubmitter) Submit(ctx context.Context, app *v1beta2.SparkApplication
...
 	if err != nil {
 		var errorMsg string
```
tariq-hasan, line 98: Also wondering if we want to have some test cases under `submission_test.go` for validation. I think the unit tests are currently bare-bones and we don't have any written for `runSparkSubmit`.

charts/spark-operator-chart/values.yaml
```diff
@@ -100,6 +100,19 @@ controller:
...
+  # reconciles finish reduces the chance of losing a status update written right after
+  # spark-submit creates the driver pod.
+  gracefulShutdownTimeout: 60s
+
+  # -- Termination grace period for the controller pod. The kubelet sends SIGKILL when this
+  # elapses. Should be set higher than `controller.gracefulShutdownTimeout` so the manager
+  # has time to drain in-flight reconciles, including any that are still writing status
+  # after a successful spark-submit.
+  terminationGracePeriodSeconds: 90
+
```
tariq-hasan, line 133: I am wondering if in-flight reconciles can really finish if a termination signal is already propagated down the controller-runtime chain to the reconcile loop before it hits the `spark-submit`.

My understanding is that the real crux of the code changes here is that JVM processes are being terminated or killed off through signal propagation so we have a better chance of avoiding driver pod creation altogether. Reconciles seem to be cancelled immediately and are simply given a grace period for cleanup rather than being allowed to finish.

tariq-hasan (commented):
Hi @a7i! I have left a few questions. Thanks!

/cc @nabuskey @ChenYi015

internal/controller/sparkapplication/submission.go
```diff
@@ -59,22 +59,25 @@ func (*SparkSubmitter) Submit(ctx context.Context, app *v1beta2.SparkApplication
 		return fmt.Errorf("failed to build spark-submit arguments: %v", err)
 	}
 
-	// Try submitting the application by running spark-submit.
+	// Try submitting the application by running spark-submit. The reconcile context is
+	// threaded into spark-submit so that controller shutdown (or any other context
+	// cancellation) terminates the child process instead of letting it run to completion
+	// and create a driver pod whose status update will never be persisted.
```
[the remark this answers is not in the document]
a7i: Reworded in b41b544 to drop the implication that cancellation prevents a driver pod from being created; the comment now says cancellation only narrows the window for losing the post-submit status update.

internal/controller/sparkapplication/submission.go
a7i: Applied in b41b544: overriding `cmd.Cancel` with a `SIGTERM` signal and setting `cmd.WaitDelay = 10s` so the JVM gets a chance to shut down gracefully before the runtime falls back to SIGKILL. Thanks for the catch!

internal/controller/sparkapplication/submission.go
a7i: Added `TestRunSparkSubmit` in b41b544 covering: missing `SPARK_HOME`, successful submission, stderr surfacing on failure, the driver-pod-already-exists case, and a SIGTERM cancellation case that asserts the trap fired (proving we send SIGTERM rather than SIGKILL).

charts/spark-operator-chart/values.yaml
a7i: Good point, you are right. The controller-runtime cancels the reconcile context as soon as SIGTERM arrives, and `updateSparkApplicationStatus` uses that same context, so the window does not let a normal reconcile finish writing status. The real value is interrupting the in-flight `spark-submit` JVM before it creates a driver pod the controller will never get to record. I reworded both `gracefulShutdownTimeout` and `terminationGracePeriodSeconds` in b41b544 to reflect that (cleanup/reap window, not "let reconciles finish"). Regenerated the chart README as well.

a7i:
## Production context

We hit the same failure mode in production on **2026-06-03** (details in [#2788](https://github.com/kubeflow/spark-operator/issues/2788#issuecomment)). Two `SparkApplication` resources were created while the controller was rolling to a new version; both followed the same sequence.

**How this PR helps:** SIGTERM during rollout interrupted an in-flight submit after the driver pod often already existed. Making `spark-submit` cancellable and widening the shutdown window should reduce orphan submissions and lost status writes on normal shutdown. That aligns with the clarified intent in review (cleanup/reap, not letting reconciles run to completion).

**What still needs #2932:** In our trace the driver was already created before cancel, and the **next** leader logged `driver pod already exist` and marked the CR `FAILED`. Adoption/idempotent submit ([#2932](https://github.com/kubeflow/spark-operator/pull/2932)) is what recovers that case; this PR and #2932 are complementary.

We are also planning `maxUnavailable: 0` / `maxSurge: 1` on the controller Deployment and chart wiring for `--graceful-shutdown-timeout` / `terminationGracePeriodSeconds` once this lands.

a7i:
Hi @tariq-hasan @nabuskey @ChenYi015 - this is a pretty low risk PR, would you consider reviewing this?

We continue to have issues in production when the operator pods are rotated or we do a new operator rollout. This causes pipelines that are mid-flight (in particular driver pod creation via spark-submit) to be cancelled.

a7i:
If you prefer, I can remove the `graceful-shutdown-timeout"` flag and instead ensure that the context is not cancelled with the `--driver-pod-creation-grace-period` as the wait delay

a7i:
> If you prefer, I can remove the `graceful-shutdown-timeout"` flag and instead ensure that the context is not cancelled with the `--driver-pod-creation-grace-period` as the wait delay

ended up doing this to reduce the changes and flags

cmd/operator/controller/start.go
```diff
@@ -312,6 +312,7 @@ func start() {
 		LeaseDuration:           &leaderElectionLeaseDuration,
 		RenewDeadline:           &leaderElectionRenewDeadline,
 		RetryPeriod:             &leaderElectionRetryPeriod,
+		GracefulShutdownTimeout: &driverPodCreationGracePeriod,
```
nabuskey: This changes the current (implicit) behaviour of `GracefulShutdownTimeout`. It currently sets to 30 seconds via`setOptionsDefaults` in controller manager.

So creation grace period setting is applied to all controller runtime managed context. 

Is this needed? As far as I can tell, it should be:  `driverPodCreationGracePeriod  <  GracefulShutdownTimeout  <  terminationGracePeriodSeconds`

If this must be in exactly this order, we could either derive it from a value or validate it at startup to protect end users.

internal/controller/sparkapplication/submission.go
```diff
@@ -44,37 +46,53 @@ type SparkApplicationSubmitter interface {
...
-	cmd := exec.Command(command, args...)
+	cmd := exec.CommandContext(ctx, command, args...)
+	// exec.CommandContext defaults to SIGKILL on cancellation. Override so we send
+	// SIGTERM first, giving spark-submit and the underlying JVM a chance to shut
+	// down gracefully (release the API server connection, flush logs). WaitDelay
+	// caps how long we wait after the signal before the process is force-killed.
+	cmd.Cancel = func() error {
+		return cmd.Process.Signal(syscall.SIGTERM)
+	}
+	cmd.WaitDelay = s.ShutdownGracePeriod
```
nabuskey, line 95: Is this a race condition? We wait for spark-submit command to exit and ShutdownGracePeriod for exactly the same time. So there's no guarantee that one happens before the other?

cmd/operator/controller/start.go
a7i: yeah initially I added a new flag called gracefulShutdownTimeout but removed it thinking that it would simplify the implementation. let me take another look

a7i: Addressed in 8a7a6a2: `GracefulShutdownTimeout` is now `driverPodCreationGracePeriod + 1s`, so the ordering is `driverPodCreationGracePeriod < GracefulShutdownTimeout < terminationGracePeriodSeconds` with defaults 10s / 11s / 60s. Chart values and flag help text document the constraint.

internal/controller/sparkapplication/submission.go
a7i: Good catch. `WaitDelay` and `GracefulShutdownTimeout` were both 10s before, so the manager could exit while `runSparkSubmit` was still in its WaitDelay window. `GracefulShutdownTimeout` is now `driverPodCreationGracePeriod + 1s`, and a test asserts we wait the full `ShutdownGracePeriod` before force-kill when SIGTERM is ignored.

cmd/operator/controller/start.go
a7i: Correction in 22469f5: we no longer override `GracefulShutdownTimeout`. It stays at controller-runtime's 30s default; `driverPodCreationGracePeriod` only sets `runSparkSubmit` WaitDelay. Startup rejects grace period >= 30s to preserve ordering.

PR #2780, opened 2025-12-16
Helm: Add toleration, affinity, and nodeselector for upgrade hook

MonkeyCanCode:
## Purpose of this PR

Currently we have toleration, affinity, and node selector for webhook/controller. However for upgrade hook, there is none. The can be pretty problematic when end-users are depending on combination of these 3 for node scaler (e.g. karpenter node pool with toleration/affinity etc.). To match the same capability, I suggest we should add support for toleration, affinity, and node selector for upgrade hook as well.

**Proposed changes:**

- Add support for toleration in upgrade hook
- Add support for affinity in upgrade hook
- Add support for nodeselector in upgrade hook

## Change Category

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [x] Documentation update

## Checklist

- [x] I have conducted a self-review of my own code.
- [x] I have updated documentation accordingly.
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.

PR #2940, opened 2026-05-12
Add labels and annotations support for upgrade hook

MonkeyCanCode:
## Purpose of this PR

Previous I added tolerations, affinity, and node selector for upgrade hook pod: https://github.com/kubeflow/spark-operator/pull/2780. For this PR, I am adding labels and annotations and here are the main reasons on why I think we should add them:
1. add annotations: this is very handy for environments where we want to control the behavior of sidecar container (which is not controlled by this operator). Take one example, assuming we have istio sidecar enabled on namespace level, the previous PR will ensure the upgrade pod will get schedule if people are using those combination for resource managements. However, if istio sidecar is enabled on namespace level, there is a chance where sidecar can start later compared the the main container (in this case, this upgrade hook pod). When istio sidecar container is not ready, this will cause the main container to not have network access thus failed the upgrade then retry. In this example, this can be solve with pod annotation to not start the main container until istio sidecar container is ready. There are many case which we may want to tweak the behavior of sidecar container via pod annotations.
2. add labels: this is more for completeness as we have labels on the job but not on the actual upgrade hook pod.

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -90,6 +90,8 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | hook.nodeSelector | object | `{}` | Node selector for the Helm hook Job. |
 | hook.affinity | object | `{}` | Affinity for the Helm hook Job. |
 | hook.tolerations | list | `[]` | List of node taints to tolerate for the Helm hook Job. |
+| hook.labels | object | `{}` | Extra labels for upgrade hook pod. |
+| hook.annotations | object | `{}` | Extra annotations for upgrade hook pod. |
 | controller.replicas | int | `1` | Number of replicas of controller. |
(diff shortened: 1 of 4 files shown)


MonkeyCanCode:
The failed integration test doesn't appear to be related to mine change.

MonkeyCanCode:
It appears DCO was missed and I just signed the commits. @tariq-hasan mind take another look?

tariq-hasan:
Changes look good to me, thanks!

The test failure looks unrelated and should be resolved if you do a rebase.

/cc @nabuskey @ChenYi015

MonkeyCanCode:
> Changes look good to me, thanks!
> 
> The test failure looks unrelated and should be resolved if you do a rebase.
> 
> /cc @nabuskey @ChenYi015

Hello,

Anything pending here on my end?

Thanks,
Yong

RobuRishabh (approved):
/lgtm

nabuskey (approved):
/approve

PR #2808, opened 2026-01-14
feat: watch namespaces based on labels

tariq-hasan:
## Purpose of this PR

This PR implements namespace selector functionality, allowing the Spark Operator to watch namespaces based on Kubernetes label selectors in addition to explicit namespace lists.

Closes #2479.

**Proposed changes:**

- **Core Implementation**
   - **New `NamespaceMatcher` utility** (`pkg/util/namespace.go`): Handles matching namespaces against explicit lists and label selectors
   - **CLI flag**: Added `--namespace-selector` flag accepting kubectl-style label selector syntax
   - **Dynamic filtering**: Event filters now check namespace labels at runtime via cache lookups
   - **Cache configuration**: When selector is used, operator caches all namespaces for dynamic filtering
- **Controller Updates**
   - **SparkApplication**: Updated to support namespace selector
   - **ScheduledSparkApplication**: Updated to support namespace selector  
   - **SparkConnect**: Updated to support namespace selector
- **Helm Chart**
   - **New value**: `spark.jobNamespaceSelector` for configuring label selector
   - **Conditional RBAC**: Namespace list/watch permissions added only when selector is configured
   - **Documentation**: Added inline documentation for selector syntax and requirements

## Change Category

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

### Rationale

In multi-tenant Kubernetes environments, managing an explicit list of namespaces can be operationally burdensome. Every time a new namespace needs to be onboarded, the operator deployment must be updated. This PR enables watching namespaces based on labels, eliminating the need for operator restarts when namespaces are added or removed.

## Checklist

- [x] I have conducted a self-review of my own code.
- [x] I have updated documentation accordingly.
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.

## Updated Implementation (Jan 25, 2026)

Force-pushed updates to clean up the implementation and add test coverage.

### Fixed
- RBAC Helm template logic for conditional cluster-scoped permissions
- Namespace matcher utility error handling and edge cases

### Added
- Integration tests covering namespace selector–based filtering

PR #2943, opened 2026-05-13
feat: Add webhook namespace-selector for multi-tenant deployments

### LINKED ISSUE #2913: Webhook namespace selector based on labels

`spark-operator` Controller can filter watched Namespaces either by their name or by label selectors. While this is very convenient for shared clusters, unfortunately this is not the case for the webhook.

The webhook can filter Namespace only based on their name. The command-line argument is there but it is never used. It would complement perfectly if webhook was filtering Namespaces based on the same labels as the controller.

### Why is this needed?

This is needed in a shared Kubernetes cluster where multiple `spark-operator` instances are running simultaneously and the namespaces are not known in advance. 

I suppose something similar to the existing implementation for the controller https://github.com/kubeflow/spark-operator/issues/2479

_No response_

### Additional context

_No response_

### Love this feature?

kouzant:
Enables running multiple spark-operator instances in one cluster, each scoped to a subset of namespaces by label.

Go:
- Add --namespace-selector flag to the webhook, mirroring the controller.
- Introduce namespaceFilteringDefaulter / namespaceFilteringValidator wrappers applied to every webhook handler at registration. Objects in unmatched namespaces skip the inner handler; non-namespaced objects pass through.
- Drop the in-process namespace map in SparkPodDefaulter -- the wrapper is now the single source of truth.
- Remove the dead --label-selector-filter flag.
- Add corev1.Namespace to the webhook cache so the wrapper's MatchesWithClient label lookup goes through the informer.
- Unit tests for both wrappers: explicit-namespace match/miss, label-selector match/miss with a fake client, NamespaceAll fast path, pass-through, and inner-error propagation.

Helm:
- Add spark.webhookNamespaceSelector as a structured LabelSelector embedded into every MutatingWebhookConfiguration / ValidatingWebhookConfiguration webhook entry's namespaceSelector field.
- Pass --namespace-selector to the webhook deployment from spark.jobNamespaceSelector.
- Grant the webhook ClusterRole namespaces get/list/watch when jobNamespaceSelector is set, mirroring the controller's pattern.

## Purpose of this PR

At the moment the only way to install multiple spark-operators in the same Kubernetes cluster is by specifying the following attribute with all the Namespaces that it will be responsible for

```yaml
spark:
  jobNamespaces:
    - ns1
    - ns2
```

The problem with this approach is that it is very static. Each time a new Namespace is created you have to `helm upgrade`.  Additionally, all namespaces declared **must** already exist as the chart will attempt creating Roles and RoleBindings in these namespaces.

Resolves https://github.com/kubeflow/spark-operator/issues/2913

**Proposed changes:**

A more dynamic approach would be for the controller and the webhook to filter events from responsible namespaces based on some labels. Luckily, https://github.com/kubeflow/spark-operator/pull/2808 already implements this but **only** for the controller. The webhook does not take the labels into consideration.

Re-use the same `util.NamespaceMatcher` struct in the webhooks responsible for
- SparkConnect
- SparkApplication
- ScheduledSparkApplication
- Pod

In addition to the webhooks above use the same logic in the `mutatingwebhookconfiguration` and `validatingwebhookconfiguration` controllers.

## Change Category

- [ ] Documentation update

### Rationale

https://github.com/kubeflow/spark-operator/issues/2913

## Checklist

### DIFF
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -456,8 +456,28 @@ spark:
   # Namespaces matching this selector will be watched in addition to those in jobNamespaces.
   # When specified, requires ClusterRole permission to list and watch namespaces.
   # Leave empty to disable namespace selector functionality.
+  # Drives the --namespace-selector flag for both the controller and the webhook deployments.
   jobNamespaceSelector: ""
 
+  # -- Structured label selector embedded into the `namespaceSelector` field of every
+  # MutatingWebhookConfiguration / ValidatingWebhookConfiguration webhook entry.
+  # When set, replaces the default name-based selector built from `jobNamespaces`.
+  # For multi-tenant deployments, this should match `jobNamespaceSelector` (which is a string)
+  # in structured form -- the Kubernetes API requires `matchLabels` / `matchExpressions`
+  # and Helm cannot reliably parse arbitrary label-selector strings.
+  # Combining this with a non-empty `jobNamespaces` is not supported.
+  webhookNamespaceSelector: {}
+  # webhookNamespaceSelector:
+  #   matchLabels:
+  #     example.com/tenant: tenantA
+
+  # Note: the chart only renders Spark ServiceAccount/Role/RoleBinding for namespaces
+  # listed explicitly in `jobNamespaces`. When `jobNamespaces` is unset, empty, or
+  # contains the "" wildcard, or when `jobNamespaceSelector` / `webhookNamespaceSelector`
+  # is used, no Spark RBAC is rendered and the cluster admin must provision a
+  # ServiceAccount + Role + RoleBinding in every namespace where Spark jobs will run.
+  # ServiceAccount is intrinsically namespaced, so there is no Helm-only way to
+  # bootstrap RBAC into a label-selected set of namespaces.
   serviceAccount:
     # -- Specifies whether to create a service account for spark applications.
     create: true
--- a/cmd/operator/webhook/start.go
+++ b/cmd/operator/webhook/start.go
@@ -45,6 +45,7 @@ import (
 	logzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	ctrlwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/kubeflow/spark-operator/v2/api/v1alpha1"
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
@@ -54,6 +55,7 @@ import (
 	"github.com/kubeflow/spark-operator/v2/pkg/certificate"
 	"github.com/kubeflow/spark-operator/v2/pkg/common"
 	operatorscheme "github.com/kubeflow/spark-operator/v2/pkg/scheme"
+	"github.com/kubeflow/spark-operator/v2/pkg/util"
 	"github.com/kubeflow/spark-operator/v2/pkg/version"
 	// +kubebuilder:scaffold:imports
 )
@@ -63,8 +65,8 @@ var (
 )
 
 var (
-	namespaces          []string
-	labelSelectorFilter string
+	namespaces        []string
+	namespaceSelector string
 
 	// Controller
 	controllerThreads int
@@ -128,7 +130,7 @@ func NewStartCommand() *cobra.Command {
 	// Controller
 	command.Flags().IntVar(&controllerThreads, "controller-threads", 10, "Number of worker threads used by the SparkApplication controller.")
 	command.Flags().StringSliceVar(&namespaces, "namespaces", []string{}, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset or contains empty string.")
-	command.Flags().StringVar(&labelSelectorFilter, "label-selector-filter", "", "A comma-separated list of key=value, or key labels to filter resources during watch and list based on the specified labels.")
+	command.Flags().StringVar(&namespaceSelector, "namespace-selector", "", "Label selector for namespaces to watch (e.g., 'spark-operator=enabled,env in (prod,staging)'). Namespaces matching this selector will be watched in addition to those specified via --namespaces. Requires ClusterRole permission to list and watch namespaces.")
 	command.Flags().DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 30*time.Second, "Informer cache sync timeout.")
 
 	command.Flags().Float32Var(&kubeAPIQPS, "kube-api-qps", 20, "Maximum QPS to the API server from the controller client.")
@@ -291,10 +293,23 @@ func start() {
 		}
 	}
 
+	namespaceMatcher, err := util.NewNamespaceMatcher(namespaces, namespaceSelector)
+	if err != nil {
+		logger.Error(err, "Failed to build namespace matcher")
+		os.Exit(1)
+	}
+
+	wrapDefaulterWithNamespaceFiltering := func(d admission.CustomDefaulter) admission.CustomDefaulter {
+		return webhook.NewNamespaceFilteringDefaulter(d, namespaceMatcher, mgr.GetClient())
+	}
+	wrapValidatorWithNamespaceFiltering := func(v admission.CustomValidator) admission.CustomValidator {
+		return webhook.NewNamespaceFilteringValidator(v, namespaceMatcher, mgr.GetClient())
+	}
+
 	if err := ctrl.NewWebhookManagedBy(mgr).
 		For(&v1alpha1.SparkConnect{}).
-		WithDefaulter(webhook.NewSparkConnectDefaulter()).
-		WithValidator(webhook.NewSparkConnectValidator()).
+		WithDefaulter(wrapDefaulterWithNamespaceFiltering(webhook.NewSparkConnectDefaulter())).
+		WithValidator(wrapValidatorWithNamespaceFiltering(webhook.NewSparkConnectValidator())).
 		WithLogConstructor(webhook.LogConstructor).
 		Complete(); err != nil {
 		logger.Error(err, "Failed to create mutating webhook for SparkConnect")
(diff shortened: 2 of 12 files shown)


cmd/operator/webhook/start.go
```diff
@@ -291,10 +293,23 @@ func start() {
 		}
 	}
 
+	namespaceMatcher, err := util.NewNamespaceMatcher(namespaces, namespaceSelector)
+	if err != nil {
+		logger.Error(err, "Failed to build namespace matcher")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
kouzant: This review comment is incorrect. If `--namespaces` is "" this is a sentinel value which will match all namespaces.
If you pass `--namespaces=""` and a Selector, then based on the documentation of the `NamespaceMatcher` struct, it should match all. I also added two test cases to cover this.

cmd/operator/webhook/start.go
```diff
@@ -394,6 +409,7 @@ func newCacheOptions() cache.Options {
 	}
 
 	byObject := map[client.Object]cache.ByObject{
+		&corev1.Namespace{}: {},
 		&corev1.Pod{}: {
 			Label: labels.SelectorFromSet(labels.Set{
```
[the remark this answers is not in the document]
kouzant: Indeed. I fixed it and I cache Namespaces conditionally

internal/webhook/namespace_filter.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return w.inner.ValidateDelete(ctx, obj)
+}
+
+func matchesNamespace(ctx context.Context, c client.Client, matcher *util.NamespaceMatcher, obj runtime.Object) (bool, error) {
+	meta, ok := obj.(metav1.Object)
+	if !ok {
+		return true, nil
+	}
+	return matcher.MatchesWithClient(ctx, c, meta.GetNamespace())
```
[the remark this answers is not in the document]
kouzant: Good call. I fixed that. I also changed the behaviour to reject object that are not `metav1.Object` or they don't have a Namespace

pkg/util/namespace_test.go
```diff
@@ -96,6 +102,13 @@ var _ = Describe("NamespaceMatcher", func() {
 			Expect(matcher.Matches("", nil)).To(BeTrue())
 		})
 
+		It("should match all namespaces when namespaces list is just an empty string and has a Selector", func() {
+			matcher, _ := util.NewNamespaceMatcher([]string{""}, "spark=true")
+			Expect(matcher.Matches("any-namespace", nil)).To(BeTrue())
+			Expect(matcher.Matches("another-namespace", nil)).To(BeTrue())
+			Expect(matcher.Matches("", nil)).To(BeTrue())
```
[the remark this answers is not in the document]
kouzant: That is not correct. Look at `pkg/util/namespace.go` If `len(explicitNamespaces) != 0` then it will add `""` to `nsMap` . `metav1.NamespaceAll` is `""` empty string.

If you pass an empty string in `--namespaces` it will match all namespaces; that's the behavior of `NamespaceMatcher`

cmd/operator/webhook/start.go
```diff
@@ -291,10 +294,23 @@ func start() {
...
 
+	namespaceMatcher, err := util.NewNamespaceMatcher(namespaces, namespaceSelector)
+	if err != nil {
+		logger.Error(err, "Failed to build namespace matcher")
+		os.Exit(1)
+	}
+
+	wrapDefaulterWithNamespaceFiltering := func(d admission.CustomDefaulter) admission.CustomDefaulter {
+		return webhook.NewNamespaceFilteringDefaulter(d, namespaceMatcher, mgr.GetClient())
+	}
```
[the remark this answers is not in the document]
kouzant: Same comment as above. `""` is the sentinel `metav1.NamespaceAll` which will match all namespaces

cmd/operator/webhook/start.go
```diff
@@ -128,7 +131,7 @@ func NewStartCommand() *cobra.Command {
 	// Controller
 	command.Flags().IntVar(&controllerThreads, "controller-threads", 10, "Number of worker threads used by the SparkApplication controller.")
 	command.Flags().StringSliceVar(&namespaces, "namespaces", []string{}, "The Kubernetes namespace to manage. Will manage custom resource objects of the managed CRD types for the whole cluster if unset or contains empty string.")
-	command.Flags().StringVar(&labelSelectorFilter, "label-selector-filter", "", "A comma-separated list of key=value, or key labels to filter resources during watch and list based on the specified labels.")
+	command.Flags().StringVar(&namespaceSelector, "namespace-selector", "", "Label selector for namespaces to watch (e.g., 'spark-operator=enabled,env in (prod,staging)'). Namespaces matching this selector will be watched in addition to those specified via --namespaces. Requires ClusterRole permission to list and watch namespaces.")
```
vjanelle, line 134: Why are you replacing the label selector?

kouzant: This was dead code. The variable was there but it was never used. I suppose the original author intended to implement that.

I changed it so it matches the flag of the controller which is also `namespace-selector` https://github.com/kubeflow/spark-operator/blob/v2.5.0/cmd/operator/controller/start.go#L165

charts/spark-operator-chart/values.yaml
```diff
@@ -456,8 +456,28 @@ spark:
...
   # Leave empty to disable namespace selector functionality.
+  # Drives the --namespace-selector flag for both the controller and the webhook deployments.
   jobNamespaceSelector: ""
 
+  # -- Structured label selector embedded into the `namespaceSelector` field of every
+  # MutatingWebhookConfiguration / ValidatingWebhookConfiguration webhook entry.
+  # When set, replaces the default name-based selector built from `jobNamespaces`.
+  # For multi-tenant deployments, this should match `jobNamespaceSelector` (which is a string)
+  # in structured form -- the Kubernetes API requires `matchLabels` / `matchExpressions`
+  # and Helm cannot reliably parse arbitrary label-selector strings.
```
tariq-hasan, line 467: Since this is advisory only should there be some kind of Helm validation on this? Ideally we ought to have a single source of truth to ensure consistency.

charts/spark-operator-chart/values.yaml
```diff
@@ -456,8 +456,28 @@ spark:
...
+  # Combining this with a non-empty `jobNamespaces` is not supported.
```
tariq-hasan, line 468:
```
# Combining this with a non-empty `jobNamespaces` is not supported.
```

This directly contradicts the behavior supported on the in-process webhook: https://github.com/kouzant/spark-operator/blob/331390a36ef82450ab10add11bbc37a5a4d150c0/pkg/util/namespace.go#L42. Do we want server-side logic to be in line with the webhook logic?

cmd/operator/webhook/start.go
tariq-hasan: Setting `defaultNamespaces` to `nil` is necessary so that, for the case when both `jobNamespaces` and `jobNamespaceSelector` are specified, the cache is able to watch all namespaces and filter by the namespace selector specifically for namespaced informers such as SparkApplication. This goes in line with how the controller sets its informer cache: https://github.com/kouzant/spark-operator/blob/331390a36ef82450ab10add11bbc37a5a4d150c0/cmd/operator/controller/start.go#L406-L416.

```suggestion
	var defaultNamespaces map[string]cache.Config
	if strings.TrimSpace(namespaceSelector) != "" {
		defaultNamespaces = nil
	} else if !slices.Contains(namespaces, cache.AllNamespaces) {
		defaultNamespaces = make(map[string]cache.Config)
		for _, ns := range namespaces {
			defaultNamespaces[ns] = cache.Config{}
		}
	}
```

tariq-hasan (commented):
Hi @kouzant! I have left a few comments. Thanks.

charts/spark-operator-chart/values.yaml
kouzant: This comment refers to `webhook/_helpers.tpl` template function

```
{{/*
Render the namespaceSelector field for a webhook entry in the
MutatingWebhookConfiguration / ValidatingWebhookConfiguration. Emits nothing
if neither `spark.webhookNamespaceSelector` is set nor `spark.jobNamespaces`
holds a concrete list (an entry of "" is the all-namespaces sentinel and
disables the name-based selector). Callers indent the result with `nindent`.
*/}}
{{- define "spark-operator.webhook.namespaceSelector" -}}
{{- if .Values.spark.webhookNamespaceSelector -}}
namespaceSelector:
{{ toYaml .Values.spark.webhookNamespaceSelector | indent 2 -}}
{{- else if and .Values.spark.jobNamespaces (not (has "" .Values.spark.jobNamespaces)) -}}
namespaceSelector:
  matchExpressions:
  - key: kubernetes.io/metadata.name
    operator: In
    values:
{{- range $jobNamespace := .Values.spark.jobNamespaces }}
    - {{ $jobNamespace }}
{{- end }}
{{- end -}}
{{- end -}}
```

If one has set both `webhookNamespaceSelector` and `jobNamespaces` then the latter is silently ignored.
The problem is that merging them becomes complicated. If one wants to include namespaces matching some labels AND metadata.name then could do it with something like the following:

```yaml
webhookNamespaceSelector:
    matchLabels:
      example.com/tenant: tenantA
    matchExpressions:
      - key: kubernetes.io/metadata.name
         operator: In
         values: [ns-a, ns-b]  
```

cmd/operator/webhook/start.go
kouzant: That's a good catch! I updated my patch with a similar change as yours. Thanks ;)

charts/spark-operator-chart/values.yaml
kouzant: The problem is that parsing and comparing something like
```
jobNamespaceSelector: "tenant=tenantA"
```
and

```yaml
webhookNamespaceSelector:
  matchLabels:
    tenant: tenantA
```

in Helm is not easy - and possibly error prone.
A possible solution would be to compare the two values in Go but it would be too late. The wrong configuration would have already been applied

RobuRishabh: can we add Helm fail when string selector set without structured one + document example @tariq-hasan

charts/spark-operator-chart/values.yaml
RobuRishabh: @kouzant Align Helm with Go (merge at API level)

RobuRishabh (changes requested):
@kouzant please rebase onto master, and take a look at new comments.

PR #2947, opened 2026-05-19
feat(controller): log when SparkApplication is marked Failing after driver-pod grace period

dineshkumar181094:
## Summary

  When the driver pod is missing past `DriverPodCreationGracePeriod`, `updateDriverState`
  transitions the `SparkApplication` to `Failing` and returns `nil`. The only signal of *why* this
  happened is `Status.AppState.ErrorMessage = "driver pod not found"` on the CR, which is invisible
  unless someone runs `kubectl describe sparkapplication`. Controller-runtime logs nothing for a
  `nil` return either, so cluster operators have no way to tell from controller logs why an app was
  marked `Failing`.

  This PR adds a single `logger.Info` line at the moment of the transition, with the namespace/name,
   expected driver pod name, previous state, last submission attempt time, and the configured grace
  period — enough to reconstruct the decision from logs alone.

  ## Changes

  - `internal/controller/sparkapplication/controller.go` — emit an `Info` log when the driver pod is
   missing after the grace period and the app is being marked `Failing`.

### DIFF
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -982,6 +982,19 @@ func (r *Reconciler) updateDriverState(ctx context.Context, app *v1beta2.SparkAp
 				app.Status.DriverInfo.PodName,
 				r.options.DriverPodCreationGracePeriod,
 			)
+			logger := log.FromContext(ctx)
+			if app.Status.AppState.State == v1beta2.ApplicationStateSubmitted {
+				logger.Info("Driver pod not found after creation grace period; marking SparkApplication as Failing",
+					"driverPodName", app.Status.DriverInfo.PodName,
+					"lastSubmissionAttemptTime", app.Status.LastSubmissionAttemptTime,
+					"gracePeriod", r.options.DriverPodCreationGracePeriod,
+				)
+			} else {
+				logger.Info("Driver pod not found; marking SparkApplication as Failing",
+					"driverPodName", app.Status.DriverInfo.PodName,
+					"previousState", app.Status.AppState.State,
+				)
+			}
 			app.Status.AppState.State = v1beta2.ApplicationStateFailing
 			app.Status.AppState.ErrorMessage = "driver pod not found"
 			app.Status.TerminationTime = metav1.Now()


internal/controller/sparkapplication/controller.go
```diff
@@ -974,6 +974,15 @@ func (r *Reconciler) updateDriverState(ctx context.Context, app *v1beta2.SparkAp
 
 	if driverPod == nil {
 		if app.Status.AppState.State != v1beta2.ApplicationStateSubmitted || metav1.Now().Sub(app.Status.LastSubmissionAttemptTime.Time) > r.options.DriverPodCreationGracePeriod {
+			logger := log.FromContext(ctx)
+			logger.Info("Marking SparkApplication as Failing: driver pod not found after grace period",
+				"namespace", app.Namespace,
```
RobuRishabh: Nice improvement for observability! One thing to note: the logger obtained via `log.FromContext(ctx)` in a controller-runtime reconciler already carries the namespace and name of the reconciled object as structured fields. Adding them explicitly here creates duplicate keys in the structured log output.

Looking at the existing `logger.Info` calls in this file, none of them explicitly pass "namespace" or "name" they rely on the context-injected values. So maybe we can remove the "namespace" and "name" key-value pairs

RobuRishabh (commented):
Hey @dineshkumar181094, thanks for the contribution! The DCO check is failing because the commit is missing a Signed-off-by line. For future commits also, always use `git commit -s` to automatically append the sign-off line.

internal/controller/sparkapplication/controller.go
dineshkumar181094: Thanks @RobuRishabh, I have updated the PR.

internal/controller/sparkapplication/controller.go
```diff
@@ -974,6 +974,13 @@ func (r *Reconciler) updateDriverState(ctx context.Context, app *v1beta2.SparkAp
 
 	if driverPod == nil {
 		if app.Status.AppState.State != v1beta2.ApplicationStateSubmitted || metav1.Now().Sub(app.Status.LastSubmissionAttemptTime.Time) > r.options.DriverPodCreationGracePeriod {
+			logger := log.FromContext(ctx)
+			logger.Info("Marking SparkApplication as Failing: driver pod not found after grace period",
+				"driverPodName", app.Status.DriverInfo.PodName,
+				"previousState", app.Status.AppState.State,
+				"lastSubmissionAttemptTime", app.Status.LastSubmissionAttemptTime,
+				"gracePeriod", r.options.DriverPodCreationGracePeriod,
+			)
```
tariq-hasan: One additional comment. Since `updateDriverState` is called from the `Submitted`, `Running` and `Unknown` reconciliation loops I am wondering if we want to restructure the logs to better reflect what's happening - like in my suggestion below. Or if we do not want to make it as verbose we can just remove the `after grace period` phrasing since it is only applicable when we are in the `Submitted` state.

```suggestion
			logger := log.FromContext(ctx)
			if app.Status.AppState.State == v1beta2.ApplicationStateSubmitted {
			    logger.Info("Driver pod not found after creation grace period; marking SparkApplication as Failing",
			        "driverPodName", app.Status.DriverInfo.PodName,
			        "lastSubmissionAttemptTime", app.Status.LastSubmissionAttemptTime,
			        "gracePeriod", r.options.DriverPodCreationGracePeriod,
			    )
			} else {
			    logger.Info("Driver pod not found; marking SparkApplication as Failing",
			        "driverPodName", app.Status.DriverInfo.PodName,
			        "previousState", app.Status.AppState.State,
			    )
			}
```

dineshkumar181094: Indeed make more sense here.

RobuRishabh (changes requested):
Thanks for the contributions @dineshkumar181094. Everything is fine. Just the code-check CI job is failing because the code isn't gofmt-formatted. Please run `make go-fmt` and push the fix.

Before pushing, it's good practice to run these locally to catch CI issues early:
```
make go-fmt
make go-lint
make unit-test
make kustomize-lint
make drift-check
```
See the [Developer Guide](https://www.kubeflow.org/docs/components/spark-operator/developer-guide/) for the full list of available targets.

dineshkumar181094:
Thanks @RobuRishabh for reviewing the PR I have squashed the commit and pushed, new learning about the project, I will make sure I keep them in mind in future PRs.

RobuRishabh (approved):
Thanks for the contribution @dineshkumar181094!

/lgtm

nabuskey (approved):
/approve

nabuskey:
@dineshkumar181094  Can you rebase your branch so that this can be merged?

dineshkumar181094:
@nabuskey I have rebased my pr.

RobuRishabh (approved):
LGTM

Issue #2949, opened 2026-05-19
feat: Dynamic PodDisruptionBudget for Spark driver


RobuRishabh:
### What feature you would like to be added?
Add a feature flag in the SparkApplication CRD that enables the operator to dynamically create and manage a PodDisruptionBudget (PDB) for the Spark driver pod, protecting it from voluntary disruptions such as node drains, cluster upgrades, and autoscaler scale-downs.
```
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: batch-job
spec:
  driver:
    podDisruptionBudget:
      enable: true
```
When enabled, the operator creates a `policy/v1` PDB with `minAvailable: 1` targeting the driver pod. The operator fully manages the PDB lifecycle creating it when the application is submitted and deleting it when the application terminates. No PDB management is done for executor pods.

### Why is this needed?
There is exactly one driver pod per SparkApplication. It is the single coordinator for the entire job. If the driver gets evicted during a voluntary disruption (e.g., kubectl drain), the entire Spark job fails immediately all executor work is lost, and the job must restart from scratch. For batch jobs that run for hours, this is catastrophic.

A PDB with `minAvailable: 1` on the driver tells Kubernetes: "do not voluntarily evict this pod." This is the simplest and most effective protection against disruption-induced job failures.

#### Important caveat: this is for batch jobs only.
A PDB on the driver means Kubernetes cannot voluntarily evict it, which means the pod stays on its node until the job completes. For streaming jobs (Spark Structured Streaming), the job never terminates on its own the driver would block node drains indefinitely. Users running streaming workloads should only enable this if they have out-of-band management to handle graceful shutdown (e.g., external orchestration that stops the application before maintenance windows).

What exists today covers only the operator itself. PR #2918 and the Helm chart's controller.podDisruptionBudget.enable protect the operator's own controller and webhook deployments. There is no mechanism to protect the Spark driver pod that actually runs the user's job.

### Describe the solution you would like

Feature flag in the CRD, Add an enable boolean under a new podDisruptionBudget field on DriverSpec. When true, the operator creates a PDB with `minAvailable: 1` targeting the driver pod. When false or unset, no PDB is created (backward compatible).

Operator manages the full PDB lifecycle:
- Create the PDB when the SparkApplication enters the SUBMITTED state (same phase where Services and Ingresses are already created)
[...]

PR #2954, opened 2026-05-22
feat: add per-application driver PodDisruptionBudget (#2949)

vjanelle:
## Purpose of this PR

Adds an opt-in per-`SparkApplication` driver PodDisruptionBudget (PDB) so
operators can prevent voluntary disruption (node drains, autoscaler
evictions) of long-running drivers without resorting to cluster-wide
policy.

The feature is gated on two switches that both default to off, so the
change is a no-op for installations that do not opt in:

1. A controller-level flag, `--enable-driver-pdb` (chart value
   `controller.driverPodDisruptionBudget.enable`), gives cluster operators
   a single permission point.
2. A per-application field, `spec.driverPodDisruptionBudget` (`*bool`),
   lets app authors opt in once the operator gate is open.

When both gates are true, the controller maintains a `policy/v1`
PodDisruptionBudget with `minAvailable: 1` and a `matchLabels` selector
that targets the driver pod by name. The PDB is owner-referenced to the
`SparkApplication` so it is GC'd at app deletion, and the reconcile loop
also cleans it up when the app reaches a terminal state or when the spec
is flipped back off.

PDB existence is converged on every non-terminal reconcile (Submitted and
Running state handlers), so:

- Transient API errors during PDB create are retried via standard
  controller-runtime backoff.
- An operator restart with `--enable-driver-pdb` turned on retroactively
  creates PDBs for already-running apps that opted in.
- Flipping `spec.driverPodDisruptionBudget` from false to true (or true
  to false) on a running app takes effect on the next reconcile cycle.

**Proposed changes:**

- API (`api/v1beta2`): add `SparkApplicationSpec.DriverPodDisruptionBudget *bool`,
  regenerate CRDs, deepcopy, OpenAPI, Python API, and chart CRDs.
- Controller (`internal/controller/sparkapplication`): new `driver_pdb.go`
  with idempotent `ensureDriverPDB` / `createDriverPDB` / `deleteDriverPDB`
  helpers; `ensureDriverPDB` wired into the Submitted and Running reconcile
  branches; PDB cleanup in `cleanUpOnTermination`.
- Controller flag: `--enable-driver-pdb` (default false), plumbed through
  `cmd/operator/controller` into `Reconciler.Options.EnableDriverPDB`.
- RBAC: add `policy/poddisruptionbudgets` create/get/list/watch/update/patch/delete
  to the controller role (Helm and Kustomize, in sync).
- Helm chart: `controller.driverPodDisruptionBudget.enable` value, gated
  RBAC rule, and helper template that emits the `--enable-driver-pdb`
  flag when set. Helm-unittest specs cover the on/off matrix.
- Tests: 13 envtest specs in `driver_pdb_test.go` covering both gates,
  spec-flip lifecycle, terminal-state cleanup, OwnerReferences GC, and
  reconcile-loop convergence after operator restart or spec flip; one
  Ginkgo e2e spec in `test/e2e/pdb_test.go` exercising the happy path
  on Kind.
- Tooling: two Makefile fixes that the new e2e suite depends on -
  `make deploy` loads the Spark runtime image into Kind alongside the
  operator image, and `make e2e-test` creates the Kind cluster on demand
  instead of assuming one exists.

## Change Category

- [ ] Documentation update

### Rationale

A missing-driver event (node drain, autoscaler eviction, kubelet upgrade)
restarts the entire Spark application from scratch. For long-running
drivers this is an expensive failure mode that PDBs were designed to
prevent, but cluster-wide PDB policy is too blunt - most users want PDBs
only for opt-in drivers, not for every Spark driver in the cluster.

The two-gate design (operator flag plus per-app field) keeps the cluster
operator in control of whether the feature exists at all, while letting
individual app authors opt their long-running pipelines in. Defaulting
both gates to off avoids any behavior change for installations that do
not opt in.

Reconcile-loop convergence (rather than a one-shot create on submit)
keeps PDB state correct across operator restarts, transient API failures,
and spec edits - a missing PDB silently defeats the entire feature, so
the controller errs on the side of repeatedly ensuring it exists.

## Checklist

### Additional Notes

- All generated artifacts (CRDs in `config/crd/bases` and
  `charts/spark-operator-chart/crds`, deepcopy, OpenAPI, Python API,
  api-docs.md) are committed. `make detect-crds-drift`,
  `make verify-codegen`, `make manifests`, `make helm-unittest`,
  `make kustomize-lint`, `make go-lint`, and `make unit-test` all pass.
- The feature is split across five commits: API + codegen, operator
  code, tests, and two Makefile chores that the e2e suite depends on.
- The e2e spec was verified manually on Kind with `DEPLOY_METHOD=helm`.

---

Co-authored with [Claude Code](https://claude.com/claude-code).


internal/controller/sparkapplication/controller.go
```diff
@@ -1148,6 +1175,10 @@ func (r *Reconciler) deleteSparkResources(ctx context.Context, app *v1beta2.Spar
 		return err
 	}
 
+	if err := r.deleteDriverPDB(ctx, app); err != nil {
```
dineshkumar181094: `deleteDriverPDB` should be called before `deleteDriverPod` else you will never be able to delete the driver pod.

internal/controller/sparkapplication/driver_pdb.go
```diff
@@ -0,0 +1,130 @@
...
+		return fmt.Errorf("failed to set controller reference on driver PDB: %v", err)
+	}
+
+	logger.Info("Creating driver PodDisruptionBudget", "name", pdb.Name, "namespace", pdb.Namespace)
+	if err := r.client.Create(ctx, pdb); err != nil {
+		if errors.IsAlreadyExists(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to create driver PDB: %v", err)
+	}
```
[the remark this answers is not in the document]
dineshkumar181094: +1 to this and it can slow down the reconciliation, better to add pdb object in cache.
SetControllerReference.
https://github.com/kubeflow/spark-operator/blob/master/cmd/operator/controller/start.go#L420-L444

internal/controller/sparkapplication/driver_pdb.go
```diff
@@ -0,0 +1,130 @@
...
+// to false (or to nil) on a running app must still drop the existing PDB,
+// otherwise it would orphan until the SparkApplication is deleted and kube
+// GC runs via OwnerReferences.
+func (r *Reconciler) deleteDriverPDB(ctx context.Context, app *v1beta2.SparkApplication) error {
+	if !r.options.EnableDriverPDB {
+		return nil
+	}
+
+	logger := log.FromContext(ctx)
+	pdb := r.createDriverPDBObject(app)
```
dineshkumar181094: Why create DriverPDB on deleteDriverPDB?

internal/controller/sparkapplication/controller.go
```diff
@@ -348,6 +354,13 @@ func (r *Reconciler) reconcileSubmittedSparkApplication(ctx context.Context, req
 				return err
 			}
 
+			// Converge the driver PDB to spec on every Submitted-state pass.
+			// Bubbles up so controller-runtime backoff retries transient
+			// API failures.
+			if err := r.ensureDriverPDB(ctx, app); err != nil {
```
dineshkumar181094, line 378: Is there any reason why this is not a part of [submitSparkApplication](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/controller.go#L897).
You create a spark application then later   pdb not working. I think this should be the part of submitSparkApplication.

vjanelle: For when we later replace submitSparkApplication - I'm thinking of making that configurable

internal/controller/sparkapplication/controller.go
tariq-hasan, line 305: I am wondering if it would make sense to also have `Owns(&policyv1.PodDisruptionBudget{}).` to ensure PDB changes are propagated not only to the cache but also to the controller workqueue for immmediate reconciliation.

charts/spark-operator-chart/values.yaml
```diff
@@ -103,6 +103,10 @@ controller:
   # -- Specifies the maximum number of Executor pods that can be tracked by the controller per SparkApplication.
   maxTrackedExecutorPerApp: 1000
 
+  # -- Driver PDB feature gate. When true, the controller creates a PodDisruptionBudget for each SparkApplication that sets spec.driverPodDisruptionBudget=true. Default false.
+  driverPodDisruptionBudget:
+    enable: false
+
```
tariq-hasan, line 130: Since Kustomize does not have this flag an RBAC mismatch shows up when running `make drift-check`. I am wondering if it would be ideal to add a flag in `test/drift/drift-check-values.yaml` to ensure Helm and Kustomize parity for the PDB configs.

tariq-hasan (commented):
Hi @vjanelle! I left a few comments.

internal/controller/sparkapplication/controller.go
RobuRishabh: I don't think it should be part of submitsparkapplication. Putting PDB creation inside submitSparkApplication would mean a PDB Create failure would cause the app to flip to FailedSubmission state (via the deferred error handler), even though spark-submit already succeeded and the driver is already running. The reconcile-loop convergence model is strictly better here, it decouples PDB errors from submission state transitions.

charts/spark-operator-chart/values.yaml
vjanelle: Always generating the PDB rbac now, couldn't quite figure out how to handle between it, kustomize, and kubebuilder

vjanelle:
Thanks - I think I addressed the comments

vjanelle:
> /ok-to-test

Not sure what to do here

RobuRishabh:
@nabuskey @vara-bonthu could you please trigger the CI here, when you get a chance. Thank You

RobuRishabh (changes requested):
Hi @vjanelle, Thanks for the contribution. The kustomize e2e matrix deploys via this manifest, but it doesn't pass --enable-driver-pdb=true. Helm gets the flag from controller.driverPodDisruptionBudget.enable: true in ci-values.yaml; kustomize has no equivalent, so createDriverPDB is a no-op (driver_pdb.go:79) and the PDB e2e tests time out waiting for a PDB that will never be created.

Please add the flag here to match Helm CI behavior (RBAC in config/rbac/role.yaml is already in place):
```
            - --driver-pod-creation-grace-period=10s
            - --enable-driver-pdb=true
            - --max-tracked-executor-per-app=1000
 ```

**Review by RobuRishabh — CHANGES_REQUESTED** [review]
Hi @vjanelle, one small fix left: drift-check is failing on TestDrift/DeploymentSpecs/ControllerArgs because kustomize injects --enable-driver-pdb=true but test/drift/drift-check-values.yaml does not. Please add:
```
controller:
  driverPodDisruptionBudget:
    enable: true
```
 After that, this looks good.

vjanelle:
Ah I see what I missed, the e2e tests take external environment variables.

Is there a reason to not just test both?

RobuRishabh:
> Ah I see what I missed, the e2e tests take external environment variables.
> 
> Is there a reason to not just test both?

@vjanelle I get what you're saying, testing both flag states, the operator-level flag (`--enable-driver-pdb`) is global config; spec opt-in is per app. Flag-off behavior is covered by unit tests (`enableGate: false`). Doubling the e2e matrix (8 → 16 jobs) for flag-off feels heavy when opt-out e2e (spec omitted, no PDB) already runs with the flag on.

Since kustomize e2e already does deploy-time overrides (IMAGE_TAG), could we inject `--enable-driver-pdb=true` in installViaKustomize() for tests only (like helm’s ci-values.yaml), instead of patching config/default? That keeps production defaults unchanged and avoids the drift-check failure. I'm open to either approach, just thinking this is simpler for the PR. wdyt?

RobuRishabh:
Hi @vjanelle I think the drift-check will still fail. This comment still needs to be addressed https://github.com/kubeflow/spark-operator/pull/2954#pullrequestreview-4556946176
After that, this looks good to go.

vjanelle:
@RobuRishabh yup I'm looking at it

vjanelle:
@RobuRishabh should've addressed it I think - it can be selectively enabled via kustomize, defaults to off (because of streaming jobs), can be enabled via helm/kustomize, and the e2e tests enable it so they can test the scenario.

RobuRishabh (approved):
/lgtm

nabuskey (approved):
/approve

PR #2956, opened 2026-05-25
fix(controller): avoid invalidating apps on TTL updates

dineshkumar181094:
## Purpose of this PR
This PR prevents `SparkApplication` updates that only change `spec.timeToLiveSeconds` from moving the application into the `INVALIDATING` state.
**Proposed changes:**
- Treat `spec.timeToLiveSeconds` like `spec.suspend`, allowing it to be updated without invalidating the running application.
- Add regression coverage to verify TTL-only `SparkApplication` updates do not trigger invalidation.
- Add `ScheduledSparkApplication` coverage for updates to `spec.template.timeToLiveSeconds`.
## Change Category
- [ ] Documentation update
### Rationale
`timeToLiveSeconds` only controls garbage collection after a SparkApplication has terminated. Updating this field should not require rerunning or invalidating the application because it does not affect the running driver or executor pods.
## Checklist
### Additional Notes
Tested locally with:
`go test ./internal/controller/sparkapplication ./internal/controller/scheduledsparkapplication`

### DIFF
--- a/internal/controller/scheduledsparkapplication/event_filter_test.go
+++ b/internal/controller/scheduledsparkapplication/event_filter_test.go
@@ -22,7 +22,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
@@ -339,39 +338,6 @@ func TestEventFilter_Update_NonScheduledSparkApplication(t *testing.T) {
 
 	if result != false {
 		t.Errorf("Update() with non-ScheduledSparkApplication = %v, expected false", result)
-	}
-}
-
-func TestEventFilter_Update_TemplateTimeToLiveSeconds(t *testing.T) {
-	scheme := runtime.NewScheme()
-	_ = corev1.AddToScheme(scheme)
-	_ = v1beta2.AddToScheme(scheme)
-	fakeClient := fake.NewClientBuilder().WithScheme(scheme).Build()
-
-	filter, err := NewEventFilter(fakeClient, []string{"default"}, "")
-	if err != nil {
-		t.Fatalf("NewEventFilter() unexpected error: %v", err)
-	}
-
-	oldApp := &v1beta2.ScheduledSparkApplication{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-app",
-			Namespace: "default",
-		},
-		Spec: v1beta2.ScheduledSparkApplicationSpec{
-			Template: v1beta2.SparkApplicationSpec{
-				Type:         v1beta2.SparkApplicationTypeScala,
-				SparkVersion: "3.5.0",
-			},
-		},
-	}
-	newApp := oldApp.DeepCopy()
-	newApp.Spec.Template.TimeToLiveSeconds = ptr.To[int64](60)
-
-	result := filter.Update(event.UpdateEvent{ObjectOld: oldApp, ObjectNew: newApp})
-
-	if result != true {
-		t.Errorf("Update() = %v, expected true", result)
 	}
 }
 
--- a/internal/controller/scheduledsparkapplication/event_filter_test.go
+++ b/internal/controller/scheduledsparkapplication/event_filter_test.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


dineshkumar181094:
@nabuskey we need to enforce ttl on all our spark app without restarting them. This could be good bug fix.

internal/controller/sparkapplication/event_filter.go
```diff
@@ -179,9 +179,11 @@ func (f *EventFilter) Update(e event.UpdateEvent) bool {
 	// This is currently best effort as we can potentially miss updates and end up in an inconsistent state.
 	if !equality.Semantic.DeepEqual(oldApp.Spec, newApp.Spec) {
 
-		// Only Spec.Suspend can be updated without any action
+		// Spec.Suspend and Spec.TimeToLiveSeconds can be updated without invalidating
+		// the running application.
 		oldAppCopy := oldApp.DeepCopy()
 		oldAppCopy.Spec.Suspend = newApp.Spec.Suspend
+		oldAppCopy.Spec.TimeToLiveSeconds = newApp.Spec.TimeToLiveSeconds
```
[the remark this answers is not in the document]
nabuskey: This is valid. Just need to treat it like the suspend field.

nabuskey (commented):
Thanks for the PR! Just a small fix and it's good to go

internal/controller/sparkapplication/event_filter.go
```diff
@@ -179,9 +179,11 @@ func (f *EventFilter) Update(e event.UpdateEvent) bool {
 	// This is currently best effort as we can potentially miss updates and end up in an inconsistent state.
```
RobuRishabh: Nit: This comment is now stale. Since TimeToLiveSeconds is also exempted from invalidation (just like Suspend), please update it
```
	// The spec has changed except for Spec.Suspend and Spec.TimeToLiveSeconds.
	// This is currently best effort as we can potentially miss updates and end up in an inconsistent state.
```

internal/controller/sparkapplication/event_filter.go
RobuRishabh: Bug (edge case): `isWebhookPatchedFieldsOnlyChange` zeroes out `Suspend` here but not `TimeToLiveSeconds`.

Consider this scenario: a user updates both `timeToLiveSeconds` and a webhook-patched executor field (like tolerations) in a single patch.

What happens:
- Check 1 (line 185): "Only TTL/Suspend?" → No (tolerations also changed) → doesn't return early
- Check 2 (line 272-294): "Only webhook fields?" → Clears webhook fields + Suspend, but TTL is still different in old vs new → `DeepEqual` returns `false`
- Falls through to INVALIDATING — app gets killed/restarted unnecessarily
Fix: Add TTL zeroing alongside the Suspend zeroing:
```
// Also zero out Suspend and TimeToLiveSeconds fields as they're handled separately
oldCopy.Spec.Suspend = nil
newCopy.Spec.Suspend = nil
oldCopy.Spec.TimeToLiveSeconds = nil
newCopy.Spec.TimeToLiveSeconds = nil
```
This ensures TTL is consistently treated as "harmless" across ALL code paths in the event filter.

internal/controller/scheduledsparkapplication/event_filter_test.go
```diff
@@ -341,6 +342,39 @@ func TestEventFilter_Update_NonScheduledSparkApplication(t *testing.T) {
...
+			Template: v1beta2.SparkApplicationSpec{
+				Type:         v1beta2.SparkApplicationTypeScala,
+				SparkVersion: "3.5.0",
+			},
+		},
+	}
+	newApp := oldApp.DeepCopy()
+	newApp.Spec.Template.TimeToLiveSeconds = ptr.To[int64](60)
+
+	result := filter.Update(event.UpdateEvent{ObjectOld: oldApp, ObjectNew: newApp})
```
RobuRishabh: This calls the Update method in `scheduledsparkapplication/event_filter.go` which basically does only namespace filtering. This test would pass even if you changed SparkVersion, MainClass, or literally any field, because the filter has no spec-diff or invalidation logic. It's essentially just testing namespace matching, not TTL exemption.

Not blocking, but consider either:
- Adding a comment in the test explaining it's a basic smoke test (not invalidation coverage), or
- Removing it to avoid implying coverage that doesn't exist

RobuRishabh (changes requested):

internal/controller/scheduledsparkapplication/event_filter_test.go
dineshkumar181094: Removed this test case.

internal/controller/sparkapplication/event_filter.go
dineshkumar181094: I have updated the pr.

internal/controller/sparkapplication/event_filter.go
dineshkumar181094: Updated the coment.

RobuRishabh (approved):
LGTM

nabuskey (approved):
/approve

PR #2973, opened 2026-06-02
 Use driverIngressOptions annotations and TLS instead of sparkUIOptions

### LINKED ISSUE #2970: driverIngressOptions.ingressAnnotations are ignored when creating Ingress resources
### What happened?

## Description

  The Spark Operator ignores annotations specified in `spec.driverIngressOptions[].ingressAnnotations` when creating Ingress resources. Instead, it only reads annotations from
  `spec.sparkUIOptions.ingressAnnotations`.

  ## Impact

  Users cannot configure Ingress-specific annotations where they logically belong. This affects:
  - TLS termination configuration
  - Ingress controller-specific settings (nginx, HAProxy, etc.)
  - Custom routing policies
  - Timeout configurations

  **Example:** Users cannot set ingress class or TLS annotations in `driverIngressOptions`, making it impossible to properly configure the driver UI Ingress without using the incorrect workaround of placing these
   annotations in `sparkUIOptions`.

### 1. Create a SparkApplication with annotations in driverIngressOptions

  ```yaml
  apiVersion: sparkoperator.k8s.io/v1beta2
  kind: SparkApplication
  metadata:
    name: spark-pi-test
    namespace: test
  spec:
    type: Scala
    mode: cluster
    image: spark:3.5.7
    mainClass: org.apache.spark.examples.SparkPi
    mainApplicationFile: local:///opt/spark/examples/jars/spark-examples.jar
    driverIngressOptions:
      - servicePort: 4040
        servicePortName: "spark-driver-ui-port"
        ingressURLFormat: "{{$appName}}-{{$appNamespace}}.example.com"
        ingressAnnotations:
          kubernetes.io/ingress.class: "nginx"
          nginx.ingress.kubernetes.io/ssl-redirect: "true"
          custom.test/annotation: "test-value"
    driver:
      cores: 1
      memory: "1024m"
    executor:
      cores: 1
      instances: 1
      memory: "512m"
  ```

  ### 2. Check the created Ingress

  ```bash
  kubectl get ingress spark-pi-test-ui -o yaml
  ```

### Expected behavior

 Ingress has the annotations specified in `driverIngressOptions.ingressAnnotations`

### Actual behavior

Ingress has NO annotations (or only operator-generated ones)

### Environment & Versions

- Kubernetes Version: 1.33
- Spark Operator Version:2.4.0
- Apache Spark Version:3.5.7

### Additional context

This bug creates confusion because:
  1. The CRD schema correctly defines `ingressAnnotations` under `driverIngressOptions`
  2. The field is documented and validated by the admission webhook
  3. But the implementation completely ignores this field
  4. Users must use the workaround of putting annotations in `sparkUIOptions` (which is semantically for the Spark History Server UI, not the driver ingress)

  The disconnect between the API schema and implementation makes the operator difficult to use correctly.

### Impacted by this bug?

shruthis4:
## What this PR does

  Fixes #2970
  
  The `createDriverIngress` function was passing empty values for annotations and TLS to `createDriverIngressV1`, causing it to ignore user-specified values in `driverIngressOptions.ingressAnnotations` and
  `driverIngressOptions.ingressTLS`. 
  
  This fix passes the correct values from `driverIngressConfiguration`, allowing users to properly configure:
  - Ingress annotations (e.g., TLS termination, ingress class)
  - TLS configuration (certificates, hosts)

 ## Testing

  Tested on Kubernetes 1.33 :
  - **Annotations:** ✅ Verified `ingressAnnotations` now properly applied to Ingress/Route
  - **IngressTLS:** Code review only - not tested with actual certificates

  **Before:** Annotations in `driverIngressOptions` were ignored
  **After:** Annotations properly applied

  ## Changes

  - Modified `createDriverIngress` (line 94) to pass `driverIngressConfiguration.IngressAnnotations` and `driverIngressConfiguration.IngressTLS` instead of empty values

  Fixes #2970

### DIFF
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -386,7 +386,7 @@ func (r *Reconciler) reconcileSubmittedSparkApplication(ctx context.Context, req
 					if err != nil {
 						return fmt.Errorf("failed to get driver ingress url: %v", err)
 					}
-					_, err = r.createDriverIngress(ctx, app, &driverIngressConfiguration, *service, ingressURL, r.options.IngressClassName)
+					_, err = r.createDriverIngress(ctx, app, &driverIngressConfiguration, *service, ingressURL, r.options.IngressClassName, r.options.IngressTLS, r.options.IngressAnnotations)
 					if err != nil {
 						return fmt.Errorf("failed to create driver ingress: %v", err)
 					}
--- a/internal/controller/sparkapplication/driveringress.go
+++ b/internal/controller/sparkapplication/driveringress.go
@@ -85,27 +85,40 @@ func getDriverIngressURL(ingressURLFormat string, app *v1beta2.SparkApplication)
 	return parsedURL, nil
 }
 
-func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.SparkApplication, driverIngressConfiguration *v1beta2.DriverIngressConfiguration, service SparkService, ingressURL *url.URL, ingressClassName string) (*SparkIngress, error) {
+func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.SparkApplication, driverIngressConfiguration *v1beta2.DriverIngressConfiguration, service SparkService, ingressURL *url.URL, ingressClassName string, defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
 	if driverIngressConfiguration.ServicePort == nil {
 		return nil, fmt.Errorf("cannot create Driver Ingress for application %s/%s due to empty ServicePort on driverIngressConfiguration", app.Namespace, app.Name)
 	}
 	ingressName := fmt.Sprintf("%s-ing-%d", app.Name, *driverIngressConfiguration.ServicePort)
-	if util.IngressCapabilities.Has("networking.k8s.io/v1") {
-		return r.createDriverIngressV1(ctx, app, service, ingressName, ingressURL, ingressClassName, []networkingv1.IngressTLS{}, map[string]string{})
+
+	// Resolve annotations and TLS: per-entry values take precedence, falling back
+	// to the operator-wide defaults when the per-entry fields are empty.
+	ingressAnnotations := driverIngressConfiguration.IngressAnnotations
+	if len(ingressAnnotations) == 0 {
+		ingressAnnotations = defaultIngressAnnotations
 	}
-	return r.createDriverIngressLegacy(ctx, app, service, ingressName, ingressURL)
+	ingressTLS := driverIngressConfiguration.IngressTLS
+	if len(ingressTLS) == 0 {
+		ingressTLS = defaultIngressTLS
+	}
+
+	if util.IngressCapabilities.Has("networking.k8s.io/v1") {
+		return r.createDriverIngressV1(ctx, app, service, ingressName, ingressURL, ingressClassName, ingressTLS, ingressAnnotations)
+	}
+	return r.createDriverIngressLegacy(ctx, app, service, ingressName, ingressURL, ingressTLS, ingressAnnotations)
 }
 
-func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string, defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
+func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string, ingressTLS []networkingv1.IngressTLS, ingressAnnotations map[string]string) (*SparkIngress, error) {
 	logger := log.FromContext(ctx)
-	ingressResourceAnnotations := util.GetWebUIIngressAnnotations(app)
-	if len(ingressResourceAnnotations) == 0 && len(defaultIngressAnnotations) != 0 {
-		ingressResourceAnnotations = defaultIngressAnnotations
+
+	// Copy the annotations into a new map so that the rewrite-target mutation below
+	// never modifies the caller's source map (either the user's SparkApplication
+	// spec or the operator-wide default options).
+	ingressResourceAnnotations := make(map[string]string, len(ingressAnnotations))
+	for key, value := range ingressAnnotations {
+		ingressResourceAnnotations[key] = value
 	}
-	ingressTLSHosts := util.GetWebUIIngressTLS(app)
-	if len(ingressTLSHosts) == 0 && len(defaultIngressTLS) != 0 {
-		ingressTLSHosts = defaultIngressTLS
-	}
+	ingressTLSHosts := ingressTLS
 
 	ingressURLPath := ingressURL.Path
 	// If we're serving on a subpath, we need to ensure we create capture groups
@@ -199,12 +212,19 @@ func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.Spa
 	}, nil
 }
 
-func (r *Reconciler) createDriverIngressLegacy(ctx context.Context, app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL) (*SparkIngress, error) {
+func (r *Reconciler) createDriverIngressLegacy(ctx context.Context, app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressTLS []networkingv1.IngressTLS, ingressAnnotations map[string]string) (*SparkIngress, error) {
 	logger := log.FromContext(ctx)
-	ingressResourceAnnotations := util.GetWebUIIngressAnnotations(app)
+
+	// Copy the annotations into a new map so that the rewrite-target mutation below
+	// never modifies the caller's source map (either the user's SparkApplication
+	// spec or the operator-wide default options).
+	ingressResourceAnnotations := make(map[string]string, len(ingressAnnotations))
+	for key, value := range ingressAnnotations {
+		ingressResourceAnnotations[key] = value
+	}
 	// var ingressTLSHosts networkingv1.IngressTLS[]
 	// That we convert later for extensionsv1beta1, but return as is in SparkIngress.
-	ingressTLSHosts := util.GetWebUIIngressTLS(app)
+	ingressTLSHosts := ingressTLS
 
 	ingressURLPath := ingressURL.Path
 	// If we're serving on a subpath, we need to ensure we create capture groups.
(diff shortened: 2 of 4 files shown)


internal/controller/sparkapplication/driveringress.go
```diff
@@ -91,7 +91,7 @@ func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.Spark
 	}
 	ingressName := fmt.Sprintf("%s-ing-%d", app.Name, *driverIngressConfiguration.ServicePort)
 	if util.IngressCapabilities.Has("networking.k8s.io/v1") {
-		return r.createDriverIngressV1(ctx, app, service, ingressName, ingressURL, ingressClassName, []networkingv1.IngressTLS{}, map[string]string{})
+		return r.createDriverIngressV1(ctx, app, service, ingressName, ingressURL, ingressClassName, driverIngressConfiguration.IngressTLS, driverIngressConfiguration.IngressAnnotations)
 	}
 	return r.createDriverIngressLegacy(ctx, app, service, ingressName, ingressURL)
 }
```
tariq-hasan: The behavior we'd like to ensure is that the per-entry spec fields take precedence for driver ingress annotations and TLS config and that there is a fallback to operator flags if none are provided. The driver-ingress call site in `internal/controller/sparkapplications/controller.go` will need to pass in the operator-enabled TLS and annotations as well.

```suggestion
func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.SparkApplication, driverIngressConfiguration *v1beta2.DriverIngressConfiguration, service SparkService, ingressURL *url.URL, ingressClassName string, defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
	if driverIngressConfiguration.ServicePort == nil {
		return nil, fmt.Errorf("cannot create Driver Ingress for application %s/%s due to empty ServicePort on driverIngressConfiguration", app.Namespace, app.Name)
	}
	ingressName := fmt.Sprintf("%s-ing-%d", app.Name, *driverIngressConfiguration.ServicePort)

	// Resolve annotations and TLS: per-entry values take precedence, falling back
	// to the operator-wide defaults when the per-entry fields are empty.
	ingressAnnotations := driverIngressConfiguration.IngressAnnotations
	if len(ingressAnnotations) == 0 {
		ingressAnnotations = defaultIngressAnnotations
	}
	ingressTLS := driverIngressConfiguration.IngressTLS
	if len(ingressTLS) == 0 {
		ingressTLS = defaultIngressTLS
	}

	if util.IngressCapabilities.Has("networking.k8s.io/v1") {
		return r.createDriverIngressV1(ctx, app, service, ingressName, ingressURL, ingressClassName, ingressTLS, ingressAnnotations)
	}
	return r.createDriverIngressLegacy(ctx, app, service, ingressName, ingressURL, ingressTLS, ingressAnnotations)
}
```

internal/controller/sparkapplication/driveringress.go
tariq-hasan: Two things need to happen here. First we move the `ingressResourceAnnotations := util.GetWebUIIngressAnnotations(app)` logic to `internal/controller/sparkapplication/web_ui.go` so that the driver ingress builder becomes reusable. Secondly, we prevent mutation on the source map for the annotations by doing a copy.

A similar change needs to happen for `createDriverIngressLegacy` as well.

```suggestion
func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.SparkApplication, service SparkService, ingressName string, ingressURL *url.URL, ingressClassName string, ingressTLS []networkingv1.IngressTLS, ingressAnnotations map[string]string) (*SparkIngress, error) {
	logger := log.FromContext(ctx)

	// Copy the annotations into a new map so that the rewrite-target mutation below
	// never modifies the caller's source map (either the user's SparkApplication
	// spec or the operator-wide default options).
	ingressResourceAnnotations := make(map[string]string, len(ingressAnnotations))
	for key, value := range ingressAnnotations {
		ingressResourceAnnotations[key] = value
	}
	ingressTLSHosts := ingressTLS
```

tariq-hasan (changes requested):

internal/controller/sparkapplication/controller.go
```diff
@@ -386,7 +386,11 @@ func (r *Reconciler) reconcileSubmittedSparkApplication(ctx context.Context, req
 					if err != nil {
 						return fmt.Errorf("failed to get driver ingress url: %v", err)
 					}
-					_, err = r.createDriverIngress(ctx, app, &driverIngressConfiguration, *service, ingressURL, r.options.IngressClassName)
+					// Extract user-specified annotations and TLS from SparkApplication
+					userIngressAnnotations := util.GetWebUIIngressAnnotations(app)
+					userIngressTLS := util.GetWebUIIngressTLS(app)
+
+					_, err = r.createDriverIngress(ctx, app, &driverIngressConfiguration, *service, ingressURL, r.options.IngressClassName, userIngressTLS, userIngressAnnotations, r.options.IngressTLS, r.options.IngressAnnotations)
```
RobuRishabh: The driver ingress call site passes `util.GetWebUIIngressAnnotations(app)` (which reads from sparkUIOptions) as the user annotations. This should not be here -- `driverIngressConfiguration.IngressAnnotations` and `.IngressTLS` are never used. The precedence resolution should happen inside `createDriverIngress` using the `driverIngressConfiguration` fields, with operator-wide flags as the fallback, matching @tariq-hasan suggestion
```
ingressAnnotations := driverIngressConfiguration.IngressAnnotations
if len(ingressAnnotations) == 0 {
    ingressAnnotations = defaultIngressAnnotations
}
```
The fix might be like 
```
_, err = r.createDriverIngress(ctx, app, &driverIngressConfiguration, *service, ingressURL, r.options.IngressClassName, r.options.IngressTLS, r.options.IngressAnnotations)
```

internal/controller/sparkapplication/driveringress.go
```diff
@@ -85,28 +85,49 @@ func getDriverIngressURL(ingressURLFormat string, app *v1beta2.SparkApplication)
 	return parsedURL, nil
 }
 
-func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.SparkApplication, driverIngressConfiguration *v1beta2.DriverIngressConfiguration, service SparkService, ingressURL *url.URL, ingressClassName string) (*SparkIngress, error) {
+func (r *Reconciler) createDriverIngress(ctx context.Context, app *v1beta2.SparkApplication, driverIngressConfiguration *v1beta2.DriverIngressConfiguration, service SparkService, ingressURL *url.URL, ingressClassName string, userIngressTLS []networkingv1.IngressTLS, userIngressAnnotations map[string]string, defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
```
RobuRishabh: `createDriverIngress` function now takes 4 extra parameters. It just passes all 4 through to the V1/Legacy functions. But it never uses `driverIngressConfiguration.IngressAnnotations` or `.IngressTLS`.

@tariq-hasan suggestion was for this function to resolve the precedence itself with just 2 extra params
```
func (r *Reconciler) createDriverIngress(... defaultIngressTLS []networkingv1.IngressTLS, defaultIngressAnnotations map[string]string) (*SparkIngress, error) {
    // ...
    ingressAnnotations := driverIngressConfiguration.IngressAnnotations
    if len(ingressAnnotations) == 0 {
        ingressAnnotations = defaultIngressAnnotations
    }
    ingressTLS := driverIngressConfiguration.IngressTLS
    if len(ingressTLS) == 0 {
        ingressTLS = defaultIngressTLS
    }
    // pass resolved values down
}
```
This is cleaner (fewer params) and correct (uses the right source).

internal/controller/sparkapplication/driveringress.go
```diff
@@ -85,28 +85,49 @@ func getDriverIngressURL(ingressURLFormat string, app *v1beta2.SparkApplication)
...
-		ingressResourceAnnotations = defaultIngressAnnotations
+
+	// Resolve annotations: per-entry values take precedence, falling back
+	// to the operator-wide defaults when the per-entry fields are empty.
+	var ingressAnnotations map[string]string
+	if len(userIngressAnnotations) != 0 {
+		ingressAnnotations = userIngressAnnotations
+	} else if len(defaultIngressAnnotations) != 0 {
+		ingressAnnotations = defaultIngressAnnotations
+	}
```
RobuRishabh: Good fix, but it receives the wrong data because of the upstream bug. The function also now takes 4 params (user + defaults) and does precedence resolution inside:
```
if len(userIngressAnnotations) != 0 {
    ingressAnnotations = userIngressAnnotations
} else if len(defaultIngressAnnotations) != 0 {
    ingressAnnotations = defaultIngressAnnotations
}
```
Tariq's suggestion had this function take only 2 params (already-resolved values), with no precedence logic inside. The 4-param approach duplicates the resolution in both V1 and Legacy, which is unnecessary.

internal/controller/sparkapplication/driveringress.go
```diff
@@ -199,12 +220,39 @@ func (r *Reconciler) createDriverIngressV1(ctx context.Context, app *v1beta2.Spa
...
-	ingressResourceAnnotations := util.GetWebUIIngressAnnotations(app)
+
+	// Resolve annotations: per-entry values take precedence, falling back
+	// to the operator-wide defaults when the per-entry fields are empty.
+	var ingressAnnotations map[string]string
+	if len(userIngressAnnotations) != 0 {
+		ingressAnnotations = userIngressAnnotations
+	} else if len(defaultIngressAnnotations) != 0 {
+		ingressAnnotations = defaultIngressAnnotations
+	}
```
RobuRishabh: ditto, Same pattern applied: 4 extra params, deep copy, precedence resolution. Previously this function was untouched -- now it matches V1. Good that the legacy path is fixed, but it still receives the wrong "user" annotations.

RobuRishabh (changes requested):
@shruthis4 thanks for the contribution, I am not sure but should we add a test case with `IngressAnnotations` and `IngressTLS` set on `DriverIngressConfiguration` to verify the fix and prevent regression.

RobuRishabh (commented):
Two things to fix before this can merge: 
(1) the existing test on line 94 of `driveringress_test.go` will fail to compile because `createDriverIngress` now takes 2 extra params -- please update that call
(2) adding a test case that sets `IngressAnnotations`/`IngressTLS` on `DriverIngressConfiguration` would protect this fix from regression.

vara-bonthu (commented):
/retest

internal/controller/sparkapplication/driveringress.go
shruthis4: Done

internal/controller/sparkapplication/driveringress.go
shruthis4: Done

nabuskey (approved):
/approve

RobuRishabh (approved):
/lgtm

PR #2975, opened 2026-06-04
(chore) use `command` in entrypoint.sh

alimaredia:
## Purpose of this PR

`which` is not POSIX compliant and thus might not exist in all base images.

## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -24,5 +24,5 @@ if ! getent passwd "$myuid" &> /dev/null; then
 fi
 
 # Path for catatonit may differ depending on base container OS
-CATATONIT=$(which catatonit)
+CATATONIT=$(command -v catatonit) || { echo "error: catatonit not found in PATH" >&2; exit 1; }
 exec "$CATATONIT" -- /usr/bin/spark-operator "$@"


entrypoint.sh
```diff
@@ -24,5 +24,5 @@ if ! getent passwd "$myuid" &> /dev/null; then
 fi
 
 # Path for catatonit may differ depending on base container OS
-CATATONIT=$(which catatonit)
++CATATONIT=$(command -v catatonit) || { echo "error: catatonit not found in PATH" >&2; exit 1; }
 exec "$CATATONIT" -- /usr/bin/spark-operator "$@"
```
tariq-hasan, line 28: This is somewhat unrelated but just wondering if we want to have some Makefile target run as part of the CI to lint and format the bash scripts.

As it stands errors in the shell manifest as e2e errors and a Makefile target would help catch these beforehand. I suppose this could go as a separate PR.

RobuRishabh: Its a good suggestion, we can have it in a follow-up PR.

tariq-hasan: Sounds good. I will open a separate PR in that case once this is merged.

tariq-hasan:
`+CATATONIT=$(command -v catatonit) || { echo "error: catatonit not found in PATH" >&2; exit 1; }`

The + at the beginning needs to be removed to make it work. It's a shell error.

RobuRishabh (approved):
LGTM

nabuskey (approved):
/approve

Issue #2337, opened 2024-11-26
Enable the Spark Operator to launch applications using user-defined mechanisms beyond the default spark-submit 


c-h-afzal:
### What feature you would like to be added?

Ability for Spark Operator to adopt a provider-pattern/pluggable-mechanism to launch spark applications. A user can specify an option other than spark-submit to launch spark applications.

### Why is this needed?

In the current implementation Spark Operator invokes spark-submit to launch a spark application in the cluster. From our testing we have determined that the penalty for the JVM spin-up causes significant increase in job latency when the cluster is under stress/heavy-load, i.e. the rate of spark applications being enqueued is higher than the rate of applications being dequeued causing the spark operator’s internal queue to swell-up and affect job latencies. We want to be able to launch spark application using native Go, without the JVM spin-up as part of spark-submit.

### Describe the solution you would like

The solution we are proposing (and willing to contribute to, if consensus can be reached) is for the Spark Operator to allow changing the only mechanism (spark-submit) of launching spark applications to a user specified one. The default mechanism remains to be spark-submit. Users can specify their own plugin to launch spark applications a different way. Specifically, (here at Salesforce Big Data Org), in our fork, we create driver pods using Go and skip the JVM penalty.  The [work-around](https://github.com/kubeflow/spark-operator/compare/master...gangahiremath:spark-on-k8s-operator:master) was devised by @ gangahiremath and mentioned in the [issue#1574](https://github.com/kubeflow/spark-operator/issues/1574)

Our work-around ports the functionality of spark-submit to Golang and significantly reduces the time it takes for a SparkApplication CRD object to be CREATED and then transition to the SUBMITTED state. If there’s enough interest in our approach, we plan to open-source Ganga's work-around too.

### Describe alternatives you have considered

For improving latencies we have considered the [pending PR](https://github.com/kubeflow/spark-operator/pull/1990) which claims of performance boost by have a single queue per app. However, we have not realized the claimed performance enhancements in our testing. We still find JVM spin-up times to be the bottle-neck and hence the proposal.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

yuchaoran2011:
Thanks for the thoughtful proposal. If the Go native way of creating Spark driver is inherently faster than Spark-submit, I would argue that it makes sense to use that as the default. Do you see any limitations compared to spark-submit?

c-h-afzal:
[...]

PR #2976, opened 2026-06-04
REST-based spark submitter service as pluggable submission strategy

venkomirisetti:
## Summary

Adds a REST submitter client to the operator as an alternative to the default `spark-submit` CLI. When the `RestSubmitter` feature gate is enabled, the operator submits SparkApplications via an external REST service instead of spawning a JVM per submission.

This PR contains only the **operator-side client** — the Helm chart, CI, Dockerfile, and submitter deployment will follow in a separate PR.

### Problem
- Each `spark-submit` spawns a full JVM (~2-8s startup, ~200-400MB memory)
- Degrades under high submission rates (1000s+ SparkApplications)
- Related: #2337

### Solution
- REST client that submits to a long-running service: ~200ms latency vs 2-8s
- Fire-and-forget: returns immediately after driver pod creation
- mTLS support for secure communication with the submitter service

## Changes

| File | Description |
|------|-------------|
| `rest_submission.go` | REST client with structured error codes, retry/backoff, TLS transport, idempotent submission handling |
| `rest_submission_test.go` | Unit tests for retry, idempotency, response parsing, TLS, and connection verification |
| `start.go` | `newSparkSubmitter()` helper with feature gate switch |
| `submission.go` | `skipPodTemplates` param to omit podTemplateFile confs when using REST path |
| `submission_test.go` | Tests for skipPodTemplates |
| `features.go` | `RestSubmitter` feature gate (alpha, default off) |

## Key Design Decisions

- **Feature gate**: `RestSubmitter` (alpha, default off) — zero impact on existing users
- **Error-code-based retry**: retries on `SUBMITTER_OVERLOADED` and `INTERNAL_SERVER_ERROR` only; network/TLS errors always retried (covers cert reload)
- **Operator-side idempotency**: on `DRIVER_POD_ALREADY_EXISTS`, compares `submission_id` — match = success (retry of completed submission), mismatch = genuine conflict
- **Startup verification**: `WaitForConnection` does full HTTP+TLS request, not just TCP dial
- **No factory abstraction**: feature gate check and wiring live in `start.go`

## Test plan

- [x] Unit tests for all retry scenarios (overloaded, internal error, bad request, max retries)
- [x] Unit tests for idempotent submission (matching ID, differing ID, other 422)
- [x] Unit tests for response parsing (success, structured error, non-JSON, truncation)
- [x] Unit tests for TLS transport (valid certs, invalid certs, partial config, whitespace paths)
- [x] Unit tests for connection wait (reachable, timeout)
- [x] Unit tests for context cancellation
- [x] `go build ./cmd/operator/...` passes
- [x] Feature gate defaults to off — no behavior change without explicit opt-in

### DIFF
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -173,6 +173,10 @@ func NewStartCommand() *cobra.Command {
 				return fmt.Errorf("invalid value %q for --scheduled-spark-application-timestamp-precision, valid values: %v", scheduledSparkApplicationTimestampPrecision, validPrecisions)
 			}
 
+			if submitterMaxRetries < 1 {
+				return fmt.Errorf("invalid value %d for --submitter-max-retries, must be at least 1", submitterMaxRetries)
+			}
+
 			return nil
 		},
 		Run: func(_ *cobra.Command, args []string) {
@@ -345,7 +349,13 @@ func start() {
 		}
 	}
 
-	sparkSubmitter := newSparkSubmitter()
+	ctx := ctrl.SetupSignalHandler()
+
+	sparkSubmitter, err := newSparkSubmitter(ctx)
+	if err != nil {
+		logger.Error(err, "Failed to create spark submitter")
+		os.Exit(1)
+	}
 
 	// Setup controller for SparkApplication.
 	if err = sparkapplication.NewReconciler(
@@ -398,7 +408,7 @@ func start() {
 	}
 
 	logger.Info("Starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		logger.Error(err, "Failed to start manager")
 		os.Exit(1)
 	}
@@ -518,14 +528,9 @@ func newSparkConnectReconcilerOptions() sparkconnect.Options {
 	return options
 }
 
-func newSparkSubmitter() sparkapplication.SparkApplicationSubmitter {
+func newSparkSubmitter(ctx context.Context) (sparkapplication.SparkApplicationSubmitter, error) {
 	if !features.Enabled(features.RestSubmitter) {
-		return &sparkapplication.SparkSubmitter{}
-	}
-
-	if submitterMaxRetries < 1 {
-		logger.Info("WARNING: --submitter-max-retries is less than 1, defaulting to 1", "configured", submitterMaxRetries)
-		submitterMaxRetries = 1
+		return &sparkapplication.SparkSubmitter{}, nil
 	}
 
 	var tlsCfg *sparkapplication.TLSConfig
@@ -546,17 +551,15 @@ func newSparkSubmitter() sparkapplication.SparkApplicationSubmitter {
 		TLS:             tlsCfg,
 	})
 	if err != nil {
-		logger.Error(err, "Failed to initialize REST submitter")
-		os.Exit(1)
+		return nil, fmt.Errorf("failed to initialize RestSubmitter: %w", err)
 	}
 
-	startupCtx, cancel := context.WithTimeout(context.Background(), submitterStartupTimeout)
+	startupCtx, cancel := context.WithTimeout(ctx, submitterStartupTimeout)
 	defer cancel()
 	if err := submitter.WaitForConnection(startupCtx); err != nil {
-		logger.Error(err, "REST submitter service not reachable at startup")
-		os.Exit(1)
+		return nil, fmt.Errorf("RestSubmitter service not reachable at startup: %w", err)
 	}
 
 	logger.Info("Using REST submitter service", "url", submitterServiceURL)
-	return submitter
+	return submitter, nil
 }
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -141,10 +141,10 @@ var (
 	submitterRequestTimeout time.Duration
 	submitterMaxRetries     int
 	submitterInitialBackoff time.Duration
-	submitterTLSEnabled  bool
-	submitterTLSCertFile string
-	submitterTLSKeyFile  string
-	submitterTLSCAFile   string
+	submitterTLSEnabled     bool
+	submitterTLSCertFile    string
+	submitterTLSKeyFile     string
+	submitterTLSCAFile      string
 )
 
 func NewStartCommand() *cobra.Command {
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -22,7 +22,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"path/filepath"
 	"slices"
 	"time"
 
@@ -142,10 +141,10 @@ var (
 	submitterRequestTimeout time.Duration
 	submitterMaxRetries     int
 	submitterInitialBackoff time.Duration
-	submitterTLSCertDir     string
-	submitterTLSCertName    string
-	submitterTLSKeyName     string
-	submitterTLSCAName      string
+	submitterTLSEnabled  bool
+	submitterTLSCertFile string
+	submitterTLSKeyFile  string
+	submitterTLSCAFile   string
 )
 
 func NewStartCommand() *cobra.Command {
@@ -173,8 +172,18 @@ func NewStartCommand() *cobra.Command {
 				return fmt.Errorf("invalid value %q for --scheduled-spark-application-timestamp-precision, valid values: %v", scheduledSparkApplicationTimestampPrecision, validPrecisions)
 			}
 
-			if submitterMaxRetries < 1 {
-				return fmt.Errorf("invalid value %d for --submitter-max-retries, must be at least 1", submitterMaxRetries)
+			if features.Enabled(features.RestSubmitter) {
+				if submitterServiceURL == "" {
+					return fmt.Errorf("--submitter-service-url is required when RestSubmitter feature gate is enabled")
+				}
+				if submitterMaxRetries < 1 {
+					return fmt.Errorf("invalid value %d for --submitter-max-retries, must be at least 1", submitterMaxRetries)
+				}
+				if submitterTLSEnabled {
+					if submitterTLSCertFile == "" || submitterTLSKeyFile == "" || submitterTLSCAFile == "" {
+						return fmt.Errorf("--submitter-tls-enabled requires --submitter-tls-cert-file, --submitter-tls-key-file, and --submitter-tls-ca-file")
+					}
+				}
 			}
 
 			return nil
@@ -249,10 +258,10 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().DurationVar(&submitterRequestTimeout, "submitter-request-timeout", 2*time.Minute, "HTTP request timeout per spark submission attempt.")
 	command.Flags().IntVar(&submitterMaxRetries, "submitter-max-retries", 3, "Max retry attempts for transient spark submission failures.")
 	command.Flags().DurationVar(&submitterInitialBackoff, "submitter-initial-backoff", 2*time.Second, "Initial backoff duration before the first retry.")
-	command.Flags().StringVar(&submitterTLSCertDir, "submitter-tls-cert-dir", "", "Directory containing TLS files for the submitter. When set, TLS is enabled.")
-	command.Flags().StringVar(&submitterTLSCertName, "submitter-tls-cert-name", "tls.crt", "File name of the client certificate in the TLS cert directory.")
-	command.Flags().StringVar(&submitterTLSKeyName, "submitter-tls-key-name", "tls.key", "File name of the client private key in the TLS cert directory.")
-	command.Flags().StringVar(&submitterTLSCAName, "submitter-tls-ca-name", "ca.crt", "File name of the CA certificate in the TLS cert directory.")
+	command.Flags().BoolVar(&submitterTLSEnabled, "submitter-tls-enabled", false, "Enable mTLS for communication with the submitter service.")
+	command.Flags().StringVar(&submitterTLSCertFile, "submitter-tls-cert-file", "", "Path to the client certificate file for mTLS with the submitter service.")
+	command.Flags().StringVar(&submitterTLSKeyFile, "submitter-tls-key-file", "", "Path to the client private key file for mTLS with the submitter service.")
+	command.Flags().StringVar(&submitterTLSCAFile, "submitter-tls-ca-file", "", "Path to the CA certificate file for verifying the submitter service.")
 
 	flagSet := flag.NewFlagSet("controller", flag.ExitOnError)
 	ctrl.RegisterFlags(flagSet)
@@ -534,12 +543,12 @@ func newSparkSubmitter(ctx context.Context) (sparkapplication.SparkApplicationSu
 	}
 
 	var tlsCfg *sparkapplication.TLSConfig
-	if submitterTLSCertDir != "" {
+	if submitterTLSEnabled {
 		tlsCfg = &sparkapplication.TLSConfig{
 			Enabled:    true,
-			CertFile:   filepath.Join(submitterTLSCertDir, submitterTLSCertName),
-			KeyFile:    filepath.Join(submitterTLSCertDir, submitterTLSKeyName),
-			CACertFile: filepath.Join(submitterTLSCertDir, submitterTLSCAName),
+			CertFile:   submitterTLSCertFile,
+			KeyFile:    submitterTLSKeyFile,
+			CACertFile: submitterTLSCAFile,
 		}
 	}
 
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -17,10 +17,12 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"slices"
 	"time"
 
@@ -29,7 +31,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	// Import features package to register feature gates.
-	_ "github.com/kubeflow/spark-operator/v2/pkg/features"
+	"github.com/kubeflow/spark-operator/v2/pkg/features"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/rest"
 
@@ -133,6 +135,17 @@ var (
 	scheduledSparkApplicationTimestampPrecision string
 	development                                 bool
 	zapOptions                                  = logzap.Options{}
+
+	// REST submitter (behind RestSubmitter feature gate)
+	submitterServiceURL     string
+	submitterStartupTimeout time.Duration
+	submitterRequestTimeout time.Duration
+	submitterMaxRetries     int
+	submitterInitialBackoff time.Duration
+	submitterTLSCertDir     string
+	submitterTLSCertName    string
+	submitterTLSKeyName     string
+	submitterTLSCAName      string
 )
 
 func NewStartCommand() *cobra.Command {
@@ -225,6 +238,17 @@ func NewStartCommand() *cobra.Command {
 		"If not set, it will be 0 in order to disable the pprof server")
 
 	command.Flags().StringVar(&scheduledSparkApplicationTimestampPrecision, "scheduled-spark-application-timestamp-precision", "nanos", "Timestamp precision for ScheduledSparkApplication run names. Valid values: nanos, micros, millis, seconds, minutes.")
+
+	// REST submitter flags (used when RestSubmitter feature gate is enabled)
+	command.Flags().StringVar(&submitterServiceURL, "submitter-service-url", "", "Full submit endpoint URL of the submitter service (required when RestSubmitter feature gate is enabled).")
+	command.Flags().DurationVar(&submitterStartupTimeout, "submitter-startup-timeout", 5*time.Minute, "How long the controller waits for the submitter service to become reachable at startup.")
+	command.Flags().DurationVar(&submitterRequestTimeout, "submitter-request-timeout", 2*time.Minute, "HTTP request timeout per spark submission attempt.")
+	command.Flags().IntVar(&submitterMaxRetries, "submitter-max-retries", 3, "Max retry attempts for transient spark submission failures.")
+	command.Flags().DurationVar(&submitterInitialBackoff, "submitter-initial-backoff", 2*time.Second, "Initial backoff duration before the first retry.")
+	command.Flags().StringVar(&submitterTLSCertDir, "submitter-tls-cert-dir", "", "Directory containing TLS files for the submitter. When set, TLS is enabled.")
+	command.Flags().StringVar(&submitterTLSCertName, "submitter-tls-cert-name", "tls.crt", "File name of the client certificate in the TLS cert directory.")
+	command.Flags().StringVar(&submitterTLSKeyName, "submitter-tls-key-name", "tls.key", "File name of the client private key in the TLS cert directory.")
+	command.Flags().StringVar(&submitterTLSCAName, "submitter-tls-ca-name", "ca.crt", "File name of the CA certificate in the TLS cert directory.")
 
 	flagSet := flag.NewFlagSet("controller", flag.ExitOnError)
 	ctrl.RegisterFlags(flagSet)
@@ -321,7 +345,7 @@ func start() {
 		}
 	}
 
-	sparkSubmitter := &sparkapplication.SparkSubmitter{}
+	sparkSubmitter := newSparkSubmitter()
 
 	// Setup controller for SparkApplication.
 	if err = sparkapplication.NewReconciler(
@@ -492,4 +516,47 @@ func newSparkConnectReconcilerOptions() sparkconnect.Options {
 		NamespaceSelector: namespaceSelector,
 	}
 	return options
+}
+
+func newSparkSubmitter() sparkapplication.SparkApplicationSubmitter {
+	if !features.Enabled(features.RestSubmitter) {
+		return &sparkapplication.SparkSubmitter{}
+	}
+
+	if submitterMaxRetries < 1 {
+		logger.Info("WARNING: --submitter-max-retries is less than 1, defaulting to 1", "configured", submitterMaxRetries)
+		submitterMaxRetries = 1
+	}
+
+	var tlsCfg *sparkapplication.TLSConfig
+	if submitterTLSCertDir != "" {
+		tlsCfg = &sparkapplication.TLSConfig{
+			Enabled:    true,
+			CertFile:   filepath.Join(submitterTLSCertDir, submitterTLSCertName),
+			KeyFile:    filepath.Join(submitterTLSCertDir, submitterTLSKeyName),
+			CACertFile: filepath.Join(submitterTLSCertDir, submitterTLSCAName),
+		}
+	}
+
+	submitter, err := sparkapplication.NewRestSparkSubmitter(sparkapplication.RestSparkSubmitterConfig{
+		URL:             submitterServiceURL,
+		RetryMaxRetries: submitterMaxRetries,
+		RequestTimeout:  submitterRequestTimeout,
+		InitialBackoff:  submitterInitialBackoff,
+		TLS:             tlsCfg,
+	})
+	if err != nil {
+		logger.Error(err, "Failed to initialize REST submitter")
+		os.Exit(1)
+	}
+
+	startupCtx, cancel := context.WithTimeout(context.Background(), submitterStartupTimeout)
+	defer cancel()
+	if err := submitter.WaitForConnection(startupCtx); err != nil {
+		logger.Error(err, "REST submitter service not reachable at startup")
+		os.Exit(1)
+	}
+
+	logger.Info("Using REST submitter service", "url", submitterServiceURL)
+	return submitter
 }
--- a/internal/controller/sparkapplication/rest_submission.go
+++ b/internal/controller/sparkapplication/rest_submission.go
@@ -317,8 +317,8 @@ func newSubmitRequest(args []string, app *v1beta2.SparkApplication) *submitReque
 	return &submitRequest{
 		SubmissionID:        app.Status.SubmissionID,
 		SparkSubmitArgs:     args,
-		DriverPodTemplate:   app.Spec.Driver.Template,
-		ExecutorPodTemplate: app.Spec.Executor.Template,
+		DriverPodTemplate:   buildDriverPodTemplate(app),
+		ExecutorPodTemplate: buildExecutorPodTemplate(app),
 	}
 }
 
@@ -378,7 +378,7 @@ func (c *RestSparkSubmitter) canRetry(attempt int, err error) bool {
 	if errors.As(err, &submitErr) {
 		return submitErr.IsRetryable()
 	}
-	// Network and TLS errors (connection refused, timeout, cert reload) are transient and retryable.
+	// Network and TLS errors (connection refused, timeout) are transient and retryable.
 	return true
 }
 
--- a/internal/controller/sparkapplication/rest_submission.go
+++ b/internal/controller/sparkapplication/rest_submission.go
@@ -1,0 +1,486 @@
+/*
+Copyright 2024 The Kubeflow authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// REST Submitter Client
+//
+// Submits SparkApplications to the k8s-spark-submitter service (POST /api/v1/spark-submit).
+//
+// # Contract
+//
+//  1. Submission-only, stateless. The submitter creates the driver pod and returns its identity.
+//     It does not track, monitor, restart, or delete applications. All lifecycle and status
+//     ownership remains with the operator via Kubernetes owner references.
+//  2. The request carries the same spark-submit argument vector the operator already produces
+//     via buildSparkSubmitArgs. There is no parallel spec schema.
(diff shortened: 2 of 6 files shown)


internal/controller/sparkapplication/submission.go
```diff
@@ -127,6 +127,10 @@ func buildSparkSubmitArgs(app *v1beta2.SparkApplication) ([]string, error) {
 		applicationOption,
 	}
 
+	if !skipPodTemplates {
+		optionFuncs = append(optionFuncs, driverPodTemplateOption, executorPodTemplateOption)
+	}
```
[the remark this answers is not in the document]
venkomirisetti: Addressed.

Dockerfile
```diff
@@ -51,6 +51,7 @@ RUN mkdir -p /etc/k8s-webhook-server/serving-certs /home/spark && \
     chmod -R g+rw /etc/k8s-webhook-server/serving-certs && \
     chown -R spark /etc/k8s-webhook-server/serving-certs /home/spark
 
+
```
nabuskey: let's not touch this file

Makefile
```diff
@@ -40,6 +40,9 @@ IMAGE_REPOSITORY ?= kubeflow/spark-operator/controller
 IMAGE_TAG ?= $(VERSION)
 IMAGE ?= $(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY):$(IMAGE_TAG)
 
+# Submitter service image (separate from operator image)
+SUBMITTER_IMAGE ?= docker.io/venkomirisetti/k8s-spark-submitter:4.0.1-7
```
nabuskey: Let's remove this.

Makefile
nabuskey, line 1: I don't think we need to test this in our repo. It should be tested in the implementation repository

test/e2e/sparkapplication_test.go
```diff
@@ -276,7 +276,6 @@ var _ = Describe("Example SparkApplication", func() {
 			By("Should eventually fail")
 			finalApp := apps[len(apps)-1]
 			Expect(finalApp.Status.AppState.State).To(Equal(v1beta2.ApplicationStateFailed))
-			Expect(finalApp.Status.AppState.ErrorMessage).To(ContainSubstring("failed to run spark-submit"))
```
nabuskey, line 279: why was this removed?

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+	submitURL      string
+	hostAddr       string
+	maxRetries     int
+	initialBackoff time.Duration
+}
+
+var _ SparkApplicationSubmitter = &RestSparkSubmitter{}
+
+// submitRequest is the submit API request payload.
+type submitRequest struct {
```
nabuskey, line 204: Let's add submission_id

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+	SparkSubmitArgs     []string                `json:"spark_submit_args"`
+	DriverPodTemplate   *corev1.PodTemplateSpec `json:"driver_pod_template,omitempty"`
+	ExecutorPodTemplate *corev1.PodTemplateSpec `json:"executor_pod_template,omitempty"`
+}
+
+// submitResponse is the submit API success response.
+type submitResponse struct {
```
nabuskey, line 211: let's add submission_id

cmd/operator/controller/start.go
```diff
@@ -216,6 +219,9 @@ func NewStartCommand() *cobra.Command {
 
 	command.Flags().StringVar(&scheduledSparkApplicationTimestampPrecision, "scheduled-spark-application-timestamp-precision", "nanos", "Timestamp precision for ScheduledSparkApplication run names. Valid values: nanos, micros, millis, seconds, minutes.")
 
+	command.Flags().StringVar(&submissionStrategy, "submission-strategy", sparkapplication.SubmissionStrategySparkSubmit, "Submission strategy: 'spark-submit' (default) or 'rest-submitter'.")
```
nabuskey: we need to use `pkg/features/features.go`

internal/controller/sparkapplication/submission.go
```diff
@@ -109,24 +111,29 @@ func buildSparkSubmitArgs(app *v1beta2.SparkApplication) ([]string, error) {
...
 		dynamicAllocationOption,
-		proxyUserOption,
-		mainApplicationFileOption,
-		applicationOption,
 	}
 
+	// Pod template options write files and add --conf args; skipped when the submission
+	// strategy sends templates inline (e.g., REST submitter).
+	if !skipPodTemplates {
+		optionFuncs = append(optionFuncs, driverPodTemplateOption, executorPodTemplateOption)
```
nabuskey, line 130: This makes rendering of pod templates a responsibility of the submitter. I would rather make this the responsibility of the operator.

When a pod template is sent inline, the operator omits the *.podTemplateFile --conf entries from spark_submit_args. All other driver/executor --conf values remain in the args.

internal/controller/sparkapplication/submitter_factory.go
nabuskey, line 1: Let's avoid using the word `factory`.

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+type RestSparkSubmitterConfig struct {
+	URL             string
+	RetryMaxRetries int
+	RequestTimeout  time.Duration
+	InitialBackoff  time.Duration
+	TLS             *TLSConfig
+}
+
+// RestSparkSubmitter submits a SparkApplication via the REST submitter service.
+type RestSparkSubmitter struct {
```
nabuskey, line 192: Let's document what it is:

1. Submission-only, stateless. The submitter creates the driver pod and returns its identity. It does not track, monitor, restart, or delete applications. All lifecycle and status ownership remains with the operator via Kubernetes owner references.
2.  The request carries the same spark-submit argument vector the operator already produces via buildSparkSubmitArgs. There is no parallel spec schema.
3. Every submission carries a client-supplied submission_id. The submitter MUST treat repeat submissions of the same submission_id as idempotent.
4. When a pod template is sent inline, the operator omits the *.podTemplateFile --conf entries from spark_submit_args. All other driver/executor --conf values remain in the args.
5. The submitter MUST merge the inline template with the args exactly as Spark would when loading a podTemplateFile, i.e. the template is the base and --conf values are layered on per Spark's precedence rules.


In addition:


`submission_id` is the idempotency key. Submission is **at-least-once** from the controller's
perspective, so the submitter MUST make it **effectively exactly-once**:

1. On `POST`, the submitter checks whether a driver pod already exists for this
   `submission_id` (e.g. via a label `sparkoperator.k8s.io/submission-id=<id>` it stamps on
   creation).
2. **No existing pod** → create it, return `201` with `idempotent_replay: false`.
3. **Existing pod created for the same `submission_id`** → do **not** create a second pod;
   return the existing pod's identity with `idempotent_replay: true` (status `200` or `201`,
   chosen consistently).
4. **Name collision with a pod NOT owned by this `submission_id`** → return `409`
   `POD_ALREADY_EXISTS` (a genuine conflict).

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+	httpClient     *http.Client
+	submitURL      string
+	hostAddr       string
+	maxRetries     int
+	initialBackoff time.Duration
+}
+
+var _ SparkApplicationSubmitter = &RestSparkSubmitter{}
+
+// submitRequest is the submit API request payload.
```
nabuskey: Document

```
Copy to clipboard{
  "submission_id": "spark-pi-1718553600000000000",
  "spark_submit_args": ["--master", "k8s://...", "--deploy-mode", "cluster", "..."],
  "driver_pod_template": { "metadata": {}, "spec": {} },
  "executor_pod_template": { "metadata": {}, "spec": {} }
}
```

| Field                  | Type              | Required | Notes |
| ---------------------- | ----------------- | -------- | ----- |
| `submission_id`        | string            | **yes**  | Primary correlation key **and idempotency key** (§7). Sourced from `SparkApplication.Status.SubmissionID`. |
| `spark_submit_args`    | array of string   | **yes**  | Verbatim `spark-submit` args. The submitter parses them with Spark's own argument resolution. |
| `driver_pod_template`  | object (PodTemplateSpec) | no | Sent inline instead of `spark.kubernetes.driver.podTemplateFile`. Omitted when the user supplied no template; the submitter then synthesizes the default driver container. |
| `executor_pod_template`| object (PodTemplateSpec) | no | As above, for executors. |

internal/controller/sparkapplication/rest_submission.go
nabuskey, line 211: `201 Created`
`Content-Type: application/json`

```json
{
  "submission_id": "spark-pi-1718553600000000000",
  "spark_app_id": "spark-b992db7da52c42298736dcbb3c9142be",
  "driver_pod_name": "spark-pi-abc123-driver",
  "driver_pod_uid": "0c1f...",
  "idempotent_replay": false,
  "submitted_at": "2026-06-16T08:30:00Z"
}
```

| Field               | Type    | Notes |
| ------------------- | ------- | ----- |
| `submission_id`     | string  | Echoes the request value. |
| `spark_app_id`      | string  | Spark application id assigned to the submission. |
| `driver_pod_name`   | string  | Name of the created driver pod. |
| `driver_pod_uid`    | string  | UID of the created driver pod; lets the operator bind to the exact object. |
| `idempotent_replay` | bool    | `true` when this response describes a pod that already existed for this `submission_id` (§7). |

**Semantics of `201`:** the driver pod **object was persisted** to the API server. It does
**not** assert the driver was scheduled, started, or succeeded. The operator observes
subsequent state by watching the driver pod.

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+	DriverPodName string `json:"driver_pod_name"`
+	DriverPodUID  string `json:"driver_pod_uid"`
+	Namespace     string `json:"namespace"`
+}
+
+// submitErrorResponse is the submitter error payload.
+type submitErrorResponse struct {
+	Timestamp string `json:"timestamp"`
+	Status    int    `json:"status"`
+	Error     string `json:"error"`
```
nabuskey: This has to be enum.

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+	Message       string `json:"message"`
+	SubmittedAt   string `json:"submitted_at"`
+	SparkAppID    string `json:"spark_app_id"`
+	DriverPodName string `json:"driver_pod_name"`
+	DriverPodUID  string `json:"driver_pod_uid"`
+	Namespace     string `json:"namespace"`
+}
+
+// submitErrorResponse is the submitter error payload.
+type submitErrorResponse struct {
```
nabuskey, line 223:
```json
{
  "submission_id": "spark-pi-1718553600000000000",
  "status": 422,
  "error_code": "POD_TEMPLATE_INVALID",
  "message": "Driver template missing 'spark-kubernetes-driver' container",
  "details": "optional sanitized detail / truncated stderr",
  "timestamp": "2026-06-16T08:30:00Z"
}
```

| Field           | Type   | Notes |
| --------------- | ------ | ----- |
| `submission_id` | string | Echoes the request value (may be empty if the body was unparseable). |
| `status`        | int    | Mirrors the HTTP status. |
| `error_code`    | string | Stable enum. **This drives the controller's retry decision**, not the HTTP status. |
| `message`       | string | returned by spark-submit. |
| `timestamp`     | string | RFC 3339. |



| `error_code`             | HTTP | Retry?      | Meaning |
| ------------------------ | ---- | ----------- | ------- |
| `INVALID_REQUEST`        | 400  | No          | Malformed JSON or missing required fields. |
| `ARGS_INVALID`           | 400  | No          | `spark_submit_args` failed Spark argument parsing. |
| `POD_TEMPLATE_INVALID`   | 422  | No          | Template missing required container / structurally invalid. |
| `LOCAL_FILE_UNSUPPORTED` | 422  | No          | Submission references local files unsupported in v1. |
| `POD_ALREADY_EXISTS`     | 409  | No (treat as success-ish) | A different object already owns this name. |
| `UNAUTHORIZED`           | 401/403 | No       | Caller not authorized (§8). |
| `SUBMITTER_OVERLOADED`   | 429/503 | **Yes**  | Submitter at capacity; back off and retry. |

.github/workflows/integration.yaml
nabuskey, line 1: Updating workflows is out of scope for this PR. We need to get contract first.

charts/spark-operator-chart/ci/ci-values-submitter-tls.yaml
nabuskey, line 1: chart updates are also out of scope

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,349 @@
...
+
+	transport.TLSClientConfig = tlsConfig
+	return transport, nil
+}
+
+// WaitForConnection blocks until the submitter service is reachable or the context expires.
+func (c *RestSparkSubmitter) WaitForConnection(ctx context.Context) error {
+	restLogger.Info("Waiting for submitter service to be ready", "addr", c.hostAddr)
+	for {
+		conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", c.hostAddr)
```
nabuskey: TPC does not grantee service readiness. So this needs to be more robust. Need to check http and tls otherwise any http or cert issues will not be surfaced until the first app is submitted.

internal/controller/sparkapplication/submitter_factory.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Retry with exponential backoff
+	cmd.Flags().IntVar(&submitterMaxRetries, "submitter-max-retries", 3, "Max retry attempts for transient spark submission failures.")
+	cmd.Flags().DurationVar(&submitterInitialBackoff, "submitter-initial-backoff", 2*time.Second, "Initial backoff duration before the first retry. Doubles on each subsequent attempt.")
+
+	// TLS
+	cmd.Flags().BoolVar(&submitterTLSEnabled, "submitter-tls-enabled", false, "Enable mTLS for submitter connections.")
+	cmd.Flags().StringVar(&submitterTLSCertFile, "submitter-tls-cert-file", "", "Client certificate PEM file path for mTLS to the submitter service.")
+	cmd.Flags().StringVar(&submitterTLSKeyFile, "submitter-tls-key-file", "", "Client private key PEM file path for mTLS to the submitter service.")
+	cmd.Flags().StringVar(&submitterTLSCACertFile, "submitter-tls-ca-cert-file", "", "CA certificate PEM file path to verify the submitter server.")
```
nabuskey: We should not specify single files here. Use directories instead. We use dir for other tls related configs.

internal/controller/sparkapplication/submitter_factory.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// TLS
+	submitterTLSEnabled    bool
+	submitterTLSCertFile   string
+	submitterTLSKeyFile    string
+	submitterTLSCACertFile string
+)
+
+// RegisterSubmitterFlags registers CLI flags for all submission strategies.
+func RegisterSubmitterFlags(cmd *cobra.Command) {
```
nabuskey: This needs to be with other flags. I do not want to introduce another place to look for flags.

nabuskey (commented):
I think this PR does too much. This PR need to work on abstraction and interface only, then gate it behind feature flag.
We also need to agree on the contract. I put my thoughts in comments.

test/e2e/sparkapplication_test.go
venkomirisetti: Removed because it's a brittle string match specific to `spark-submit`. With submission now pluggable, different submitters produce different error messages. The state and retry-count assertions still validate the actual behavior.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Could you clarify how submission_id in the submitter request/response would be useful here? The submission ID is already propagated to driver and executor pods via spark conf as spark.kubernetes.driver.label.spark-app-selector — Spark applies this label natively regardless of the submitter.

[Driver label](https://github.com/kubeflow/spark-operator/blob/f043c2a/internal/controller/sparkapplication/submission.go#L347-L348)
[Executor label](https://github.com/kubeflow/spark-operator/blob/f043c2a/internal/controller/sparkapplication/submission.go#L720-L721)

Are you looking for it for logging/debugging purposes on the submitter service side, or is there another use case in mind?

Makefile
venkomirisetti: Sure, is there any specific repo for implementation?

charts/spark-operator-chart/ci/ci-values-submitter-tls.yaml
venkomirisetti: Do you mean all helm changes?

Makefile
nabuskey: Not really at this point since it's an experimental feature. In this repo, we validate the controller side logic only.

internal/controller/sparkapplication/rest_submission.go
nabuskey: Mostly around idempotency and a clear contract between the controller and submitter, it's a good idea to have something that both sides agrees on explicitly instead of having it buried inside of a string that is meant to be used for something else.

charts/spark-operator-chart/ci/ci-values-submitter-tls.yaml
nabuskey: Yes, I like to have small changes in PRs. We can do helm changes in a future PR but I'd like this PR to focus on the contract and code changes only.

internal/controller/sparkapplication/submission.go
venkomirisetti: **spark-submit path (local):**
- `buildSparkSubmitArgs(app, skipPodTemplates=false)` → operator renders templates to local disk, adds `--conf spark.kubernetes.*.podTemplateFile=...` to args
- Operator invokes spark-submit subprocess with all args

**REST submitter path:**
- `buildSparkSubmitArgs(app, skipPodTemplates=true)` → operator renders `PodTemplateSpec` but does NOT write to disk or add `podTemplateFile` conf to args
- Operator sends rendered templates inline in JSON payload
- Submitter writes inline content to disk and builds the spark-submit args with its local file paths for submission to Spark's K8s client library directly (no spark-submit subprocess)

The operator owns rendering in both paths. The only difference is who writes to disk — necessarily the submitter in the REST path since the file must exist on the submitter's filesystem.

venkomirisetti (commented):
Submitting pending review

Dockerfile
venkomirisetti: Done — removed from this PR. Will follow in a separate Helm/deployment PR.

Makefile
venkomirisetti: Done — removed from this PR.

Makefile
venkomirisetti: Agreed. Removed Makefile changes. Submitter tests live in the implementation repo.

test/e2e/sparkapplication_test.go
venkomirisetti: Reverted — the e2e test line is restored.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — submission_id added to request payload, sourced from app.Status.SubmissionID (UUID). Used as the idempotency key on the operator side.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — submission_id added to both success and error response structs.

cmd/operator/controller/start.go
venkomirisetti: Done — using `RestSubmitter` feature gate in `pkg/features/features.go`. The gate check is in `start.go` via `newSparkSubmitter()` helper.

internal/controller/sparkapplication/submission.go
venkomirisetti: Addressed — pod templates are sent inline as JSON in the request. The submitter writes them to temp files and merges with spark confs per Spark's precedence rules. Operator omits podTemplateFile --conf entries when using REST path.

internal/controller/sparkapplication/submitter_factory.go
venkomirisetti: Done — removed `submitter_factory.go` entirely. Feature gate check and wiring now live directly in `start.go`.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — request payload documented in the file-level doc with field table.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — success response documented with field table. Note: `idempotent_replay` is not included as it's not implementable on the submitter side (stateless, uses Spark libs directly).

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — error_code is now a typed enum (`SubmitErrorCode`) with constants for each code.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — error response documented with field table and error_code enum table showing HTTP status, retryability, and meaning for each code.

.github/workflows/integration.yaml
venkomirisetti: Agreed — removed from this PR. Will add in a follow-up with Helm/deployment changes.

charts/spark-operator-chart/ci/ci-values-submitter-tls.yaml
venkomirisetti: Done — all chart changes removed. Will follow in a separate PR.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — `WaitForConnection` now does a full HTTP GET through the TLS-configured client (not TCP dial). This verifies network, HTTP, and TLS cert chain on startup.

internal/controller/sparkapplication/submitter_factory.go
venkomirisetti: Done — switched to directory-based TLS config (`--submitter-tls-cert-dir`). Individual file paths are constructed from the dir + standard names (tls.crt, tls.key, ca.crt).

internal/controller/sparkapplication/submitter_factory.go
venkomirisetti: Done — submitter flags are now in `start.go` alongside all other controller flags.

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Done — added file-level documentation covering the contract (4 points), idempotency semantics, request/response payloads with field tables, and the full error code enum table.

Key differences from what you suggested:
- The submitter is stateless — it does NOT check for existing pods or track prior submissions. It simply uses Spark libs to create the pod and returns `DRIVER_POD_ALREADY_EXISTS` if K8s rejects the create.
- Idempotency is operator-side: the client compares the returned `submission_id` with its own to decide success vs conflict.
- No `idempotent_replay` field — not feasible given stateless design.

venkomirisetti:
@nabuskey Thanks for the thorough review! I've addressed all your feedback.

**Scope update:** This PR is now reduced to the operator-side REST client only (6 files). Helm charts, CI workflows, Dockerfile, and Makefile changes are removed and will follow in separate PRs once the client is merged.

**Changes in this revision:**
- Documented the full contract, request/response payloads, error code enum table, and idempotency semantics in the file-level doc comment
- Error codes are now a typed enum (`SubmitErrorCode`)
- `WaitForConnection` does a full HTTP+TLS request (not TCP dial)
- TLS uses directory-based config (`--submitter-tls-cert-dir`)
- All submitter flags moved to `start.go` alongside other controller flags
- Feature gate check moved to `start.go` with `newSparkSubmitter()` helper (no factory)
- Fixed `gofmt` and added DCO sign-off

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: **Update:** The submitter now supports server-side idempotency. It labels the driver pod with `submission_id` and on conflict returns 200 with `duplicate_submission=true` and existing pod details if same `submission_id`, or 409 `DRIVER_POD_ALREADY_EXISTS` for genuine conflicts. The operator simply treats both 201 and 200 as success — no client-side comparison needed.

cmd/operator/controller/start.go
```diff
@@ -493,3 +517,46 @@ func newSparkConnectReconcilerOptions() sparkconnect.Options {
 	}
 	return options
 }
+
+func newSparkSubmitter() sparkapplication.SparkApplicationSubmitter {
```
nabuskey:
```suggestion
func newSparkSubmitter(ctx context.Context) (sparkapplication.SparkApplicationSubmitter, error) {
```

1. Use a passed in context from `ctrl.SetupSignalHandler()`. This way, we don't have a dangling context.
2. Return an error instead of exiting inside the function. ox.exit does not run deferred functions. Current convention in this file is to os.exit in start()

cmd/operator/controller/start.go
nabuskey, line 278:
```suggestion

ctx := ctrl.SetupSignalHandler()
```

Set the context here

cmd/operator/controller/start.go
```diff
@@ -493,3 +517,46 @@ func newSparkConnectReconcilerOptions() sparkconnect.Options {
...
+		RequestTimeout:  submitterRequestTimeout,
+		InitialBackoff:  submitterInitialBackoff,
+		TLS:             tlsCfg,
+	})
+	if err != nil {
+		logger.Error(err, "Failed to initialize REST submitter")
+		os.Exit(1)
+	}
+
+	startupCtx, cancel := context.WithTimeout(context.Background(), submitterStartupTimeout)
```
nabuskey:
```suggestion
	startupCtx, cancel := context.WithTimeout(ctx, submitterStartupTimeout)
```

cmd/operator/controller/start.go
nabuskey:
```suggestion
	if err := mgr.Start(ctx); err != nil {
```
functionally the same. We just move it above.

cmd/operator/controller/start.go
```diff
@@ -321,7 +345,7 @@ func start() {
 		}
 	}
 
-	sparkSubmitter := &sparkapplication.SparkSubmitter{}
+	sparkSubmitter := newSparkSubmitter()
```
nabuskey:
```suggestion
	err = sparkSubmitter := newSparkSubmitter(ctx)
	if err != nil {
		logger.Error(err, "failed to create spark submitter")
	}
```
pass context, then handle error

internal/controller/sparkapplication/submission.go
nabuskey: Sorry I was not clear here.

`driverPodTemplateOption` and `executorPodTemplateOption` set the special non-controller owner ref specifically for Kueue.  So pod templates must be rendered by the operator first, then passed to the submitter in the REST payload to preserve exact current behaviour, instead of trusting the client to render it correctly.

cmd/operator/controller/start.go
```diff
@@ -493,3 +517,46 @@ func newSparkConnectReconcilerOptions() sparkconnect.Options {
...
+
+func newSparkSubmitter() sparkapplication.SparkApplicationSubmitter {
+	if !features.Enabled(features.RestSubmitter) {
+		return &sparkapplication.SparkSubmitter{}
+	}
+
+	if submitterMaxRetries < 1 {
+		logger.Info("WARNING: --submitter-max-retries is less than 1, defaulting to 1", "configured", submitterMaxRetries)
+		submitterMaxRetries = 1
+	}
```
nabuskey: This is unexpected by end users. We should fail and surface the info to the user instead of inferring what they want.

cmd/operator/controller/start.go
```diff
@@ -226,6 +239,17 @@ func NewStartCommand() *cobra.Command {
...
+	command.Flags().DurationVar(&submitterRequestTimeout, "submitter-request-timeout", 2*time.Minute, "HTTP request timeout per spark submission attempt.")
+	command.Flags().IntVar(&submitterMaxRetries, "submitter-max-retries", 3, "Max retry attempts for transient spark submission failures.")
+	command.Flags().DurationVar(&submitterInitialBackoff, "submitter-initial-backoff", 2*time.Second, "Initial backoff duration before the first retry.")
+	command.Flags().StringVar(&submitterTLSCertDir, "submitter-tls-cert-dir", "", "Directory containing TLS files for the submitter. When set, TLS is enabled.")
+	command.Flags().StringVar(&submitterTLSCertName, "submitter-tls-cert-name", "tls.crt", "File name of the client certificate in the TLS cert directory.")
+	command.Flags().StringVar(&submitterTLSKeyName, "submitter-tls-key-name", "tls.key", "File name of the client private key in the TLS cert directory.")
+	command.Flags().StringVar(&submitterTLSCAName, "submitter-tls-ca-name", "ca.crt", "File name of the CA certificate in the TLS cert directory.")
+
```
nabuskey: Do you have real need for these flags? 
My point about using dir instead was that assuming k8s secret type of `kubernetes.io/tls`, k8s api validates the tls.key and the tls.crt keys exist in data. Cert manager also does operates this way with additional ca.crt (same as SA CA bundle).
Not having these flags help reduce the number of flags introduced too.

internal/controller/sparkapplication/rest_submission.go
```diff
@@ -0,0 +1,479 @@
...
+
+func (c *RestSparkSubmitter) canRetry(attempt int, err error) bool {
+	if attempt >= c.maxRetries-1 {
+		return false
+	}
+	var submitErr *SubmitError
+	if errors.As(err, &submitErr) {
+		return submitErr.IsRetryable()
+	}
+	// Network and TLS errors (connection refused, timeout, cert reload) are transient and retryable.
```
nabuskey: Looks like cert reload isn't handled. Did you mean to add it here or in a later PR?

nabuskey (commented):
Just three things we need to handle:
1. Context consolidation for handling signals correctly.
2. Reduce the number of flags for tls config.
3. podTemplate rendering to ensure rendered templates are consistent across submitters (rest and local).

internal/controller/sparkapplication/rest_submission.go
venkomirisetti: Yes, will come with a new PR for certs reload feature. The current setup should work without any issue for long valid certificates like webhook service.

cmd/operator/controller/start.go
venkomirisetti: I followed the existing webhook TLS pattern in this codebase which also exposes dir + individual filename flags (e.g. `--webhook-cert-name`, `--webhook-key-name`). The defaults already match the `kubernetes.io/tls` convention (`tls.crt`, `tls.key`, `ca.crt`), so users using cert-manager or standard secrets won't need to touch them. The flags are there for edge cases where someone mounts certs from a non-standard source with different filenames. Happy to remove if you feel strongly — just wanted to call out the precedent.

cmd/operator/controller/start.go
venkomirisetti: Done-- addressed

cmd/operator/controller/start.go
venkomirisetti: done

cmd/operator/controller/start.go
venkomirisetti: Done

cmd/operator/controller/start.go
venkomirisetti: Done

cmd/operator/controller/start.go
venkomirisetti: Done

cmd/operator/controller/start.go
venkomirisetti: Addressed

internal/controller/sparkapplication/submission.go
venkomirisetti: Gotcha missed it! Good point — addressed in ac598b4. I extracted `buildDriverPodTemplate`/`buildExecutorPodTemplate` as pure functions that render the template with the owner ref injected (DeepCopy to avoid mutating the original). Both the spark-submit path (`driverPodTemplateOption`) and the REST path (`newSubmitRequest`) now call these helpers, so the Kueue non-controller owner ref is always present in the pod templates regardless of submission strategy.

venkomirisetti:
Thanks again for the thorough review! I've addressed the feedback in ac598b4:

- **Context/error propagation**: `newSparkSubmitter` now takes `ctx` from `ctrl.SetupSignalHandler()` and returns an error instead of calling `os.Exit`.
- **Fail loudly on invalid config**: `--submitter-max-retries` validation moved to `PreRunE` — fails early with a clear error message.
- **Pod templates (Kueue owner ref)**: Extracted `buildDriverPodTemplate`/`buildExecutorPodTemplate` as pure functions used by both spark-submit and REST paths. Templates now always include the owner ref.
- **Cert reload comment**: Removed the misleading mention — cert reload is planned for a follow-up PR.
- **TLS filename flags**: Replied inline with rationale (follows existing webhook TLS pattern).

Ready for another look when you get a chance!

venkomirisetti:
Addressed the TLS flag simplification as discussed:

- Removed `--submitter-tls-cert-dir` and the filename flags (`--submitter-tls-cert-name`, `--submitter-tls-key-name`, `--submitter-tls-ca-name`)
- Added `--submitter-tls-enabled` bool flag to explicitly enable/disable mTLS
- Added 3 direct path flags: `--submitter-tls-cert-file`, `--submitter-tls-key-file`, `--submitter-tls-ca-file`
- Grouped all REST submitter validations (including `--submitter-service-url` required check) under the `RestSubmitter` feature gate in `PreRunE`

The cert-dir + filename approach added indirection — users still needed to know the filenames, and the dir flag doubled as the TLS enable signal which was implicit. Direct path flags are explicit, easier to validate, and align better with how cert-manager mounts secrets (each file at a known path). The `--submitter-tls-enabled` flag makes the intent clear without relying on "dir is set" as a proxy.

nabuskey (approved):
/approve

PR #2998, opened 2026-06-22
Bump Go to 1.25.11

nabuskey:
## Purpose of this PR

This PR bumps Go to 1.25.11 from 1.25.0

**Proposed changes:**

- Bump Go to 1.25.11
- Pin Go and Spark images to hashes

## Change Category

- [ ] Documentation update

### Rationale

We need to keep Go version updated

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,9 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=docker.io/library/spark:4.0.1
+ARG SPARK_IMAGE=docker.io/library/spark:4.0.1@sha256:a552a335e0aedb44fa28b20ed7e9dc52c2e856faf87fb0409fb57dbf37b45bf7
 
-FROM golang:1.25.9 AS builder
+FROM docker.io/library/golang:1.25.11@sha256:00feed335fe561979f2cdcc30a5191231977c5631fe79c40f7d3ab63b4fa222f AS builder
 
 WORKDIR /workspace
 
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/kubeflow/spark-operator/v2
 
 go 1.25.0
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -14,9 +14,9 @@
 # limitations under the License.
 #
 
-ARG SPARK_IMAGE=docker.io/library/spark:4.0.1
+ARG SPARK_IMAGE=docker.io/library/spark:4.0.1@sha256:a552a335e0aedb44fa28b20ed7e9dc52c2e856faf87fb0409fb57dbf37b45bf7
 
-FROM golang:1.25.9 AS builder
+FROM docker.io/library/golang:1.25.11@sha256:00feed335fe561979f2cdcc30a5191231977c5631fe79c40f7d3ab63b4fa222f AS builder AS builder
```
RobuRishabh: a mistype is here remove the AS builder its present twice.

RobuRishabh (changes requested):

RobuRishabh (changes requested):
@nabuskey The Dockerfile bump to `golang:1.25.11` looks good, but `go.mod` is unchanged ([here](https://github.com/kubeflow/spark-operator/blob/master/go.mod#L3) still has `go 1.25.0` only).

CI installs Go via `go-version-file: go.mod` ([integration.yaml#L32-L35](https://github.com/kubeflow/spark-operator/blob/master/.github/workflows/integration.yaml#L32-L35)), so tests/lint still run on 1.25.0 while the container image is built with 1.25.11.

To align CI with the Dockerfile builder without changing the language version line format, will need to add:

```go
toolchain go1.25.11

wdyt?

nabuskey:
Ah good point. We probably want to enforce patch level for Go but it's fine to use toolchain for now. We'll move to it next bump.

PR #2988, opened 2026-06-13
docs: Add dedicated Spark Operator documentation website

vara-bonthu:
## Purpose of this PR

This PR adds a **dedicated documentation website** for Kubeflow Spark Operator, built with **Sphinx + Furo + MyST Markdown** and configured for **Read the Docs** similar to the new [Kubeflow Trainer doc site](https://trainer-doc-website.readthedocs.io/en/latest/).

The project's docs currently live on kubeflow.org (Hugo). This gives Spark Operator its own versioned, searchable docs site with a custom landing page, light/dark themes, and a mobile-responsive layout — with all content migrated as-is from the existing kubeflow.org docs

The site lives under `docs/website/`, and all Python tooling (Read the Docs, CI, and the local `make` targets) is managed with `uv`, consistent with other Kubeflow repos.

Linked issue #2987

### Demo

https://github.com/user-attachments/assets/08b5d0bd-e9f2-4920-abf7-0f9e4c900e2e

## Change Category

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [x] Documentation update

### Rationale

A standalone, versioned docs site (like Kubeflow Trainer's) improves discoverability, search, and the new-user experience, and decouples the operator's docs from the broader kubeflow.org Hugo site. Content is migrated verbatim from the existing docs, so this is additive and non-breaking — no Go code or runtime behavior is touched.

## How to test

Prerequisite: [uv](https://docs.astral.sh/uv/) installed.  uv manages Python and all docs dependencies automatically (no venv or pip step needed).

### from a clone of kubeflow/spark-operator

```bash
git fetch origin pull/2988/head:pr-2988
git checkout pr-2988
```

### build and preview with live reload → http://localhost:8000

```bash
make docs-serve
```

Other targets:

    make docs            # one-off build into docs/website/_build/html
    make docs-test       # strict build (warnings = errors), same as CI
    make docs-linkcheck  # check external links

## Checklist

- [x] I have conducted a self-review of my own code.
- [x] I have updated documentation accordingly.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.

## Next Steps

1. **Merge this PR** — adds the Sphinx documentation site, the `.readthedocs.yaml` build config, and a CI workflow that builds + link-checks the docs on every PR.
2. **Set up Read the Docs (after merge):**
[...]

PR #2999, opened 2026-06-22
feat(helm): add RestSubmitter feature gate with controller integration

venkomirisetti:
## Summary
- Add `RestSubmitter` feature gate to the helm chart (disabled by default)
- Configure the controller to connect to an external REST submitter service when the gate is enabled
- Add submitter connection config in `values.yaml` (serviceUrl, timeouts, retries)
- Add startup probe to controller when submitter is enabled
- Add unit tests for enabled/disabled states

## Scope
- Controller deployment template changes only (submitter args, startup probe)
- `values.yaml` feature gate + submitter connection config
- Drift-check alignment (`config/manager/manager.yaml`)

## Out of scope
- Submitter service helm templates (deployed separately from the [submitter repo](https://github.com/venkomirisetti/k8s-spark-submitter))
- TLS/mTLS configuration (will be documented separately)
- E2E tests or CI workflow changes
- Documentation (to be added under `doc/website` as a follow-up)

## Test plan
- [x] Helm unit tests pass (216 tests)
- [x] Drift-check passes
- [x] `helm template` validated with feature gate enabled/disabled

### DIFF
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -84,6 +84,8 @@ controller:
     enabled: false
   - name: LoadSparkDefaults
     enabled: false
+  - name: RestSubmitter
+    enabled: false
 
   # -- The number of old history to retain to allow rollback.
   revisionHistoryLimit: 10
@@ -297,6 +299,34 @@ controller:
       enable: true
       # -- Specifies the maximum delay duration for the workqueue rate limiter.
       duration: 6h
+
+# REST Submitter Service configuration.
+# The submitter service itself is deployed separately (see submitter service repository).
+# These values configure the controller's connection to the submitter.
+# Enable via controller.featureGates RestSubmitter=true.
+submitter:
+  # -- URL of the REST submitter service endpoint.
+  serviceUrl: http://spark-operator-submitter-svc:8080/api/v1/spark-submit
+
+  # -- HTTP request timeout per spark submission attempt.
+  requestTimeout: 2m
+
+  # -- How long the controller waits for the submitter service to become reachable at startup.
+  startupTimeout: 5m
+
+  # -- Max retry attempts for transient submission failures.
+  maxRetries: 3
+
+  # -- Initial backoff duration before the first retry. Doubles on each subsequent attempt.
+  initialBackoff: 2s
+
+  # Controller startup probe settings. Allows the controller to block during startup
+  # without being killed by liveness probe. Should cover startupTimeout.
+  startupProbe:
+    # -- Seconds between probe attempts.
+    periodSeconds: 5
+    # -- Number of failures before restart. (failureThreshold × periodSeconds) should >= startupTimeout.
+    failureThreshold: 60
 
 webhook:
   # -- Specifies whether to enable webhook.
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -93,7 +93,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | hook.labels | object | `{}` | Extra labels for the Helm hook Job pod. |
 | hook.annotations | object | `{}` | Extra annotations for the Helm hook Job pod. |
 | controller.replicas | int | `1` | Number of replicas of controller. |
-| controller.featureGates | list | `[{"enabled":false,"name":"PartialRestart"},{"enabled":false,"name":"LoadSparkDefaults"}]` | Feature gates to enable or disable specific features. |
+| controller.featureGates | list | `[{"enabled":false,"name":"PartialRestart"},{"enabled":false,"name":"LoadSparkDefaults"},{"enabled":false,"name":"RestSubmitter"}]` | Feature gates to enable or disable specific features. |
 | controller.revisionHistoryLimit | int | `10` | The number of old history to retain to allow rollback. |
 | controller.leaderElection.enable | bool | `true` | Specifies whether to enable leader election for controller. |
 | controller.leaderElection.leaseDuration | string | `"15s"` | Leader election lease duration. |
@@ -147,6 +147,13 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.workqueueRateLimiter.bucketSize | int | `500` | Specifies the maximum number of items that can be in the workqueue at any given time. |
 | controller.workqueueRateLimiter.maxDelay.enable | bool | `true` | Specifies whether to enable max delay for the workqueue rate limiter. This is useful to avoid losing events when the workqueue is full. |
 | controller.workqueueRateLimiter.maxDelay.duration | string | `"6h"` | Specifies the maximum delay duration for the workqueue rate limiter. |
+| submitter.serviceUrl | string | `"http://spark-operator-submitter-svc:8080/api/v1/spark-submit"` | URL of the REST submitter service endpoint. |
+| submitter.requestTimeout | string | `"2m"` | HTTP request timeout per spark submission attempt. |
+| submitter.startupTimeout | string | `"5m"` | How long the controller waits for the submitter service to become reachable at startup. |
+| submitter.maxRetries | int | `3` | Max retry attempts for transient submission failures. |
+| submitter.initialBackoff | string | `"2s"` | Initial backoff duration before the first retry. Doubles on each subsequent attempt. |
+| submitter.startupProbe.periodSeconds | int | `5` | Seconds between probe attempts. |
+| submitter.startupProbe.failureThreshold | int | `60` | Number of failures before restart. (failureThreshold × periodSeconds) should >= startupTimeout. |
 | webhook.enable | bool | `true` | Specifies whether to enable webhook. |
 | webhook.replicas | int | `1` | Number of replicas of webhook server. |
 | webhook.revisionHistoryLimit | int | `10` | The number of old history to retain to allow rollback. |
--- a/charts/spark-operator-chart/templates/_helpers.tpl
+++ b/charts/spark-operator-chart/templates/_helpers.tpl
@@ -76,3 +76,12 @@ Spark Operator image
 {{- define "spark-operator.image" -}}
 {{ printf "%s/%s:%s" .Values.image.registry .Values.image.repository (.Values.image.tag | default .Chart.AppVersion | toString) }}
 {{- end -}}
+
+{{/*
+Whether the RestSubmitter feature gate is enabled.
+*/}}
+{{- define "spark-operator.submitter.enabled" -}}
+{{- range .Values.controller.featureGates -}}
+{{- if and (eq .name "RestSubmitter") .enabled -}}true{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -140,6 +140,13 @@ spec:
         {{- if .Values.controller.featureGates }}
         - --feature-gates={{ range $index, $gate := .Values.controller.featureGates }}{{ if $index }},{{ end }}{{ $gate.name }}={{ $gate.enabled }}{{ end }}
         {{- end }}
+        {{- if include "spark-operator.submitter.enabled" . }}
+        - --submitter-service-url={{ .Values.submitter.serviceUrl }}
+        - --submitter-request-timeout={{ .Values.submitter.requestTimeout }}
+        - --submitter-startup-timeout={{ .Values.submitter.startupTimeout }}
+        - --submitter-max-retries={{ .Values.submitter.maxRetries }}
+        - --submitter-initial-backoff={{ .Values.submitter.initialBackoff }}
+        {{- end }}
         {{- if or .Values.prometheus.metrics.enable .Values.controller.pprof.enable }}
         ports:
         {{- if .Values.controller.pprof.enable }}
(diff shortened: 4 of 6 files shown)


charts/spark-operator-chart/templates/submitter/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+      {{- end }}
+    spec:
+      containers:
+      - name: spark-submitter
+        image: {{ include "spark-operator.submitter.image" . }}
+        imagePullPolicy: {{ .Values.submitter.image.pullPolicy }}
+        ports:
+        - name: http
+          containerPort: {{ .Values.submitter.port }}
+          protocol: TCP
```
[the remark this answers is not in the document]
venkomirisetti: Addressed.

charts/spark-operator-chart/templates/submitter/service.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: {{ include "spark-operator.submitter.serviceName" . }}
+  labels:
+    {{- include "spark-operator.submitter.labels" . | nindent 4 }}
+spec:
+  selector:
+    {{- include "spark-operator.submitter.selectorLabels" . | nindent 4 }}
+  ports:
+  - port: {{ .Values.submitter.port }}
+    targetPort: http
+    name: http
```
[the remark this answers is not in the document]
venkomirisetti: Addressed

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
```diff
@@ -0,0 +1,96 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/ -}}
+
+{{- if include "spark-operator.submitter.enabled" . }}
+{{- if .Values.submitter.tls.enabled }}
+{{- if .Values.certManager.enable }}
+{{- if not (.Capabilities.APIVersions.Has "cert-manager.io/v1/Certificate") }}
+{{- fail "The cluster does not support the required API version `cert-manager.io/v1` for `Certificate`." }}
```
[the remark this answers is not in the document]
venkomirisetti: Addressed

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
```diff
@@ -0,0 +1,96 @@
...
+  isCA: true
+  secretName: {{ include "spark-operator.submitter.tls.secretName" . }}-ca
+  issuerRef:
+    {{- if not .Values.certManager.issuerRef }}
+    group: cert-manager.io
+    kind: Issuer
+    name: {{ include "spark-operator.certManager.issuer.name" . }}
+    {{- else }}
+    {{- toYaml .Values.certManager.issuerRef | nindent 4 }}
+    {{- end }}
```
tariq-hasan: Since we fall to the default issuer if no custom `issuerRef` is provided I am wondering if the issuer should be [gating](https://github.com/venkomirisetti/spark-operator/blob/879b96322292863cb0561208b6fe7ce82a5ba3af/charts/spark-operator-chart/templates/certmanager/issuer.yaml#L17) on webhook being exclusively enabled.

charts/spark-operator-chart/values.yaml
```diff
@@ -298,6 +300,133 @@ controller:
...
+# Spark Submitter Service configuration.
+# Enable via controller.featureGates RestSubmitter=true.
+submitter:
+  image:
+    # -- Submitter service image repository.
+    repository: docker.io/venkomirisetti/k8s-spark-submitter
+    # -- Submitter service image tag.
+    tag: 4.0.1-15
+    # -- Submitter service image pull policy.
+    pullPolicy: IfNotPresent
```
tariq-hasan: I was curious if the goal is to update with an image published from the org-owned registry once the submitter service and its image publish workflow are merged or if that would happen sooner.

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
```diff
@@ -0,0 +1,96 @@
+{{- /*
+Copyright 2025 The Kubeflow authors.
```
tariq-hasan: This is more of a nit but I'm thinking maybe we should be using the current year for all new files as per the convention in other subprojects.

On this note I have noticed that the increasingly prevalent practice in the [CNCF ecosystem](https://github.com/kubernetes/steering/issues/299) is to leave out the year altogether and some Kubeflow [subprojects](https://github.com/kubeflow/trainer/issues/3275) have already begun to adopt this practice.

charts/spark-operator-chart/templates/submitter/rbac.yaml
```diff
@@ -0,0 +1,68 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "spark-operator.submitter.name" . }}
+subjects:
+- kind: ServiceAccount
+  name: {{ include "spark-operator.submitter.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+{{- end }}
+{{- end }}
```
tariq-hasan: I am wondering if we want to enable namespace scoping here given that this is already supported on the controller.

charts/spark-operator-chart/templates/submitter/_helpers.tpl
```diff
@@ -0,0 +1,112 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+
+{{/*
+Submitter service image
+*/}}
+{{- define "spark-operator.submitter.image" -}}
+{{ printf "%s:%s" .Values.submitter.image.repository .Values.submitter.image.tag }}
+{{- end -}}
```
tariq-hasan: I am wondering if we want to have [parity](https://github.com/venkomirisetti/spark-operator/blob/879b96322292863cb0561208b6fe7ce82a5ba3af/charts/spark-operator-chart/templates/_helpers.tpl#L73-L78) in the image helpers in regards to supporting also a registry config.

charts/spark-operator-chart/values.yaml
```diff
@@ -298,6 +300,133 @@ controller:
...
+  resources: {}
+  # requests:
+  #   cpu: 100m
+  #   memory: 256Mi
+  # limits:
+  #   cpu: 500m
+  #   memory: 512Mi
+
+  # -- Security context for the submitter container.
+  securityContext: {}
```
tariq-hasan: I was wondering if we want to have parity with the controller and webhook in terms of security context.

charts/spark-operator-chart/values.yaml
```diff
@@ -298,6 +300,133 @@ controller:
...
+  tolerations: []
+
+  # -- Topology spread constraints for submitter pods.
+  topologySpreadConstraints: []
+
+  # -- Priority class for submitter pods.
+  priorityClassName: ""
+
+  # -- Security context for submitter pods.
+  podSecurityContext: {}
```
tariq-hasan: I was wondering if we want to have parity with the controller and webhook in terms of pod security context.

tariq-hasan (commented):
Hi @venkomirisetti! I have added a few questions.

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
venkomirisetti: Great catch! Fixed — the issuer now renders when either webhook or submitter TLS is enabled, so it's no longer exclusively gated on webhook.

charts/spark-operator-chart/values.yaml
venkomirisetti: Yes, the plan is to move this to the kubeflow org registry. I've requested repo creation (https://cloud-native.slack.com/archives/C074588U7EG/p1781632881850969) — once that's done, we'll update the default registry. For now using my personal registry as a placeholder.

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
venkomirisetti: Updated all new files to use `Copyright 2026`.

charts/spark-operator-chart/templates/submitter/rbac.yaml
venkomirisetti: The submitter needs cluster-level RBAC since it creates pods, configmaps, and services across all namespaces for driver creation. Namespace-scoped RBAC would break multi-namespace Spark job submissions.

charts/spark-operator-chart/templates/submitter/_helpers.tpl
venkomirisetti: Added — the submitter image helper now uses `registry/repository:tag` format, matching the controller pattern.

charts/spark-operator-chart/values.yaml
venkomirisetti: Added hardened defaults matching the controller/webhook: `readOnlyRootFilesystem`, `runAsNonRoot`, `runAsUser: 185`, `allowPrivilegeEscalation: false`, `drop ALL` capabilities, and `RuntimeDefault` seccomp profile.

charts/spark-operator-chart/values.yaml
venkomirisetti: Added `fsGroup: 185` matching controller/webhook defaults.

venkomirisetti:
@tariq-hasan Thanks for the thorough review! All comments have been addressed in the latest commit:

1. **Issuer gating** — Fixed to render when either webhook or submitter TLS is enabled
2. **Image registry** — Pending org repo creation; using personal registry as placeholder
3. **Copyright year** — Updated to 2026 for all new files
4. **Namespace scoping** — Submitter needs cluster-level RBAC for cross-namespace driver creation
5. **Image helper** — Added `registry` field matching controller pattern
6. **Security context** — Added hardened defaults matching controller/webhook
7. **Pod security context** — Added `fsGroup: 185`

Ready for re-review when you get a chance!

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
```diff
@@ -0,0 +1,96 @@
...
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "spark-operator.certManager.submitter.certificate.name" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "spark-operator.submitter.labels" . | nindent 4 }}
+spec:
+  secretName: {{ include "spark-operator.submitter.tls.secretName" . }}
```
tariq-hasan: I am curious if we should share the same certificate between the [submitter](https://github.com/venkomirisetti/spark-operator/blob/666fb0a7a727bf116dedd338cfdd99e9c1511294/charts/spark-operator-chart/templates/submitter/deployment.yaml#L132-L134) and the [controller](https://github.com/venkomirisetti/spark-operator/blob/666fb0a7a727bf116dedd338cfdd99e9c1511294/charts/spark-operator-chart/templates/controller/deployment.yaml#L229-L231) or if distinct certificates would be more appropriate in the mTLS setup.

charts/spark-operator-chart/templates/submitter/rbac.yaml
tariq-hasan: Since the operator supports namespace-related [flags](https://github.com/venkomirisetti/spark-operator/blob/666fb0a7a727bf116dedd338cfdd99e9c1511294/charts/spark-operator-chart/values.yaml#L609-L621) my understanding was coming from how those are wired into the [cluster-level](https://github.com/venkomirisetti/spark-operator/blob/666fb0a7a727bf116dedd338cfdd99e9c1511294/charts/spark-operator-chart/templates/controller/rbac.yaml#L30-L47) and [namespace-level](https://github.com/venkomirisetti/spark-operator/blob/666fb0a7a727bf116dedd338cfdd99e9c1511294/charts/spark-operator-chart/templates/controller/rbac.yaml#L140-L182) permissions for the controller (and by extension, spark-submit). This was actually just more about ensuring parity with how the controller wires those permissions and enabling the smallest set of privileges.

tariq-hasan (commented):
I just have these two additional comments. Thanks.

charts/spark-operator-chart/templates/certmanager/submitter-certificate.yaml
venkomirisetti: Great catch on highlighting this critical aspect of the mTLS design!

Both deployments are part of the same operator, same namespace, and same trust boundary — the communication is between different components of a single logical service. A shared certificate keeps things simple while still providing TLS protection for payloads in transit. Distinct certificates would add complexity without meaningful security benefit in this context, since the blast radius is identical either way.

That said, I'm open to splitting into distinct client/server certificates if you feel it's warranted — happy to make the change.

charts/spark-operator-chart/templates/submitter/rbac.yaml
venkomirisetti: Good point — updated the submitter RBAC to mirror the controller pattern.

venkomirisetti (commented):
@tariq-hasan Addressed your latest review comments.

tariq-hasan:
> @tariq-hasan Addressed your latest review comments.

Sounds good! No further questions from my end.
/cc @nabuskey @RobuRishabh

venkomirisetti:
@nabuskey Reworked the PR based on your feedback — stripped out all submitter service templates, TLS/cert-manager changes, and kept only the controller integration with simple HTTP communication.

**What's in this PR now:**
- `RestSubmitter` feature gate in `values.yaml` (disabled by default)
- Controller deployment renders submitter args (service URL, timeouts, retries) when gate is enabled
- Startup probe on controller when submitter is enabled
- Unit tests for both states

**Separate follow-ups:**
- Submitter service helm chart → submitter source repo
- Documentation → `doc/website` (will follow the Sphinx/Furo pattern from PR #2988)

Ready for your review when you get a chance!

nabuskey (approved):
/approve

PR #3000, opened 2026-06-22
fix: use API reader for namespace label checks in event filters

RobuRishabh:
## Purpose of this PR

Fixes a race condition in namespace selector filtering that causes flaky e2e failures (`Namespace Filtering > With namespace selector > should process SparkApp in labeled namespace`).

When `jobNamespaceSelector` is configured, event filters evaluate namespace labels via the controller's **cached client**. If a SparkApplication create event arrives before the namespace informer has indexed the new namespace, the lookup fails, the predicate returns `false`, and the event is **silently dropped** with no retry. The SparkApplication stays in `New` state until the test times out.

### Observed CI failures

This race has been seen across multiple PRs and Kubernetes versions:

| CI run | K8s version | Symptom |
|---|---|---|
| [job/79602618303](https://github.com/kubeflow/spark-operator/actions/runs/26975877379/job/79602618303?pr=2975) | v1.32.11 | Namespace filtering flake — SparkApp stuck in `New`, 5min timeout |
| [job/79746346579](https://github.com/kubeflow/spark-operator/actions/runs/26976934082/job/79746346579?pr=2975) | v1.33.7 | BeforeSuite timeout (likely separate Helm readiness issue) |
| [job/79606205486](https://github.com/kubeflow/spark-operator/actions/runs/26976934082/job/79606205486?pr=2975) | v1.34.3 | BeforeSuite timeout (likely separate Helm readiness issue) |

Also reproduced on PR #2990 and other unrelated PRs (e.g. REST submitter).

**Proposed changes:**

- Use `mgr.GetAPIReader()` for live namespace label lookups in SparkApplication, ScheduledSparkApplication, and SparkConnect event filters/predicates
- Update `MatchesWithClient` to accept `client.Reader` instead of `client.Client`
- Keep the cached `client.Client` on SparkApplication event filters only where writes are needed (e.g. `Status().Update()` in the Update predicate)

## Change Category

- [ ] Documentation update

### Rationale

The operator maintains two independent informer caches (namespaces and SparkApplications). A test (or user) can create a labeled namespace and immediately submit a SparkApplication — the test client sees the namespace, but the controller cache may not yet.

Using the uncached API reader for label checks eliminates this lag while preserving correct fail-closed behavior: unlabeled namespaces still return `false`, and non-existent namespaces still return a real `NotFound`.

This is preferred over returning `true` on lookup error, which would fix the flake but incorrectly allow events through for unlabeled namespaces during the same race window.

## Checklist

### DIFF
--- a/internal/controller/scheduledsparkapplication/controller.go
+++ b/internal/controller/scheduledsparkapplication/controller.go
@@ -237,7 +237,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, options controller.Optio
 
 	// Create event filter with error handling
 	eventFilter, err := NewEventFilter(
-		mgr.GetClient(),
+		mgr.GetAPIReader(),
 		r.options.Namespaces,
 		r.options.NamespaceSelector,
 	)
--- a/internal/controller/scheduledsparkapplication/event_filter.go
+++ b/internal/controller/scheduledsparkapplication/event_filter.go
@@ -31,7 +31,7 @@ import (
 
 // EventFilter filters out ScheduledSparkApplication events.
 type EventFilter struct {
-	client           client.Client
+	namespaceReader  client.Reader
 	namespaceMatcher *util.NamespaceMatcher
 	logger           logr.Logger
 }
@@ -40,14 +40,14 @@ type EventFilter struct {
 var _ predicate.Predicate = &EventFilter{}
 
 // NewEventFilter creates a new EventFilter instance.
-func NewEventFilter(client client.Client, namespaces []string, namespaceSelector string) (*EventFilter, error) {
+func NewEventFilter(namespaceReader client.Reader, namespaces []string, namespaceSelector string) (*EventFilter, error) {
 	matcher, err := util.NewNamespaceMatcher(namespaces, namespaceSelector)
 	if err != nil {
 		return nil, err
 	}
 
 	return &EventFilter{
-		client:           client,
+		namespaceReader:  namespaceReader,
 		namespaceMatcher: matcher,
 		logger:           log.Log.WithName("scheduled-spark-application-event-filter"),
 	}, nil
@@ -88,7 +88,7 @@ func (f *EventFilter) Generic(e event.GenericEvent) bool {
 
 func (f *EventFilter) filter(app *v1beta2.ScheduledSparkApplication) bool {
 	// Check if namespace matches using the matcher
-	matched, err := f.namespaceMatcher.MatchesWithClient(context.TODO(), f.client, app.Namespace)
+	matched, err := f.namespaceMatcher.MatchesWithClient(context.TODO(), f.namespaceReader, app.Namespace)
 	if err != nil {
 		f.logger.Error(err, "failed to check namespace match", "namespace", app.Namespace, "app", app.Name)
 		return false
--- a/internal/controller/sparkapplication/controller.go
+++ b/internal/controller/sparkapplication/controller.go
@@ -248,7 +248,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, options controller.Optio
 
 	// Create event filters with error handling
 	podEventFilter, err := newSparkPodEventFilter(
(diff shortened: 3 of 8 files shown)


nabuskey (commented):
whenever `GetAPIReader` is used, I think of API costs.

In this case with namespace selector used, no caching is used? So this mean we will hit the API for pod updates and app updates?

also combining ctx TODO with the changes are problematic. It can block indefinitely.

RobuRishabh:
> whenever `GetAPIReader` is used, I think of API costs.
> 
> In this case with namespace selector used, no caching is used? So this mean we will hit the API for pod updates and app updates?

@nabuskey I traced the code paths to clarify scope.

**Caching is not removed.** This PR only changes namespace *label lookups* in predicates from cached client → `GetAPIReader()`. Pod/SparkApp reconciles, event handlers, and informer caches are unchanged (`newCacheOptions` untouched).

**live API is wired (4 sites only):**
- `sparkapplication/controller.go` — pod + app event filters
- `scheduledsparkapplication/controller.go` — event filter
- `sparkconnect/reconciler.go` — namespace predicate

**the namespace GET actually happens:** only `pkg/util/namespace.go` → `MatchesWithClient()`, and only when `jobNamespaceSelector` is set and the namespace is not in the explicit `--namespaces` list. Three fast-paths return without any GET.

**Not every pod/app event hits the API:**
- Reconcile: `r.client.Get()` (cache) — see `getSparkApplication()`
- Pod updates: filtered out if phase unchanged or pod is not Spark-operator-labeled before namespace check runs

Before this PR we already called `Get(namespace)` on these paths, the difference is cache vs live API. The live lookup fixes the informer cache race that silently dropped create events.

nabuskey:
When `namespaceSelector` is used, caching is effectively removed, correct? It doesn't use cache for anything and it fires every single event. Namespaces do not change often in most clusters. Far fewer than pods. I don't think we should introduce this just to fix a test. 
The use of todo context with direct GET is also a big concern. We now hit the api directly without bounded by the cache. Previously, the use of todo context was not an issue because it hits the cache, but now it hits the API directly with no constraints as to when it exists. This can be quite bad when API is under high load which is often the case for production spark clusters due to the nature of how Spark works in k8s.

RobuRishabh:
> When `namespaceSelector` is used, caching is effectively removed, correct? It doesn't use cache for anything and it fires every single event. Namespaces do not change often in most clusters. Far fewer than pods. I don't think we should introduce this just to fix a test. The use of todo context with direct GET is also a big concern. We now hit the api directly without bounded by the cache. Previously, the use of todo context was not an issue because it hits the cache, but now it hits the API directly with no constraints as to when it exists. This can be quite bad when API is under high load which is often the case for production spark clusters due to the nature of how Spark works in k8s.

@nabuskey Your both concerns make sense to me.

**On API cost:** You're right that with the current PR, namespace label lookups in predicates always use the API reader when `jobNamespaceSelector` is set. I should have been clearer: workload caching (pods, SparkApps, reconciler reads) is unchanged, but namespace *label checks in predicates* no longer use cache, so yes, those paths can hit the apiserver on every filtered event. That's heavier than I'd like for production Spark clusters.

**On `context.TODO()`:** Agree this is problematic once we call the live API. Predicate callbacks don't receive a context today, but we can use a bounded timeout (e.g. `context.WithTimeout`) for the API fallback path rather than unbounded `context.TODO()`.

**On "just to fix a test":** The e2e test exposes a real production race,  create labeled namespace + SparkApp before the namespace informer indexes it → predicate returns false → create event is silently dropped (no retry) → app stuck in `New`. Same pattern can happen with CI applying both resources in one pass. So I'd like to fix the bug, but I agree pure `GetAPIReader()` everywhere isn't the right tradeoff.

What I can do, I can 
1. Try cached client first for namespace label lookup
2. On `NotFound` only, fall back to API reader with a short timeout
3. Use `%w` for error wrapping 

if you prefer a different approach (e.g. namespace watch + requeue). or just close this PR. I am open to anything.

PR #3001, opened 2026-06-22
Add deploymentAnnotations to controller and webhook deployments

cHiv0rz:
## Purpose of this PR

The chart supports `controller.annotations` and `webhook.annotations`, but these target the pod template (`spec.template.metadata.annotations`), not the Deployment resource itself (`metadata.annotations`).

Tools like [Reloader](https://github.com/stakater/Reloader) watch for annotations on the Deployment object to trigger rolling restarts when ConfigMaps or Secrets change. There is currently no way to set Deployment-level annotations without forking the chart templates.

**Proposed changes:**

- Add controller.deploymentAnnotations value applied to metadata.annotations of the controller Deployment
- Add webhook.deploymentAnnotations value applied to metadata.annotations of the webhook Deployment
- Add helm unit tests for both new values

## Change Category

- [ ] Documentation update

### Rationale

The existing `annotations` keys are already documented as pod annotations. Adding a separate `deploymentAnnotations` key mirrors that pattern and keeps the distinction explicit, consistent with how other charts (e.g. ingress-nginx, cert-manager) handle this separation.

## Checklist

### Additional Notes

No breaking changes — `deploymentAnnotation`s defaults to `{}` in both cases, so existing deployments are unaffected.

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -124,6 +124,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.rbac.annotations | object | `{}` | Extra annotations for the controller RBAC resources. |
 | controller.labels | object | `{}` | Extra labels for controller pods. |
 | controller.annotations | object | `{}` | Extra annotations for controller pods. |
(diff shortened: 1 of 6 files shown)


nabuskey (approved):
Thank you very much!

/approve

PR #3003, opened 2026-06-23
Preserve user-supplied service port

nabuskey:
`mutateServerService` overwrote the entire `Spec.Ports` slice on creation,
  discarding ports supplied via the `Server.Service` override.

It now merges instead: user-supplied ports are kept, and any missing required
  ports are appended by name. Existing Services are left untouched.
## Change Category

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/controller/sparkconnect/reconciler.go
+++ b/internal/controller/sparkconnect/reconciler.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/events"
 	"k8s.io/client-go/util/workqueue"
-	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


internal/controller/sparkconnect/reconciler.go
```diff
@@ -498,40 +498,28 @@ func (r *Reconciler) createOrUpdateServerService(ctx context.Context, conn *v1al
...
+			{Name: "blockmanager", Port: 7079, TargetPort: intstr.FromInt32(7079), Protocol: corev1.ProtocolTCP, AppProtocol: ptr.To("tcp")},
+			{Name: "driver-rpc", Port: 7078, TargetPort: intstr.FromInt32(7078), Protocol: corev1.ProtocolTCP, AppProtocol: ptr.To("tcp")},
+			{Name: "spark-connect-server", Port: 15002, TargetPort: intstr.FromInt32(15002), Protocol: corev1.ProtocolTCP, AppProtocol: ptr.To("grpc")},
+			{Name: "web-ui", Port: 4040, TargetPort: intstr.FromInt32(4040), Protocol: corev1.ProtocolTCP, AppProtocol: ptr.To("http")},
+		}
+
+		ports := map[string]struct{}{}
+		for _, port := range svc.Spec.Ports {
+			ports[port.Name] = struct{}{}
+		}
```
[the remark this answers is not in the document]
RobuRishabh: I agree with copilot here, consider failing early here for clearer reconcile errors.

internal/controller/sparkconnect/reconciler.go
```diff
@@ -498,40 +498,28 @@ func (r *Reconciler) createOrUpdateServerService(ctx context.Context, conn *v1al
...
+		for _, port := range requiredPorts {
+			_, ok := ports[port.Name]
+			if !ok {
+				svc.Spec.Ports = append(svc.Spec.Ports, port)
+			}
 		}
```
tariq-hasan: I am wondering if we want to start with the required ports as the base and only add in user-provided ports which do not conflict on the port name and which do not conflict on the port number and protocol combination.

I mention this because otherwise we might have a user-provided name targeting 15002, for example, and `spark-connect-server` targeting 15002 as well which would then cause an API server error.

Also the ports are hardcoded either in the [operator](https://github.com/nabuskey/spark-operator/blob/0e625a94d4f8548e317d36d7d62cc44018236c7e/internal/controller/sparkconnect/options.go#L180-L181) or natively in Spark. So if a user overrides the port number for `web-ui` to 4000, for example, then it could be pointing at a port on the driver pod that may not be serving any requests.

internal/controller/sparkconnect/reconciler.go
tariq-hasan, line 568: I am wondering if webhook validation on the user-provided service would be useful as well.

This would be for the four protected ports as well as other user-provided ports in case the spec for those are invalid as well.

tariq-hasan (commented):
Hi @nabuskey! I have added a few comments.

internal/controller/sparkconnect/reconciler.go
nabuskey: Great points. I agree completely.

nabuskey: Actually thinking about it more, what does it mean when a user provides `spec.server.service` field? To me, it means three things:
1. If required ports are missing, the operator is expected to add these to ensure it works.
2. If additional ports are compatible with required ports, the operator add them.
3. If additional ports conflicts with required ports, we reject explicitly.

With these in mind,  I wonder if we should be exposing service object as-is in the spec. Follow the spark application convention instead. This would be a breaking spec change so we need to this in the next version.

tariq-hasan: That makes sense. The long-term fix would also enable better parity with `SparkApplication`.

internal/webhook/sparkconnect_validator.go
```diff
@@ -262,9 +264,51 @@ func (v *SparkConnectValidator) validateServerSpec(sc *v1alpha1.SparkConnect) er
...
+	for _, up := range ports {
+		req, ok := requiredByName[up.Name]
+		if !ok {
+			continue
+		}
+		if !IsCompatibleServerServicePort(up, req) {
+			return fmt.Errorf(
+				"service port %q conflicts with the required Spark Connect server port (required targetPort %q, protocol %q)",
+				up.Name, req.TargetPort.String(), req.Protocol,
+			)
```
RobuRishabh, line 290: I was wondering if a user-supplied port with empty Name, it will slip past validation but produce a multi-port Service with an unnamed entry, which the API server rejects. we can add an early check: `if up.Name == "" { return error }` before the requiredByName lookup. Also if add this we should also add a test case Add a table entry for an unnamed port `{Name: "", Port: 9999}` with `wantError: true` to cover the unnamed-port edge case.

internal/controller/sparkconnect/reconciler.go
RobuRishabh, line 53: this is stale now. It is now duplicated by common.SparkConnectServerPort.

RobuRishabh (commented):
@nabuskey left some minor comments. After that, this is LGTM from my side.

PR #3015, opened 2026-06-25
feat(chart): make Helm hook Job resources configurable

maxpschonder:
## What this does

Makes the resource requests/limits of the CRD-upgrade hook Job configurable through a new `hook.resources` value.

## Why

We run the operator on OpenShift and, for compliance reasons, have to build our own hook image with the OpenShift client baked in. That `oc`-client simply doesn't fit into the 64Mi memory limit the chart currently hardcodes in `templates/hook/job.yaml`, so the hook pod gets OOM-killed and the install/upgrade fails. There's currently no way to bump that limit from values.

I went with the same approach already used for `controller.resources` and `webhook.resources` so it feels consistent with the rest of the chart. The default for `hook.resources` keeps the existing 100m/64Mi requests and limits, so nothing changes for anyone who doesn't touch it.

## Changes

- `templates/hook/job.yaml` – render the hook container resources from `.Values.hook.resources`
- `values.yaml` – add `hook.resources`, defaulting to the previous hardcoded values
- `README.md` – regenerated with `helm-docs`
- `tests/hook/job_test.yaml` – tests for the default and an overridden value

`make helm-unittest` is green.

### DIFF
--- a/charts/spark-operator-chart/README.md
+++ b/charts/spark-operator-chart/README.md
@@ -87,6 +87,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | hook.image.registry | string | `"ghcr.io"` | Image registry. |
 | hook.image.repository | string | `"kubeflow/spark-operator/kubectl"` | Image repository. |
 | hook.image.tag | string | If not set, the chart appVersion will be used. | Image tag. |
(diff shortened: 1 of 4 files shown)


nabuskey (approved):
Thank you very much!

/approve

PR #3016, opened 2026-06-26
fix(webhook): enforce ResourceQuota using Status.Hard instead of Spec…

saddamr3e:
While reviewing the quota validation logic, I noticed that `validateResourceQuota` checks whether a resource exists in `Status.Hard` but performs the final comparison against `Spec.Hard`.

This change makes the validation consistent by using `Status.Hard` for both the lookup and the comparison. A regression test has also been added to cover the case where `Spec.Hard` and `Status.Hard` contain different values.

**Proposed changes:**

* Update `validateResourceQuota` to compare resource usage against `resourceQuota.Status.Hard`.
* Add a regression test (`TestValidateResourceQuota_SpecHardDivergence`) covering divergent `Spec.Hard` and `Status.Hard` values.

## Change Category

Bug fix (non-breaking change)
This change updates the resource quota validation logic to consistently use Status.Hard when checking resource limits and adds a regression test to verify the behavior when Spec.Hard and Status.Hard differ.

### Rationale

The function already uses `Status.Hard` to determine whether a resource is tracked. Using the same field for the limit comparison keeps the validation logic consistent and avoids mixing two different quota fields during the same validation.

## Checklist

[ ] I have conducted a self-review of my own code.
[ ] I have updated documentation accordingly.
[ ] I have added tests that prove my changes are effective.
 [ ] Existing unit tests pass locally with my changes.

### Additional Notes

The regression test covers the scenario where `Spec.Hard` and `Status.Hard` differ. Running the full webhook test suite locally requires the Kubernetes envtest binaries, which are not available in my current environment.

### DIFF
--- a/internal/webhook/resourcequota.go
+++ b/internal/webhook/resourcequota.go
@@ -251,7 +251,7 @@ func validateResourceQuota(resourceList corev1.ResourceList, resourceQuota corev
 			continue
 		}
 		quantity.Add(resourceQuota.Status.Used[key])
-		if quantity.Cmp(resourceQuota.Spec.Hard[key]) > 0 {
+		if quantity.Cmp(resourceQuota.Status.Hard[key]) > 0 {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


saddamr3e:
any update?

PR #3004, opened 2026-06-23
Scope make e2e-test to a dedicated Kind cluster

nabuskey:
## Purpose of this PR

**Proposed changes:**

- make e2e-test no longer targets the active kubectl context, preventing accidental changes to the current (possibly production) cluster
- Added kind-create-cluster as a prerequisite of e2e-test so the cluster is provisioned automatically 
- Exported KUBECONFIG=$(KIND_KUBE_CONFIG) to the go test invocation so the suite connects only to the Kind cluster
- Changed KIND_KUBE_CONFIG default from $(HOME)/.kube/config to $(CURDIR)/.kube/config to isolate cluster state from the user's global kubeconfig

## Change Category

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

## Checklist

- [ ] I have conducted a self-review of my own code.
- [ ] I have updated documentation accordingly.
- [ ] I have added tests that prove my changes are effective or that my feature works.
- [ ] Existing unit tests pass locally with my changes.

PR #3019, opened 2026-06-26
(#2963) feat(webhook): reject disallowed URL schemes in SparkApplication fetch fields

### LINKED ISSUE #2963: [FEATURE] Reject configured URL schemes in spec.sparkConf at admission
---
name: Feature request
about: Suggest an idea for this project
title: '[FEATURE] Reject configured URL schemes in spec.sparkConf at admission'
labels: enhancement
---

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?

`sparkConfOption` in `internal/controller/sparkapplication/submission.go` forwards every `spec.sparkConf` entry to `spark-submit` as `--conf key=value`. `spark-submit` runs in the operator pod, so any `http`/`https`/`ftp` URL in a fetch-capable conf (`spark.kubernetes.{driver,executor}.podTemplateFile`, `spark.jars`, `spark.files`, `spark.archives`, `spark.submit.pyFiles`, `spark.kubernetes.file.upload.path`, and any future additions) is dereferenced by the operator's principal: its ServiceAccount, mounted secrets, Workload Identity / IRSA, and VPC reachability. The submitter is a different and lower-privilege principal. This is SSRF.

`SparkApplicationValidator.validateSpec` does not inspect `sparkConf`. There is no operator-side knob to restrict it.

#### Describe the solution you would like

Add a controller flag `--sparkconf-denied-url-schemes` (and matching Helm value), default empty (no behavior change). When set, `validateSpec` rejects any `SparkApplication` whose `spec.sparkConf` contains a value whose `url.Parse` scheme is in the deny list. Comma-separated values (`spark.jars`, etc.) are split and each element checked. Empty/relative values are unaffected.

Recommended hardening: `--sparkconf-denied-url-schemes=http,https,ftp`.

Value-based, not key-based: maintaining a list of fetch-capable Spark conf keys drifts as Spark evolves and any miss is a silent bypass. Value-scheme checks are Spark-version-independent. False positives (e.g. an HTTP URL deliberately embedded in `spark.driver.extraJavaOptions` for a JVM-side fetch) take the same migration path as legitimate fetches: switch to an allowed scheme.

Acceptance:
- [ ] Flag + Helm value.
- [ ] `validateSpec` rejection with an error naming the offending key and scheme.
- [ ] Comma-separated values split and checked.
- [ ] Unit tests + e2e covering rejection and the unset-flag default.
- [ ] README/values docs.

#### Describe alternatives you have considered

- Key-based allowlist: rejected, drift problem above.
- Operator-side pre-fetch with a hardened client: larger surface, host/IP allowlists are hard to get right (DNS rebinding, redirects, IPv6, alternate IP literals).
- NetworkPolicy only: useful as defense-in-depth, can't produce admission-time errors or carry per-CR context.
- Mutating webhook rewriting URLs: surprising; rejection is clearer.
- External OPA/Kyverno: re-implements the same check per consumer; no shipped default.

#### Additional context

The validator alone isn't sufficient if `spark-submit` ever puts fetched response bytes in stderr: `runSparkSubmit` returns raw `exitErr.Stderr`, `submitSparkApplication` writes it to `app.Status.AppState.ErrorMessage`, and `recordSparkApplicationEvent` propagates it to a Warning Event - readable by anyone with `get sparkapplications`/`get events`. The admission-time deny list closes the input vector; sanitizing those side channels closes the exfil vector. Both are needed.

vjanelle:
## Purpose of this PR

spark-submit runs inside the operator pod, so any http/https/etc. URL in a
fetch-capable SparkApplication field (spec.mainApplicationFile, spec.sparkConf
values, spec.deps.{jars,files,pyFiles,archives,repositories}) is dereferenced by
the operator's principal: its ServiceAccount, mounted secrets, Workload Identity
/ IRSA, and VPC reachability. That is SSRF. The validating webhook previously did
not inspect these fields and there was no operator-side knob to restrict them.

This adds an admission-time check that validates the URL scheme of every value in
those fields. Schemeless paths and file://, local:// URIs are always allowed (they
reference local or in-image files the operator never fetches); any other scheme
must appear in the operator's allow list or the SparkApplication is rejected with
an error naming the offending field, scheme, and value. The check is value-based
rather than key-based so it stays Spark-version independent and cannot be silently
bypassed by a newly fetch-capable conf key.

Fixes #2963.

**Proposed changes:**

- Add --spark-allowed-url-schemes webhook flag and matching
  webhook.sparkAllowedURLSchemes Helm value (default empty: only schemeless,
  file://, local:// pass; add gs, s3a, hdfs, etc. as needed).
- Check spec.mainApplicationFile, spec.sparkConf (comma-split), and the
  fetch-capable spec.deps.* lists in SparkApplicationValidator.validateSpec.
  spec.deps.packages/excludePackages are excluded (Maven coordinates, not URLs).
- Fail closed: reject values url.Parse cannot handle and scheme-relative
  //host/path references that are not local paths.
- Cover the behavior with unit tests, a deps-field coverage guard, and an e2e
  rejection/acceptance spec; add a kind-e2e-test Make target.

## Change Category

- [x] Documentation update

### Rationale

A key-based allowlist drifts as Spark evolves and any miss is a silent bypass;
value-scheme checks are Spark-version independent. The default allows only local
references so the hardening is on by default, while operators opt into the remote
schemes their workloads actually need.

## Checklist

### Additional Notes

spec.hadoopConf is intentionally out of scope: its values become spark.hadoop.*
config consumed by the driver/executor at runtime, not URLs the operator fetches
at submit time, and many are legitimately host/endpoint-shaped. The issue also
notes an exfil vector (spark-submit stderr propagated to Status.ErrorMessage and
Warning events); sanitizing those side channels is tracked separately.

Signed-off-by: Vincent Janelle <randomfrequency@gmail.com>
Co-authored-by: Claude <claude@anthropic.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,6 +202,10 @@ e2e-test: envtest kind-load-image kind-load-spark-image ## Run the e2e tests aga
 	@echo "Running e2e tests (deploy_method=$(DEPLOY_METHOD))..."
 	DEPLOY_METHOD=$(DEPLOY_METHOD) IMAGE_TAG=$(IMAGE_TAG) KUBECONFIG=$(KIND_KUBE_CONFIG) go test ./test/e2e/ -v -ginkgo.v -timeout 30m
 
+.PHONY: kind-e2e-test
+kind-e2e-test: IMAGE_TAG=local
+kind-e2e-test: kind-load-image e2e-test ## Build image, load into Kind cluster, and run the e2e suite.
+
 ##@ Kustomize
 
 .PHONY: kustomize-set-image
--- a/Makefile
+++ b/Makefile
@@ -202,10 +202,6 @@ e2e-test: envtest kind-load-image kind-load-spark-image ## Run the e2e tests aga
 	@echo "Running e2e tests (deploy_method=$(DEPLOY_METHOD))..."
 	DEPLOY_METHOD=$(DEPLOY_METHOD) IMAGE_TAG=$(IMAGE_TAG) KUBECONFIG=$(KIND_KUBE_CONFIG) go test ./test/e2e/ -v -ginkgo.v -timeout 30m
 
-.PHONY: kind-e2e-test
-kind-e2e-test: IMAGE_TAG=local
-kind-e2e-test: kind-load-image e2e-test ## Build image, load into Kind cluster, and run the e2e suite.
-
 ##@ Kustomize
 
 .PHONY: kustomize-set-image
--- a/charts/spark-operator-chart/templates/webhook/deployment.yaml
+++ b/charts/spark-operator-chart/templates/webhook/deployment.yaml
@@ -75,8 +75,11 @@ spec:
         {{- with .Values.webhook.resourceQuotaEnforcement.enable }}
         - --enable-resource-quota-enforcement=true
         {{- end }}
-        {{- with .Values.webhook.sparkAllowedURLSchemes }}
+        {{- if .Values.webhook.urlSchemeValidation.enable }}
+        - --enable-url-scheme-validation=true
+        {{- with .Values.webhook.urlSchemeValidation.allowedSchemes }}
         - --spark-allowed-url-schemes={{ . | join "," }}
+        {{- end }}
         {{- end }}
         {{- if .Values.certManager.enable }}
         - --enable-cert-manager=true
--- a/charts/spark-operator-chart/templates/webhook/deployment.yaml
+++ b/charts/spark-operator-chart/templates/webhook/deployment.yaml
@@ -75,6 +75,9 @@ spec:
         {{- with .Values.webhook.resourceQuotaEnforcement.enable }}
         - --enable-resource-quota-enforcement=true
         {{- end }}
+        {{- with .Values.webhook.sparkAllowedURLSchemes }}
+        - --spark-allowed-url-schemes={{ . | join "," }}
+        {{- end }}
         {{- if .Values.certManager.enable }}
         - --enable-cert-manager=true
         {{- end }}
--- a/charts/spark-operator-chart/templates/webhook/deployment.yaml
+++ b/charts/spark-operator-chart/templates/webhook/deployment.yaml
@@ -75,6 +75,21 @@ spec:
         {{- with .Values.webhook.resourceQuotaEnforcement.enable }}
         - --enable-resource-quota-enforcement=true
         {{- end }}
+        {{- if .Values.webhook.urlSchemeValidation.enable }}
+        - --enable-url-scheme-validation=true
+        {{- with .Values.webhook.urlSchemeValidation.allowedSchemes }}
+        - --allowed-url-schemes={{ . | join "," }}
+        {{- end }}
+        {{- with .Values.webhook.urlSchemeValidation.allowedSchemesAnyHost }}
+        - --allowed-url-schemes-any-host={{ . | join "," }}
+        {{- end }}
+        {{- with .Values.webhook.urlSchemeValidation.allowedHosts }}
+        - --allowed-url-hosts={{ . | join "," }}
+        {{- end }}
+        {{- with .Values.webhook.urlSchemeValidation.allowedWildcardHosts }}
+        - --allowed-wildcard-url-hosts={{ . | join "," }}
+        {{- end }}
+        {{- end }}
         {{- if .Values.certManager.enable }}
         - --enable-cert-manager=true
         {{- end }}
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -388,6 +388,12 @@ webhook:
     # -- Specifies whether to enable the ResourceQuota enforcement for SparkApplication resources.
     enable: false
 
+  # -- URL schemes permitted in fetch-capable SparkApplication fields (spec.sparkConf values,
+  # spec.deps.*, spec.mainApplicationFile) at admission time, in addition to the always-allowed
+  # local schemes (schemeless, file://, local://). Any other URL scheme is rejected.
+  # Add schemes your workloads require (e.g. gs, s3a, hdfs).
+  sparkAllowedURLSchemes: []
+
   serviceAccount:
     # -- Specifies whether to create a service account for the webhook.
     create: true
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -388,11 +388,16 @@ webhook:
     # -- Specifies whether to enable the ResourceQuota enforcement for SparkApplication resources.
     enable: false
 
-  # -- URL schemes permitted in fetch-capable SparkApplication fields (spec.sparkConf values,
-  # spec.deps.*, spec.mainApplicationFile) at admission time, in addition to the always-allowed
-  # local schemes (schemeless, file://, local://). Any other URL scheme is rejected.
-  # Add schemes your workloads require (e.g. gs, s3a, hdfs).
-  sparkAllowedURLSchemes: []
+  urlSchemeValidation:
+    # -- Specifies whether to enable URL-scheme validation of fetch-capable SparkApplication
+    # fields (submit-time spec.sparkConf keys, spec.deps.*, spec.mainApplicationFile) at admission
+    # time. Opt-in and disabled by default so existing SparkApplications using remote schemes are
+    # not rejected on upgrade.
+    enable: false
+    # -- URL schemes permitted in fetch-capable SparkApplication fields when validation is enabled,
+    # in addition to the always-allowed local schemes (schemeless, file://, local://). Any other
+    # URL scheme is rejected. Add schemes your workloads require (e.g. gs, s3a, hdfs).
+    allowedSchemes: []
 
   serviceAccount:
     # -- Specifies whether to create a service account for the webhook.
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -413,14 +413,17 @@ webhook:
     # -- Specifies whether to enable URL-scheme validation of fetch-capable SparkApplication and
     # ScheduledSparkApplication template fields (submit-time spec.sparkConf keys, spec.deps.*,
     # spec.mainApplicationFile) at admission time. Opt-in and disabled by default so existing
-    # applications using remote schemes are not rejected on upgrade.
+    # applications using remote schemes are not rejected on upgrade. This reduces SSRF risk for
+    # the checked fields, but it is not a complete SSRF defense. Use restricted egress and
+    # least-privilege identities with this feature.
     enable: false
     # -- URL schemes permitted in fetch-capable SparkApplication fields when validation is enabled.
     # Remote URLs must also match allowedHosts or allowedWildcardHosts unless their scheme is in
     # allowedSchemesAnyHost.
     allowedSchemes: []
-    # -- Allowed URL schemes permitted to access any host, such as `s3a` where a VPC endpoint and
-    # network policy constrain reachable buckets. Each scheme must also appear in allowedSchemes.
+    # -- Allowed URL schemes permitted to access any host. Each scheme must also appear in
+    # allowedSchemes. Use only when separate network and service authorization policies constrain
+    # the endpoints and resources that the scheme can access.
     allowedSchemesAnyHost: []
     # -- Exact scheme-qualified URL hosts permitted in fetch-capable SparkApplication fields when
     # validation is enabled, such as `https://repo.example.com` or `s3a://bucket`. Empty host
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -409,6 +409,28 @@ webhook:
     # -- Specifies whether to enable the ResourceQuota enforcement for SparkApplication resources.
     enable: false
 
+  urlSchemeValidation:
+    # -- Specifies whether to enable URL-scheme validation of fetch-capable SparkApplication and
+    # ScheduledSparkApplication template fields (submit-time spec.sparkConf keys, spec.deps.*,
+    # spec.mainApplicationFile) at admission time. Opt-in and disabled by default so existing
+    # applications using remote schemes are not rejected on upgrade.
+    enable: false
+    # -- URL schemes permitted in fetch-capable SparkApplication fields when validation is enabled.
+    # Remote URLs must also match allowedHosts or allowedWildcardHosts unless their scheme is in
+    # allowedSchemesAnyHost.
+    allowedSchemes: []
+    # -- Allowed URL schemes permitted to access any host, such as `s3a` where a VPC endpoint and
+    # network policy constrain reachable buckets. Each scheme must also appear in allowedSchemes.
+    allowedSchemesAnyHost: []
+    # -- Exact scheme-qualified URL hosts permitted in fetch-capable SparkApplication fields when
+    # validation is enabled, such as `https://repo.example.com` or `s3a://bucket`. Empty host
+    # lists deny all remote URLs.
+    allowedHosts: []
+    # -- Scheme-qualified leftmost wildcard URL hosts permitted in fetch-capable SparkApplication
+    # fields when validation is enabled, such as `https://*.maven.apache.org`. Empty host lists
+    # deny all remote URLs.
+    allowedWildcardHosts: []
+
   serviceAccount:
     # -- Specifies whether to create a service account for the webhook.
     create: true
--- a/internal/webhook/sparkapplication_validator.go
+++ b/internal/webhook/sparkapplication_validator.go
@@ -19,6 +19,9 @@ package webhook
 import (
 	"context"
 	"fmt"
+	"net/url"
+	"reflect"
+	"slices"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -40,14 +43,31 @@ type SparkApplicationValidator struct {
 	client client.Client
 
 	enableResourceQuotaEnforcement bool
+	// allowedURLSchemes is the set of additional URL schemes permitted in fetch-capable spec
+	// fields (sparkConf values, deps.*, mainApplicationFile) beyond the always-allowed local
+	// schemes (schemeless, file://, local://). Any value whose URL scheme is not always-allowed
+	// and not in this set is rejected at admission time. Add schemes your workloads require
+	// (e.g. gs, s3a, hdfs).
+	allowedURLSchemes map[string]struct{}
 }
 
 // NewSparkApplicationValidator creates a new SparkApplicationValidator instance.
-func NewSparkApplicationValidator(client client.Client, enableResourceQuotaEnforcement bool) *SparkApplicationValidator {
+// allowedURLSchemes lists URL schemes permitted in fetch-capable spec fields (sparkConf values,
+// deps.*, mainApplicationFile) in addition to the always-allowed local schemes (schemeless,
+// file://, local://). Any other scheme is rejected.
+func NewSparkApplicationValidator(client client.Client, enableResourceQuotaEnforcement bool, allowedURLSchemes []string) *SparkApplicationValidator {
+	allowed := make(map[string]struct{}, len(allowedURLSchemes))
+	for _, s := range allowedURLSchemes {
+		s = strings.TrimSpace(strings.ToLower(s))
+		if s != "" {
+			allowed[s] = struct{}{}
+		}
+	}
 	return &SparkApplicationValidator{
 		client: client,
 
 		enableResourceQuotaEnforcement: enableResourceQuotaEnforcement,
+		allowedURLSchemes:              allowed,
 	}
 }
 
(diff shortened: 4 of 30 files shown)


charts/spark-operator-chart/values.yaml
```diff
@@ -349,6 +340,12 @@ webhook:
     # -- Specifies whether to enable the ResourceQuota enforcement for SparkApplication resources.
     enable: false
 
+  # -- URL schemes permitted in fetch-capable SparkApplication fields (spec.sparkConf values,
+  # spec.deps.*, spec.mainApplicationFile) at admission time, in addition to the always-allowed
+  # local schemes (schemeless, file://, local://). Any other URL scheme is rejected.
+  # Add schemes your workloads require (e.g. gs, s3a, hdfs).
+  sparkAllowedURLSchemes: []
+
```
tariq-hasan: As this is a breaking change for operators already deployed and using remote targets I am wondering if this should be backported or included as part of the next release.

Makefile
```diff
@@ -199,6 +176,10 @@ e2e-test: envtest ## Run the e2e tests against a Kind k8s instance that is spun
 	@echo "Running e2e tests (deploy_method=$(DEPLOY_METHOD))..."
 	DEPLOY_METHOD=$(DEPLOY_METHOD) IMAGE_TAG=$(IMAGE_TAG) go test ./test/e2e/ -v -ginkgo.v -timeout 30m
 
+.PHONY: kind-e2e-test
+kind-e2e-test: IMAGE_TAG=local
+kind-e2e-test: kind-load-image e2e-test ## Build image, load into Kind cluster, and run the e2e suite.
```
tariq-hasan: I am wondering if `kind-e2e-test` would be necessary now that https://github.com/kubeflow/spark-operator/pull/3004 has made `kind-load-image` a [dependency](https://github.com/kubeflow/spark-operator/blob/master/Makefile#L197-L201) for `e2e-test`.

internal/webhook/sparkapplication_validator.go
```diff
@@ -152,6 +172,159 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+	// (spark.jars.packages, spark.jars.excludePackages, spark.jars.ivySettings) hold Maven
+	// coordinates or Ivy XML paths that parse as invalid URL schemes and are exempt.
+	for key, value := range app.Spec.SparkConf {
+		if _, exempt := sparkConfKeysExempt[key]; exempt {
+			continue
+		}
+		field := fmt.Sprintf("spec.sparkConf[%q]", key)
+		if err := v.checkURLSchemes(field, strings.Split(value, ",")); err != nil {
+			return err
+		}
```
tariq-hasan: I am wondering if this check might be a bit too restrictive and if it would block many false positives as well.

For example, Spark configs such as `spark.eventLog.dir`, `spark.sql.warehouse.dir` and `spark.kubernetes.driverEnv.* / executorEnv.*` are not dereferenced by the operator at submission time and are instead leveraged during the runtime execution in the driver or executors.

The list of operator-dereferenced fields at submission time is rather small: the main application file, the `deps.*` lists, the pod-template-file keys, `spark.kubernetes.file.upload.path`, etc.

Would it be a worthwhile trade-off, in the interest of future-proofing this feature, to run a blanket check and block on every field with a scheme not in the allowlist?

tariq-hasan: I am also wondering if setting up this allowlist with a blanket check across all Spark configs could ironically lead to weakening the control and make it more permissive.

More specifically, if one wants to allow the `s3a://` scheme for a runtime config such as `spark.eventLog.dir` then this becomes an allowed scheme for all configs - both submission-time and runtime.

This naturally leads to a coupling between submission-time concerns and runtime concerns. We would want our focus to be specifically for submission-time operator-dereferenced fields yet runtime concerns dictate what goes in the allowlist and could end up enabling more permissive behavior on submission-time than what we may want.

Would that be an acceptable behavior for the feature?

tariq-hasan: In light of the above two comments I am thinking if what would be more feasible could be to have a specific set of spark config keys to validate. If not this approach then perhaps we could use a deny list with the broad values scan across all the spark configs.

internal/webhook/sparkapplication_validator.go
```diff
@@ -152,6 +172,159 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+
+// sparkConfKeysExempt are sparkConf keys whose comma-separated values are NOT dereferenced as
+// URLs: they hold Maven coordinates (groupId:artifactId:version) or Ivy XML paths resolved by
+// the Spark Ivy subsystem, not operator-fetched URLs. spark.jars.packages is the primary case:
+// a value like "org.postgresql:postgresql:42.7.3" parses as scheme "org.postgresql" but is not
+// a URL. spark.jars.excludePackages and spark.jars.ivySettings follow the same pattern.
+var sparkConfKeysExempt = map[string]struct{}{
+	"spark.jars.packages":        {},
+	"spark.jars.excludePackages": {},
+	"spark.jars.ivySettings":     {},
```
tariq-hasan: I'm thinking that `spark.jars.ivySettings` is redundant here since the [official docs](https://spark.apache.org/docs/latest/configuration.html) mention that only the file scheme is supported for this config and we already exempt this scheme under `alwaysAllowedURLSchemes`.

internal/webhook/sparkapplication_validator.go
```diff
@@ -152,6 +172,159 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+		// "//host/x" has an empty scheme but a host - it's a network path, not a local one,
+		// so don't let the empty-scheme exemption cover it.
+		if scheme == "" && u.Host != "" {
+			return fmt.Errorf("%s contains a scheme-relative URL with host %q which is not a local path: %q", field, u.Host, value)
+		}
+		return nil
+	}
+	if _, allowed := v.allowedURLSchemes[scheme]; !allowed {
+		return fmt.Errorf("%s contains a value with URL scheme %q which is not in the allowed list: %q", field, scheme, value)
+	}
```
tariq-hasan, line 545: This is more of a general UX issue but I have been observing that an error is returned on the first validation error. This could be an issue in that multiple URL scheme errors may take multiple tries to figure out. What can make it worse is that the map ordering is inherently non-deterministic so for multiple errors each time a different error would surface.

I am thinking if we want to just collect all the validation errors in a slice and return them at once (in a deterministic order) to ensure a better debugging experience and reproducible error messages.

tariq-hasan (commented):
Hi @vjanelle! I have left a few questions.

Two additional observations:
- I believe there's an inherent assumption here that the Go parsing logic is the same as the Spark-native URL parsing logic.
- There's no webhook validation on these configs for `ScheduledSparkApplication` so no validation will be done during apply time and the errors will be detected on the first actual run of the app.

internal/webhook/sparkapplication_validator.go
nabuskey: I agree that our focus is on submit-time configuration validations and not run-time configuration validations because we don't know what custom plugins users are using and their configurations values are. The goal of this PR is to restrict the operator from making requests to external services using its privileges. I think we should restrict validations to spark-submit script  configurations.
Additionally, this feature needs to be strictly opt-in and not break the current behaviour.

charts/spark-operator-chart/values.yaml
vjanelle: It's a security thing, tbh there should also be a CVE associated with it.

Makefile
vjanelle: Probably not anymore, I'll remove it.

internal/webhook/sparkapplication_validator.go
vjanelle: The scope of the vulnerability is around what the spark-submit process runs as, because the operator could fetch things that someone's user might not be able to - assuming they're allowed to submit as that user that is.

vjanelle: Is opt-in for security behaviours the best idea?

vjanelle: @tariq-hasan @nabuskey The reason for not having a specific list is having to maintain it forever - for instance log4j can take http urls as a configuration (Not actually sure that this would address that, need to add it as a test I think).

However for some keys like spark.ui.reverseProxyUrl, we would need to allow that by default (not sure these tests catch that either).

I'll look into the feasibility either way.

vjanelle: I can also add it so that it's configurable either way, default deny w/allow list or default allow w/deny list

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,6 +187,197 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+//
+// spark.jars / spark.files / spark.submit.pyFiles / spark.archives mirror spec.deps.{jars,files,
+// pyFiles,archives}. spark.jars.repositories mirrors spec.deps.repositories: in KUBERNETES cluster
+// mode (the operator's mode) spark-submit calls DependencyUtils.resolveMavenDependencies at submit
+// time (SparkSubmit.scala, the !isStandAloneCluster branch), which queries these repositories from
+// the operator pod to resolve --packages, so a remote repo URL is fetched with the operator's
+// principal. spark.kubernetes.file.upload.path is the destination the operator's spark-submit
+// uploads local deps to: BasicDriverFeatureStep/DriverCommandFeatureStep call
+// KubernetesUtils.uploadFileUri inside KubernetesDriverBuilder, which runs submit-side in
+// KubernetesClientApplication, again with the operator's credentials.
```
tariq-hasan: I'm actually thinking the Spark-internal details could be removed as these might drift over time from the actual source.

```suggestion
```

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,6 +187,197 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+	}
+	u, err := url.Parse(value)
+	if err != nil {
+		return []error{fmt.Errorf("%s contains a value that is not a valid URL: %q: %w", field, value, err)}
+	}
+	scheme := strings.ToLower(u.Scheme)
+	if slices.Contains(alwaysAllowedURLSchemes, scheme) {
+		// "//host/x" has an empty scheme but a host - it's a network path, not a local one,
+		// so don't let the empty-scheme exemption cover it.
+		if scheme == "" && u.Host != "" {
```
tariq-hasan: Just for defensiveness I am wondering if we want to make the check for all schemes under `alwaysAllowedURLSchemes`. Otherwise we allow hostnames that have either the `file://` or `local://` schemes.

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,6 +187,197 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+			continue
+		}
+		tag := strings.Split(f.Tag.Get("json"), ",")[0]
+		if tag == "" || tag == "-" || slices.Contains(depsURLFieldsExempt, tag) {
+			continue
+		}
+		fields = append(fields, depsURLField{index: i, label: "spec.deps." + tag})
+	}
+	return fields
+}
```
tariq-hasan, line 415: I'm wondering if the heavy machinery of reflections is overkill here - given that we are just doing this for.a small `Dependencies` struct.

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,6 +187,197 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+// local /tmp path it writes itself (submission.go driver/executorPodTemplateOption, appended after
+// the user's sparkConf so it wins), so a user value never reaches spark-submit; the keys stay here
+// so the check still bites if that override is ever removed or bypassed.
+var sparkConfURLKeys = map[string]struct{}{
+	"spark.jars":                                  {},
+	"spark.files":                                 {},
+	"spark.submit.pyFiles":                        {},
+	"spark.archives":                              {},
+	"spark.jars.repositories":                     {},
+	"spark.kubernetes.file.upload.path":           {},
```
tariq-hasan: I'm thinking probably it would be nice to follow the same pattern as [here](https://github.com/kubeflow/spark-operator/blob/master/internal/webhook/sparkconf_validator.go#L35-L47) and define the constants the same way.

charts/spark-operator-chart/templates/webhook/deployment.yaml
```diff
@@ -75,6 +75,12 @@ spec:
         {{- with .Values.webhook.resourceQuotaEnforcement.enable }}
         - --enable-resource-quota-enforcement=true
         {{- end }}
+        {{- if .Values.webhook.urlSchemeValidation.enable }}
+        - --enable-url-scheme-validation=true
+        {{- with .Values.webhook.urlSchemeValidation.allowedSchemes }}
+        - --spark-allowed-url-schemes={{ . | join "," }}
+        {{- end }}
+        {{- end }}
```
tariq-hasan, line 92: Just for consistency I'm thinking it would be nice to have `--allowed-url-schemes` instead of `--spark-allowed-url-schemes`.

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,6 +187,197 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+		if scheme == "" && u.Host != "" {
+			return []error{fmt.Errorf("%s contains a scheme-relative URL with host %q which is not a local path: %q", field, u.Host, value)}
+		}
+		return nil
+	}
+	if _, allowed := v.allowedURLSchemes[scheme]; !allowed {
+		return []error{fmt.Errorf("%s contains a value with URL scheme %q which is not in the allowed list: %q", field, scheme, value)}
+	}
 	return nil
 }
```
tariq-hasan, line 548: It could be better to use typed errors throughout as exhibited [here](https://github.com/kubeflow/spark-operator/blob/master/internal/webhook/sparkconf_validator.go#L25-L32).

tariq-hasan (commented):
Hi @vjanelle! I have added a few additional comments.

Also wanted to mention two additional notes on the approach here.
- The scheme probably is not enough to fully close the SSRF issue altogether. For example, Maven repos and other jar hosting by default need the https scheme so if the https scheme were to be enabled this would bring back the SSRF issue once more - given that harmful remote targets with the https scheme could potentially be leveraged with operator privileges. Perhaps it’s host/CIDR allowlisting that could be the mitigation that fully closes this gap.
- Also a secondary thing but the policy as it stands is not actually configurable per spec field so it’s a blanket allow or deny on any given scheme.

vjanelle:
@tariq-hasan do you have an example of the maven https requirements? I thought that was only just maven coordinates.

NVM, figured it out.

hack/url-scheme-parity/go.mod
```diff
@@ -0,0 +1,5 @@
+// Standalone module: isolates this verification harness from the operator module
```
vjanelle: @tariq-hasan hack to try to compare url schemes between spark and golang, can drop this but including it for review purposes

internal/webhook/sparkapplication_validator.go
vjanelle: We're always going to be chasing spark upstream I think, without a default enforcement regime somehow

internal/webhook/sparkapplication_validator.go
vjanelle: file and local are allowed explicitly, since they're in the pod.

internal/webhook/sparkapplication_validator.go
vjanelle: Given that the data is joined back in when submitted to spark-submit, there's potential for abuse (comma separated values) without additional pre-validation I think

internal/webhook/sparkapplication_validator.go
vjanelle: missed this one, implementing

hack/url-scheme-parity/go.mod
tariq-hasan: Makes sense. I think better to remove`hack/url-scheme-parity/go.mod` altogether from the PR.

internal/webhook/sparkapplication_validator.go
nabuskey: Is reflection necessary here? The API currently has five fetch-capable dependency fields. An explicit field list would make the security boundary easier to maintain. `validateDepsURLSchemes` already takes care of comma separated values?

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,9 +313,257 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+// executor templates are normally fetched later by the driver. The operator currently overwrites
+// both user values with local /tmp paths before spark-submit, so user URLs do not reach either
+// fetch path. Keep validating them in case that override is removed or bypassed.
+const (
+	sparkJarsConfigKey                     = "spark.jars"
+	sparkFilesConfigKey                    = "spark.files"
+	sparkSubmitPyFilesConfigKey            = "spark.submit.pyFiles"
+	sparkArchivesConfigKey                 = "spark.archives"
+	sparkJarsRepositoriesConfigKey         = "spark.jars.repositories"
+	sparkKubernetesFileUploadPathConfigKey = "spark.kubernetes.file.upload.path"
```
nabuskey, line 375: Hmm `spark.kubernetes.file.upload.path` is not safe. 

```yaml
 spec:
   mainApplicationFile: file:///var/run/secrets/kubernetes.io/serviceaccount/token
   sparkConf:
     spark.kubernetes.file.upload.path: s3a://not-supposed-to-be-here
```

Is this something we want to enforce and guard against? No, I don't think so. But we should document the implication and what people need to do to guard against it.

hack/url-scheme-parity/go.mod
tariq-hasan: Makes sense. I was thinking better to remove`hack/url-scheme-parity/go.mod` altogether from the PR.

tariq-hasan:
Hi @vjanelle! The scheme allowlist with the host policy seems to the correct shape. @nabuskey Would appreciate your thoughts on this as well.

At the same time though I am thinking it's ideal to postpone the host policy to a subsequent PR. The error sanitization could also go as a follow-up PR. That way we can reduce the scope to just having the scheme validation for now and are able to move this forward. What do you think? @vjanelle @nabuskey

hack/url-scheme-parity/go.mod
nabuskey: I think this should stay as review purposes only and be removed before merge.

nabuskey:
>The scheme allowlist with the host policy seems to the correct shape. 

Looks right to me. This PR is definitely comprehensive and took me a while to review, but I think it can go in as long as comments are resolved.

internal/webhook/sparkapplication_validator.go
vjanelle: @nabuskey just future proofing - any string arrays that were added in the future would be automatically scanned.

vjanelle: I can add a test as a guard, which would negate the reflection?

internal/webhook/sparkapplication_validator.go
vjanelle: You could export the token to an s3 bucket (that the operator can write to)

internal/webhook/sparkapplication_validator.go
```diff
@@ -156,9 +313,257 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+)
+
+var sparkConfURLKeys = map[string]struct{}{
+	sparkJarsConfigKey:                            {},
+	sparkFilesConfigKey:                           {},
+	sparkSubmitPyFilesConfigKey:                   {},
+	sparkArchivesConfigKey:                        {},
+	sparkJarsRepositoriesConfigKey:                {},
+	common.SparkKubernetesDriverPodTemplateFile:   {},
+	common.SparkKubernetesExecutorPodTemplateFile: {},
```
tariq-hasan, line 385: One quick question: we are restricting the list of allowed spark configuration keys. At the same time though we enable operators to configure the host policy.

But because the transport mechanism is not controlled can't this lead to a security bypass and weakening of these spark conf key controls?

For example `spark.jars` may be allowlisted to some S3 resources but `spec.hadoop.*` is not controlled so the actual S3A endpoint could be redirected to a harmful destination using `spark.hadoop.fs.s3a.endpoint`.

I am wondering if it makes sense to also have a small deny-set for transport config keys - `fs.s3a.endpoint`, `fs.s3a.bucket.*.endpoint`, `fs.gs.storage.root.url`, `fs.defaultFS`, `fs.s3a.aws.credentials.provider`, ABFS equivalents, etc. - that we check under both `spec.hadoopConf` and `spec.hadoop.*` entries in `sparkConf`.

charts/spark-operator-chart/values.yaml
```diff
@@ -395,6 +395,28 @@ webhook:
...
+    allowedSchemesAnyHost: []
+    # -- Exact scheme-qualified URL hosts permitted in fetch-capable SparkApplication fields when
+    # validation is enabled, such as `https://repo.example.com` or `s3a://bucket`. Empty host
+    # lists deny all remote URLs.
+    allowedHosts: []
+    # -- Scheme-qualified leftmost wildcard URL hosts permitted in fetch-capable SparkApplication
+    # fields when validation is enabled, such as `https://*.maven.apache.org`. Empty host lists
+    # deny all remote URLs.
+    allowedWildcardHosts: []
+
```
tariq-hasan, line 436: I'm wondering if we should be clear in the comments that host allowlisting is not a complete SSRF defence but only one part of a layered stack of defence. For example, redirects and DNS rebinding are outside the scope of what this operator can configure. At the same time an admin can configure NetworkPolicy and that would be the stronger security boundary ultimately.

internal/webhook/sparkapplication_validator.go
nabuskey: I think trying to enforce all these become too much. I honestly don't know how many settings there are and how they interact with each other.

I think it comes down to deciding where the operator's responsibilities end. How much do we hand hold end users and enforce our security policy? Configuration keys like `fs.s3a.endpoint, fs.s3a.bucket.*.endpoint` are very useful because it gives end users the flexibility to configure things specific to their needs at job level. We have always given users the flexibility and ease of use with this interface. 

My take on this is we should document the implications of these settings and steer end users to a more secure direction instead of enforcing it. Good thing about custom resources is that you can use policy engines like Kyverno against them too.

charts/spark-operator-chart/values.yaml
nabuskey: I agree. We need a documentation website overhaul. It's been on my back burner for a while...

internal/webhook/sparkapplication_validator.go
nabuskey: @vjanelle @tariq-hasan @roburishabh I would love your opinions on this.

tariq-hasan: Makes sense to me. It was unrealistic to have thought the operator can be made to close the gap fully.

vjanelle: My opinion basically is that if you turn this on, anything that could be an SSRF gets blocked and it's up to you to ensure that doesn't become a problem. And provide enough knobs that you can open it up, but not so many that it gets insane to configure.

There's a few things in this PR where we will need to make sure CRD changes don't add things, etc.

charts/spark-operator-chart/values.yaml
vjanelle: A good chunk of the problems will just be the error sanitisation tbh

nabuskey:
I think this is good to go after `hack/url-scheme-parity` code is removed.  @vjanelle

vjanelle:
@tariq-hasan @nabuskey sorry - missed the doc check, that's added now.

PR #3021, opened 2026-06-30
Scope codegen paths to workload-operator packages

### LINKED ISSUE #3022: Scope controller-gen and gen_register paths to explicit package trees
## Problem

The `Makefile` targets `manifests` and `generate` currently use `paths="./..."`, and `hack/update-codegen.sh` passes the repo root (`${SCRIPT_ROOT}`) to `gen_register`. This tells `controller-gen` and `kube_codegen` to recursively scan the entire repository tree.

This works today because spark-operator is a single Go module, but it becomes a problem as soon as any nested Go module is introduced under the repo root — a pattern that's increasingly common in the Kubernetes operator ecosystem for platform integration, plugin scaffolding, or multi-component builds.

When that happens:
- `controller-gen` does **not** respect `go.mod` boundaries — it walks into the nested module, picks up its types, and produces stray CRD/RBAC manifests under `config/`
- `gen_register` scans API types from the nested module and generates register boilerplate that doesn't belong to the parent operator
- `make verify-codegen` breaks because the generated output no longer matches the checked-in files

## Why this matters

- Organizations building platform layers commonly nest integration modules alongside the upstream operator to avoid maintaining a separate repository

Explicit path scoping is a low-cost defensive practice — it documents *what* the codegen is supposed to cover and prevents silent breakage if the repo structure evolves.

## Proposed change

**`Makefile`** — Replace `paths="./..."` with an explicit variable:

```makefile
WORKLOAD_OPERATOR_CODEGEN_PATHS := ./api/...;./cmd/...;./internal/...;./pkg/...

manifests: controller-gen
	$(CONTROLLER_GEN) ... paths="$(WORKLOAD_OPERATOR_CODEGEN_PATHS)" ...

generate: controller-gen manifests
	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="$(WORKLOAD_OPERATOR_CODEGEN_PATHS)"
```

**`hack/update-codegen.sh`** — Scope `gen_register` to the API directory:

```bash
kube::codegen::gen_register \
  --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.go.txt" \
  "${SCRIPT_ROOT}/api"
```

No functional change for the current single-module layout. All existing CRDs, RBAC, and generated code remain identical.

cc: @nabuskey @vara-bonthu

RobuRishabh:
## Purpose of this PR

Fixes #3022  `controller-gen paths="./..."` and `gen_register "${SCRIPT_ROOT}"` scan the entire repository tree. When nested Go modules are added under the repo root, codegen picks up their types, produces stray CRD/RBAC artifacts under `config/`, and breaks `make verify-codegen`.

**Proposed changes:**

- Add `WORKLOAD_OPERATOR_CODEGEN_PATHS` variable to scope `manifests` and `generate` targets to the four workload-operator package trees (`api`, `cmd`, `internal`, `pkg`)
- Scope `gen_register` in `hack/update-codegen.sh` to `api/` instead of the entire repo root

## Change Category

- [ ] Documentation update

### Rationale

No functional change for the current repo layout. This is a forward-compatibility guard for monorepo-style module additions — it ensures `controller-gen` and `gen_register` only process the workload operator's own packages and don't accidentally scan into sibling Go modules that may be added at the repo root.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,13 +109,15 @@ print-%: ; @echo $*=$($*)
 
 ##@ Development
 
+CONTROLLER_GEN_PATHS := ./api/...;./internal/...
+
 .PHONY: manifests
 manifests: controller-gen ## Generate CustomResourceDefinition, RBAC and WebhookConfiguration manifests.
-	$(CONTROLLER_GEN) crd:generateEmbeddedObjectMeta=true rbac:roleName=spark-operator-controller webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) crd:generateEmbeddedObjectMeta=true rbac:roleName=spark-operator-controller webhook paths="$(CONTROLLER_GEN_PATHS)" output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
 generate: controller-gen manifests ## Generate Go code and Python APIs.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="$(CONTROLLER_GEN_PATHS)"
 	$(MAKE) python-api
 
 .PHONY: update-crd
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -35,6 +35,7 @@ kube::codegen::gen_register \
 # Generate client code: client, lister, informer in client-go directory
 kube::codegen::gen_client \
   --with-watch \
+  --one-input-api api \
   --output-dir "${SCRIPT_ROOT}/pkg/client" \
   --output-pkg "${SPARK_OPERATOR_PKG}/pkg/client" \
   --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.go.txt" \
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -30,7 +30,7 @@ source "${CODEGEN_PKG}/kube_codegen.sh"
 # Generate register code
 kube::codegen::gen_register \
   --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.go.txt" \
-  "${SCRIPT_ROOT}"
+  "${SCRIPT_ROOT}/api"
 
 # Generate client code: client, lister, informer in client-go directory
 kube::codegen::gen_client \
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -20,15 +20,13 @@ set -o pipefail
 
 SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-_tmp="${SCRIPT_ROOT}/_tmp"
+_tmp="$(mktemp -d)"
 DIFF_ROOTS=("${SCRIPT_ROOT}/api" "${SCRIPT_ROOT}/pkg")
 
 cleanup() {
   rm -rf "${_tmp}"
 }
 trap "cleanup" EXIT SIGINT
-
-cleanup
 
 for root in "${DIFF_ROOTS[@]}"; do
   rel="${root#"${SCRIPT_ROOT}/"}"
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -20,9 +20,8 @@ set -o pipefail
 
 SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-DIFFROOT="${SCRIPT_ROOT}/pkg"
-TMP_DIFFROOT="${SCRIPT_ROOT}/_tmp/pkg"
 _tmp="${SCRIPT_ROOT}/_tmp"
+DIFF_ROOTS=("${SCRIPT_ROOT}/api" "${SCRIPT_ROOT}/pkg")
 
 cleanup() {
   rm -rf "${_tmp}"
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -109,13 +109,15 @@ print-%: ; @echo $*=$($*)
 
 ##@ Development
 
+WORKLOAD_OPERATOR_CODEGEN_PATHS := ./api/...;./cmd/...;./internal/...;./pkg/...
```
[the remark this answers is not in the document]
nabuskey: I think this is correct: https://book.kubebuilder.io/reference/controller-gen.html

hack/update-codegen.sh
tariq-hasan, line 42: I notice that we are scoping scheme registration to `/api` but we are not scoping client generation. This is despite the fact that both scheme registration and client generation are part of the code generation mechanism.

tariq-hasan, line 42: Add to that currently the diff for code generation only [checks](https://github.com/RobuRishabh/spark-operator/blob/1f665cee37e4dcbf3ff7c37ce20642be297a77bd/hack/verify-codegen.sh#L23-L25) it for `gen_client` and not for `gen_register`. I can infer this from the fact that the generated code for `gen_client` is written into `pkg/client` whereas the generated code for `gen_register` goes in `/api`.

tariq-hasan, line 42: We therefore have an asymmetry in the current code whereby we are diffing client generation but scoping scheme registration.

I would think the correct approach would be to diff both scheme registration in `/api` and client generation in `/pkg/client` as part of `verify-codegen.sh` and scope both `gen_register` and `gen_client` in `update-codegen.sh` to `/api`.

Makefile
```diff
@@ -109,13 +109,15 @@ print-%: ; @echo $*=$($*)
...
 .PHONY: manifests
 manifests: controller-gen ## Generate CustomResourceDefinition, RBAC and WebhookConfiguration manifests.
-	$(CONTROLLER_GEN) crd:generateEmbeddedObjectMeta=true rbac:roleName=spark-operator-controller webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) crd:generateEmbeddedObjectMeta=true rbac:roleName=spark-operator-controller webhook paths="$(WORKLOAD_OPERATOR_CODEGEN_PATHS)" output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
 generate: controller-gen manifests ## Generate Go code and Python APIs.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="$(WORKLOAD_OPERATOR_CODEGEN_PATHS)"
 	$(MAKE) python-api
```
tariq-hasan, line 121: I'm wondering if we want to rethink the `WORKLOAD_OPERATOR_CODEGEN_PATHS` naming for these paths. I have two points to make on this note:

- Technically code generation is done (almost) exclusively by the `kube::codegen::*` shell functions. We use two of these in the spark operator repo - `gen_register` and `gen_client`. These functions essentially read the `+k8s`, `+genclient` and `+groupName` markers and generate boilerplate code. In this case all these markers live in the `/api` package and we've hardcoded this package name in `hack/update-codegen.sh`. On the other hand manifest (and deepcopy) generation is done by the controller-gen tool by reading all the `+kubebuilder` markers. In this case these markers live in the `/api` and `/internal` packages. So one would expect these paths to be called "controller gen paths" as opposed to "code gen paths" - a naming that's reflective of the tool we leverage. But as it stands we are including them in `WORKLOAD_OPERATOR_CODEGEN_PATHS`.
- Spark operator as it stands does not have a native concept of a workload operator. So if we use the "workload operator" phrasing we might be intending meaning that is not currently supported in the project.

Because of these two points I would think that `CONTROLLER_GEN_PATHS` is a more accurate naming that reflects the scope and the intended usage of this variable.

tariq-hasan, line 121: On this note I also want to add that, other than the `+kubebuilder` markers in the `/api` and `/internal` packages that are read by the `controller-gen` tool, there also exist `+kubebuilder` markers in `cmd/operator/controller/start.go`, `cmd/operator/webhook/start.go`,  `config/crd/kustomization.yaml`, `config/samples/kustomization.yaml`, `pkg/certificate/suite_test.go`, `pkg/scheme/scheme.go` and `test/e2e/suite_test.go`. However, these other markers are read exclusively by the Kubebuilder CLI since these are of the form `+kubebuilder:scaffold:*`.

Therefore, I am thinking that we can scope the paths exclusively to `/api` and `/internal` packages, and leave out the `/cmd` and `/pkg` packages from the list.

Makefile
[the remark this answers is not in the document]
tariq-hasan: Actually both happen to be correct as per the [source code](https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/markers/parse.go#L413-L417) and [package docs](https://pkg.go.dev/sigs.k8s.io/controller-tools/pkg/markers#hdr-Definitions_and_Parsing). The semicolon just happens to be a legacy form.

tariq-hasan (commented):
Hi @RobuRishabh! I have added a few comments.

Makefile
RobuRishabh: Valid point, make sense. I will rename it to `CONTROLLER_GEN_PATHS`.

RobuRishabh: Thanks for the thorough review @tariq-hasan! All points addressed:

1. Renamed WORKLOAD_OPERATOR_CODEGEN_PATHS → CONTROLLER_GEN_PATHS
2. Reduced paths to ./api/...;./internal/... 
3. gen_client stays at "${SCRIPT_ROOT}" — I tried scoping it to api/ for consistency, but gen_client resolves API types relative to the input root, so it produced api/api/v1beta2 (double path). Unlike gen_register, it needs the repo root to correctly construct Go package paths. (Let me know what you think  of this)
4. Updated verify-codegen.sh to diff both api/ and pkg/ this closes the pre-existing gap where gen_register output drift was never caught.

hack/update-codegen.sh
RobuRishabh: Addressed. gen_register is scoped to api/. For gen_client, I tried scoping it to `api/` too, but it broke `gen_client` uses the input root to resolve package paths relative to it, so `api/` as root produced `api/api/v1beta2`. Left it at "${SCRIPT_ROOT}". Also updated `verify-codegen.sh` to diff both api/ and pkg/ per your suggestion.

hack/verify-codegen.sh
```diff
@@ -31,17 +30,25 @@ trap "cleanup" EXIT SIGINT
 
 cleanup
```
tariq-hasan: I suppose the bare `cleanup` can be removed as `mktemp -d` now creates a fresh temporary directory each time the script runs. `trap "cleanup" EXIT SIGINT` for final cleanup still seems to be useful before the script ends for cleanup purposes.

```suggestion
cleanup() {
  rm -rf "${_tmp}"
}
trap "cleanup" EXIT SIGINT
```

tariq-hasan (commented):
Just one comment. Otherwise /lgtm

tariq-hasan:
Actually, looking back I think this can be useful to anchor kube::codegen::gen_client: https://github.com/kubernetes/code-generator/blob/master/kube_codegen.sh#L502-L505.

So something like

```sh
kube::codegen::gen_client \
  --with-watch \
  --one-input-api api \
  --output-dir "${SCRIPT_ROOT}/pkg/client" \
  --output-pkg "${SPARK_OPERATOR_PKG}/pkg/client" \
  --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.go.txt" \
  "${SCRIPT_ROOT}"
  ```
  
  might do the trick I think.

tariq-hasan:
Looks good now I suppose. 👍

RobuRishabh:
> Looks good now I suppose. 👍 

Thank you for all the help. I learned new topics through this.

RobuRishabh:
@nabuskey can you please give this PR a final review.

nabuskey (approved):
/approve

PR #3027, opened 2026-07-05
fix(chart): grant webhook leader election event permissions

### LINKED ISSUE #3024: Helm chart webhook Role missing events permission for leader election
### What happened?

- [x] I have searched the open/closed issues and my issue is not listed.

When `webhook.leaderElection.enable=true` (the default) and the release namespace is **not** listed in `spark.jobNamespaces`, the webhook Role in the operator namespace gets lease permissions but **not** `events` create/update/patch. controller-runtime leader election then logs RBAC errors when recording `LeaderElection` events.

The controller Role template already handles this case: when the release namespace is not in `spark.jobNamespaces`, it falls back to an `events` rule (see `templates/controller/rbac.yaml`). The webhook Role template (`templates/webhook/rbac.yaml`) has no equivalent fallback.

We hit this with chart **2.5.1**, release namespace `zendesk-spark-operator`, and explicit `spark.jobNamespaces` (operator namespace excluded; scoped via `spark.jobNamespaceSelector` at runtime).

Example log:

```
events is forbidden: User "system:serviceaccount:zendesk-spark-operator:spark-operator-v2-webhook" cannot create resource "events" in API group "" in the namespace "zendesk-spark-operator"
```

Leader election still works via leases; the error is noisy and suggests incomplete RBAC.

### Reproduction Code

1. Install chart 2.5.1 with defaults except:
   - `webhook.leaderElection.enable: true`
   - `webhook.replicas: 2`
   - `spark.jobNamespaces` set to application namespaces only (release namespace **not** included)
   - `spark.jobNamespaceSelector` used for runtime scoping
2. Start webhook pods.
3. Observe forbidden errors when a replica acquires the leader lease.

### Expected behavior

Webhook Role in the release namespace should grant `events` create/update/patch when leader election is enabled, matching the controller Role fallback pattern.

Suggested template change in `templates/webhook/rbac.yaml` after the leader-election lease rules:

```yaml
{{- else if .Values.webhook.leaderElection.enable }}
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - update
  - patch
{{- end }}
```

(parallel to `templates/controller/rbac.yaml` lines 106-117)

### Actual behavior

Webhook Role only includes secrets + leases. Leader election event recording fails with forbidden errors.

### Environment & Versions

- Kubernetes Version: 1.28+ (EKS)
- Spark Operator Version: 2.5.1 (Helm chart)
- Apache Spark Version: N/A (operator RBAC only)

### Additional context

Kustomize manifests in `config/webhook/role.yaml` also omit `events`; a prior kustomize PR added events for leader election but Helm parity was called out as follow-up. There is no Helm values knob (e.g. `webhook.rbac.extraRules`) to work around this without post-render patches.

### Impacted by this bug?

immanuwell:
## Purpose of this PR

Fixes #3024.

Add `events` create/update/patch to the webhook Role when leader election is enabled.
Keep the kustomize webhook Role in sync with the default deployment.
Add Helm tests for enabled and disabled leader election.

## Change Category

- [ ] Documentation update

### Rationale

The chart enables webhook leader election by default. `client-go` records `LeaderElection` events when a pod acquires or releases the lease, so the missing `events` verbs cause RBAC errors in the normal startup path. This is easy to hit, even with 1 replica, since the first lease acquire records an event too. No weird quota edge here, just a plain default path bug, pretty noisy in logs.

## Checklist

### Additional Notes

Repro:
1. Install the chart with `webhook.leaderElection.enable=true`.
2. Keep the release namespace out of `spark.jobNamespaces`.
3. Start the webhook and check logs.
4. Before this fix you get `events is forbidden ... cannot create resource "events"`.

### DIFF
--- a/charts/spark-operator-chart/templates/webhook/rbac.yaml
+++ b/charts/spark-operator-chart/templates/webhook/rbac.yaml
@@ -84,6 +84,16 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 rules:
+{{- if .Values.webhook.leaderElection.enable }}
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - update
+  - patch
+{{- end }}
 - apiGroups:
   - ""
   resources:
--- a/charts/spark-operator-chart/tests/webhook/rbac_test.yaml
+++ b/charts/spark-operator-chart/tests/webhook/rbac_test.yaml
@@ -85,6 +85,17 @@ tests:
           kind: Role
           name: spark-operator-webhook
           namespace: spark-operator
+      - contains:
+          path: rules
+          content:
+            apiGroups:
+              - ""
+            resources:
+              - events
+            verbs:
+              - create
+              - update
+              - patch
 
   - it: Should create role and rolebinding for webhook in release namespace
     documentIndex: 3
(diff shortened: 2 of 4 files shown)


nabuskey:
@RobuRishabh can you take a look please?

RobuRishabh (changes requested):
@immanuwell Thank you for the contribution. 
1. Tests are good. Minor suggestion: consider adding a third test case where `leaderElection.enable=true` AND release namespace is in `jobNamespaces`, to verify events and policyRules coexist without conflict.
2. Two CI checks are currently failing:
-  **`build-spark-operator`** (unit tests) 
-  **`kustomize-lint`**

Both fail on `TestKustomizeBuild/WebhookRole` in `test/kustomize/kustomize_build_test.go`.

So the Kustomize manifests are designed to stay in parity with Helm -- the Go-based kustomize build tests enforce that the rendered output matches the expected RBAC shape. Right now, the existing test asserts that the kustomize webhook Role should **not** contain `events`, but your PR adds `events` to `config/webhook/role.yaml`. You need to update this assertion to match the new expected behavior.

In `test/kustomize/kustomize_build_test.go`, change line 238-239 from:
```go
assert.True(t, rulesHaveResource(role.Rules, "events"),
    "webhook Role should have 'events' for leader election event recording")
```
This aligns the kustomize build test with the intended change in `config/webhook/role.yaml`. Once you push that update, both CI checks should pass.

immanuwell:
@RobuRishabh addressed your suggestions, please take a look

RobuRishabh:
@nabuskey Could you please approve the CI workflow runs for this PR? The author has addressed all the review feedback once CI passes, this is LGTM from my side. You can also take a look.

RobuRishabh (approved):
/lgtm

PR #3029, opened 2026-07-07
test(e2e): add ScheduledSparkApplication workload e2e test

sahana-sreeram:
This test adds e2e functionality confirming that a ScheduledSparkApplication actually completes a real run on its configured schedule. It runs a workload for the ScheduledSparkApplication CRD instead of just creating the CRD. It waits for the cron schedule to fire, verifies the spawned child SparkApplication completes, and then checks the driver logs for the expected output.

## Purpose of this PR

**Proposed changes:**

- Add `test/e2e/scheduledsparkapplication_test.go`, a new e2e test that exercises the full ScheduledSparkApplication scheduling lifecycle

## Change Category

- [ ] Documentation update

### Rationale

This test adds missing coverage to confirm that a ScheduledSparkApplication produces a completed, correct Spark run on its configured schedule, adding to the existing unit-level coverage of controller logic.

## Checklist

- [ ] I have updated documentation accordingly (N/A)

### Additional Notes

Verified locally: `make e2e-test` passing 19/19, including both specs in this new file.

### DIFF
--- a/test/e2e/scheduledsparkapplication_test.go
+++ b/test/e2e/scheduledsparkapplication_test.go
@@ -1,0 +1,152 @@
+package e2e_test
+
+import (
+	"bufio"
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
+	"github.com/kubeflow/spark-operator/v2/pkg/common"
+	"github.com/kubeflow/spark-operator/v2/pkg/util"
+)
+
+var _ = Describe("Example ScheduledSparkApplication", func() {
+	Context("spark-pi-scheduled", func() {
+		ctx := context.Background()
+		path := os.Getenv("SCHEDULED_APP_YAML")
+		if path == "" {
+			path = filepath.Join("..", "..", "examples", "spark-pi-scheduled.yaml")
+		}
+		var scheduledApp *v1beta2.ScheduledSparkApplication
+
+		BeforeEach(func() {
+			By("Parsing ScheduledSparkApplication from file")
+			scheduledApp = &v1beta2.ScheduledSparkApplication{}
+			file, err := os.Open(path)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
+			defer func() { Expect(file.Close()).To(Succeed()) }()
+
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 100)
+			Expect(decoder).NotTo(BeNil())
+			Expect(decoder.Decode(scheduledApp)).NotTo(HaveOccurred())
+
+			// Use GenerateName so each spec gets a unique CR name,
+			// avoiding AlreadyExists when CLEANUP=false skips deletion.
+			scheduledApp.GenerateName = scheduledApp.Name + "-"
+			scheduledApp.Name = ""
+
+			// Override schedule to fire quickly for testing.
+			scheduledApp.Spec.Schedule = "@every 1m"
+
+			By("Creating ScheduledSparkApplication")
+			Expect(k8sClient.Create(ctx, scheduledApp)).To(Succeed())
+		})
+
+		AfterEach(func() {
+			if strings.EqualFold(os.Getenv("CLEANUP"), "false") && CurrentSpecReport().Failed() {
+				return
+			}
+
+			key := types.NamespacedName{Namespace: scheduledApp.Namespace, Name: scheduledApp.Name}
+			if err := k8sClient.Get(ctx, key, scheduledApp); err == nil {
+				By("Deleting ScheduledSparkApplication")
+				Expect(k8sClient.Delete(ctx, scheduledApp)).To(Succeed())
+			}
+
+			By("Verifying child SparkApplications are cascade-deleted via owner references")
+			Eventually(func(g Gomega) {
+				appList := &v1beta2.SparkApplicationList{}
+				g.Expect(k8sClient.List(ctx, appList,
+					client.InNamespace(scheduledApp.Namespace),
+					client.MatchingLabels{common.LabelScheduledSparkAppName: scheduledApp.Name},
+				)).To(Succeed())
+				g.Expect(appList.Items).To(BeEmpty())
+			}).WithTimeout(WaitTimeout).WithPolling(PollInterval).Should(Succeed())
+		})
+
+		It("Should reach Scheduled state and populate NextRun", func() {
+			key := types.NamespacedName{Namespace: scheduledApp.Namespace, Name: scheduledApp.Name}
+
+			By("Waiting for ScheduledSparkApplication to reach Scheduled state")
+			Eventually(func(g Gomega) {
+				app := &v1beta2.ScheduledSparkApplication{}
+				g.Expect(k8sClient.Get(ctx, key, app)).To(Succeed())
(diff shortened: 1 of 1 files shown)


test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/kubeflow/spark-operator/v2/pkg/util"
+)
+
+var _ = Describe("Example ScheduledSparkApplication", func() {
+	Context("spark-pi-scheduled", func() {
+		ctx := context.Background()
+		path := os.Getenv("SCHEDULED_APP_YAML")
+		if path == "" {
+			path = filepath.Join("..", "..", "examples", "spark-pi-scheduled.yaml")
+		}
```
RobuRishabh: It should be simplified to match the existing convention:
```
path := filepath.Join("..", "..", "examples", "spark-pi-scheduled.yaml")
```
We don't need `SCHEDULED_APP_YAML` env var, its likely a carry-over. We haven't set `SCHEDULED_APP_YAML` anywhere

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
+package e2e_test
```
RobuRishabh, line 17: Every file in this repo has the standard Apache 2.0 license header. Please add the copyright header at the top of the file.

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+			// Override schedule to fire quickly for testing.
+			scheduledApp.Spec.Schedule = "@every 1m"
+
+			By("Creating ScheduledSparkApplication")
+			Expect(k8sClient.Create(ctx, scheduledApp)).To(Succeed())
+		})
+
+		AfterEach(func() {
+			if strings.EqualFold(os.Getenv("CLEANUP"), "false") && CurrentSpecReport().Failed() {
```
RobuRishabh: The `CLEANUP=false` conditional cleanup skip is not used by any existing e2e test in the upstream repo. Every `AfterEach` in `sparkapplication_test.go` cleans up unconditionally. probably remove CLEANUP Environment Variable Guard

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
+			defer func() { Expect(file.Close()).To(Succeed()) }()
+
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 100)
+			Expect(decoder).NotTo(BeNil())
+			Expect(decoder.Decode(scheduledApp)).NotTo(HaveOccurred())
+
+			// Use GenerateName so each spec gets a unique CR name,
+			// avoiding AlreadyExists when CLEANUP=false skips deletion.
```
RobuRishabh: With the CLEANUP guard removed in my other comment, the comment's rationale no longer applies. GenerateName is still valuable because this Context has two It specs sharing one BeforeEach, if the first spec's AfterEach fails to delete, GenerateName prevents the second spec from colliding.
Replace the comment to:
```
// Use GenerateName so each spec gets a unique CR name,
// preventing collisions if a prior AfterEach fails to delete.
```

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
...
+			key := types.NamespacedName{Namespace: scheduledApp.Namespace, Name: scheduledApp.Name}
+
+			By("Waiting for a child SparkApplication to be created")
+			var childName string
+			Eventually(func(g Gomega) {
+				app := &v1beta2.ScheduledSparkApplication{}
+				g.Expect(k8sClient.Get(ctx, key, app)).To(Succeed())
+				g.Expect(app.Status.LastRunName).NotTo(BeEmpty())
+				childName = app.Status.LastRunName
+			}).WithTimeout(2 * time.Minute).WithPolling(PollInterval).Should(Succeed())
```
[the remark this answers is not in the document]
RobuRishabh: agree here, replace it with
```
}).WithTimeout(WaitTimeout).WithPolling(PollInterval).Should(Succeed())
```

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
...
+			for scanner.Scan() {
+				if strings.Contains(scanner.Text(), "Pi is roughly 3") {
+					foundPi = true
+					break
+				}
+			}
+			Expect(scanner.Err()).NotTo(HaveOccurred())
+			Expect(foundPi).To(BeTrue(), "expected driver logs to contain 'Pi is roughly 3'")
+		})
+	})
```
RobuRishabh, line 170: The streaming `bufio.Scanner` approach for reading driver logs is inconsistent with every other e2e test in the repo. All existing tests in `sparkapplication_test.go` use the simpler `.Do(ctx).Raw()` + `strings.Contains` pattern. The streaming approach also introduces the `bufio.Scanner` 64KB default token size limitation, which could cause false failures on long log lines.
Replace it with
```
driverPodName := util.GetDriverPodName(childApp)
bytes, err := clientset.CoreV1().Pods(scheduledApp.Namespace).GetLogs(driverPodName, &corev1.PodLogOptions{}).Do(ctx).Raw()
Expect(err).NotTo(HaveOccurred())
Expect(bytes).NotTo(BeEmpty())
Expect(strings.Contains(string(bytes), "Pi is roughly 3")).To(BeTrue())
```

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,152 @@
+package e2e_test
+
+import (
+	"bufio"
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
```
RobuRishabh, line 24: after fixing all the comments I think some import will become unused like bufio, io, k8s.io/utils/ptr, os. PLease check thoroughly.

RobuRishabh (changes requested):

sahana-sreeram:
Thanks for the review @RobuRishabh ! Addressed all above comments in [1efe149](https://github.com/kubeflow/spark-operator/pull/3029/commits/1efe1496d1c3e387493345e3304bbddf7881d850), confirmed e2e tests still pass

RobuRishabh:
@nabuskey Could you please run the CI here. If CI passes its ready to go in, its lgtm

nabuskey (commented):
Thanks for the PR! So this tests just one run of scheduled app? For an E2E test, I think it makes sense to test and make sure at least two tests run at the expected interval.

sahana-sreeram:
@nabuskey, definitely makes sense. Fixed in [0497190](https://github.com/kubeflow/spark-operator/pull/3029/commits/0497190412ab76cff200f680dd1af93bb3f1e31f). I've added extra steps to wait for a second run, verify it doesn't start before the schedule interval has elapsed, and wait for the second run to complete, which should verify E2E behavior. Please lmk in case of any other changes to make!

test/e2e/scheduledsparkapplication_test.go
```diff
@@ -0,0 +1,171 @@
...
+				g.Expect(app.Status.LastRunName).NotTo(Equal(childName))
+				secondChildName = app.Status.LastRunName
+			}).WithTimeout(WaitTimeout).WithPolling(PollInterval).Should(Succeed())
+
+			By("Verifying the second run starts at correct interval after the first run")
+			secondChildKey := types.NamespacedName{Namespace: scheduledApp.Namespace, Name: secondChildName}
+			secondChildApp := &v1beta2.SparkApplication{}
+			Expect(k8sClient.Get(ctx, secondChildKey, secondChildApp)).To(Succeed())
+			actualInterval := secondChildApp.CreationTimestamp.Sub(childApp.CreationTimestamp.Time)
+			Expect(actualInterval).To(BeNumerically(">=", expectedInterval),
```
nabuskey:
```suggestion
			Expect(actualInterval).To(BeNumerically(">=", expectedInterval - 2*time.Second)),
```

Because of precision and rounding, it may not satisfy `>=`. I would add a bit of leniency.

RobuRishabh (approved):

nabuskey (approved):
/approve

PR #3031, opened 2026-07-07
test(e2e): add SparkConnect query e2e test

sahana-sreeram:
This test adds E2E coverage for SparkConnect, confirming that it produces a correct result for a client query, extending beyond unit-level coverage of the controller's logic. It opens a PySpark Connect session against a running SparkConnect server, executes a dataframe query, and validates that the returned row count and computed values are correct.

## Purpose of this PR

**Proposed changes:**

- Add test/e2e/sparkconnect_query_test.go, an e2e test that actually executes a PySpark Connect
  query against a SparkConnect server and validates the computed result
## Change Category

- [ ] Documentation update

### Rationale

The current SparkConnect test only verifies that the server and executor pods reach a ready state, but this one executes an actual client query against a live server and validates the computed result. Since the new test performs every check the old one did as a precondition before running the query, `sparkconnect_test.go` is now redundant and has been removed as a part of this PR.

## Checklist

### Additional Notes

Verified locally: `make e2e-test` passing 19/19, including specs in this new file.


test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
+package e2e_test
```
RobuRishabh, line 17: please add the copyright header

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
...
+			conn = &v1alpha1.SparkConnect{}
+			Expect(decoder.Decode(conn)).NotTo(HaveOccurred())
+			conn.Name = "spark-connect-query"
+
+			By("Creating SparkConnect")
+			Expect(k8sClient.Create(ctx, conn)).To(Succeed())
+		})
+
+		AfterEach(func() {
+			if strings.EqualFold(os.Getenv("CLEANUP"), "false") && CurrentSpecReport().Failed() {
```
RobuRishabh: Remove the CLEANUP env var guard. No existing upstream e2e test uses this pattern

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
...
+			Expect(controllerutil.SetOwnerReference(conn, clientPod, k8sClient.Scheme())).To(Succeed())
+			Expect(k8sClient.Create(ctx, clientPod)).To(Succeed())
+
+			By("Waiting for PySpark client pod to be ready")
+			Eventually(func() bool {
+				key := types.NamespacedName{Namespace: conn.Namespace, Name: clientPod.Name}
+				pod := &corev1.Pod{}
+				if err := k8sClient.Get(ctx, key, pod); err != nil {
+					return false
+				}
```
RobuRishabh: The PySpark client pod is created and polled for readiness, but never used. Both kubectl exec commands (lines 153 and 172) target serverPodName, not the client pod. I think we can remove the entire client pod block (lines 122-148). This also removes the metav1 and controllerutil imports.

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
...
+			Expect(output).To(ContainSubstring("VALIDATE=14"),
+				"expected doubled value of 14 for id=7, got: %s", output)
+		})
+	})
+})
+
+func runCommand(name string, args ...string) (string, error) {
+	out, err := exec.Command(name, args...).CombinedOutput()
+	return string(out), err
+}
```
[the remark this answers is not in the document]
RobuRishabh: agree here

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
...
+		ctx := context.Background()
+		path := filepath.Join("..", "..", "examples", "sparkconnect", "spark-connect.yaml")
+
+		var conn *v1alpha1.SparkConnect
+
+		BeforeEach(func() {
+			By("Parsing SparkConnect from file")
+			file, err := os.Open(path)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
```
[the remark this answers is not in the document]
RobuRishabh: agree with copilot here

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,191 @@
+package e2e_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
```
RobuRishabh, line 25: After applying changes that I asked, we can remove unused imports: os, strings, metav1, controllerutil.

RobuRishabh (changes requested):
@alimaredia I have a question here from you, please guide us here, can we consider keeping `sparkconnect_test.go` as a fast smoke test. The new test depends on `pip install` from PyPI inside the cluster, which can fail due to network restrictions in CI. If pip fails, all SparkConnect e2e coverage is lost.

Alternatively, split the new test into two `It` blocks: one for readiness checks (fast, no external deps) and one for the query (depends on pip).

sahana-sreeram:
Fixes made in [3802562](https://github.com/kubeflow/spark-operator/pull/3031/commits/3802562619baad50d22dc7aa46d2a5ef9e94b4ab)

test/e2e/sparkconnect_query_test.go
sahana-sreeram: Fixed the client pod to actually be used, simulating real E2E Spark Connect - remote connectivity to Spark cluster via a thin client. Both kubectl exec calls were targeting serverPodName instead of clientPod.Name, so the client pod was created and never touched. Now it installs pyspark-client==4.0.1 (a minimal Spark Connect client) and runs the query from there, connecting to the server over the network via the Service — a real external client instead of the server talking to itself. Fixed in [91c6132](https://github.com/kubeflow/spark-operator/pull/3031/commits/91c613220fc9d09e70f2b157b150b8f996b6e800)

sahana-sreeram:
> @alimaredia I have a question here from you, please guide us here, can we consider keeping `sparkconnect_test.go` as a fast smoke test. The new test depends on `pip install` from PyPI inside the cluster, which can fail due to network restrictions in CI. If pip fails, all SparkConnect e2e coverage is lost.
> 
> Alternatively, split the new test into two `It` blocks: one for readiness checks (fast, no external deps) and one for the query (depends on pip).

Happy to re-add sparkconnect_test.go if that's preferred. The first three steps of the new test (server pod ready, executor pods ready, Service created) already do everything the `sparkconnect_test.go` did, so nothing would be lost in the path by removing it. The only real gap could be if pip install fails mid-test, which might be a CI/ network-reliability question rather than in the SparkConnect E2E. Open to whichever is preferred!

RobuRishabh:
@nabuskey Could you run the CI for this PR as well. Thank You

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,205 @@
...
+			clientPod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "pyspark-client-",
+					Namespace:    conn.Namespace,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:    "pyspark-client",
+							Image:   "quay.io/fedora/python-312:latest",
```
nabuskey: Any reason you want to use `quay.io/fedora/python-312` instead of an official spark image? If we use an official image, we can just run one of examples instead of creating our own.  Also we need to use immutable tags for all container image references.

sahana-sreeram: Thank you for the review! I've switched the client pod to the official Spark image and corrected the image tags. I kept the Python client script because the test is meant to simulate a user workflow of spinning up a separate client, connecting to the server through the service, and running a PySpark query - which would be a different testing client stack than the built-in Scala examples.

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,208 @@
...
+			clientPod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "pyspark-client-",
+					Namespace:    conn.Namespace,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:    "pyspark-client",
+							Image:   "docker.io/library/spark:4.0.1",
```
nabuskey: We need to avoid using mutable tags. Let's use a hash like this: https://github.com/kubeflow/spark-operator/blob/master/Dockerfile

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,208 @@
...
+				if err := k8sClient.Get(ctx, key, pod); err != nil {
+					return false
+				}
+				return util.IsPodReady(pod)
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(BeTrue())
+
+			By("Installing PySpark Connect dependencies")
+			output, err := runCommand(
+				"kubectl", "exec", "-n", conn.Namespace, clientPod.Name, "--",
+				"bash", "-c", "export HOME=/tmp && pip install --quiet --disable-pip-version-check pandas pyarrow grpcio grpcio-status",
```
nabuskey: With the spark image used, we shouldn't need to install packages like this correct? I would really like to avoid installing version unlocked packages like this in E2E, especially in Python. The less we install things, the better.

nabuskey:
Thinking about this a bit more, we should avoid creating pods with an image like that. It pulls the image every time with no caching used because nodes run in Kind. So either run the query locally or load it.

sahana-sreeram:
[0c8648f](https://github.com/kubeflow/spark-operator/pull/3031/commits/0c8648f611ef54c75776a3a8b871e320d25fe673) - fixed the client pod to reuse the operator's own controller image instead of a separate one, since it's already loaded into Kind by `make e2e-test`, so nothing new to pull or maintain.

For the Python side, looks like we'd still have to install the SparkConnect dependencies. From SparkConnect's docs, these deps are hard requirements for pyspark.sql.connect regardless of the query. I've stabilized that by pinning exact versions instead of leaving them unlocked!

To avoid installs entirely, the test would need to use the SparkConnect JVM/Scala client, but keeping this test Python would mirror how users/data scientists would likely use Spark Connect.

RobuRishabh:
@sahana-sreeram can you sign-off commit.

test/e2e/sparkconnect_query_test.go
```diff
@@ -0,0 +1,220 @@
...
+				return util.IsPodReady(pod)
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(BeTrue())
+
+			By("Installing PySpark Connect dependencies")
+			output, err := runCommand(
+				"kubectl", "exec", "-n", conn.Namespace, clientPod.Name, "--",
+				"sh", "-c",
+				"export HOME=/tmp && python3 -m pip install --quiet --disable-pip-version-check --no-cache-dir "+
+					"pandas==2.2.3 pyarrow==18.1.0 grpcio==1.68.1 grpcio-status==1.68.1 googleapis-common-protos==1.66.0",
+			)
```
RobuRishabh: @nabuskey, wanted to follow up on your pip comment ([discussion](https://github.com/kubeflow/spark-operator/pull/3031#discussion_r3590953216)) before we send @sahana-sreeram back for another round.

The PR still runs `pip install` inside the client pod at test time. to fix that, wanted your call:

**1. Pre-built client image + load into Kind** (your "load it" idea)
Build a small Spark client image with deps baked in, `kind load` it like we do for the operator, use `PullNever` no pip during the test.
- Pros: No PyPI in cluster; realistic pod → Service → server path
- Cons: Extra Dockerfile + CI step

**2. Run the query from the CI host** (your "run locally" idea)
Install deps on the runner, port-forward the Connect service, run Python from there.
- Pros: No pip in cluster; simpler test
- Cons: Doesn't test an in-cluster client; pip still runs on the runner

**3. JVM/Scala Connect client**
Replace the inline PySpark client with a JVM/Scala Spark Connect client. For operator e2e we don't need Python coverage  we only need to prove the Connect endpoint accepts a remote session and returns correct results. Keep the in-cluster client → Service → server path, and rebase onto latest `master`.
- Pros: No pip at all; simplest path, Spark image already has Java/Scala
- Cons: Doesn't mirror a PySpark user workflow (fine for our purposes)

A or B avoids pip at test runtime. C avoids pip entirely.

Everything else in the PR looks reasonable (operator image reuse, remote client via Service, pinned versions).

@sahana-sreeram still outstanding from earlier review: rebase onto latest `master` (Spark 4.0.4). Splitting readiness/query into separate `It` blocks is still open depending on which approach we take.

@nabuskey which option do you want @sahana-sreeram to go with? Once you confirm, we'll ask her to implement that.

RobuRishabh (changes requested):

test/e2e/sparkconnect_query_test.go
tariq-hasan: I'm actually thinking avoiding PyPI dependency would help avoid flakiness caused by external dependencies during CI. In addition, as already mentioned before, the downloads would be expensive in the sense that they would run on every CI. So option 3 looks to be the most feasible one from my understanding.

The issue with Option 1 is the additional image creation pipeline and with Option 2 it's that the actual client-to-server path is bypassed altogether.

Will hold off on @nabuskey's thoughts.

nabuskey: Yes, avoiding PyPI is ideal for two reasons:
1. Avoid CI failures because of PyPI. We do not have dependencies on PyPI in the operator itself. I'd hate to see it fail like this.
2. Security. Pulling packages without pinning to hashes at runtime is not a good idea. Considering what has been happening in JS / Python ecosystems with regard to supply chain attacks, I certainly do not want this repo to become an attack surface somehow.

So what exactly is missing in the base image that is not allowing us to use spark connect client?

tariq-hasan: If JVM client is the intention here then probably the [base](https://github.com/kubeflow/spark-operator/blob/master/Dockerfile#L17) image may have everything we need I suppose - but it would be good to test it out to confirm.

I suppose the real crux of the PR here is to expand coverage beyond simple reconciliation and readiness checks to full connectivity checks so the simplest path - the JVM client - would probably do in this case. The overhead of the PyPI approach does not seem justifiable (even though that may be the more common use case for interactive workloads in general).

RobuRishabh: @sahana-sreeram can you please verify the operator base Spark image ([Dockerfile](https://github.com/kubeflow/spark-operator/blob/master/Dockerfile#L17)) already has what you need, you're reusing it for the client pod with `PullNever`.

1. Rebase onto latest `master` (Spark 4.0.4)
2. From that image, connect to Connect via the Service (`sc://...:15002`) with a JVM/Scala client
3. Keep the same assertions (row count + computed value)

If it works: remove `pip install`, keep client pod → Service → server, re-run Integration CI.
If something's missing from the base image, comment back before adding pip or a new image.

sahana-sreeram:
Verified the operator base image had everything I needed for this and switched to the JVM/Scala approach to avoid PyPI/pip. Confirmed this test and other E2E tests still pass. @nabuskey @tariq-hasan @RobuRishabh

nabuskey (approved):
/approve

PR #3032, opened 2026-07-07
fix: harden controller-runtime cache against OOMKill vulnerabilities

### LINKED ISSUE #3028: Harden controller-runtime cache configuration against OOMKill (unfiltered ByObject entries and invisible informers)
## Summary

The Spark Operator's controller-runtime cache configuration contains several anti-patterns that can lead to **OOMKill in production** when the operator watches high-volume namespaces. A standard cluster user with the `edit` ClusterRole can create enough objects (e.g., 700 ConfigMaps at 900KB each = 630MB) to exceed the operator pod's memory limit and force a crash.

This issue catalogs the findings from running the [Operator Anti-Pattern Scanner](https://github.com/eformat/operator-anti-patterns) against the current `main` branch (`2078e74`). The scanner is based on the anti-pattern taxonomy documented in two Red Hat Developer articles:

- [Protect your Kubernetes Operator from OOMKill](https://developers.redhat.com/articles/2026/06/01/protect-your-kubernetes-operator-oomkill) — Describes how unfiltered `ByObject` entries (empty `{}`) cause the informer cache to hold all objects of a type cluster-wide, and provides a 4-step cache remediation flow.
- [5 anti-patterns that cause Kubernetes operator vulnerabilities](https://developers.redhat.com/articles/2026/07/06/5-anti-patterns-cause-kubernetes-operator-vulnerabilities) — Catalogs five hidden ways unfiltered informers silently enter production code, even when predicates or cache configurations appear to provide protection.

---

### CRITICAL — AP-3: Invisible informer from `client.Get()`/`client.List()` (5 findings)

When the cached client calls `Get()` or `List()` for a type **not registered in `ByObject`**, controller-runtime silently creates a **cluster-wide informer** on first invocation. This is the most dangerous pattern because there is no visible indication in the cache configuration that these informers exist.

| File | Line | Type | Code |
|---|---|---|---|
| [`internal/controller/sparkapplication/controller.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/internal/controller/sparkapplication/controller.go#L1162) | 1162 | `corev1.PodList` | `r.client.List(ctx, pods, client.InNamespace(app.Namespace), client.MatchingLabels(matchLabels))` |
| [`internal/controller/sparkapplication/driveringress.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/internal/controller/sparkapplication/driveringress.go#L180) | 180 | `networkingv1.Ingress` | `r.client.Get(ctx, client.ObjectKeyFromObject(ingress), existingIngress)` |
| [`internal/controller/sparkapplication/driveringress.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/internal/controller/sparkapplication/driveringress.go#L278) | 278 | `extensionsv1beta1.Ingress` | `r.client.Get(ctx, client.ObjectKeyFromObject(ingress), existingLegacyIngress)` |
| [`internal/controller/sparkconnect/reconciler.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/internal/controller/sparkconnect/reconciler.go#L645) | 645 | `corev1.PodList` | `r.client.List(ctx, ...)` |
| [`internal/webhook/sparkapplication_validator.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/internal/webhook/sparkapplication_validator.go#L185) | 185 | `corev1.ResourceQuotaList` | `v.client.List(ctx, resourceQuotaList, client.InNamespace(app.Namespace))` |

**Note:** The `PodList` finding at `controller.go:1162` is likely a **false positive** — `corev1.Pod` is already registered in `ByObject` with a label selector, and controller-runtime reuses the same informer for `Pod` and `PodList`. However, the Ingress and ResourceQuota findings are real.

**Fix:** Add these types to `DisableFor` (to bypass the cache and use direct API reads), or add them to `ByObject` with appropriate label/field selectors.

---

### CRITICAL — AP-6: Unfiltered `ByObject` entries (5 findings)

A `ByObject` entry with an empty config `{}` caches **all objects of that type** across all watched namespaces with no label or field selector filtering.

| File | Line | Type | Risk Level |
|---|---|---|---|
| [`cmd/operator/controller/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/controller/start.go#L471) | 471 | `corev1.Namespace{}` | **Low** — cluster-scoped, low volume, unprivileged users cannot create namespaces |
| [`cmd/operator/controller/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/controller/start.go#L482) | 482 | `corev1.PersistentVolumeClaim{}` | **High** — tenants can flood PVCs in watched namespaces |
| [`cmd/operator/controller/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/controller/start.go#L483) | 483 | `corev1.Service{}` | **High** — tenants can flood Services in watched namespaces |
| [`cmd/operator/webhook/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/webhook/start.go#L391) | 391 | `SparkApplication{}` | **False positive** — operator-owned CRD |
| [`cmd/operator/webhook/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/webhook/start.go#L392) | 392 | `ScheduledSparkApplication{}` | **False positive** — operator-owned CRD |

**Note:** The `SparkApplication` and `ScheduledSparkApplication` entries are operator-owned CRDs that need to watch all instances — these are expected to have empty configs. The `Namespace` entry is intentional for namespace-selector-based filtering and carries low risk.

**PVC and Service are the actionable findings.** These should have label selectors (e.g., `common.LabelLaunchedBySparkOperator`) to only cache operator-managed resources, consistent with how `Pod` and `ConfigMap` are already configured in the same block.

---

### HIGH — AP-1: Predicate filters don't limit cache (9 findings)

`builder.WithPredicates()` and `WithEventFilter()` only filter which events reach the reconciler — the underlying informer still performs a full LIST+WATCH. These predicates provide a false sense of cache scoping.

Found across 5 files:
- `internal/controller/sparkapplication/controller.go` (lines 282, 287)
- `internal/controller/scheduledsparkapplication/controller.go` (line 253)
- `internal/controller/sparkconnect/reconciler.go` (lines 114, 122, 176, 182)
- `internal/controller/mutatingwebhookconfiguration/controller.go` (line 71)
- `internal/controller/validatingwebhookconfiguration/controller.go` (line 71)

**Note:** Many of these are mitigated by the `ByObject` label selectors already present for `Pod` and `ConfigMap`. The webhook configuration controllers use field selectors. These are informational rather than urgent, but worth documenting so future contributors don't rely on predicates as a cache-limiting mechanism.

---

### MEDIUM — AP-10: No `DefaultTransform` to strip `managedFields` (2 findings)

Neither the controller nor the webhook cache configuration includes `DefaultTransform: cache.TransformStripManagedFields()`. Every cached object carries ~2-5KB of `managedFields` metadata that the operator never reads. Across thousands of cached objects, this adds non-trivial memory overhead.

| File | Line |
|---|---|
| [`cmd/operator/controller/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/controller/start.go#L467) | 467 |
| [`cmd/operator/webhook/start.go`](https://github.com/kubeflow/spark-operator/blob/2078e740f9273d6cdcb8e2bf5894dc9d3d93b983/cmd/operator/webhook/start.go#L409) | 409 |

**Fix:** One-line addition to each `cache.Options`:
```go
options := cache.Options{
    DefaultTransform: cache.TransformStripManagedFields(),
    // ... existing config ...
}
```

---

## Recommended Remediation Priority

| Priority | Finding | Effort | Impact |
|---|---|---|---|
| 1 | AP-6: Add label selectors to PVC and Service `ByObject` entries | Low | Prevents tenant-induced OOMKill |
| 2 | AP-3: Add Ingress types to `DisableFor` (or `ByObject` with selector) | Low | Eliminates invisible cluster-wide informers |
| 3 | AP-3: Add ResourceQuota to `DisableFor` in webhook | Low | Eliminates invisible informer when quota enforcement is on |
| 4 | AP-10: Add `DefaultTransform: cache.TransformStripManagedFields()` | Trivial | Reduces per-object memory by ~2-5KB |
| 5 | AP-1: Document predicate vs. cache-filter distinction | N/A | Prevents future regressions |

I'm happy to submit a PR with these fixes if the approach looks acceptable to the maintainers.

---

## References

- **Blog Part 1:** [Protect your Kubernetes Operator from OOMKill](https://developers.redhat.com/articles/2025/06/11/protect-your-kubernetes-operator-oomkill)
- **Blog Part 2:** [5 anti-patterns that cause Kubernetes operator vulnerabilities](https://developers.redhat.com/articles/2025/07/03/5-anti-patterns-cause-kubernetes-operator-vulnerabilities)

RobuRishabh:
## Purpose of this PR

Harden controller-runtime cache against OOMKill by eliminating unfiltered informers that allow standard `edit` users to flood the operator's memory.

Fixes: https://github.com/kubeflow/spark-operator/issues/3028

**Changes:**

- Remove unused `PersistentVolumeClaim` from `ByObject` (no `client.Get/List` or `Watches` for PVCs exist anywhere in the codebase)
- Add label selector to `Service` `ByObject` entry, matching the existing `Pod`/`ConfigMap` pattern, and stamp `LabelCreatedBySparkOperator` on all operator-created Services
- Add `DisableFor` for Ingress types to prevent invisible cluster-wide informers from `client.Get()` calls
- Add `DisableFor` for `ResourceQuota` in webhook — quota validation needs all ResourceQuotas, so cache bypass is correct over label filtering
- Add `DefaultTransform: cache.TransformStripManagedFields()` to both caches

**References:**
- [Protect your Kubernetes Operator from OOMKill](https://developers.redhat.com/articles/2025/06/11/protect-your-kubernetes-operator-oomkill)
- [5 anti-patterns that cause Kubernetes operator vulnerabilities](https://developers.redhat.com/articles/2025/07/03/5-anti-patterns-cause-kubernetes-operator-vulnerabilities)
- Scanner: [eformat/operator-anti-patterns](https://github.com/eformat/operator-anti-patterns)

## Change Category

## Checklist

### DIFF
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -40,6 +40,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -305,6 +306,14 @@ func start() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: operatorscheme.ControllerScheme,
 		Cache:  newCacheOptions(),
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&networkingv1.Ingress{},
+					&extensionsv1beta1.Ingress{},
+				},
+			},
+		},
 		Metrics: metricsserver.Options{
 			BindAddress:   metricsBindAddress,
 			SecureServing: secureMetrics,
@@ -467,6 +476,7 @@ func newCacheOptions() cache.Options {
 	options := cache.Options{
 		Scheme:            operatorscheme.ControllerScheme,
 		DefaultNamespaces: defaultNamespaces,
+		DefaultTransform:  cache.TransformStripManagedFields(),
 		ByObject: map[client.Object]cache.ByObject{
 			&corev1.Namespace{}: {},
 			&corev1.Pod{}: {
@@ -479,8 +489,11 @@ func newCacheOptions() cache.Options {
 					common.LabelCreatedBySparkOperator: "true",
 				}),
 			},
-			&corev1.PersistentVolumeClaim{}:      {},
-			&corev1.Service{}:                    {},
+			&corev1.Service{}: {
+				Label: labels.SelectorFromSet(labels.Set{
+					common.LabelCreatedBySparkOperator: "true",
+				}),
+			},
 			&v1beta2.SparkApplication{}:          {},
 			&v1beta2.ScheduledSparkApplication{}: {},
 			&v1alpha1.SparkConnect{}:             {},
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -308,6 +308,8 @@ func start() {
 		Cache:  newCacheOptions(),
 		Client: client.Options{
 			Cache: &client.CacheOptions{
+				// TODO: If Ingress reconciliation is added, move these to ByObject with a label selector.
+				// Currently write-only (created with OwnerReferences, no Watches/Owns registered).
 				DisableFor: []client.Object{
 					&networkingv1.Ingress{},
 					&extensionsv1beta1.Ingress{},
--- a/cmd/operator/webhook/start.go
+++ b/cmd/operator/webhook/start.go
@@ -210,13 +210,7 @@ func start() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: operatorscheme.WebhookScheme,
 		Cache:  newCacheOptions(),
-		Client: client.Options{
-			Cache: &client.CacheOptions{
-				DisableFor: []client.Object{
-					&corev1.ResourceQuota{},
-				},
-			},
-		},
+		Client: client.Options{},
 		Metrics: metricsserver.Options{
 			BindAddress:   metricsBindAddress,
 			SecureServing: secureMetrics,
@@ -395,6 +389,7 @@ func newCacheOptions() cache.Options {
 				common.LabelLaunchedBySparkOperator: "true",
 			}),
 		},
+		&corev1.ResourceQuota{}:              {},
 		&v1beta2.SparkApplication{}:          {},
 		&v1beta2.ScheduledSparkApplication{}: {},
 		&admissionregistrationv1.MutatingWebhookConfiguration{}: {
--- a/cmd/operator/webhook/start.go
+++ b/cmd/operator/webhook/start.go
@@ -210,6 +210,13 @@ func start() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: operatorscheme.WebhookScheme,
 		Cache:  newCacheOptions(),
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&corev1.ResourceQuota{},
+				},
+			},
+		},
 		Metrics: metricsserver.Options{
 			BindAddress:   metricsBindAddress,
 			SecureServing: secureMetrics,
@@ -402,13 +409,10 @@ func newCacheOptions() cache.Options {
 		},
 	}
 
-	if enableResourceQuotaEnforcement {
-		byObject[&corev1.ResourceQuota{}] = cache.ByObject{}
-	}
-
 	options := cache.Options{
 		Scheme:            operatorscheme.WebhookScheme,
 		DefaultNamespaces: defaultNamespaces,
+		DefaultTransform:  cache.TransformStripManagedFields(),
 		ByObject:          byObject,
 	}
 
--- a/internal/controller/sparkapplication/driveringress.go
+++ b/internal/controller/sparkapplication/driveringress.go
@@ -332,11 +332,14 @@ func (r *Reconciler) createDriverIngressService(
 	serviceLabels map[string]string,
 ) (*SparkService, error) {
 	logger := log.FromContext(ctx)
+	resourceLabels := util.GetResourceLabels(app)
+	resourceLabels[common.LabelCreatedBySparkOperator] = "true"
+
 	service := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            serviceName,
 			Namespace:       app.Namespace,
-			Labels:          util.GetResourceLabels(app),
+			Labels:          resourceLabels,
 			OwnerReferences: []metav1.OwnerReference{util.GetOwnerReference(app)},
 		},
 		Spec: corev1.ServiceSpec{
@@ -361,6 +364,7 @@ func (r *Reconciler) createDriverIngressService(
 	if len(serviceLabels) != 0 {
 		service.Labels = serviceLabels
 	}
+	service.Labels[common.LabelCreatedBySparkOperator] = "true"
 
 	if len(serviceAnnotations) != 0 {
 		service.Annotations = serviceAnnotations
--- a/internal/controller/sparkapplication/driveringress.go
+++ b/internal/controller/sparkapplication/driveringress.go
@@ -332,14 +332,11 @@ func (r *Reconciler) createDriverIngressService(
 	serviceLabels map[string]string,
 ) (*SparkService, error) {
 	logger := log.FromContext(ctx)
-	resourceLabels := util.GetResourceLabels(app)
-	resourceLabels[common.LabelCreatedBySparkOperator] = "true"
-
 	service := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            serviceName,
 			Namespace:       app.Namespace,
-			Labels:          resourceLabels,
+			Labels:          util.GetResourceLabels(app),
 			OwnerReferences: []metav1.OwnerReference{util.GetOwnerReference(app)},
 		},
 		Spec: corev1.ServiceSpec{
@@ -381,11 +378,26 @@ func (r *Reconciler) createDriverIngressService(
 			if !errors.IsAlreadyExists(err) {
 				return nil, err
 			}
-			// Lost a race — another reconciliation created it; fetch it.
-			if err := r.client.Get(ctx, client.ObjectKeyFromObject(service), existing); err != nil {
+			// Handle upgrade scenario: Service exists in the cluster but is not
+			// visible in the filtered cache because it was created before the
+			// LabelCreatedBySparkOperator label selector was added to the informer.
+			// Use Patch (merge patch) instead of Update because we cannot Get the
+			// object from the filtered cache to obtain its resourceVersion.
+			base := &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      service.Name,
+					Namespace: service.Namespace,
+				},
+			}
+			desired := base.DeepCopy()
+			desired.Labels = service.Labels
+			desired.Annotations = service.Annotations
+			desired.OwnerReferences = service.OwnerReferences
+			desired.Spec = service.Spec
+			if err := r.client.Patch(ctx, desired, client.MergeFrom(base)); err != nil {
 				return nil, err
 			}
-			service = existing
+			service = desired
 		}
 		logger.Info("Created service for SparkApplication", "name", service.Name)
 	} else {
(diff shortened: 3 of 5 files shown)


cmd/operator/controller/start.go
```diff
@@ -305,6 +306,14 @@ func start() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: operatorscheme.ControllerScheme,
 		Cache:  newCacheOptions(),
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
```
nabuskey, line 313: I am a bit concerned here from maintenance perspective. We currently do not issue many API calls for ingress so disabling it likely ok from performance perspective. I am worried about impact in the future. Disabling cache here can be missed easily and could cause increased load to API if we decide to do more comprehensive reconciliation against operator-created ingress objects. Is it a better idea to cache operator owned ingress only? What do you think?

cmd/operator/webhook/start.go
```diff
@@ -210,6 +210,13 @@ func start() {
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: operatorscheme.WebhookScheme,
 		Cache:  newCacheOptions(),
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
```
nabuskey: This is likely an issue. If the resource quota option is enabled, it reads it every reconciliation from the API. This would be particularly bad for short-lived spark applications.

cmd/operator/controller/start.go
RobuRishabh: the maintenance concern is fair. Here's why I chose DisableFor over a label-filtered cache:

Ingress was never in ByObject to begin with — the current code creates invisible cluster-wide informers on the first client.Get() call ([first - driveringress.go](https://github.com/kubeflow/spark-operator/blob/be5bb94aa27491f029470976e542fdef23934c7b/internal/controller/sparkapplication/driveringress.go#L180), [second - driveringress.go](https://github.com/kubeflow/spark-operator/blob/be5bb94aa27491f029470976e542fdef23934c7b/internal/controller/sparkapplication/driveringress.go#L278), [controller.go:1294](https://github.com/kubeflow/spark-operator/blob/be5bb94aa27491f029470976e542fdef23934c7b/internal/controller/sparkapplication/controller.go#L1345)). DisableFor eliminates those invisible informers while preserving the existing behavior of direct API calls.

The operator's relationship with Ingress is write-only, we create them with OwnerReferences and never reconcile in response to Ingress events (no Watches/Owns for Ingress exists). A label-filtered cache would create a permanent informer that no controller ever consumes, wasting a Watch connection and memory for objects nothing reads from.

There are only 3 client.Get() calls for Ingress in the entire codebase, all in low-frequency create/check paths — the API load is negligible.

I'll add a comment above DisableFor explaining why Ingress is excluded from the cache, so future contributors are aware before adding Ingress reconciliation. If someone does add comprehensive Ingress reconciliation in the future, that would be the right time to move Ingress from DisableFor into ByObject with a label selector.

Does that address your concern?

cmd/operator/webhook/start.go
RobuRishabh: You're right. Will revert it. ResourceQuotas are admin-controlled, so the OOM attack vector shouldn't apply here. Maybe, I will have to fix this the [scanner](https://github.com/eformat/operator-anti-patterns) too it need admin-only type awareness.

internal/controller/sparkapplication/driveringress.go
tariq-hasan, line 402: I am wondering if, for the case of the filtered `Service` objects, we would want to follow the same [approach](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkapplication/monitoring_config.go#L53-L72) as for the ConfigMap fix in https://github.com/kubeflow/spark-operator/pull/2878.

internal/controller/sparkconnect/reconciler.go
```diff
@@ -587,6 +587,7 @@ func (r *Reconciler) mutateServerService(_ context.Context, conn *v1alpha1.Spark
 	for key, val := range GetCommonLabels(conn) {
 		svc.Labels[key] = val
 	}
+	svc.Labels[common.LabelCreatedBySparkOperator] = "true"
```
tariq-hasan: I'm actually thinking we can remove `svc.Labels[common.LabelCreatedBySparkOperator] = "true"` since [`GetCommonLabels`](https://github.com/RobuRishabh/spark-operator/blob/1ec597787f7ab23540231b35e65eff9a22f16e16/internal/controller/sparkconnect/util.go#L26-L33) already adds these labels.

internal/controller/sparkapplication/driveringress.go
```diff
@@ -332,11 +332,14 @@ func (r *Reconciler) createDriverIngressService(
 	serviceLabels map[string]string,
 ) (*SparkService, error) {
 	logger := log.FromContext(ctx)
+	resourceLabels := util.GetResourceLabels(app)
+	resourceLabels[common.LabelCreatedBySparkOperator] = "true"
+
```
tariq-hasan: I'm thinking we won't need to add this snippet here since we already do `service.Labels[common.LabelCreatedBySparkOperator] = "true"` below.

tariq-hasan (commented):
Hi @RobuRishabh! I have added a few comments.

cmd/operator/controller/start.go
tariq-hasan: Actually one more thing I wanted to point out is that the [RBAC](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/templates/controller/_helpers.tpl#L156-L162) for `Ingress` is missing `list` and `watch` permissions so technically the code as it stands needs fixing on a bug fix I guess - given that the invisible informer will get a permission denied error on list and watch.

cmd/operator/webhook/start.go
tariq-hasan: Same issue with the [RBAC](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/templates/webhook/_helpers.tpl#L132-L137) for `resourcequotas`.

RobuRishabh:
Hi @tariq-hasan, Thank you for the review, I addressed your comments.

cmd/operator/controller/start.go
RobuRishabh: Thank you for pointing this out, @tariq-hasan.  It actually reinforces why DisableFor is the right approach here. The Ingress RBAC intentionally grants only get, create, update, delete — no list or watch. The Helm chart test [explicitly asserts this](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/tests/controller/rbac_test.yaml#L331) ("Should grant ingresses without list or watch in ClusterRole"). So any informer-based approach (ByObject, Owns, Watches) would fail with permission denied on the LIST/WATCH calls.

With DisableFor, client.Get() goes directly to the API server using only the get verb, which the RBAC already allows. So DisableFor doesn't just help with the OOM risk, it also aligns with the existing RBAC design. No RBAC change should be needed here.

I believe the pre-existing invisible informers from client.Get() (before this PR) were likely hitting permission denied errors on LIST/WATCH behind the scenes, controller-runtime retries silently, which would mask the failure. DisableFor should eliminate that entirely.

cmd/operator/webhook/start.go
RobuRishabh: Thank you @tariq-hasan another great observation. The webhook RBAC for resourcequotas only grants list ([_helpers.tpl](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/templates/webhook/_helpers.tpl#L135-L137)), not watch. The ByObject informer would need both list and watch to function correctly.

That said, I believe this is a pre-existing gap, ResourceQuota was already in ByObject before this PR. I think it might make sense to track the missing watch permission as a separate issue to keep this cache-hardening PR focused. Happy to file one would that work ?
cc: @nabuskey

nabuskey:
Looks good to me. @tariq-hasan I would love your review as well.

tariq-hasan:
No further questions from me.
/lgtm

cmd/operator/webhook/start.go
tariq-hasan: I guess that makes sense to leave the RBAC fix to a separate issue.

nabuskey (approved):
/approve

PR #3042, opened 2026-07-13
feat: garbage-collect terminated SparkApplications via operator default TTL

dineshkumar181094:
## Purpose of this PR

Terminated `SparkApplication` objects (`COMPLETED`/`FAILED`) are only garbage-collected when the user sets `spec.timeToLiveSeconds`. When that field is unset, the object — and its driver pod could — linger indefinitely. This is a real hazard when driver pods carry sidecars (logging agents, service-mesh proxies) that do not reliably exit after the primary Spark container finishes: at scale the lingering pods cause pod-IP exhaustion and resource crunch.

This PR adds an operator-configurable **default TTL** that is applied to terminated SparkApplications **only when the user has not set their own** `timeToLiveSeconds`. It is applied purely as a runtime fallback for the cleanup decision — the user's object is never mutated — and it is off by default, gated behind a new alpha feature gate.

**Proposed changes:**

- Add a new alpha feature gate `DefaultTimeToLive` (default `false`).
- Add a `--default-time-to-live-seconds` controller flag (default `0` = disabled) plumbed through to the reconciler `Options`.
- Add `util.EffectiveTimeToLiveSeconds`, which resolves the TTL governing cleanup: user spec wins; else the operator default when the gate is enabled and the value is `> 0`; else `nil` ("never expire"). The pure `util.IsExpired` is left unchanged.
- Use the resolved effective TTL in `reconcileTerminatedSparkApplication` (applies to both `COMPLETED` and `FAILED`), with an info log emitted when the operator default is the source.
- Expose the setting via the Helm chart: `controller.defaultTimeToLiveSeconds` value, a `DefaultTimeToLive` entry in `controller.featureGates`, the rendered flag in the controller deployment, chart unit tests, and regenerated chart docs.

## Change Category

- [x] Documentation update

### Rationale

The behavior is opt-in and non-destructive: an explicit user TTL always wins, the object's spec is never rewritten, and two independent guards (feature gate enabled **and** a positive value) must both hold for the fallback to activate. This makes the upgrade path a no-op for existing clusters while giving operators a cluster-wide safety net against resource exhaustion.

## Checklist

### Additional Notes

- The feature gate follows the repo's existing alpha→beta→GA promotion model in `pkg/features`.
- Tests: unit coverage for `EffectiveTimeToLiveSeconds`, a controller integration test verifying deletion of a terminated app past the default TTL (gate on, no spec TTL), and Helm chart tests asserting the flag renders only when the value is set.

### DIFF
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -95,6 +95,8 @@ controller:
     enabled: false
   - name: RestSubmitter
     enabled: false
+  - name: DefaultTimeToLive
+    enabled: false
 
   # -- The number of old history to retain to allow rollback.
   revisionHistoryLimit: 10
@@ -123,6 +125,11 @@ controller:
 
   # -- Specifies the maximum number of Executor pods that can be tracked by the controller per SparkApplication.
   maxTrackedExecutorPerApp: 1000
+
+  # -- Default Time-To-Live (in seconds) applied to terminated SparkApplications that do not
+  # set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate to be enabled.
+  # 0 (default) disables it.
+  defaultTimeToLiveSeconds: 0
 
   # -- Driver PDB feature gate. When true, the controller creates a PodDisruptionBudget for each SparkApplication that sets spec.driverPodDisruptionBudget=true. Default false.
   driverPodDisruptionBudget:
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -193,6 +193,10 @@ func NewStartCommand() *cobra.Command {
 				}
 			}
 
+			if features.Enabled(features.DefaultTimeToLive) && defaultTimeToLiveSeconds <= 0 {
+				return fmt.Errorf("invalid value %d for --default-time-to-live-seconds, must be a positive value when the DefaultTimeToLive feature gate is enabled", defaultTimeToLiveSeconds)
+			}
+
 			return nil
 		},
 		Run: func(_ *cobra.Command, args []string) {
@@ -289,6 +293,13 @@ func NewStartCommand() *cobra.Command {
 
 func start() {
 	setupLog()
+
+	// The flag is ignored unless the DefaultTimeToLive feature gate is enabled. Warn
+	// here (rather than in PreRunE) because the logger is only initialized by setupLog.
+	if defaultTimeToLiveSeconds > 0 && !features.Enabled(features.DefaultTimeToLive) {
+		logger.Info("Ignoring --default-time-to-live-seconds because the DefaultTimeToLive feature gate is disabled",
+			"defaultTimeToLiveSeconds", defaultTimeToLiveSeconds)
+	}
 
 	// Create the client rest config. Use kubeConfig if given, otherwise assume in-cluster.
 	cfg, err := ctrl.GetConfig()
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -85,6 +85,7 @@ var (
 	controllerThreads        int
 	cacheSyncTimeout         time.Duration
 	maxTrackedExecutorPerApp int
+	defaultTimeToLiveSeconds int64
 
 	// Driver PDB feature gate. When enabled, the controller creates a
 	// PodDisruptionBudget for each SparkApplication that sets
@@ -205,6 +206,10 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().StringVar(&namespaceSelector, "namespace-selector", "", "Label selector for namespaces to watch (e.g., 'spark-operator=enabled,env in (prod,staging)'). Namespaces matching this selector will be watched in addition to those specified via --namespaces. Requires ClusterRole permission to list and watch namespaces.")
 	command.Flags().DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 30*time.Second, "Informer cache sync timeout.")
 	command.Flags().IntVar(&maxTrackedExecutorPerApp, "max-tracked-executor-per-app", 1000, "The maximum number of tracked executors per SparkApplication.")
+	command.Flags().Int64Var(&defaultTimeToLiveSeconds, "default-time-to-live-seconds", 0,
+		"Default Time-To-Live in seconds applied to terminated SparkApplications that do "+
+			"not set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate. "+
+			"0 (default) or negative disables it.")
 	command.Flags().BoolVar(&enableDriverPDB, "enable-driver-pdb", false,
 		"Enable creation of a PodDisruptionBudget for Spark driver pods. "+
 			"Each SparkApplication must additionally opt in via "+
@@ -528,6 +533,7 @@ func newSparkApplicationReconcilerOptions() sparkapplication.Options {
 		SparkExecutorMetrics:         sparkExecutorMetrics,
 		MaxTrackedExecutorPerApp:     maxTrackedExecutorPerApp,
 		EnableDriverPDB:              enableDriverPDB,
+		DefaultTimeToLiveSeconds:     defaultTimeToLiveSeconds,
 	}
 	if enableBatchScheduler {
 		options.KubeSchedulerNames = kubeSchedulerNames
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -296,11 +296,14 @@ func NewStartCommand() *cobra.Command {
 func start() {
 	setupLog()
 
-	// The flag is ignored unless the DefaultTimeToLive feature gate is enabled. Warn
-	// here (rather than in PreRunE) because the logger is only initialized by setupLog.
-	if defaultTimeToLiveSeconds > 0 && !features.Enabled(features.DefaultTimeToLive) {
-		logger.Info("Ignoring --default-time-to-live-seconds because the DefaultTimeToLive feature gate is disabled",
-			"defaultTimeToLiveSeconds", defaultTimeToLiveSeconds)
+	// Normalize the configured TTL before passing it to the controller so downstream
+	// cleanup logic does not need to depend on the global feature gate.
+	if !features.Enabled(features.DefaultTimeToLive) {
+		if defaultTimeToLiveSeconds > 0 {
+			logger.Info("Ignoring --default-time-to-live-seconds because the DefaultTimeToLive feature gate is disabled",
+				"defaultTimeToLiveSeconds", defaultTimeToLiveSeconds)
+		}
+		defaultTimeToLiveSeconds = 0
 	}
 
 	// Create the client rest config. Use kubeConfig if given, otherwise assume in-cluster.
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -193,8 +193,10 @@ func NewStartCommand() *cobra.Command {
 				}
 			}
 
-			if features.Enabled(features.DefaultTimeToLive) && defaultTimeToLiveSeconds <= 0 {
-				return fmt.Errorf("invalid value %d for --default-time-to-live-seconds, must be a positive value when the DefaultTimeToLive feature gate is enabled", defaultTimeToLiveSeconds)
+			// A negative TTL is never meaningful; reject it regardless of the gate.
+			// Zero is the valid "off" sentinel and is left untouched.
+			if defaultTimeToLiveSeconds < 0 {
+				return fmt.Errorf("invalid value %d for --default-time-to-live-seconds, must not be negative", defaultTimeToLiveSeconds)
 			}
 
 			return nil
@@ -213,7 +215,7 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().Int64Var(&defaultTimeToLiveSeconds, "default-time-to-live-seconds", 0,
 		"Default Time-To-Live in seconds applied to terminated SparkApplications that do "+
 			"not set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate. "+
-			"0 (default) or negative disables it.")
+			"0 (default) disables it; a negative value is rejected.")
 	command.Flags().BoolVar(&enableDriverPDB, "enable-driver-pdb", false,
 		"Enable creation of a PodDisruptionBudget for Spark driver pods. "+
 			"Each SparkApplication must additionally opt in via "+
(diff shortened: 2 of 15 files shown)


dineshkumar181094:
@RobuRishabh / @Tabrizian  can you please take a look at this feature.

dineshkumar181094:
@nabuskey  can you please take a look at this feature and this pr.

charts/spark-operator-chart/values.yaml
```diff
@@ -124,6 +126,11 @@ controller:
   # -- Specifies the maximum number of Executor pods that can be tracked by the controller per SparkApplication.
   maxTrackedExecutorPerApp: 1000
 
+  # -- Default Time-To-Live (in seconds) applied to terminated SparkApplications that do not
+  # set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate to be enabled.
+  # 0 (default) disables it.
+  defaultTimeToLiveSeconds: 0
```
tariq-hasan, line 132: Would there be some way to link the dependency between `controller.defaultTimeToLiveSeconds` and the `DefaultTimetoLive` entry in the `featuresGates` list?

At this time passing `defaultTimeToLiveSeconds` while having the feature turned off is essentially a silent no-op on the Helm install (as well as controller startup).

Maybe we can use the [pattern](https://github.com/dineshkumar181094/spark-operator/blob/41f0c426954589a3edad95928591832895848e71/charts/spark-operator-chart/templates/_helpers.tpl#L80-L87) from the REST submitter service here as well.

cmd/operator/controller/start.go
tariq-hasan, line 194: Similar to how we do it for the REST submission flags do we want to have validation for the operator default TTL - erroring when the flag is non-positive but the feature gate is on and logging a warning at the very least when the flag is set but the feature gate is turned off?

internal/controller/sparkapplication/controller.go
```diff
@@ -714,7 +720,13 @@ func (r *Reconciler) reconcileTerminatedSparkApplication(ctx context.Context, re
 		return ctrl.Result{}, nil
 	}
 
-	if util.IsExpired(app) {
+	effectiveTTLSeconds, usedDefault := util.EffectiveTimeToLiveSeconds(app, r.options.DefaultTimeToLiveSeconds)
+	if usedDefault {
+		logger.Info("Applying operator default TTL to terminated SparkApplication",
+			"ttlSeconds", *effectiveTTLSeconds, "state", app.Status.AppState.State)
+	}
```
tariq-hasan: I'm thinking probably better to move this logging to when the actual deletion occurs so it does not need to fire on every reconcile of the terminated app.

internal/controller/sparkapplication/controller.go
tariq-hasan, line 768: The terminal reconciliation loop is now using a repurposed concept of effective TTL - either user-provided or operator-defaulted - but the [event filter](https://github.com/dineshkumar181094/spark-operator/blob/41f0c426954589a3edad95928591832895848e71/internal/controller/sparkapplication/event_filter.go#L174) is still using just the user-provided TTL if there is one.

Should we update the event filter as well to be consistent on the changes?

tariq-hasan (commented):
Hi @dineshkumar181094! I have added a few comments.

charts/spark-operator-chart/values.yaml
dineshkumar181094: Valid point I have update the helm chart and the controller code as well to handle this case.

cmd/operator/controller/start.go
dineshkumar181094: Added a soft warning if user turn on the feature but feature is disabled.

internal/controller/sparkapplication/controller.go
dineshkumar181094: Indeed move the log line into the actual deletion logic.

internal/controller/sparkapplication/controller.go
dineshkumar181094: Yeah Now the Expired definition has changed. That part in event filter was considering no-op in case the app version is same and app has not expired and no retry just to avoid reconciliation.
But now the app expired definition has changed  with the default operator side it make sense to update there so if the user has not put ttl and operator is applying controller side ttl and it should not result in no-op and start the deletion logic instead.

dineshkumar181094:
> Hi @dineshkumar181094! I have added a few comments.

@tariq-hasan Thank you for review all points are valid addressed all points.

nabuskey:
@RobuRishabh Could you please review this?

cmd/operator/controller/start.go
```diff
@@ -205,6 +210,10 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().StringVar(&namespaceSelector, "namespace-selector", "", "Label selector for namespaces to watch (e.g., 'spark-operator=enabled,env in (prod,staging)'). Namespaces matching this selector will be watched in addition to those specified via --namespaces. Requires ClusterRole permission to list and watch namespaces.")
 	command.Flags().DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 30*time.Second, "Informer cache sync timeout.")
 	command.Flags().IntVar(&maxTrackedExecutorPerApp, "max-tracked-executor-per-app", 1000, "The maximum number of tracked executors per SparkApplication.")
+	command.Flags().Int64Var(&defaultTimeToLiveSeconds, "default-time-to-live-seconds", 0,
+		"Default Time-To-Live in seconds applied to terminated SparkApplications that do "+
+			"not set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate. "+
+			"0 (default) or negative disables it.")
```
RobuRishabh: Nit: the `--default-time-to-live-seconds` flag help says "0 (default) or negative disables it", but when the `DefaultTimeToLive` gate is enabled, 0 or negative actually causes a hard startup error from `PreRunE`.
I would say lets modify the last line to `"Must be a positive value when the gate is enabled; 0 (default) when the gate is disabled.")`

charts/spark-operator-chart/values.yaml
```diff
@@ -124,6 +126,11 @@ controller:
   # -- Specifies the maximum number of Executor pods that can be tracked by the controller per SparkApplication.
   maxTrackedExecutorPerApp: 1000
 
+  # -- Default Time-To-Live (in seconds) applied to terminated SparkApplications that do not
+  # set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate to be enabled.
+  # 0 (default) disables it.
```
RobuRishabh, line 131: Worth adding a note that when the `DefaultTimeToLive` gate is enabled, this value must be set to a positive integer or the controller will fail to start. Something like:

```
# -- Default Time-To-Live (in seconds) applied to terminated SparkApplications that do not
# set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate to be enabled.
# Must be set to a positive value when the gate is enabled; the controller will refuse to
# start otherwise. 0 (default) means the feature is inactive.
```

internal/controller/sparkapplication/controller.go
RobuRishabh, line 767: Consider adding a test case where the terminated app has _not yet_ outlived the operator default TTL e.g., `TerminationTime: -30s` with `DefaultTimeToLiveSeconds: 60`. This would verify the requeue-after calculation with the effective TTL and prove the app is _not_ prematurely deleted.

RobuRishabh (commented):
@dineshkumar181094 left some minor comment.

cmd/operator/controller/start.go
dineshkumar181094: Updated the helper,
0 -> means disabled and default behaviour. Though ideally default should be a infinite number to keep the behaviour intact. So if user has not supplied any ttl the app should have infinite ttl.
this value 0 it will mimic the backward compatible behaviour.
-ve value -> hardstop error not starting the controller just to be explicit.

charts/spark-operator-chart/values.yaml
dineshkumar181094: Now I have update the 0 means default disabled.

internal/controller/sparkapplication/controller.go
dineshkumar181094: added the test case .

cmd/operator/controller/start.go
dineshkumar181094:
```go
func EffectiveTimeToLiveSeconds(app *v1beta2.SparkApplication, defaultSeconds int64) (*int64, bool) {
	if app.Spec.TimeToLiveSeconds != nil {
		return app.Spec.TimeToLiveSeconds, false
	}
	if features.Enabled(features.DefaultTimeToLive) && defaultSeconds > 0 {
		seconds := defaultSeconds
		return &seconds, true
	}
	return nil, false
}
```
Zero results in backward compatibility.

RobuRishabh:
@tariq-hasan any extra comments from your side. It's good to go from my side. @nabuskey can you please trigger the CI when you get a chance. Thanks

pkg/util/sparkapplication.go
```diff
@@ -60,14 +61,26 @@ func IsTerminated(app *v1beta2.SparkApplication) bool {
 		app.Status.AppState.State == v1beta2.ApplicationStateFailed
 }
 
-// IsExpired returns whether the given SparkApplication is expired.
+// IsExpired returns whether the given SparkApplication is expired according to its
+// own spec.timeToLiveSeconds.
 func IsExpired(app *v1beta2.SparkApplication) bool {
+	return IsExpiredWithTTL(app, app.Spec.TimeToLiveSeconds)
+}
```
tariq-hasan: I'm wondering if we need `IsExpired` since it is not leveraged anywhere anymore.

Rather maybe we just remove `IsExpired` and replace all instances of `IsExpiredWithTTL` with `IsExpired`.

This would also require removing [these](https://github.com/dineshkumar181094/spark-operator/blob/80263ac06f75e360537853916a195e1c3a7866ad/pkg/util/sparkapplication_test.go#L127-L196) tests.

```suggestion
```

pkg/util/sparkapplication.go
```diff
@@ -76,6 +89,29 @@ func IsExpired(app *v1beta2.SparkApplication) bool {
...
+//
+// It never mutates the SparkApplication.
+func EffectiveTimeToLiveSeconds(app *v1beta2.SparkApplication, defaultSeconds int64) (*int64, bool) {
+	if app.Spec.TimeToLiveSeconds != nil {
+		return app.Spec.TimeToLiveSeconds, false
+	}
+	if features.Enabled(features.DefaultTimeToLive) && defaultSeconds > 0 {
+		seconds := defaultSeconds
+		return &seconds, true
+	}
```
tariq-hasan: I'm thinking it might be better to use something like the following in [`start.go`](https://github.com/dineshkumar181094/spark-operator/blob/80263ac06f75e360537853916a195e1c3a7866ad/cmd/operator/controller/start.go#L299-L304) and propagate this value of `defaultTimeToLiveSeconds` into the controller logic all the way to the `EffectiveTimeToLiveSeconds` helper.

```
func start() {
    setupLog()

    // The flag is ignored unless the DefaultTimeToLive feature gate is enabled.
    if !features.Enabled(features.DefaultTimeToLive) {
        if defaultTimeToLiveSeconds > 0 {
            logger.Info("Ignoring --default-time-to-live-seconds because the DefaultTimeToLive feature gate is disabled",
                "defaultTimeToLiveSeconds", defaultTimeToLiveSeconds)
        }
        defaultTimeToLiveSeconds = 0
    }

    // Create the client rest config...
```

That way, we avoid dependency on the global feature gate within what are effectively helper functions.

```suggestion
	if defaultSeconds > 0 {
		seconds := defaultSeconds
		return &seconds, true
	}
```

Ideally we'd want helper functions in a utility package to be pure functions and for these to not have side effects from global external dependencies. The above approach could help avoid a form of code smell where we couple components that are best left decoupled.

cmd/operator/controller/start.go
```diff
@@ -285,6 +296,13 @@ func NewStartCommand() *cobra.Command {
 func start() {
 	setupLog()
 
+	// The flag is ignored unless the DefaultTimeToLive feature gate is enabled. Warn
+	// here (rather than in PreRunE) because the logger is only initialized by setupLog.
+	if defaultTimeToLiveSeconds > 0 && !features.Enabled(features.DefaultTimeToLive) {
+		logger.Info("Ignoring --default-time-to-live-seconds because the DefaultTimeToLive feature gate is disabled",
+			"defaultTimeToLiveSeconds", defaultTimeToLiveSeconds)
+	}
+
```
tariq-hasan, line 308: I believe that the log here is meant to provide info on the unintended user behavior of not enabling the feature gate when the user intention is to use the feature by setting a positive operator TTL.

That said this path would be unexercised in that case since the Helm chart silently [drops](https://github.com/dineshkumar181094/spark-operator/blob/80263ac06f75e360537853916a195e1c3a7866ad/charts/spark-operator-chart/templates/controller/deployment.yaml#L135-L136) this case and does not pass the config to the controller binary in that case.

Seems like, from a UX perspective, a user receiving a fail-fast on the Helm chart deployment, would be better off troubleshooting than having a silent fallthrough that does not exercise the above code.

tariq-hasan, line 308: The bigger issue seems to be some form of redundancy in the overall design in the sense that the feature gate being off and `--default-time-to-live-seconds` being 0 mean the same thing.

The disabled case happens if `--default-time-to-live-seconds` is 0 regardless of the state of the feature gate - so that is one form of redundancy.

The other redundancy is when `--default-time-to-live-seconds` is positive but the gate is off - like in the code above. 

Not exactly sure how effective the feature gate really is here because `--default-time-to-live-seconds` being 0 already handles the gate on/off case.

tariq-hasan (commented):
Just added a few additional comments. Overall lgtm. Thanks!

pkg/util/sparkapplication.go
dineshkumar181094: This make sense.

cmd/operator/controller/start.go
dineshkumar181094: Here is the explanation for 0 value https://github.com/kubeflow/spark-operator/pull/3042#discussion_r3673752414.
regarding the feature flag. I mean we can do it without feature flag as well but making sure I'm following convention  while adding new feature ie explicit flag for feature.

tariq-hasan: That makes sense. I suppose we can keep as is for now then.

pkg/util/sparkapplication.go
dineshkumar181094: addressed this.

pkg/util/sparkapplication.go
dineshkumar181094: done.

dineshkumar181094: I removed the IsExpiredWithTTL and updated the signature  of IsExpired added defaulttimetolive as second arg.

dineshkumar181094:
@tariq-hasan  addressed all feedback and refactored the code.

tariq-hasan:
Thanks!
/lgtm

nabuskey (approved):
/approve

Issue #3006, opened 2026-06-23
Server-side authorization for spec.sparkConf keys


vjanelle:
### What feature you would like to be added?

Per-key validation of sparkConf values passed in against a potential allow/deny list to prevent potentially specifying keys in sparkConf that get passed to spark-submit verbatim.

### Why is this needed?

Any tenant with create/update on SparkApplication can set arbitrary Spark config keys via spec.sparkConf. The submission path (internal/controller/sparkapplication/submission.go:295-308, sparkConfOption) forwards every key/value to spark-submit as --conf k=v verbatim. 

Sensitive vectors that currently pass admission unchanged:
  - spark.ssl.enabled=false — TLS downgrade
  - spark.kubernetes.authenticate.driver.oauthTokenFile=/path — arbitrary token path
  - spark.kubernetes.authenticate.driver.serviceAccountName=cluster-admin — SA override / privilege escalation
  - spark.kubernetes.namespace=<other> — namespace override
  - spark.master / spark.kubernetes.driver.master — redirect to external cluster (note: k8s://... happens to be blocked today only because k8s isn't an allowed URL scheme, not because the key is guarded)

Potentially more.

### Describe the solution you would like

Add a key-level policy to the validating webhook, alongside the existing value-scheme check.

1. New validation method in SparkApplicationValidator (internal/webhook/sparkapplication_validator.go). Iterate app.Spec.SparkConf.

1. Policy model — propose/spec one of:
  - Denylist of exact keys + key prefixes that are rejected outright.
  - Pinned keys the operator force-sets/overrides regardless of user input (e.g. namespace must equal app.Namespace)
  - Value constraints (e.g. spark.ssl.enabled must not be false).
  - Probably a mix. Spec the data structure and where defaults live.
 
2. Keys to cover — constants already exist in pkg/common/spark.go, reuse them, don't hardcode strings: - SparkKubernetesAuthenticateDriverServiceAccountName (spark.kubernetes.authenticate.driver.serviceAccountName)
  - SparkKubernetesAuthenticateExecutorServiceAccountName
  - SparkKubernetesNamespace (spark.kubernetes.namespace)
  - SparkKubernetesDriverMaster (spark.kubernetes.driver.master)
  - spark.master and the spark.kubernetes.authenticate.* family + spark.ssl.* — no constants exist yet
  - Decide policy for prefixes like spark.kubernetes.authenticate. wholesale vs. specific leaf keys.

3. Configuration surface
[...]

PR #3043, opened 2026-07-14
feat(webhook): add sparkConf key validation

sahana-sreeram:
Blocks dangerous sparkConf keys in the validating webhook to prevent security-sensitive overrides due to last-value-wins in spark-submit. Applied to SparkApplication, ScheduledSparkApplication, and SparkConnect validators
This addresses #3006. 

## Purpose of this PR

The operator builds spark-submit arguments in a fixed order (defined in `submission.go`), which is last-value-wins. The user could set entries in sparkConf that applied after the operator-set values like namespace, service account setup, and container image config. Therefore, a tenant could silently override these by setting the same key in sparkConf (e.g. a user could escalate privileges with just a config change in sparkConf).

This PR adds key-level validation to the webhook so these keys are rejected at admission before they reach the controller. Every blocked key already has a dedicated CRD field alternative, and the error message tells the user exactly which one to use. This is applied to SparkApplication, ScheduledSparkApplication, and SparkConnect, none of which had this validation before.

**Proposed changes:**

- Add validateSparkConf to block 14 sparkConf keys that can override operator-managed values due to spark-submit argument ordering
- Blocked keys: service account names (driver, executor), OAuth token/tokenFile (submission client, driver, executor), master URLs, container image keys, and namespace (pinned to app namespace)
- Wire validation into all three CRD validators
- Unit tests covering all 14 attack vectors, update rejection, and benign-key pass-through across all three CRDs 
- E2E tests verify the webhook rejects a denied key and accepts a safe key against a live cluster.

## Change Category

- [x] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

Users who currently set any of these keys via sparkConf will need to move the value to the corresponding CRD field (a one-line manifest change per key, with clear instructions in the error msg), but allowing these keys to pass through is the vulnerability itself, so a breaking change seems to be the only way to fix it.

### Rationale

The vulnerability exists because sparkConfOption runs after the namespaceOption (in position 6), imageOption(position 7), and other operator-related options in optionFuncs. The denylist uses specific key enumeration instead of prefix matching to avoid blocking cert-related authenticate keys that would not have a CRD alternative. SSL blocking was excluded dude Istio users would need to legitimately disable the built-in SSL, so it wouldn't fit into this current architecture. 

## Checklist

### Additional Notes

Denylist:
- spark.kubernetes.authenticate.driver.serviceAccountName / executor.serviceAccountName - SA override (use spec.driver.serviceAccount)
- spark.kubernetes.authenticate.oauthToken / oauthTokenFile (submission, driver, executor variants) - credential injection (use Kubernetes service accounts)
- spark.master / spark.kubernetes.driver.master - cluster redirect (managed by operator)
- spark.kubernetes.container.image / driver / executor variants - image override (use spec.image)
- spark.kubernetes.namespace - value allowed only if value matches the app's namespace


internal/webhook/sparkapplication_validator.go
```diff
@@ -175,6 +180,32 @@ func (v *SparkApplicationValidator) validateSparkVersion(app *v1beta2.SparkAppli
 	return nil
 }
 
+func validateSparkConf(sparkConf map[string]string, namespace string) error {
+	for key, value := range sparkConf {
+		switch key {
+		case common.SparkKubernetesAuthenticateDriverServiceAccountName,
+			common.SparkKubernetesAuthenticateExecutorServiceAccountName:
+			return fmt.Errorf("sparkConf key %q is not allowed: use the serviceAccount field on the CRD instead", key)
```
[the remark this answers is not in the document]
vjanelle: Instead of fmt.Errorf, can you use a typed Error?

internal/webhook/sparkapplication_validator.go
```diff
@@ -175,6 +180,32 @@ func (v *SparkApplicationValidator) validateSparkVersion(app *v1beta2.SparkAppli
...
+		case common.SparkKubernetesAuthenticateOAuthTokenFile,
+			common.SparkKubernetesAuthenticateOAuthToken,
+			common.SparkKubernetesAuthenticateDriverOAuthTokenFile,
+			common.SparkKubernetesAuthenticateDriverOAuthToken,
+			common.SparkKubernetesAuthenticateExecutorOAuthTokenFile,
+			common.SparkKubernetesAuthenticateExecutorOAuthToken:
+			return fmt.Errorf("sparkConf key %q is not allowed: use the serviceAccount field on the CRD instead", key)
```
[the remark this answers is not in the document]
vjanelle: If we're grouping keys with error messages, it might be better to use a data structure for mapping these out. Otherwise this case statement is going to potentially grow over time.

It could also allow for configurable keys.

pkg/common/spark.go
```diff
@@ -51,8 +51,24 @@ const (
...
 )
 
+// Spark core properties.
+const (
+	SparkMaster = "spark.master"
+)
+
 // Spark on Kubernetes properties.
 const (
+	SparkKubernetesAuthenticateOAuthTokenFile = "spark.kubernetes.authenticate.oauthTokenFile"
```
vjanelle: Potentially put this into a different file

vjanelle (changes requested):
Typed errors are the main thing.

sahana-sreeram:
Addressed in [159212a](https://github.com/kubeflow/spark-operator/pull/3043/commits/159212a1f515394f2974dfe2b492812955e1350b)!

vjanelle (approved):

pkg/common/spark_auth.go
nabuskey, line 1: Any particular reason these are in a dedicated file? I much prefer to keep constants in spark.go

sahana-sreeram: Fixed in [e6debd2](https://github.com/kubeflow/spark-operator/pull/3043/commits/e6debd2f6daa2255764f20b2a02fe8cc24a69a17) to match constants staying in spark.go

nabuskey (approved):
Thank you very much for the PR! This is definitely needed.

/approve

PR #3047, opened 2026-07-17
test(e2e): add Prometheus metrics and Spark UI e2e tests

sahana-sreeram:
## Purpose of this PR

Adds two E2E tests covering operator functionality that previously had no coverage - verifying the Prometheus metrics endpoint reports correct data and that the Spark UI is actually reachable while a job runs. Both are fully platform-agnostic, with just Kubernetes primitives. 

**Proposed changes:**

- Add `test/e2e/prometheus_metrics_test.go`which finds the controller pod, verifies `/metrics` serves valid Prometheus-formatted data, then runs a real SparkApplication and confirms that Spark counters increment correctly
- Add `test/e2e/spark_ui_test.go`, which runs a SparkApplication, verifies the operator creates a UI Service on port 4040, then port-forwards directly to the driver pod and makes a real HTTP request to confirm the Spark UI actually renders

## Change Category

- [ ] Documentation update

### Rationale

The existing tests check that Kubernetes objects get created, exists, and are Ready; but they don't verify that the features behind them work E2E. Adding these two tests will close that gap for observability via Prometheus metrics and the Spark UI for job monitoring. Neither test requires any new cluster, image, or build setup and should not add much overhead to the CI. 

## Checklist

### Additional Notes

Verified test behavior locally with `make e2e-test`. The Spark UI test's port-forward step includes a bounded timeout on both the connection-ready wait and the HTTP request, so a stalled port-forward fails and retries cleanly within the existing `WaitTimeout` window instead of risking an indefinite hang.

### DIFF
--- a/test/e2e/prometheus_metrics_test.go
+++ b/test/e2e/prometheus_metrics_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2026 The Kubeflow authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e_test
+
+import (
+	"context"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	coordinationv1 "k8s.io/api/coordination/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
+)
+
+var _ = Describe("Prometheus Metrics", func() {
+	Context("Controller metrics endpoint", func() {
+		ctx := context.Background()
+		path := filepath.Join("..", "..", "examples", "spark-pi.yaml")
+
+		var app *v1beta2.SparkApplication
+
+		BeforeEach(func() {
+			By("Parsing SparkApplication from file")
+			file, err := os.Open(path)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
+			defer func() { Expect(file.Close()).To(Succeed()) }()
+
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 100)
+			Expect(decoder).NotTo(BeNil())
+
+			app = &v1beta2.SparkApplication{}
+			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
+			app.Name = "spark-pi-metrics-test"
+		})
+
+		AfterEach(func() {
+			key := types.NamespacedName{Namespace: app.Namespace, Name: app.Name}
+			if err := k8sClient.Get(ctx, key, app); err == nil {
+				By("Deleting SparkApplication")
+				Expect(k8sClient.Delete(ctx, app)).To(Succeed())
+			}
+		})
+
+		It("Should serve Prometheus metrics including Spark application metrics", func() {
+			By("Finding the controller pod")
+			pods := &corev1.PodList{}
+			Expect(k8sClient.List(ctx, pods,
+				client.InNamespace(ReleaseNamespace),
+				client.MatchingLabels{
+					"app.kubernetes.io/name":      "spark-operator",
+					"app.kubernetes.io/component": "controller",
+				},
+			)).To(Succeed())
+			Expect(pods.Items).NotTo(BeEmpty(), "controller pod not found")
+
+			By("Identifying the leader controller pod via Lease")
+			controllerPod := pods.Items[0]
+			lease := &coordinationv1.Lease{}
+			if err := k8sClient.Get(ctx, types.NamespacedName{
+				Namespace: ReleaseNamespace,
+				Name:      "spark-operator-controller-lock",
+			}, lease); err == nil && lease.Spec.HolderIdentity != nil {
+				for _, pod := range pods.Items {
+					if pod.Name == *lease.Spec.HolderIdentity {
+						controllerPod = pod
+						break
+					}
+				}
+			}
+
+			By("Detecting metrics scheme and port from controller deployment")
+			deploy := &appsv1.Deployment{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{
+				Namespace: ReleaseNamespace,
+				Name:      "spark-operator-controller",
+			}, deploy)).To(Succeed())
+
+			metricsScheme := "http"
+			metricsPort := "8080"
+			for _, arg := range deploy.Spec.Template.Spec.Containers[0].Args {
+				if arg == "--secure-metrics" || arg == "--secure-metrics=true" {
+					metricsScheme = "https"
+				}
+				if strings.HasPrefix(arg, "--metrics-bind-address=") {
+					addr := strings.TrimPrefix(arg, "--metrics-bind-address=")
+					if _, port, err := net.SplitHostPort(addr); err == nil && port != "" {
+						metricsPort = port
+					}
+				}
+			}
+
+			By("Verifying the metrics endpoint serves Prometheus-formatted data")
+			data, err := clientset.CoreV1().Pods(ReleaseNamespace).
+				ProxyGet(metricsScheme, controllerPod.Name, metricsPort, "metrics", nil).
+				DoRaw(ctx)
+			Expect(err).NotTo(HaveOccurred(), "failed to proxy GET /metrics from controller pod")
+
+			metricsOutput := string(data)
+			Expect(metricsOutput).To(ContainSubstring("# HELP"))
+			Expect(metricsOutput).To(ContainSubstring("# TYPE"))
+			Expect(metricsOutput).To(ContainSubstring("go_goroutines"))
+			Expect(metricsOutput).To(ContainSubstring("process_cpu_seconds_total"))
+
+			By("Creating SparkApplication to exercise the metrics pipeline")
+			Expect(k8sClient.Create(ctx, app)).To(Succeed())
+
+			By("Waiting for SparkApplication to complete")
+			key := types.NamespacedName{Namespace: app.Namespace, Name: app.Name}
+			Expect(waitForSparkApplicationCompleted(ctx, key)).NotTo(HaveOccurred())
+
+			By("Verifying Spark application metrics are present after app completion")
+			Eventually(func() string {
+				data, err := clientset.CoreV1().Pods(ReleaseNamespace).
+					ProxyGet(metricsScheme, controllerPod.Name, metricsPort, "metrics", nil).
+					DoRaw(ctx)
+				if err != nil {
+					return ""
+				}
+				return string(data)
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(And(
+				ContainSubstring("spark_application_count"),
+				ContainSubstring("spark_application_submit_count"),
+				ContainSubstring("spark_application_success_count"),
+			))
+		})
+	})
+})
--- a/test/e2e/prometheus_metrics_test.go
+++ b/test/e2e/prometheus_metrics_test.go
@@ -17,10 +17,14 @@ limitations under the License.
 package e2e_test
 
 import (
+	"bufio"
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -34,6 +38,29 @@ import (
 
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
 )
+
+// sumMetricValues parses Prometheus text-exposition output and returns the
+// sum of all sample values for the given metric name, across all label combinations.
+func sumMetricValues(metricsText, name string) float64 {
+	pattern := regexp.MustCompile(`^` + regexp.QuoteMeta(name) + `(\{[^}]*\})?\s+(\S+)$`)
+
+	var total float64
+	scanner := bufio.NewScanner(strings.NewReader(metricsText))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+		matches := pattern.FindStringSubmatch(line)
+		if matches == nil {
+			continue
+		}
+		if value, err := strconv.ParseFloat(matches[2], 64); err == nil {
+			total += value
+		}
+	}
+	return total
+}
 
 var _ = Describe("Prometheus Metrics", func() {
 	Context("Controller metrics endpoint", func() {
@@ -54,7 +81,7 @@ var _ = Describe("Prometheus Metrics", func() {
 
 			app = &v1beta2.SparkApplication{}
 			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
-			app.Name = "spark-pi-metrics-test"
+			app.Name = fmt.Sprintf("spark-pi-metrics-test-%d", GinkgoRandomSeed())
 		})
 
 		AfterEach(func() {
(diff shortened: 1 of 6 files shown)


test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+			By("Identifying the leader controller pod via Lease")
+			controllerPod := pods.Items[0]
+			lease := &coordinationv1.Lease{}
+			if err := k8sClient.Get(ctx, types.NamespacedName{
+				Namespace: ReleaseNamespace,
+				Name:      "spark-operator-controller-lock",
+			}, lease); err == nil && lease.Spec.HolderIdentity != nil {
+				for _, pod := range pods.Items {
+					if pod.Name == *lease.Spec.HolderIdentity {
```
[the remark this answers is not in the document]
RobuRishabh: Copilot is right here, Leader election identity mismatch. `pod.Name == *lease.Spec.HolderIdentity` will never match because controller-runtime sets the identity to `<hostname>_<UUID>`. Use `strings.HasPrefix(*lease.Spec.HolderIdentity, pod.Name+"_")` instead. Currently works by accident in single-replica mode (falls through to `pods.Items[0]`), but the dead code is misleading.

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,151 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
+			defer func() { Expect(file.Close()).To(Succeed()) }()
+
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 100)
+			Expect(decoder).NotTo(BeNil())
+
+			app = &v1beta2.SparkApplication{}
+			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
+			app.Name = "spark-pi-metrics-test"
```
RobuRishabh: Static test name risks collision. `app.Name = "spark-pi-metrics-test"` is a hardcoded name. If Ginkgo parallelism is ever enabled, this could collide. Consider using `GinkgoRandomSeed()` like the UI test does, or accept the tradeoff and note it.

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,151 @@
...
+					DoRaw(ctx)
+				if err != nil {
+					return ""
+				}
+				return string(data)
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(And(
+				ContainSubstring("spark_application_count"),
+				ContainSubstring("spark_application_submit_count"),
+				ContainSubstring("spark_application_success_count"),
+			))
```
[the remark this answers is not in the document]
RobuRishabh: Copilot is right here, Consider asserting non-zero values, e.g., MatchRegexp("spark_application_success_count\\{[^}]*\\} [1-9]"), or capture values before/after and assert a delta.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
+			Expect(decoder).NotTo(BeNil())
+
+			app = &v1beta2.SparkApplication{}
+			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
+
+			// Use a unique name and large partition count so SparkPi runs long enough
+			// for the test to query the UI while it's still in the Running state.
+			app.Name = fmt.Sprintf("spark-pi-ui-test-%d", GinkgoRandomSeed())
+			app.Spec.Arguments = []string{"1000000"}
```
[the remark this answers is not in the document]
RobuRishabh: 1M partition can be little excessive I think. This will be very slow and could cause timeouts. Try "100000" or "50000" and test in KIND to find the minimum value that keeps the job running long enough for the port-forward check (~10-15 seconds).

RobuRishabh (changes requested):

RobuRishabh:
@tariq-hasan @nabuskey any final comments from your side on this PR ?

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+			Expect(hasUIPort).To(BeTrue(), "UI service should expose port 4040")
+
+			By("Verifying the Spark UI is responding on port 4040")
+			Eventually(func(g Gomega) {
+				transport, upgrader, err := spdy.RoundTripperFor(cfg)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				reqURL := clientset.CoreV1().RESTClient().Post().
+					Resource("pods").
```
nabuskey: are you testing the pod or service? this looks like port forwarding to pod instead of service?

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
+			Expect(decoder).NotTo(BeNil())
+
+			app = &v1beta2.SparkApplication{}
+			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
+
+			// Use a unique name and a partition count high enough that SparkPi keeps
+			// running for the few seconds it takes the test to query the UI.
+			app.Name = fmt.Sprintf("spark-pi-ui-test-%d", GinkgoRandomSeed())
+			app.Spec.Arguments = []string{"50000"}
```
nabuskey: Instead of trying to run a large number of partitions which isn't a reliable way to keep it running, can you use a more explicit waiting with time.sleep in a pyspark script?

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+				g.Expect(err).NotTo(HaveOccurred())
+
+				reqURL := clientset.CoreV1().RESTClient().Post().
+					Resource("pods").
+					Namespace(app.Namespace).
+					Name(driverPodName).
+					SubResource("portforward").
+					URL()
+
+				dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, http.MethodPost, reqURL)
```
nabuskey: we need to set a http client timeout here. It's possible `case <-time.After(15 * time.Second):` hits, but the dialer is still blocked. so the go routine keeps running.

nabuskey (commented):
I think it makes sense to use the API proxy instead of port-forwarding for these tests.

https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls

This way we don't have to worry about race. You can use context timeout + http client. Easier to reason about too.

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,187 @@
+/*
+Copyright 2026 The Kubeflow authors.
+
```
tariq-hasan, line 3: As per [CNCF/LF guidance](https://github.com/kubernetes/steering/issues/299) new files should not have year in the copyright header.

```suggestion
/*
Copyright The Kubeflow Authors.

```

There has already been push in the adjacent Trainer and SDK projects to remove year from all future copyright texts. The Trainer project already has a canonical copyright enforcement [workflow](https://github.com/kubeflow/trainer/pull/3359) that could be leveraged through a common infra setup for Kubeflow projects.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2026 The Kubeflow authors.
+
```
tariq-hasan, line 3:
```suggestion
/*
Copyright The Kubeflow Authors.

```

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,187 @@
...
+				Namespace: ReleaseNamespace,
+				Name:      "spark-operator-controller",
+			}, deploy)).To(Succeed())
+
+			metricsScheme := "http"
+			metricsPort := "8080"
+			for _, arg := range deploy.Spec.Template.Spec.Containers[0].Args {
+				if arg == "--secure-metrics" || arg == "--secure-metrics=true" {
+					metricsScheme = "https"
+				}
```
tariq-hasan: The `--secure-metrics` branch is never really exercised in CI and exercising this path would also require auth setup.

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,187 @@
...
+					addr := strings.TrimPrefix(arg, "--metrics-bind-address=")
+					if _, port, err := net.SplitHostPort(addr); err == nil && port != "" {
+						metricsPort = port
+					}
+				}
+			}
+
+			By("Verifying the metrics endpoint serves Prometheus-formatted data")
+			data, err := clientset.CoreV1().Pods(ReleaseNamespace).
+				ProxyGet(metricsScheme, controllerPod.Name, metricsPort, "metrics", nil).
```
tariq-hasan: I'm thinking it may make sense to not hardcode the endpoint name or parse the container arguments but rather use the pod [annotations](https://github.com/sahana-sreeram/spark-operator-testing/blob/306426e1fba9dc9879ce396b2fd7c4edb7f11ab3/pkg/common/prometheus.go#L31-L32) that the chart [uses](https://github.com/sahana-sreeram/spark-operator-testing/blob/306426e1fba9dc9879ce396b2fd7c4edb7f11ab3/charts/spark-operator-chart/templates/controller/deployment.yaml#L44-L45) directly?

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,187 @@
...
+				data, err := clientset.CoreV1().Pods(ReleaseNamespace).
+					ProxyGet(metricsScheme, controllerPod.Name, metricsPort, "metrics", nil).
+					DoRaw(ctx)
+				if err != nil {
+					return false, err
+				}
+				metricsOutput := string(data)
+				return sumMetricValues(metricsOutput, "spark_application_count") > countBefore &&
+					sumMetricValues(metricsOutput, "spark_application_submit_count") > submitCountBefore &&
+					sumMetricValues(metricsOutput, "spark_application_success_count") > successCountBefore, nil
```
tariq-hasan: I'm thinking best to reuse the constants: https://github.com/sahana-sreeram/spark-operator-testing/blob/306426e1fba9dc9879ce396b2fd7c4edb7f11ab3/pkg/common/metrics.go#L21-L29.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			Eventually(func() bool {
+				if err := k8sClient.Get(ctx, key, app); err != nil {
+					return false
+				}
+				return app.Status.AppState.State == v1beta2.ApplicationStateRunning &&
+					app.Status.DriverInfo.WebUIServiceName != ""
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(BeTrue())
+
+			By("Verifying the WebUI status fields are populated")
+			Expect(app.Status.DriverInfo.WebUIPort).To(Equal(int32(4040)))
```
tariq-hasan: Same here for the port: https://github.com/sahana-sreeram/spark-operator-testing/blob/306426e1fba9dc9879ce396b2fd7c4edb7f11ab3/pkg/common/spark.go#L360.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			hasUIPort := false
+			for _, port := range svc.Spec.Ports {
+				if port.Port == 4040 {
+					hasUIPort = true
+					break
+				}
+			}
+			Expect(hasUIPort).To(BeTrue(), "UI service should expose port 4040")
+
+			By("Verifying the Spark UI is responding on port 4040")
```
tariq-hasan: Can we use the [ProxyGet](https://github.com/sahana-sreeram/spark-operator-testing/blob/420ebb42e70fc57d37eb26f5fc6464d30881bb89/test/e2e/prometheus_metrics_test.go#L149) approach and simplify the logic and dependencies here - adapting it for the service as asked below?

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+				if err := k8sClient.Get(ctx, key, app); err != nil {
+					return false
+				}
+				return app.Status.AppState.State == v1beta2.ApplicationStateRunning &&
+					app.Status.DriverInfo.WebUIServiceName != ""
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(BeTrue())
+
+			By("Verifying the WebUI status fields are populated")
+			Expect(app.Status.DriverInfo.WebUIPort).To(Equal(int32(4040)))
+			Expect(app.Status.DriverInfo.WebUIAddress).NotTo(BeEmpty())
```
tariq-hasan: This assertion can never really fail since the address is constructed as with both the service IP and service port: https://github.com/sahana-sreeram/spark-operator-testing/blob/420ebb42e70fc57d37eb26f5fc6464d30881bb89/internal/controller/sparkapplication/controller.go#L390.

We can either assert this structure or just parse and assert just the service IP.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+			By("Parsing SparkApplication from file")
+			file, err := os.Open(path)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(file).NotTo(BeNil())
+			defer func() { Expect(file.Close()).To(Succeed()) }()
+			decoder := yaml.NewYAMLOrJSONDecoder(file, 4096)
+			Expect(decoder).NotTo(BeNil())
+
+			app = &v1beta2.SparkApplication{}
+			Expect(decoder.Decode(app)).NotTo(HaveOccurred())
```
tariq-hasan: I'm thinking probably ideal to standardize this logic as it seems to appear in other places as well: https://github.com/sahana-sreeram/spark-operator-testing/blob/420ebb42e70fc57d37eb26f5fc6464d30881bb89/test/e2e/pdb_test.go#L38-L54.

test/e2e/spark_ui_test.go
```diff
@@ -0,0 +1,159 @@
...
+				httpClient := &http.Client{Timeout: 15 * time.Second}
+				resp, err := httpClient.Get(fmt.Sprintf("http://localhost:%d", ports[0].Local))
+				g.Expect(err).NotTo(HaveOccurred())
+				defer func() { _ = resp.Body.Close() }()
+
+				g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+
+				body, err := io.ReadAll(resp.Body)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(string(body)).To(ContainSubstring("Spark Jobs"))
```
tariq-hasan: Do we want to make this logic more robust by parsing the JSON response from `/api/v1/applications` - which is a more stable contract than parsing the HTML?

tariq-hasan (commented):
Hi @sahana-sreeram! I have added a few additional comments.

sahana-sreeram:
Comments addressed iin [5d75271](https://github.com/sahana-sreeram/spark-operator-testing/commit/5d75271c70c13e92d1ebcf70260cb6b987ce4df1) !

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,165 @@
...
+					ProxyGet(metricsScheme, controllerPod.Name, metricsPort, metricsPath, nil).
+					DoRaw(ctx)
+				if err != nil {
+					return false, err
+				}
+				metricsOutput := string(data)
+				return sumMetricValues(metricsOutput, common.MetricSparkApplicationCount) > countBefore &&
+					sumMetricValues(metricsOutput, common.MetricSparkApplicationSubmitCount) > submitCountBefore &&
+					sumMetricValues(metricsOutput, common.MetricSparkApplicationSuccessCount) > successCountBefore, nil
+			}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(BeTrue())
```
tariq-hasan: I think better to use [this](https://github.com/sahana-sreeram/spark-operator-testing/blob/6b6b7409aae78cfc313d85c8c467b54fcbd8ef36/test/e2e/spark_ui_test.go#L135-L147) pattern here to enable easier debugging in case of errors. This would be using the `Eventually(func(g Gomega) { ... }).Should(Succeed())` pattern with separate `g.Expect(...)` assertions.

test/e2e/prometheus_metrics_test.go
```diff
@@ -0,0 +1,165 @@
...
+		matches := pattern.FindStringSubmatch(line)
+		if matches == nil {
+			continue
+		}
+		if value, err := strconv.ParseFloat(matches[2], 64); err == nil {
+			total += value
+		}
+	}
+	return total
+}
```
tariq-hasan, line 63: For the parser I think better to return `(float64, error)` or `(float64, bool)` to enable better error handling.

Also, better to delegate the regex-handling to the Prometheus parser [package](https://pkg.go.dev/github.com/prometheus/common/expfmt).

tariq-hasan (commented):
Hi @sahana-sreeram! I've added a few additional comments.

tariq-hasan:
Thanks @sahana-sreeram!
/lgtm

sahana-sreeram:
@andreyvelich @nabuskey checking if this can be merged!

nabuskey (approved):
/approve

PR #3051, opened 2026-07-23
fix: preserve shell-sensitive Spark Connect configuration values

jayanth86:
## Purpose of this PR

Spark Connect constructs its server startup command with `bash -c`, so Spark and Hadoop configuration values containing shell metacharacters are currently interpreted by the shell instead of being passed literally.

For example, setting `spark.redaction.regex=(?i)secret|password|token` can prevent a Spark Connect server from starting.

**Proposed changes:**

- Shell-quote user-supplied Spark and Hadoop configuration arguments before adding them to the server startup command.
- Preserve embedded single quotes, whitespace, regular-expression metacharacters, command-like text, newlines, and empty values.
- Add focused regression coverage to the existing Spark Connect Ginkgo suite.

## Change Category

- [ ] Documentation update

### Rationale

Spark and Hadoop configuration values should reach the Spark Connect server unchanged, regardless of characters that have special meaning to the shell.

## Checklist

### Validation

- `go test ./internal/controller/sparkconnect -ginkgo.focus='preserves shell-sensitive'`
- `go test ./internal/controller/sparkconnect/...`
- `go test -race ./internal/controller/sparkconnect/...`
- `go vet ./internal/controller/sparkconnect/...`

### DIFF
--- a/internal/controller/sparkconnect/options.go
+++ b/internal/controller/sparkconnect/options.go
@@ -121,10 +121,14 @@ func sparkConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 	for key, value := range conn.Spec.SparkConf {
 		// Configuration property for the driver pod name has already been set.
 		if key != common.SparkKubernetesDriverPodName {
-			args = append(args, "--conf", fmt.Sprintf("%s=%s", key, value))
+			args = append(args, "--conf", shellQuoteSparkConfig(key, value))
 		}
 	}
 	return args, nil
+}
+
+func shellQuoteSparkConfig(key, value string) string {
+	return "'" + strings.ReplaceAll(key+"="+value, "'", "'\"'\"'") + "'"
 }
 
 func hadoopConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
@@ -135,11 +139,11 @@ func hadoopConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 	// Add Hadoop configuration properties.
 	for key, value := range conn.Spec.HadoopConf {
 		if strings.HasPrefix(key, common.SparkHadoopPropertiesPrefix) {
-			args = append(args, "--conf", fmt.Sprintf("%s=%s", key, value))
+			args = append(args, "--conf", shellQuoteSparkConfig(key, value))
 		} else {
 			// Add prefix to the configuration key if it does not start with `spark.hadoop.`.
 			// Users will be able to use the configuration key with or without prefix.
-			args = append(args, "--conf", fmt.Sprintf("%s%s=%s", common.SparkHadoopPropertiesPrefix, key, value))
+			args = append(args, "--conf", shellQuoteSparkConfig(common.SparkHadoopPropertiesPrefix+key, value))
 		}
 	}
 	return args, nil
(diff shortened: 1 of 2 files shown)


xiaogecai:
Assigned label: `assigned-roburishabh` (most frequent contributor from past 100 commits)

RobuRishabh (approved):
/lgtm

Verified locally, sparkconnect tests (incl. race), go vet, and make unit-test all pass. Thanks for the contribution @jayanth86

RobuRishabh:
@nabuskey can you please trigger the workflow, when you get chance.

internal/controller/sparkconnect/options.go
```diff
@@ -121,12 +121,16 @@ func sparkConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
...
 		if key != common.SparkKubernetesDriverPodName {
-			args = append(args, "--conf", fmt.Sprintf("%s=%s", key, value))
+			args = append(args, "--conf", shellQuoteSparkConfig(key, value))
 		}
 	}
 	return args, nil
 }
 
+func shellQuoteSparkConfig(key, value string) string {
+	return "'" + strings.ReplaceAll(key+"="+value, "'", "'\"'\"'") + "'"
```
nabuskey: This is hard to read. Can you please use fmt instead?

tariq-hasan: I'm actually thinking that the root cause we want to fix relates to the construction of the [container arguments](https://github.com/jayanth86/spark-operator/blob/00ba077967057560ad34949906d5ff4846659ce2/internal/controller/sparkconnect/reconciler.go#L386).

The fact that the arguments are joined together leaves open the interpretation for what constitutes a single argument in the final `container.Args`.

The `SparkApplication` controller avoids this issue entirely with its [`cmd := exec.Command(command, args...)`](https://github.com/jayanth86/spark-operator/blob/00ba077967057560ad34949906d5ff4846659ce2/internal/controller/sparkapplication/submission.go#L77).

Perhaps we just want to do `container.Args = args` with no joining, quoting, etc. and move the [`${SPARK_HOME}/sbin/start-connect-server.sh`](https://github.com/jayanth86/spark-operator/blob/00ba077967057560ad34949906d5ff4846659ce2/internal/controller/sparkconnect/options.go#L44) invocation into [`container.Command`](https://github.com/jayanth86/spark-operator/blob/00ba077967057560ad34949906d5ff4846659ce2/internal/controller/sparkconnect/reconciler.go#L381) - but it would need to be verified how that shape will look like.

RobuRishabh: @tariq-hasan, the refactor you're suggesting is more invasive than it looks, we'd also need to rework `spark.driver.host` (currently relies on shell expansion of `$(POD_IP ...)` at startup) and `${SPARK_HOME}` path handling (PreStop too), plus reconciler/e2e test updates.
For this PR I'd lean toward landing the quoting fix, it addresses the reported bug with minimal risk. 
The structural change is worth a follow-up issue if @nabuskey want to eliminate shell from the startup path entirely. Happy to help sketch that separately.

tariq-hasan: No issues from me. Just wanted to flag the underlying issue.

jayanth86: fixed, please take another look

nabuskey: I agree with Tariq. It should follow the way spark application does it. We need to fix it. https://github.com/kubeflow/spark-operator/issues/3072

nabuskey (approved):

PR #3057, opened 2026-07-28
fix(webhook): fall back to admission request namespace when pod namespace is empty (Kubernetes 1.22 compatibility)

shepherd44:
## Purpose of this PR

Fixes silent no-op of the pod mutating webhook on older Kubernetes API servers (verified on 1.22): driver/executor pods created by spark-submit are admitted without any mutation (no volumes, tolerations, initContainers, etc. from the SparkApplication spec), with no error and no log line.

**Root cause**

`SparkPodDefaulter.Default()` filters pods by `pod.Namespace`:

```go
namespace := pod.Namespace
if !d.isSparkJobNamespace(namespace) {
    return nil
}
```

spark-submit (fabric8 client) POSTs pods without `metadata.namespace` in the request body (the namespace is only in the URL path). On Kubernetes 1.22, the API server does not populate `metadata.namespace` on the object sent to admission webhooks in this case, so `pod.Namespace` is an empty string inside the webhook, `isSparkJobNamespace("")` returns false, and the defaulter returns without mutating. Newer API servers populate the namespace before calling webhooks, which is why this does not reproduce there.

The result is hard to diagnose: the webhook is called (`controller_runtime_webhook_requests_total{webhook="/mutate--v1-pod",code="200"}` increments, `code="500"` stays 0), `failurePolicy: Fail` does not trigger, and nothing is logged.

**Proposed changes:**

- `SparkPodDefaulter.Default()`: when `pod.Namespace` is empty, fall back to the namespace of the admission request (`admission.RequestFromContext`), which is always populated by the API server from the request URL.
- Unit test covering the fallback (empty pod namespace + request namespace in watched/unwatched namespaces).

## Change Category

### Rationale

Environment where this was hit:

- Kubernetes v1.22.9 (on-premises), 3 API servers
- spark-operator 2.4.0 (`ghcr.io/kubeflow/spark-operator/controller:2.4.0`, digest `sha256:067ee917...`), Helm chart 2.4.0, webhook enabled, `spark.jobNamespaces` set to an explicit list. Also reproduced on 2.5.1.
- The same image digest on EKS v1.35.6 (`v1.35.6-eks-8f14419`) mutates the same SparkApplication correctly.

Observed behavior on 1.22:

- Webhook log over 4 days contains 9037 `Defaulting SparkApplication` lines and 0 `Mutating Spark pod` lines, while `/mutate--v1-pod` served 30k+ requests, all HTTP 200, zero 500.
- Driver pod is created with all operator labels present at creation time (single `fabric8-kubernetes-client` entry in `managedFields` containing `sparkoperator.k8s.io/app-name`, `.../launched-by-spark-operator`), scheduled, and runs — but without the tolerations/volumes/initContainers declared in `spec.driver`.
- Re-creating the same driver pod manifest with `kubectl create` (which puts `metadata.namespace` into the body) is mutated correctly by the same webhook instance, on the same cluster, in the same namespace.

Minimal reproduction independent of Spark, against a 1.22 API server with the webhook watching `spark-test`:

```console
# metadata.namespace omitted from the body (as fabric8 does) — silently accepted, no mutation:
$ kubectl create --raw "/api/v1/namespaces/spark-test/pods?dryRun=All" -f pod-without-namespace.json
{"kind":"Pod", ...}

# identical pod with metadata.namespace in the body — webhook processes it
# (denied here because the referenced SparkApplication does not exist):
$ kubectl create --raw "/api/v1/namespaces/spark-test/pods?dryRun=All" -f pod-with-namespace.json
Error from server (Forbidden): admission webhook "mutate--v1-pod.sparkoperator.k8s.io" denied the request:
failed to get SparkApplication spark-test/nonexistent: SparkApplication.sparkoperator.k8s.io "nonexistent" not found
```

Note that setting `spark.jobNamespaces` to include the empty string is not a workaround: `isSparkJobNamespace("")` would pass, but the subsequent `client.Get` with an empty namespace fails and, with `failurePolicy: Fail`, rejects every matching pod.

Current workaround in our cluster until this is fixed: we keep the legacy 1.x operator (chart 1.4.6, image `v1beta2-1.6.2-3.5.0`) running purely as the pod mutation engine — its webhook uses a label-based `namespaceSelector` and no in-process namespace re-check, so it is not affected — while the 2.x operator handles reconciliation. This "2.x controller + 1.x webhook" split works but blocks retiring the 1.x deployment.

## Checklist

- [x] I have updated documentation accordingly (code comment).

### DIFF
--- a/internal/webhook/sparkpod_defaulter.go
+++ b/internal/webhook/sparkpod_defaulter.go
@@ -76,6 +76,16 @@ func (d *SparkPodDefaulter) Default(ctx context.Context, pod *corev1.Pod) error 
 
 	logger := log.FromContext(ctx)
 	namespace := pod.Namespace
+	if namespace == "" {
+		// On older API servers (verified on Kubernetes 1.22), the object sent to
+		// admission webhooks does not have metadata.namespace populated when the
+		// client omitted it from the request body, which is the case for driver
+		// and executor pods created by spark-submit (fabric8 client). Fall back
+		// to the namespace of the admission request itself.
+		if req, err := admission.RequestFromContext(ctx); err == nil {
+			namespace = req.Namespace
+		}
+	}
 	if !d.isSparkJobNamespace(namespace) {
 		return nil
 	}
--- a/internal/webhook/sparkpod_defaulter_test.go
+++ b/internal/webhook/sparkpod_defaulter_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package webhook
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
@@ -24,9 +25,13 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/stretchr/testify/assert"
+	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
 	"github.com/kubeflow/spark-operator/v2/pkg/common"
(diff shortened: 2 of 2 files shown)


sxivansx:
Checked this against master. `pod.Namespace` is read exactly once, at `sparkpod_defaulter.go:78`, and nothing downstream reads it, so the fallback covers the whole path. Not setting `pod.Namespace` is the right call either way, since namespace is immutable in an admission patch.

On whether 1.22 is worth carrying a fix for: the README version matrix lists Kubernetes `1.16+` for every 2.x release, so this is inside the project's own stated support range.

`admission.RequestFromContext` is present in the controller-runtime we vendor (v0.23.3). Your test passes and the whole webhook package is green on the branch.

Two things:

`internal/webhook/sparkpod_defaulter_test.go` conflicts with master now, so this needs a rebase before it can merge.

The error from `RequestFromContext` is dropped, so if it ever fails the namespace stays empty and you get back the same silent no-op this fixes. A log line when the fallback actually triggers would also give the next person on an old API server something to grep for — by your own account that was the hardest part to diagnose.

I'm not a maintainer, so this still needs a reviewer.

RobuRishabh (commented):
Hi @shepherd44 Thanks for the contribution, can you please rebase the PR. 
I have some minor suggestions:
1. If `RequestFromContext` fails, we silently keep the old no-op behavior, so consider logging or returning the error so misconfiguration is visible.
2. A `V(1)` log when the fallback actually triggers would help the next person debugging on 1.22

Rest looks good to me.

shepherd44:
@RobuRishabh @sxivansx Thanks both. Rebased onto master and force-pushed — the conflict was just the import block in `sparkpod_defaulter_test.go`.

Both suggestions applied:

- The fallback now logs at `V(1)` when it actually triggers, with the resolved namespace.
- The `RequestFromContext` error is returned instead of dropped. controller-runtime sets the request on the context unconditionally before calling `Default` (`defaulter_custom.go`), so a missing request is a programming error rather than something an old API server can produce — returning it makes that loud instead of reproducing the silent no-op this PR fixes.

While in there I also reworded the comment to cite kubernetes/kubernetes#94637 rather than "verified on 1.22". The apiserver only started setting the object namespace before admission in v1.24, so everything below that is affected, not just 1.22. Also added tests for the `NamespaceAll` case, pod-namespace precedence over the request namespace, and the missing-request branch.

RobuRishabh:
Thanks 
/lgtm

Issue #3076, opened 2026-08-07
Restore test coverage reporting to Kubeflow standards


vjanelle:
### What happened?

- [x] I have searched the open/closed issues and my issue is not listed.

Spark Operator runs `make unit-test` in CI and generates `cover.out` plus `cover.html`, but CI does not publish the coverage profile or report coverage status. The files disappear with the runner, so maintainers cannot track coverage trends or detect regressions from pull requests.

CI also does not enforce minimum coverage for newly added or changed code. New behavior can therefore merge without meaningful test coverage even when repository-wide coverage remains unchanged or increases because of unrelated packages.

Kubeflow Trainer publishes its Go coverage profile to Coveralls from its test workflow. Spark Operator should restore equivalent Kubeflow project coverage reporting, establish an agreed repository baseline, and enforce an agreed minimum coverage threshold for new code.

The old #57 tracked improving unit test coverage and was closed in 2018. This issue tracks the current CI reporting and enforcement gap.

### Reproduction Code

1. Open `.github/workflows/integration.yaml`.
2. Observe that the `build-spark-operator` job runs `make unit-test`.
3. Observe that no later step uploads `cover.out` to Coveralls or another project-standard coverage service.
4. Open a pull request that adds untested code or reduces coverage; CI provides no coverage delta, new-code coverage result, or failing coverage status.

### Expected behavior

- CI publishes `cover.out` using the coverage service and workflow pattern adopted by Kubeflow projects.
- Pull requests show repository coverage, coverage delta, and coverage for newly added or changed code.
- CI enforces an agreed minimum coverage percentage for new code and fails pull requests that do not meet it.
- Repository documentation displays current coverage status and documents the enforced threshold where appropriate.
- Maintainers define an initial repository baseline and a policy that prevents unintentional overall coverage regressions without blocking incremental test improvements.
- Coverage enforcement supports explicit, reviewed exclusions for generated or otherwise non-testable code rather than silently lowering the threshold.
- Coverage collection remains compatible with envtest-backed packages and excludes only packages intentionally omitted today, such as e2e and drift tests.

### Actual behavior

`make unit-test` generates local coverage files, but CI neither preserves nor reports them. Coverage regressions and insufficient coverage of new code are invisible during review.

### Environment & Versions

- Kubernetes Version: N/A
- Spark Operator Version: master
- Apache Spark Version: N/A

### Additional context

[...]

PR #3077, opened 2026-08-07
ci: restore test coverage reporting

HarshPopat23:
## What this PR does / why we need it

Spark Operator already generates `cover.out` as part of `make unit-test`, but the coverage profile is currently discarded when CI finishes. This means coverage trends and regressions are not visible during review.

This PR restores coverage reporting following the existing Kubeflow Trainer workflow pattern.

### Changes

* Uploads the existing `cover.out` profile to Coveralls after Go unit tests.
* Uses the same pinned `actions-goveralls` action and configuration as Kubeflow Trainer.
* Adds a Coveralls coverage badge to the README.
* Keeps the existing unit-test coverage collection and package exclusions unchanged.

### Coverage enforcement

This PR intentionally does not hard-code an arbitrary coverage threshold.

The first successful report on `master` should establish the repository baseline. Maintainers can then enable Coveralls status checks and configure the agreed coverage/decrease thresholds in the repository's Coveralls settings, as proposed in #3076.

A local Windows run produced an approximate 34.3% profile, but I am not treating that as the repository baseline because the envtest suites fail during `AfterSuite` teardown on native Windows when stopping `kube-apiserver` and `etcd`. The test specs themselves completed successfully; the authoritative baseline should come from the Linux CI run.

### Validation

* `git diff --check`
* GitHub Actions workflow YAML parses successfully.
* Verified the coverage upload configuration against Kubeflow Trainer's current Go test workflow.
* Locally exercised the existing unit-test coverage path with Kubernetes 1.35 envtest assets; native Windows envtest teardown is unsupported, so the GitHub-hosted Linux run will provide the authoritative full result.

Related to #3076

### Scope and operational prerequisite

This PR restores coverage profile reporting only.

Before uploads and the README badge can become functional, an authorized maintainer must enable `kubeflow/spark-operator` in Coveralls. The badge may display `unknown` until the first successful upload from `master`.

Coverage thresholds and reviewed exclusions for generated or non-testable code will be handled separately after the Linux CI baseline is available.

Coveralls repository enablement is tracked in #3089

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -117,6 +117,12 @@ jobs:
       - name: Run go unit tests
         run: make unit-test
 
+      - name: Coveralls report
+        uses: shogo82148/actions-goveralls@9606dbc5ac5cf888a0e9ef901515c3cd516a2790 # v1
+        continue-on-error: true
+        with:
+          path-to-profile: cover.out
+
       - name: Build Spark operator
         run: make build-operator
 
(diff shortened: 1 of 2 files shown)


HarshPopat23:
Thanks @tariq-hasan ,
Waiting for further guide.

PR #3078, opened 2026-08-08
feat(ci): Publish Spark Operator binaries with GitHub releases

### LINKED ISSUE #3073: Publish Spark Operator binaries with GitHub releases

Build and attach versioned Spark Operator Linux binaries to every GitHub release so users can construct their own minimal or organization-approved container images without compiling the operator from source.

At minimum, publish statically linked `spark-operator` binaries for `linux/amd64` and `linux/arm64` using the same source, version metadata, and linker flags as official controller images.

### Why is this needed?

Current releases publish the Helm chart and official container images, but not the operator executable. Organizations that require a custom base image, internal image hardening, approved CA bundles, or private registry conventions must install the Go toolchain and reproduce the project build to create an image.

Publishing official binaries provides a verifiable build artifact and makes custom Dockerfiles straightforward while retaining upstream version metadata.

- Extend the release workflow to cross-build `spark-operator` for `linux/amd64` and `linux/arm64` through the existing `make build-operator` path or an equivalent reproducible target.
- Preserve current static linking and version linker flags.
- Package binaries in architecture-specific archives with predictable names.
- Attach archives and a `SHA256SUMS` file to GitHub releases before publication.
- Publish provenance and signatures according to Kubeflow supply-chain standards.
- Verify each binary with `spark-operator version` and a basic startup/help smoke test.
- Document a minimal custom Dockerfile using the released binary.
- Keep official container images as the recommended default distribution.

- Extract the binary from the official controller image with a multi-stage Docker build. This ties custom builds to Docker tooling and is less convenient for verification and non-Docker builders.
- Compile from source. This requires a Go toolchain and makes exact reproduction of release metadata the user's responsibility.
- Publish binaries for every desktop platform. Initial scope can remain Linux because the primary use case is building controller images.

### Additional context

`make build-operator` already produces a statically linked `bin/spark-operator` with release version metadata.

Current release workflow attaches only packaged Helm charts: https://github.com/kubeflow/spark-operator/blob/master/.github/workflows/release.yaml

Related release automation: #3063

Historical #1009 requested a release binary for `sparkctl`; this request is specifically for the current `spark-operator` executable.

### Love this feature?

vikas-saxena02:
## Purpose of this PR

Closes #3073

Releases currently publish Helm charts and container images but not the operator
executable. Organizations that need a custom base image, internal hardening, an approved
CA bundle, or private registry conventions must install a Go toolchain and reproduce the
project build, taking on responsibility for matching release version metadata themselves.

**Proposed changes:**

- Cross-build `spark-operator` for `linux/amd64` and `linux/arm64` in the `draft_release` job via the existing `make build-operator` path
- Package each architecture as `spark-operator_<version>_linux_<arch>.tar.gz` containing the binary and `LICENSE`
- Generate a `SHA256SUMS` file covering both archives
- Smoke-test the artifacts before publication
- Attach archives and checksums to the draft release alongside the Helm chart packages
- Document building a custom operator image from the released binary

## Change Category

- [x] Documentation update

### Rationale

Building through `make build-operator` means the published binaries carry the same
`LDFLAGS` as the official controller images, so `spark-operator version` reports
identical version, commit, tree state and build date.

**Step ordering is load-bearing.** The binary build runs before `Package Helm charts`.
That step writes `*.tgz` into the repository root, which is not gitignored, so
`git status --porcelain` becomes non-empty and the Makefile would stamp
`GIT_TREE_STATE=dirty` into every published binary. There is a comment in the workflow to
prevent this being reordered later.

**Verification is asymmetric.** The amd64 binary runs natively on the runner and is
exercised with `spark-operator version` and `--help`. arm64 cannot execute on an x86
runner without emulation, so it is checked with `file` for the correct target
architecture and static linking. Happy to add a QEMU step if reviewers would prefer a
real arm64 execution test, though it adds runner time to every release.

**Single job, no artifact passing.** Cross-compiling two `CGO_ENABLED=0` binaries takes
well under a minute, so this stays inside `draft_release` rather than adding a matrix job
plus an upload/download-artifact round trip. Happy to split it out if maintainers prefer.

**Scoped out:** the issue also asks for provenance and signatures per Kubeflow
supply-chain standards. The repository has no signing or SBOM tooling in any workflow
today, so adding cosign here would mean deciding keyless vs. keyed signing and OIDC
configuration inside an otherwise mechanical change. Suggest tracking that separately;
happy to follow up.

## Checklist

### Additional Notes

The `Release` workflow triggers on pushes to `release-*` that touch `VERSION`, so it
cannot execute on a fork PR. I have validated the packaging and checksum logic locally
and confirmed the step order parses as intended, but the job is untested end to end.
Reviewers may want to dry-run `draft_release` before merge.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -402,6 +402,78 @@ jobs:
           VERSION=$(cat VERSION)
           echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
+      - name: Set up Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version-file: go.mod
+
+      # NOTE: this must run before "Package Helm charts". That step writes
+      # *.tgz into the repository root, which is not gitignored, so
+      # `git status --porcelain` becomes non-empty and the Makefile would
+      # stamp the binaries with GIT_TREE_STATE=dirty.
+      - name: Build operator binaries
+        run: |
+          set -euo pipefail
+          for arch in amd64 arm64; do
+            stage="dist/linux_${arch}"
+            mkdir -p "${stage}"
+            echo "Building linux/${arch}..."
+            GOOS=linux GOARCH="${arch}" \
+              make build-operator SPARK_OPERATOR="${PWD}/${stage}/spark-operator"
+            cp LICENSE "${stage}/"
+            tar -czf "dist/spark-operator_${VERSION}_linux_${arch}.tar.gz" \
+              -C "${stage}" spark-operator LICENSE
+          done
+
+      - name: Generate checksums
+        run: |
+          set -euo pipefail
+          cd dist
+          sha256sum spark-operator_*.tar.gz > SHA256SUMS
+          cat SHA256SUMS
+
+      - name: Set up Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version-file: go.mod
+
+      # NOTE: this must run before "Package Helm charts". That step writes
+      # *.tgz into the repository root, which is not gitignored, so
+      # `git status --porcelain` becomes non-empty and the Makefile would
+      # stamp the binaries with GIT_TREE_STATE=dirty.
+      - name: Build operator binaries
+        run: |
+          set -euo pipefail
+          for arch in amd64 arm64; do
+            stage="dist/linux_${arch}"
+            mkdir -p "${stage}"
+            echo "Building linux/${arch}..."
+            GOOS=linux GOARCH="${arch}" \
+              make build-operator SPARK_OPERATOR="${PWD}/${stage}/spark-operator"
+            cp LICENSE "${stage}/"
+            tar -czf "dist/spark-operator_${VERSION}_linux_${arch}.tar.gz" \
+              -C "${stage}" spark-operator LICENSE
+          done
+
+      - name: Generate checksums
+        run: |
+          set -euo pipefail
+          cd dist
+          sha256sum spark-operator_*.tar.gz > SHA256SUMS
+          cat SHA256SUMS
+
+      - name: Verify binaries
+        run: |
+          set -euo pipefail
+          tar -xzf "dist/spark-operator_${VERSION}_linux_amd64.tar.gz" \
+            -C "${RUNNER_TEMP}" spark-operator
+          "${RUNNER_TEMP}/spark-operator" version
+          "${RUNNER_TEMP}/spark-operator" --help > /dev/null
+          tar -xzf "dist/spark-operator_${VERSION}_linux_arm64.tar.gz" \
+            -C "${RUNNER_TEMP}" spark-operator
+          file "${RUNNER_TEMP}/spark-operator" | grep -q "ARM aarch64"
+          file "${RUNNER_TEMP}/spark-operator" | grep -q "statically linked"
+
       - name: Get Helm version
         id: get_helm_version
         run: |
@@ -432,4 +504,7 @@ jobs:
           if [[ "$VERSION" == *rc* ]]; then
             args+=(--prerelease)
           fi
-          gh release create "${args[@]}" *.tgz
+          gh release create "${args[@]}" \
+            *.tgz \
+            dist/spark-operator_*.tar.gz \
+            dist/SHA256SUMS
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -432,36 +432,6 @@ jobs:
           sha256sum spark-operator_*.tar.gz > SHA256SUMS
           cat SHA256SUMS
 
-      - name: Set up Go
-        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
-        with:
-          go-version-file: go.mod
-
-      # NOTE: this must run before "Package Helm charts". That step writes
-      # *.tgz into the repository root, which is not gitignored, so
-      # `git status --porcelain` becomes non-empty and the Makefile would
-      # stamp the binaries with GIT_TREE_STATE=dirty.
-      - name: Build operator binaries
-        run: |
-          set -euo pipefail
-          for arch in amd64 arm64; do
-            stage="dist/linux_${arch}"
-            mkdir -p "${stage}"
-            echo "Building linux/${arch}..."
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
tariq-hasan, line 477: We'd want to run the build step and Helm packaging before releasing operator and kubectl images so we are able to fail-fast in case of errors before any images are published.

More generally speaking we'd not want to build the binary twice. We'd want to build it once and then inject it into the container image and the release artifact for consistency. This second portion could go as a separate PR.

.github/workflows/release.yaml
```diff
@@ -402,6 +402,48 @@ jobs:
...
+          sha256sum spark-operator_*.tar.gz > SHA256SUMS
+          cat SHA256SUMS
+
+      - name: Verify binaries
+        run: |
+          set -euo pipefail
+          tar -xzf "dist/spark-operator_${VERSION}_linux_amd64.tar.gz" \
+            -C "${RUNNER_TEMP}" spark-operator
+          "${RUNNER_TEMP}/spark-operator" version
+          "${RUNNER_TEMP}/spark-operator" --help > /dev/null
```
tariq-hasan: I think the verification can be more extensive, especially since the `version` [command](https://github.com/vikas-saxena02/spark-operator/blob/d1e49bbc58c3d0887cbbc2de7e8c26a7254b45db/cmd/operator/version/root.go#L34-L35) always exits 0 and therefore any errors won't be caught.

Something like the following would be nice to have.

```
test -z "$(git status --porcelain)"

# build step

"${RUNNER_TEMP}/spark-operator" version | grep -q "Git Tree State: clean"
"${RUNNER_TEMP}/spark-operator" version | grep -q "${VERSION}"
```

.github/workflows/release.yaml
```diff
@@ -402,6 +402,48 @@ jobs:
           VERSION=$(cat VERSION)
           echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
+      - name: Set up Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version-file: go.mod
```
tariq-hasan, line 408: Since this is a privileged release job with write permissions probably not a good idea to use cached dependencies.

```suggestion
      - name: Set up Go
        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
        with:
          go-version-file: go.mod
          cache: false
```

.github/workflows/release.yaml
```diff
@@ -402,6 +402,48 @@ jobs:
...
+          for arch in amd64 arm64; do
+            stage="dist/linux_${arch}"
+            mkdir -p "${stage}"
+            echo "Building linux/${arch}..."
+            GOOS=linux GOARCH="${arch}" \
+              make build-operator SPARK_OPERATOR="${PWD}/${stage}/spark-operator"
+            cp LICENSE "${stage}/"
+            tar -czf "dist/spark-operator_${VERSION}_linux_${arch}.tar.gz" \
+              -C "${stage}" spark-operator LICENSE
+          done
```
tariq-hasan, line 431: We'd want to avoid a dirty working tree as per the comment above. That said, `mkdir -p "${stage}"` and the subsequent build step creates a dirty working tree anyway since `dist/` is not gitignored by the repo.

To mitigate this issue we can stage in `$RUNNER_TEMP` or `bin/dist/linux_${arch}` since it is gitignored.

tariq-hasan (commented):

docs/website/user-guide/building-custom-images.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+## Building a custom image
+
+The operator invokes `${SPARK_HOME}/bin/spark-submit` as a subprocess, so the runtime
+image must contain a Spark distribution and a JVM. The example below layers the released
+binary onto the official Spark image; substitute your own approved base as needed.
+
+```dockerfile
+ARG SPARK_IMAGE=docker.io/apache/spark:4.0.4
```
vjanelle: make SPARK_VERSION an image argument?

HarshPopat23 (commented):
I reviewed the current head (`1df72c8`). Publishing official binaries is valuable, but I see a few blocking release-consistency issues.

The release archives cannot currently have identical metadata to the official images. The image binary and release binary are built in separate jobs, while `BUILD_DATE` is derived from the wall clock. Additionally, after the AMD64 artifact is written under the unignored `dist/` directory, the ARM64 `make` invocation can observe a dirty working tree and embed `GIT_TREE_STATE=dirty`.

Could we build the binaries once before publishing anything and reuse those exact artifacts for both the images and GitHub release? At minimum, the build metadata should be deterministic and explicitly verified.

Also, this PR uses `Closes #3073`, but provenance and signatures from that issue are intentionally excluded. Could we either implement them or change this to `Part-of #3073` and link a follow-up issue?

Finally, the documented Dockerfile requires `entrypoint.sh`, but the release archive contains only the binary and `LICENSE`. Please either package the entrypoint or provide a pinned, verifiable download step.

I agree with the existing fail-fast and verification comments as well.

.github/workflows/release.yaml
vikas-saxena02: addressed with other comemnts

.github/workflows/release.yaml
```diff
@@ -402,6 +402,62 @@ jobs:
           VERSION=$(cat VERSION)
           echo "VERSION=${VERSION}" >> $GITHUB_ENV
```
tariq-hasan, line 403: Given we are reading the version into a `VERSION` env and passing it into the `build-operator` target in one of the next steps what's happening here is that the [`VERSION`](https://github.com/vikas-saxena02/spark-operator/blob/283fa94b3f2da86fe74d8e0824c488b4cb6dbf77/Makefile#L16) could technically be `v2.5.1` and then [`GIT_VERSION`](https://github.com/vikas-saxena02/spark-operator/blob/283fa94b3f2da86fe74d8e0824c488b4cb6dbf77/Makefile#L22) would be `vv2.5.1` - and this is what would then be used during the build process.

The fix would be to make this change in the Makefile so `VERSION` is not overridden:

```
VERSION := $(shell cat VERSION | sed "s/^v//")
```

vikas-saxena02: code chnage made

nabuskey (approved):
In a follow up PR, we need to update the release process so that we use the same binaries in released assets and in Docker images.

/approve

Issue #3049, opened 2026-07-19
Add fallback mechanism for driver pod service account


tariq-hasan:
### What feature you would like to be added?

The operator currently passes the service account referenced in the `SparkApplication` CR in the `spark-submit` command. However, there is no fallback mechanism in case the CR does not have the service account.

Similarly, the operator currently creates the driver pod using the service account referenced in the pod template spec of the SparkConnect CR. However, there is no fallback mechanism in case the CR does not have the service account.

The goal of this issue is to make the service account configurable via Helm value and pass it on to the operator as a controller flag so that it can be used as a fallback mechanism when a CR does not contain the service account.

In addition, the goal is to make the service account a field on the `SparkConnect` CRD spec to enable parity with SparkApplication CRD

This is needed for downstream use cases such as for Kubeflow SDK which does not expose the service account as a configurable option.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

vikas-saxena02:
@tariq-hasan I'd like to pick this up if no one is already working on it.

I've read through the relevant paths to make sure I understand the current behaviour before proposing anything:

- `SparkApplication`: `spark.kubernetes.authenticate.driver.serviceAccountName` is only emitted when `.spec.driver.serviceAccount` is set (`internal/controller/sparkapplication/submission.go:383`), with the executor equivalent at `:758`.
- `SparkConnect`: `mutateServerPod` (`internal/controller/sparkconnect/reconciler.go:344`) copies `.spec.server.template.spec` as-is, so `serviceAccountName` only exists if the user supplied it in the pod template. Executors never get a service account conf at all.
- `SparkPodSpec` (`api/v1alpha1/sparkconnect_types.go:107`) has no `serviceAccount` field, hence the parity gap with `v1beta2`.

**Proposed shape**

Mirror the existing `--default-batch-scheduler` pattern: a `--default-service-account` controller flag, a corresponding field on the reconciler `Options` structs, and a Helm value under `controller` that defaults to `include "spark-operator.spark.serviceAccountName"`. That default works cleanly because `templates/spark/serviceaccount.yaml` creates the same SA name in every entry of `spark.jobNamespaces`.

I'd suggest splitting this into three PRs so the API change doesn't hold up the rest:

1. Flag + `Options` field + `SparkApplication` fallback (driver and executor) + Helm value and chart unittest.
2. `SparkConnect` server and executor fallback using the same flag.
[...]

PR #3092, opened 2026-08-14
feat(controller): add default service account fallback for SparkApplication driver

vikas-saxena02:
## Purpose of this PR

Part of #3049.

Adds an operator-level fallback service account for `SparkApplication` driver pods.

The operator only emits `spark.kubernetes.authenticate.driver.serviceAccountName`
when `.spec.driver.serviceAccount` is set (`submission.go:399`). Applications
submitted without it fall through to the namespace's `default` service account,
which normally lacks the RBAC to create executor pods.

Per @tariq-hasan's guidance on the issue, the scope here is **drivers only** — the
executor path (`submission.go:774`) is untouched. This is the first of three PRs;
`SparkConnect` server fallback and the `ServerSpec.serviceAccount` CRD field follow
separately, which is why this is `Part of` rather than `Closes`.

**Proposed changes:**

- Add a `--default-service-account` controller flag and a matching
  `DefaultServiceAccount` field on `sparkapplication.Options`.
- Resolve the fallback against a deep copy of the `SparkApplication` immediately
  before submission. Precedence is CR field → flag → unset.
- Surface it as an opt-in Helm value `controller.defaultServiceAccount`, defaulting
  to `""` so the flag is not rendered and behaviour is unchanged on upgrade.
- Add unit, envtest and chart unittest coverage.

## Change Category

- [ ] Documentation update

### Rationale

**Deep copy rather than in-place mutation.** The fallback is only ever visible in the
`spark-submit` invocation and is never written back to the custom resource. There is
an integration test asserting exactly this — it reconciles with a default configured,
checks the submitted application carries it, then re-fetches from the API server and
asserts the stored spec is unchanged.

**I did not widen `sparkSubmitOptionFunc` as I said I would on the issue.** The
`SparkApplicationSubmitter.Submit(ctx, app)` interface (`submission.go:41`) doesn't
carry `Options` either, so reaching `driverConfOption` that way would mean changing
28 option funcs plus the interface and both submitter implementations, for one string
field. Resolving in the reconciler keeps defaulting in the controller exactly as
requested and is a much smaller diff.

**The Helm value is opt-in, not auto-populated.** Defaulting it to the chart-created
account would break installs running `spark.jobNamespaces: [""]`, since
`templates/spark/serviceaccount.yaml` guards on `ne $jobNamespace ""` and creates no
Spark service account at all in that mode. Auto-defaulting can be a follow-up once
the namespace discussion on #3049 settles.

## Checklist

### Additional Notes

**Small behaviour change beyond the issue text:** a CR with `serviceAccount: ""` now
receives the fallback. Previously it emitted
`spark.kubernetes.authenticate.driver.serviceAccountName=` with an empty value, which
is never useful. An explicitly empty string is now treated as unspecified.

Test coverage added:
- `TestApplyDefaultDriverServiceAccount` — 5 subtests covering no default, unset,
  empty string, CR value wins, and propagation into the driver conf.
- Two envtest cases asserting the flag reaches submission and that the stored CR is
  not mutated.
- Two chart unittests asserting the arg renders when set and is absent when empty.

`make go-fmt go-vet build-operator unit-test helm-unittest` all pass locally.
`make helm-lint` reports a chart version bump is needed, but `Chart.yaml` is only
bumped by release commits in this repo, so I've left it to CI.


internal/controller/sparkapplication/submission.go
```diff
@@ -146,6 +146,28 @@ func buildSparkSubmitArgs(app *v1beta2.SparkApplication, skipPodTemplates bool)
...
+	}
+	if app.Spec.Driver.ServiceAccount != nil && *app.Spec.Driver.ServiceAccount != "" {
+		return app
+	}
+
+	copied := app.DeepCopy()
+	copied.Spec.Driver.ServiceAccount = &defaultServiceAccount
+	return copied
+}
+
```
tariq-hasan: The precedence logic should also account for service account set within the pod template. `spec.driver.serviceAccount` should have precedence over `spec.driver.template.spec.serviceAccountName` which should have precedence over controller flag.

tariq-hasan: We should also move `applyDefaultDriverServiceAccount` to `pkg/util/sparkapplication.go` to keep the submission logic self-contained.

cmd/operator/controller/start.go
tariq-hasan, line 207: We should have validation for the controller flag within the `PreRunE` block to ensure the name is not rejected by the API server.

internal/controller/sparkapplication/controller.go
```diff
@@ -1004,7 +1008,12 @@ func (r *Reconciler) submitSparkApplication(ctx context.Context, app *v1beta2.Sp
 		}
 	}()
 
-	if err := r.submitter.Submit(ctx, app); err != nil {
+	// Fall back to the operator-level default service account when the SparkApplication
+	// does not specify one for the driver. This is applied to a copy of the application
+	// so that the fallback is never written back to the custom resource.
+	submitApp := applyDefaultDriverServiceAccount(app, r.options.DefaultServiceAccount)
```
tariq-hasan: We'd want to log what the service account was set to so that normal user configuration can be differentiated from controller-resolved configuration.

charts/spark-operator-chart/README.md
```diff
@@ -106,6 +106,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.driverPodCreationGracePeriod | string | `"10s"` | Grace period after a successful spark-submit when driver pod not found errors will be retried. Useful if the driver pod can take some time to be created. |
 | controller.maxTrackedExecutorPerApp | int | `1000` | Specifies the maximum number of Executor pods that can be tracked by the controller per SparkApplication. |
 | controller.defaultTimeToLiveSeconds | int | `0` | Default Time-To-Live (in seconds) applied to terminated SparkApplications that do not set spec.timeToLiveSeconds. Requires the DefaultTimeToLive feature gate to be enabled. 0 (default) disables it. |
+| controller.defaultServiceAccount | string | `""` | Name of the service account used by the driver pod when a SparkApplication does not specify one. Leave empty to disable the fallback, in which case the driver pod uses the namespace's default service account. The service account must already exist in every namespace where Spark applications are submitted; note that `spark.serviceAccount.create` only provisions accounts for the namespaces listed in `spark.jobNamespaces`. |
```
tariq-hasan, line 109: Other than chart docs we should update the user docs as well:

https://github.com/kubeflow/spark-operator/blob/master/docs/website/user-guide/writing-sparkapplication.md?plain=1#L96.

charts/spark-operator-chart/templates/controller/deployment.yaml
tariq-hasan, line 137: Probably better to move the flag here and group all the defaults in one place.

tariq-hasan (commented):
Hi @vikas-saxena02! I have added comments on the PR.

internal/controller/sparkapplication/submission.go
vikas-saxena02: Done. `ApplyDefaultDriverServiceAccount` now checks `.spec.driver.template.spec.serviceAccountName` before falling back to the flag, so the order is `.spec.driver.serviceAccount` → pod template → controller flag. Added unit tests for each level, plus a reconciler-level test asserting the flag does not override a template.

vikas-saxena02: Done. Moved and exported as `util.ApplyDefaultDriverServiceAccount` in `pkg/util/sparkapplication.go`; tests moved to `pkg/util/sparkapplication_test.go`.

cmd/operator/controller/start.go
vikas-saxena02: Done. Added a `validation.IsDNS1123Subdomain` check on `--default-service-account` in `PreRunE`, following the same pattern as the TTL validation.

internal/controller/sparkapplication/controller.go
vikas-saxena02: Done. The controller now logs `Applied default driver service account` with the resolved value, and only when the fallback actually fires — so controller-resolved config is distinguishable from user-set config.

charts/spark-operator-chart/README.md
vikas-saxena02: Done. Added to the driver section in `docs/website/user-guide/writing-sparkapplication.md`.

charts/spark-operator-chart/templates/controller/deployment.yaml
vikas-saxena02: Done. Moved next to `--default-time-to-live-seconds` so the defaults are grouped.

nabuskey (approved):
/approve

Issue #2962, opened 2026-05-28
Workload Aware Scheduling for Spark Jobs on Kubernetes


andreyvelich:
### What feature you would like to be added?

Recently, [Kubernetes introduced support](https://kubernetes.io/docs/concepts/workloads/workload-api/) for Workload and PodGroup API: 
The community is now exploring these primitives as the foundation for advanced scheduling capabilities such as gang scheduling, topology-aware scheduling, preemption, and broader workload-aware orchestration through WAS WG:
 https://github.com/orgs/kubernetes/projects/251

We briefly discussed this topic as the recent Spark on Kubernetes call: https://youtu.be/j9lhLlyw1Uo?t=2677

I believe, integrating these Kubernetes-native primitives with Spark could unlock many benefits for users, while aligning Spark more closely with the evolving Kubernetes scheduling ecosystem. In particular, Spark’s unique capabilities, Dynamic Resource Allocation, provide an opportunity to explore additional scheduling and resource-management use cases beyond traditional batch workloads.

Similar KEP for Trainer: https://github.com/kubeflow/trainer/pull/3219

Would love to hear your feedback on this!

cc @vara-bonthu @nabuskey @ChenYi015 @tenzen-y @vjanelle @Vedant-Deshpande  @kannon92 @helayoty @wojtek-t @mm4tt @tosi3k @tariq-hasan @Shekharrajak 

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

nabuskey:
I am 100% for this. We already have support for custom schedulers (volcano and yunikorn). We should be able to extend it.

andreyvelich:
@RobuRishabh Thank you! FYI, here are some useful KEPs to take a look how we did integration across the community:

- Job: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job
- JobSet: https://github.com/kubernetes-sigs/jobset/pull/1250
- TrainJob: https://github.com/kubeflow/trainer/pull/3219

PR #3093, opened 2026-08-14
feat(api): add MinMember configuration for workload scheduling

Rajadi16:
Part of #2962

## Purpose of this PR

Adds the preparatory API surface for the upcoming native Kubernetes Workload/PodGroup (scheduling.k8s.io/v1alpha2) batch-scheduler backend. No user-facing behavior change in this release — the workload scheduler name is not yet registered.

**Proposed changes:**
- `api/v1beta2`: add `PodSchedulingGroup` type and `SchedulingGroup` field on `SparkPodSpec`, plus `MinMember` field on `BatchSchedulerConfiguration`. `SchedulingGroup` is operator-managed (set by the workload scheduler backend); `MinMember` is an optional user override for the gang minCount.
- `internal/webhook`: add `validateWorkloadSchedulerFields()` that rejects user-set `schedulingGroup` fields and emits a warning when `batchSchedulerOptions.queue` is set with `batchScheduler=workload` (queue has no effect on the Workload API). Table-driven unit tests included.
- `charts`: add `scheduling.k8s.io` RBAC rule (`workloads`, `podgroups`) gated on new `controller.batchScheduler.workload.enable` Helm value (default `false`).
- Generated: deepcopy, CRD schemas (both `SparkApplication` and `ScheduledSparkApplication`), OpenAPI spec, Python SDK models.

## Change Category

- [ ] Documentation update

### Rationale

Unblocks the follow-up `workload` scheduler backend PR by landing API types, webhook validation, and RBAC first, as additive plumbing with zero behavior change for existing Volcano/YuniKorn/kube-scheduler paths. `workload` is not yet a registered scheduler name, so none of this is reachable by users yet.

## Checklist

- [ ] I have updated documentation accordingly. *(N/A — user docs land in a follow-up PR)*

### Additional Notes

- Diff is ~490 lines, above the original ~110–144 estimate — the delta is entirely generated surface: OpenAPI spec (`swagger.json`), Python SDK client models, and CRD schema for `ScheduledSparkApplication` (which embeds the same pod spec, so it needed regen too, not just `SparkApplication`). No hand-written logic beyond the types/webhook/RBAC changes above.
- RBAC is gated behind a new `controller.batchScheduler.workload.enable` values flag (default `false`) rather than granted unconditionally, since no scheduler backend consumes it until the follow-up PR.
- `make manifests`/`make generate` re-run; generated diffs (deepcopy, OpenAPI, CRDs, Python SDK) included and verified against source types.


api/v1beta2/sparkapplication_types.go
```diff
@@ -529,6 +538,24 @@ type SparkPodSpec struct {
...
+
+// PodSchedulingGroup links a driver or executor pod template to a native Kubernetes
+// PodGroup (scheduling.k8s.io/v1alpha2) for gang scheduling. It is set internally
+// by the "workload" batch scheduler backend and must not be set directly by users.
+type PodSchedulingGroup struct {
+	// PodGroupName is the name of the PodGroup this pod belongs to.
+	// Must be a DNS subdomain.
+	//
+	// +required
+	PodGroupName string `json:"podGroupName"`
```
RobuRishabh: `schedulingGroup` becomes permanent v1beta2 surface users cannot set and [#3113](https://github.com/kubeflow/spark-operator/pull/3113) will not persist. Later removal is a breaking CRD change. Volcano already binds PodGroups via annotations. Drop `SchedulingGroup` / `PodSchedulingGroup` from the CRD. Bind PodGroup on the submission copy or pod template. Keep `MinMember`.

internal/webhook/sparkapplication_validator.go
```diff
@@ -159,6 +169,48 @@ func (v *SparkApplicationValidator) validateSpec(ctx context.Context, app *v1bet
...
+
+	if app.Spec.BatchScheduler != nil && *app.Spec.BatchScheduler == "workload" {
+		if opts != nil && opts.MinMember != nil {
+			initialExecutors := max(int32(1), util.GetInitialExecutorNumber(app))
+			if *opts.MinMember > initialExecutors {
+				return nil, fmt.Errorf(
+					"spec.batchSchedulerOptions.minMember (%d) must not exceed the initial executor count (%d)",
+					*opts.MinMember,
+					initialExecutors,
+				)
```
RobuRishabh: `workload` + `sparkConf.spark.executor.instances=4` + `minMember: 4` is rejected as “exceed … (2)”.
Compare against the same count the backend will gang (spec + SparkConf + DRA), or delay this check until that backend exists and share one function.

charts/spark-operator-chart/templates/controller/_helpers.tpl
```diff
@@ -226,4 +226,19 @@ Create the role policy rules for the controller in every Spark job namespace
   - watch
   - create
   - delete
+{{- if .Values.controller.batchScheduler.workload.enable }}
```
RobuRishabh: `workload.enable=true` grants `scheduling.k8s.io` RBAC even when `batchScheduler.enable` is `false`. Nest under `batchScheduler.enable` and add a helm unittest

RobuRishabh (changes requested):
Hi @Rajadi16 Thanks for the contribution.  There are few **ScheduledSparkApplication validation gap**

Here you added `schedulingGroup` / `minMember` to `SparkApplicationSpec` and validates them in `validateWorkloadSchedulerFields` (`sparkapplication_validator.go`), but `ScheduledSparkApplication` embeds the same spec as `spec.template` and its validator still only checks `sparkConf`:

https://github.com/kubeflow/spark-operator/blob/master/internal/webhook/scheduledsparkapplication_validator.go#L98-L100

1. User applies a `ScheduledSparkApplication` with e.g. `spec.template.driver.schedulingGroup` or invalid `minMember` → **admitted**
2. Controller copies `spec.template` into a child `SparkApplication` (`scheduledsparkapplication/controller.go` ~L289) → **SparkApplication webhook rejects**

So invalid schedules look fine until the first cron run. The PR also regenerated the `ScheduledSparkApplication` CRD with these fields, so this isn’t theoretical.

**What you can do:**
- Reuse `validateWorkloadSchedulerFields` on `spec.template`, e.g. wrap as `&SparkApplication{Spec: app.Spec.Template}` in `ScheduledSparkApplicationValidator.validate`
- Add tests in `scheduledsparkapplication_validator_test.go` mirroring the SparkApplication workload cases

Not asking for unrelated webhook refactors, just the same rules on the shared template type this PR extended.

andreyvelich (commented):
Thank you for starting this @Rajadi16!
Before updating the API, shall we start with the small KEP to discuss API design for WAS support in SparkApplication CRD?
For example, for TrainJob, JobSet, Job, Deployment, and LWS we keep APIs consistent with `.spec.scheduling` field.
If we go ahead with the similar API for Spark, that would be great.

cc @kannon92 @mm4tt

- https://github.com/kubernetes-sigs/jobset/pull/1253
- https://github.com/kubeflow/trainer/pull/3219
- https://github.com/kubernetes-sigs/lws/pull/979

PR #3096, opened 2026-08-14
feat: support physical CPU request and limit in SparkConnect CRD

### LINKED ISSUE #2957: Support physical CPU request/limit in the SparkConnect CRD

Add `CoreLimit` and `CoreRequest` to the SparkConnect API so users can set the physical Kubernetes CPU request and limit for the Connect server and executor pods independently of `spark.{driver,executor}.cores`.

### Why is this needed?

The SparkConnect `SparkPodSpec` exposes only `Cores *int32`, which the controller renders into `spark.driver.cores` / `spark.executor.cores`. That property is the number of JVM task slots and must be a whole integer - it is not the pod's physical CPU request. Spark deliberately separates the two: `spark.kubernetes.{driver,executor}.{request,limit}.cores` set the actual Kubernetes CPU request/limit and accept fractional/millicore values, while `spark.{driver,executor}.cores` set task concurrency.

Without these fields a SparkConnect user cannot request fewer (or fractional) physical CPUs than task slots (e.g. `cores=4` with `request.cores=3500m` for oversubscription/bin-packing), cannot set a hard CPU limit distinct from the request, and cannot right-size pods on constrained clusters. The gap also has a concrete downstream effect: the Kubeflow SDK's Spark client accepts Kubernetes-style CPU strings (e.g. `Driver(resources={"cpu": "500m"})`) and currently crashes, because there is no honest CRD target for a fractional CPU value ([kubeflow/sdk#466](https://github.com/kubeflow/sdk/issues/466)). This enhancement is the operator-side change that unblocks it.

_No response_

_No response_

### Additional context

Relates to https://github.com/kubeflow/sdk/issues/466.

### Love this feature?

DUKartik:
## Purpose of this PR

This PR adds support for configuring physical Kubernetes CPU requests and limits for SparkConnect through the `SparkConnect` CRD.

Currently, SparkConnect exposes the `cores` field, which configures Spark driver/executor cores. However, `cores` does not provide independent control over the physical Kubernetes CPU request and limit for the corresponding pods.

**Proposed changes:**

- Add optional `coreRequest` and `coreLimit` fields to `SparkPodSpec` for SparkConnect server and executor pods.
- Map `coreRequest` and `coreLimit` to the corresponding Spark Kubernetes CPU request/limit configuration for the driver and executors.
- Add CPU quantity validation and unit tests for CPU configuration generation and validation.
- Regenerate the CRD, OpenAPI, Swagger, and Python API artifacts.
- Update the SparkConnect example to demonstrate the new CPU request/limit fields.

### Example

```yaml
executor:
  instances: 2
  cores: 1
  coreRequest: "1000m"
  coreLimit: "1"
  memory: 512m
```

The new fields are independent of `cores`:

- `cores` → `spark.driver.cores` / `spark.executor.cores`
- `coreRequest` → physical Kubernetes CPU request
- `coreLimit` → physical Kubernetes CPU limit

## Change Category

- [ ] Documentation update

### Rationale

SparkConnect currently does not expose physical Kubernetes CPU request/limit configuration through its CRD.

This change provides independent control over Spark's core configuration and the Kubernetes CPU resources assigned to SparkConnect driver and executor pods, while preserving the existing behavior of the `cores` field.

The new fields are optional, so existing SparkConnect resources remain backward compatible.

## Checklist

### Additional Notes

The implementation maps the new fields as follows:

- `server.coreRequest` → `spark.kubernetes.driver.request.cores`
- `server.coreLimit` → `spark.kubernetes.driver.limit.cores`
- `executor.coreRequest` → `spark.kubernetes.executor.request.cores`
- `executor.coreLimit` → `spark.kubernetes.executor.limit.cores`

CPU quantity validation follows the repository's existing Kubernetes quantity validation conventions.

Generated API artifacts were regenerated and verified, including the CRD, OpenAPI/Swagger specification, and Python API models.

The SparkConnect example was updated to document usage of `coreRequest` and `coreLimit`.

Relevant SparkConnect controller and webhook tests pass locally, `go vet` passes, CRD drift checks pass, generated artifacts are up to date, and `git diff --check` passes.

Closes #2957


internal/controller/sparkconnect/options.go
```diff
@@ -178,6 +178,18 @@ func driverConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
...
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=%s", common.SparkKubernetesDriverRequestCores, *conn.Spec.Server.CoreRequest))
+	}
+
+	// Driver CPU limit
+	if conn.Spec.Server.CoreLimit != nil {
+		args = append(args, "--conf",
+			fmt.Sprintf("%s=%s", common.SparkKubernetesDriverLimitCores, *conn.Spec.Server.CoreLimit))
+	}
+
```
tariq-hasan: The request and limit cannot be set using `spark.kubernetes.driver.{request,limit}.cores` since the operator spins up the driver pod as part of the client mode setup.

api/v1alpha1/sparkconnect_types.go
```diff
@@ -106,10 +106,23 @@ type ExecutorSpec struct {
 // SparkPodSpec defines common things that can be customized for a Spark driver or executor pod.
 type SparkPodSpec struct {
 	// Cores maps to `spark.driver.cores` or `spark.executor.cores` for the driver and executors, respectively.
+	// Cores represents Spark task-slot/JVM concurrency and is independent of physical Kubernetes CPU requests and limits.
 	// +optional
 	// +kubebuilder:validation:Minimum=1
 	Cores *int32 `json:"cores,omitempty"`
 
+	// CoreRequest specifies the physical CPU request for the pod, controlling the Kubernetes CPU request.
+	// This is independent of Cores and maps to `spark.kubernetes.driver.request.cores` or `spark.kubernetes.executor.request.cores`.
```
tariq-hasan: `CoreRequest` cannot map to `spark.kubernetes.driver.request.cores`. Since the operator creates the server pod as part of the client mode setup the operator is responsible for setting up the resource request on the pod.

api/v1alpha1/sparkconnect_types.go
```diff
@@ -106,10 +106,23 @@ type ExecutorSpec struct {
...
+	// Valid values follow Kubernetes quantity format (e.g., "500m", "1", "1.5").
+	// +optional
+	CoreRequest *string `json:"coreRequest,omitempty"`
+
+	// CoreLimit specifies the physical CPU limit for the pod, controlling the Kubernetes CPU limit.
+	// This is independent of Cores and maps to `spark.kubernetes.driver.limit.cores` or `spark.kubernetes.executor.limit.cores`.
+	// Valid values follow Kubernetes quantity format (e.g., "500m", "1", "1.5").
+	// +optional
+	CoreLimit *string `json:"coreLimit,omitempty"`
+
```
tariq-hasan, line 157: Same argument as above applies for `CoreLimit`.

tariq-hasan, line 157: The precedence rule against the same specification in the pod template needs to be determined as well.

It would make sense to always override resource request or limit set on the pod template by the same setting on `spec.{server,executor}.{coreRequest,coreLimit}`.

internal/controller/sparkconnect/options_test.go
```diff
@@ -129,6 +129,174 @@ var _ = Describe("Options functions", func() {
...
+			args, err := driverConfOption(conn)
+			Expect(err).NotTo(HaveOccurred())
+			config := shellParsedSparkConfig(args)
+
+			// Verify decimal values are preserved as strings
+			Expect(config).To(HaveKeyWithValue(common.SparkKubernetesDriverRequestCores, "1.5"))
+			Expect(config).To(HaveKeyWithValue(common.SparkKubernetesDriverLimitCores, "2.5"))
+		})
+	})
 })
```
tariq-hasan, line 300: Please add/update e2e tests in `test/e2e` to exhibit that the feature works as expected.

internal/webhook/sparkconnect_validator.go
```diff
@@ -328,3 +357,19 @@ func validateMemoryString(memory string) error {
...
+		return fmt.Errorf("CPU quantity cannot be an empty string")
+	}
+
+	_, err := resource.ParseQuantity(cpu)
+	if err != nil {
+		return fmt.Errorf("invalid CPU quantity %q: %v", cpu, err)
+	}
+
+	return nil
+}
```
tariq-hasan, line 421: The validation needs to be more extensive. The `ParseQuantity` check validates the syntax to ensure that the value is a valid K8s resource but it does not validate if the value makes sense for CPU resources (e.g., check on units, sign, zero value, etc.) or if request is less than limit.

The better approach here is to use CEL validation directly on the API spec and type the new fields on the spec as `*resource.Quantity`.

api/v1alpha1/sparkconnect_types.go
tariq-hasan, line 157: We should update the docs as well to document the new API fields.

https://github.com/kubeflow/spark-operator/tree/master/docs/website/user-guide

examples/sparkconnect/spark-connect.yaml
```diff
@@ -55,6 +55,8 @@ spec:
   executor:
     instances: 2
     cores: 1
+    coreRequest: "1000m"
+    coreLimit: "1"
```
tariq-hasan: We'd want to use different values for the executor request and limit to be more comprehensive in the e2e.

tariq-hasan (commented):
Hi @DUKartik! Thank you for raising the PR. I have added some review comments.

api/v1alpha1/sparkconnect_types.go
DUKartik: Thanks, addressed. I added documentation for the new coreRequest and coreLimit fields in the Spark Connect user guide and linked the guide from the User Guide index

DUKartik: Thanks, addressed. server.coreLimit is now applied directly to the CPU limit of the operator-created SparkConnect server pod, matching the coreRequest behavior.

DUKartik: Thanks, addressed. spec.server.coreRequest/coreLimit now override the CPU request/limit from the pod template, while other template resources are preserved. Added unit and e2e coverage for this precedence behavior.

internal/controller/sparkconnect/options_test.go
DUKartik: Thanks, addressed. The e2e test now uses different executor request (500m) and limit (1500m) values and verifies both are passed independently.

internal/webhook/sparkconnect_validator.go
DUKartik: Thanks for the clarification. I’ve expanded the validation to cover CPU-specific constraints, including non-zero/non-negative values and coreRequest <= coreLimit, while keeping the fields as *string to match the existing v1beta2.SparkApplication API convention.

If you’d prefer *resource.Quantity with CEL validation here, I can make that change as well. In that case, I think we should apply the same API typing approach to the existing v1beta2 CoreRequest/CoreLimit fields for consistency

api/v1alpha1/sparkconnect_types.go
```diff
@@ -106,10 +106,55 @@ type ExecutorSpec struct {
...
+	// Precedence (server only): if both this field and
+	// `spec.server.template.spec.containers[name=spark-kubernetes-driver].resources.requests.cpu`
+	// are set, this field wins for the CPU key. Other resource keys (memory, ephemeral-storage, etc.)
+	// on the template are preserved unchanged, matching the `addMemoryLimit` resource-merge
+	// convention used elsewhere in the project.
+	//
+	// Valid values follow Kubernetes quantity format (e.g., "500m", "1", "1.5"). The value must be
+	// non-zero and, when `coreLimit` is also set, must not exceed `coreLimit`.
+	// +optional
+	CoreRequest *string `json:"coreRequest,omitempty"`
```
tariq-hasan, line 134: We should use `CoreRequest *resource.Quantity` instead of `CoreRequest *string`. This way we'd delegate resource parsing to K8s.

api/v1alpha1/sparkconnect_types.go
```diff
@@ -106,10 +106,55 @@ type ExecutorSpec struct {
...
+	// Precedence (server only): if both this field and
+	// `spec.server.template.spec.containers[name=spark-kubernetes-driver].resources.limits.cpu` are
+	// set, this field wins for the CPU key. Other resource keys (memory, ephemeral-storage, etc.)
+	// on the template are preserved unchanged, matching the `addMemoryLimit` resource-merge
+	// convention used elsewhere in the project.
+	//
+	// Valid values follow Kubernetes quantity format (e.g., "500m", "1", "1.5"). The value must be
+	// non-zero and, when `coreRequest` is also set, must be greater than or equal to `coreRequest`.
+	// +optional
+	CoreLimit *string `json:"coreLimit,omitempty"`
```
tariq-hasan, line 156: Same here. Let's use `CoreLimit *resource.Quantity` instead of `CoreLimit *string`.

docs/website/user-guide/working-with-sparkconnect.md
```diff
@@ -0,0 +1,139 @@
...
+    memory: 512m
+```
+
+## Deleting a SparkConnect
+
+A `SparkConnect` can be deleted using `kubectl delete sparkconnect <name>`. Deleting a `SparkConnect` causes the operator to garbage-collect the server pod, the `Service`, and the `ConfigMap`. Executor pods are owned by Spark and are removed when the server shuts down.
+
+## Updating a SparkConnect
+
+A `SparkConnect` can be updated using `kubectl apply -f <updated file>`. The operator reconciles the change. The admission webhook re-validates the spec on every update.
```
tariq-hasan, line 37: The operator technically does not reconcile driver resource updates since there is a guard on the creation timestamp.

https://github.com/DUKartik/kubeflow_spark-operator/blob/514d41c75bfbeae9e5bd7ca7d484c5a47ae6c447/internal/controller/sparkconnect/reconciler.go#L352

api/v1alpha1/sparkconnect_types.go
tariq-hasan, line 156: Let's keep the comments as concise as for other existing fields as they get copied into generated manifest and Python models.

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,234 @@
...
+				"expected spark-submit args to include executor limit cores 1500m, got: %s", allArgs)
+		})
+	})
+
+	Context("Precedence: spec.server.coreRequest overrides template CPU request", func() {
+		// Tariq asked that the precedence rule be deterministic and explicit.
+		// The operator's contract is: if both spec.server.coreRequest and
+		// spec.server.template.spec.containers[].resources.requests.cpu are
+		// set, spec.server.coreRequest wins for the CPU key. Other resource
+		// keys (memory, ephemeral-storage) on the template are preserved.
```
tariq-hasan, line 141: Let's remove name references.

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,234 @@
...
+		ctx := context.Background()
+
+		var conn *v1alpha1.SparkConnect
+
+		BeforeEach(func() {
+			image := "apache/spark:4.0.0"
```
tariq-hasan, line 147: This image is not used for the e2e tests in general. 4.0.4 is the one that's used exclusively.

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,234 @@
...
+				"expected template memory 1Gi to be preserved, got %s", memReq.String())
+
+			By("Asserting the template's memory limit is preserved")
+			memLim, ok := serverPod.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory]
+			Expect(ok).To(BeTrue(), "template memory limit should be preserved")
+			Expect(memLim.Equal(resource.MustParse("1Gi"))).To(BeTrue(),
+				"expected template memory 1Gi to be preserved, got %s", memLim.String())
+		})
+	})
+})
```
tariq-hasan, line 234: We should test for pod readiness not only pod creation.

docs/website/user-guide/working-with-sparkconnect.md
```diff
@@ -0,0 +1,139 @@
...
+    initialExecutors: 2
+    minExecutors: 1
+    maxExecutors: 10
+    shuffleTrackingEnabled: true
+    shuffleTrackingTimeout: 60
+```
+
+## Field Reference
+
+For the full API definition, including all fields, see the [`SparkConnect` API reference](../reference/api-docs.md).
```
tariq-hasan, line 139: The documentation is duplicated in several parts with [`docs/website/user-guide/spark-connect.md`](https://github.com/kubeflow/spark-operator/blob/761686ee016287e0c2afc856ad15d601de32cdad/docs/website/user-guide/spark-connect.md).

Let's match the docs for SparkApplication in terms of content density, formatting, layout, etc.

tariq-hasan (commented):
Hi @DUKartik! I have added a few comments.

api/v1alpha1/sparkconnect_types.go
DUKartik: Done ,CoreRequest and CoreLimit are now *resource.Quantity in api/v1alpha1/sparkconnect_types.go

api/v1alpha1/sparkconnect_types.go
DUKartik: Done. The generated CRD represents both fields using Kubernetes' standard resource.Quantity schema (x-kubernetes-int-or-string), consistent with Kubernetes resource quantities. No additional CEL rule was added.

docs/website/user-guide/working-with-sparkconnect.md
DUKartik: Done. The Updating a SparkConnect documentation now explicitly explains that spec.server.coreRequest, coreLimit, cores, memory, template, and spec.image only take effect after the server pod is recreated, matching the current reconciliation behavior.

api/v1alpha1/sparkconnect_types.go
DUKartik: Done. The CoreRequest and CoreLimit comments were reduced to single-sentence field descriptions to match the surrounding fields.

test/e2e/sparkconnect_resources_test.go
DUKartik: Done. The reviewer-specific comments were removed from the test code and controller tests.

test/e2e/sparkconnect_resources_test.go
DUKartik: Done. Both affected e2e fixtures now use SparkVersion: "4.0.4", matching the docker.io/apache/spark:4.0.4 image, the canonical SparkConnect example, and the Makefile default.

test/e2e/sparkconnect_resources_test.go
DUKartik: Done. The Precedence context now includes a readiness check using util.IsPodReady, while the resource-spec assertions that only depend on pod creation continue to wait for pod creation rather than JVM readiness.

docs/website/user-guide/working-with-sparkconnect.md
DUKartik: Done. I kept working-with-sparkconnect.md and rewrote it to follow the working-with-sparkapplication.md structure.

internal/webhook/sparkconnect_validator.go
```diff
@@ -325,3 +370,32 @@ func validateMemoryString(memory string) error {
...
+
+// validateCPURequestLELimit validates that the parsed CPU request is less than or equal to the
+// parsed CPU limit. Both inputs must already be valid Kubernetes CPU quantities.
+func validateCPURequestLELimit(request, limit *resource.Quantity) error {
+	if request.Cmp(*limit) > 0 {
+		return fmt.Errorf("coreRequest %q must not be greater than coreLimit %q", request.String(), limit.String())
+	}
+	return nil
+}
```
tariq-hasan: For this validation we need to first determine the effective request and the effective limit by using the precedence rule that CRD field is what always wins over other ways the value is passed in. The effective values should take into account the fact that any of these fields could be missing in the spec.

Once the effective values are determined then we need to validate if request is less than limit.

internal/webhook/sparkconnect_validator.go
```diff
@@ -325,3 +370,32 @@ func validateMemoryString(memory string) error {
...
+
+	if cpu.IsZero() {
+		return fmt.Errorf("invalid CPU quantity: must be greater than zero")
+	}
+	if cpu.Sign() < 0 {
+		return fmt.Errorf("invalid CPU quantity: must not be negative")
+	}
+
+	return nil
+}
```
tariq-hasan: CEL validation only checks if the quantity is a valid k8s quantity. We also need to confirm if the quantity is a valid CPU quantity. For example, "1Gi" is not valid CPU.

internal/controller/sparkconnect/reconciler.go
```diff
@@ -472,6 +480,30 @@ func (r *Reconciler) mutateServerPod(ctx context.Context, conn *v1alpha1.SparkCo
...
+
+	if conn.Spec.Server.CoreLimit != nil {
+		if container.Resources.Limits == nil {
+			container.Resources.Limits = corev1.ResourceList{}
+		}
+		container.Resources.Limits[corev1.ResourceCPU] = *conn.Spec.Server.CoreLimit
+	}
+
+	return nil
+}
```
tariq-hasan, line 501: There is no need to return an error if all we return is `nil`.

.github/workflows/osv-scanner.yaml
```diff
@@ -1,71 +1,71 @@
...
-            exit 1
-          fi
-
-      - name: Upload SARIF to GitHub Security tab
-        if: always() && hashFiles('osv-results.sarif') != ''
-        uses: github/codeql-action/upload-sarif@1b168cd39490f61582a9beae412bb7057a6b2c4e # v4.31.8
-        with:
-          sarif_file: osv-results.sarif
-          category: kubeflow-spark-operator-osv-scanner
+name: OSV-Scanner Vulnerability Scan
```
tariq-hasan: Several files in this PR like this one don't have any changes but show up in the diff.

Please keep the diff to only files that have changed.

api/v1alpha1/sparkconnect_types.go
```diff
@@ -111,6 +112,14 @@ type SparkPodSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	Cores *int32 `json:"cores,omitempty"`
 
+	// CoreRequest is the physical CPU core request for the pod, applied to the container's resources.requests.cpu for the server pod or passed to Spark as spark.kubernetes.executor.request.cores for executors.
```
tariq-hasan: Let's keep to this otherwise it may be confusing individually for `server.coresRequest` or `executor.coresRequest`.

```suggestion
	// CoreRequest is the physical CPU core request for the pod.
```

api/v1alpha1/sparkconnect_types.go
```diff
@@ -111,6 +112,14 @@ type SparkPodSpec struct {
...
+	// +optional
+	CoreRequest *resource.Quantity `json:"coreRequest,omitempty"`
+
+	// CoreLimit is the physical CPU core limit for the pod, applied to the container's resources.limits.cpu for the server pod or passed to Spark as spark.kubernetes.executor.limit.cores for executors.
```
tariq-hasan: Same here.

```suggestion
	// CoreLimit is the physical CPU core limit for the pod.
```

tariq-hasan (changes requested):
Hi @DUKartik! Thanks for addressing the feedback. The PR is coming into shape. I have added a few more comments.

.github/workflows/osv-scanner.yaml
DUKartik: done

internal/webhook/sparkconnect_validator.go
DUKartik: Done in b0e872b. Effective request/limit values now respect CRD-over-template precedence, with validation covering partial/missing fields. Added tests for the relevant permutations and updated the error message/docs in 36b6c50.

internal/controller/sparkconnect/reconciler.go
DUKartik: Done in b0e872b. Removed the unnecessary error return from setupServerContainerResources and updated its call sites/tests accordingly

DUKartik:
Hi @tariq-hasan, just following up on #3096. All the latest review comments have been addressed, and it’s been over a week since the last review clearance.

When you get a chance, could you please take another look? Thanks!

docs/website/user-guide/index.md
```diff
@@ -37,6 +37,13 @@ Restart policies, failure handling, and managing running applications
 Use `ScheduledSparkApplication` to run Spark jobs on a cron schedule
 ::::
 
+::::{grid-item-card} Working with SparkConnect
+:link: working-with-sparkconnect
+:link-type: doc
+
+Run a long-lived Spark Connect server and connect Spark clients to it
+::::
+
```
tariq-hasan, line 46: Please move this below the **Spark Connect** heading.

examples/sparkconnect/spark-connect.yaml
```diff
@@ -54,7 +85,14 @@ spec:
...
+    cores: 2
+    # Maps to `spark.kubernetes.executor.{request,limit}.cores` and is
+    # applied by Spark when it creates the executor pods. The values
+    # `500m` and `1500m` differ to illustrate that coreRequest can be
+    # smaller than coreLimit.
+    coreRequest: "500m"
+    coreLimit: "1500m"
     memory: 512m
     template:
       metadata:
```
tariq-hasan, line 98: Please leave this example untouched and create a separate example named `spark-connect-custom-resource.yaml` similar to [`spark-pi-custom-resource.yaml`](https://github.com/kubeflow/spark-operator/blob/master/examples/spark-pi-custom-resource.yaml).

You can add in these snippets in the new example to showcase the precedence behavior.

https://github.com/kubeflow/spark-operator/blob/a0f33cccc54858516d8954e7aef2c2bad0eecc76/examples/spark-pi-pod-template.yaml#L109-L121

https://github.com/kubeflow/spark-operator/blob/a0f33cccc54858516d8954e7aef2c2bad0eecc76/examples/spark-pi-pod-template.yaml#L183-L195

internal/controller/sparkconnect/reconciler.go
```diff
@@ -376,6 +376,12 @@ func (r *Reconciler) mutateServerPod(ctx context.Context, conn *v1alpha1.SparkCo
 			pod.Spec.Containers,
 			common.SparkDriverContainerName,
 		)
+
+		// Setup Kubernetes CPU resources for the Connect server container.
+		// The server pod is created by the operator as part of the client mode setup, so
+		// server.coreRequest/server.coreLimit are applied directly to the pod spec instead of
+		// being mapped to spark.kubernetes.driver.{request,limit}.cores Spark configuration.
+		setupServerContainerResources(container, conn)
 		// Setup image.
```
tariq-hasan, line 386: Just a nit for spacing.

```suggestion

		// Setup Kubernetes CPU resources for the Connect server container.
		// The server pod is created by the operator as part of the client mode setup, so
		// server.coreRequest/server.coreLimit are applied directly to the pod spec instead of
		// being mapped to spark.kubernetes.driver.{request,limit}.cores Spark configuration.
		setupServerContainerResources(container, conn)

		// Setup image.
```

internal/controller/sparkconnect/reconciler.go
tariq-hasan: Actually, on second thought, I am thinking it is better to move the function to `pkg/util/container.go` and return `container *corev1.Container` as the function counts as a helper.

internal/controller/sparkconnect/reconciler.go
tariq-hasan: It is better to deepcopy since we are mutating the container resources in the spec.

```suggestion
			pod.Spec = template.Spec.DeepCopy()
```

internal/controller/sparkconnect/reconciler.go
```diff
@@ -472,6 +478,28 @@ func (r *Reconciler) mutateServerPod(ctx context.Context, conn *v1alpha1.SparkCo
...
+	}
+
+	if conn.Spec.Server.CoreLimit != nil {
+		if container.Resources.Limits == nil {
+			container.Resources.Limits = corev1.ResourceList{}
+		}
+		container.Resources.Limits[corev1.ResourceCPU] = *conn.Spec.Server.CoreLimit
+	}
+}
+
```
tariq-hasan: Given that executors, and drivers in cluster mode, default their request to cores when there is no request provided we should do the same here for server request.

internal/webhook/sparkconnect_validator.go
```diff
@@ -325,3 +374,89 @@ func validateMemoryString(memory string) error {
...
+			if limit == nil {
+				if v, ok := container.Resources.Limits[corev1.ResourceCPU]; ok {
+					limit = &v
+				}
+			}
+		}
+	}
+
+	return request, limit
+}
```
tariq-hasan: The logic for `effectiveExecutorCPUResources` would need to be adapted - since Spark always overwrites the executor template's CPU request, using `request.cores` if set and `spark.executor.cores` otherwise. It sets the limit only from `limit.cores`.

This is the reference: https://github.com/apache/spark/blob/203017c434785baf66fb6bb00805520c350cc6dc/docs/running-on-kubernetes.md#L2343-L2346

This is the extract: _The cpu limits are set by `spark.kubernetes.{driver,executor}.limit.cores`. The cpu (request) is set by `spark.{driver,executor}.cores`._

The fix is to compute the effective request as `coreRequest`, then sparkConf `request.cores`, then `executor.cores`, then 1. The effective limit should be `coreLimit`, then sparkConf `limit.cores`, then the template limit.

internal/webhook/sparkconnect_validator.go
```diff
@@ -325,3 +374,89 @@ func validateMemoryString(memory string) error {
 
 	return nil
 }
+
+// validateCPUQuantity validates a Kubernetes CPU quantity.
+//
+// It enforces that the value represents a positive, non-zero CPU resource. A nil pointer is
+// treated as "not specified" and passes validation; the caller is responsible for guarding
+// against nil before invoking.
```
tariq-hasan: The comment needs to be updated since `nil` is now guarded against in `validateCPUQuantity`.

internal/webhook/sparkconnect_validator.go
```diff
@@ -325,3 +374,89 @@ func validateMemoryString(memory string) error {
...
+	if cpu == nil {
+		return fmt.Errorf("CPU quantity cannot be nil")
+	}
+
+	if cpu.IsZero() {
+		return fmt.Errorf("invalid CPU quantity: must be greater than zero")
+	}
+	if cpu.Sign() < 0 {
+		return fmt.Errorf("invalid CPU quantity: must not be negative")
+	}
```
tariq-hasan:
```suggestion
	if cpu.Sign() <= 0 {
		return fmt.Errorf("invalid CPU quantity: must be greater than zero")
	}
```

internal/webhook/sparkconnect_validator.go
```diff
@@ -263,6 +264,30 @@ func (v *SparkConnectValidator) validateServerSpec(sc *v1alpha1.SparkConnect) er
...
+	// Cross-validate that the effective coreRequest is less than or equal to the effective
+	// coreLimit. This is enforced by Kubernetes itself for container resources, but rejecting
+	// it here gives a clearer error at admission time.
+	request, limit := effectiveServerCPUResources(sc)
+	if request != nil && limit != nil {
+		if err := validateCPURequestLELimit(request, limit); err != nil {
+			return fmt.Errorf("invalid server CPU request/limit: %v", err)
+		}
+	}
+
```
tariq-hasan, line 296: Let's wrap all the errors in the validator using `%w` instead of `%v` to preserve the error.

internal/webhook/sparkconnect_validator.go
tariq-hasan, line 483: Spark conf validation is missing as well. We'd want to ensure that something like sparkConf limit.cores=1 together with coreRequest: "2" does not pass validation.

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,255 @@
+/*
+Copyright 2026 The Kubeflow authors.
+
```
tariq-hasan, line 3: This is the new copyright template being used in Kubeflow subprojects.

```suggestion
/*
Copyright The Kubeflow Authors.

```

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,255 @@
...
+		It("applies server CPU resources to the server pod and emits executor CPU conf", func() {
+			By("Creating the SparkConnect")
+			Expect(k8sClient.Create(ctx, conn)).To(Succeed())
+
+			By("Waiting for the operator to create the server pod")
+			serverPod := waitForServerPod(ctx, conn)
+
+			By("Asserting the server container CPU request matches spec.server.coreRequest")
+			cpuReq, ok := serverPod.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]
```
tariq-hasan, line 98: You can look up the container by name.

test/e2e/sparkconnect_resources_test.go
```diff
@@ -0,0 +1,255 @@
...
+		Name:      sparkconnect.GetServerPodName(conn),
+	}
+	serverPod := &corev1.Pod{}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, serverPod)
+	}).WithPolling(PollInterval).WithTimeout(WaitTimeout).Should(Succeed())
+
+	Expect(serverPod.Spec.Containers).NotTo(BeEmpty(), "server pod should have at least one container")
+	return serverPod
+}
```
tariq-hasan, line 255: You'd need to inspect an executor pod as well.

docs/website/user-guide/working-with-sparkconnect.md
```diff
@@ -0,0 +1,33 @@
...
+A `SparkConnect` can be checked using the `kubectl describe sparkconnect <name>` command. The output shows the specification and status of the `SparkConnect` as well as the events associated with it. The events communicate the overall process and errors of the `SparkConnect`.
+
+## Specifying CPU Resources
+
+`SparkPodSpec` exposes `cores` (Spark task-slot count, mapped to `spark.driver.cores` or `spark.executor.cores`) and `coreRequest` / `coreLimit` (physical Kubernetes CPU request/limit, mapped to the container's `resources.{requests,limits}.cpu`). `cores` and `coreRequest` / `coreLimit` are independent.
+
+The server pod is created directly by the operator, so `spec.server.coreRequest` / `coreLimit` are applied to the operator-created server pod's container resources. The executor pods are created by Spark, so `spec.executor.coreRequest` / `coreLimit` are passed via `spark.kubernetes.executor.{request,limit}.cores`. The admission webhook enforces a positive value for both fields and validates `coreRequest <= coreLimit` on the effective values: a CRD field wins over the same field set in the pod template, and the check is skipped when the effective request or limit is not set.
```
tariq-hasan, line 33: For docs please also update the table in `spark-connect.md` with the new fields.

https://github.com/kubeflow/spark-operator/blob/a0f33cccc54858516d8954e7aef2c2bad0eecc76/docs/website/user-guide/spark-connect.md#L112-L124

internal/controller/sparkconnect/reconciler.go
tariq-hasan, line 356: It is better to deepcopy labels, annotations and spec as we later mutate them on the pod. The comment applies originally because we mutate container resources in this PR but the recommendation applies generically for labels, annotations and spec.

tariq-hasan (changes requested):
Hi @DUKartik! Apologies for the delay. I have added a few more comments. Overall lgtm.

internal/webhook/sparkconnect_validator.go
DUKartik: thanks for the pointer to the Container spec table — I checked it against Spark's
own executor feature step before changing anything, and the asymmetry you describe is exactly what
happens: the request is added unconditionally from `request.cores`/`spark.executor.cores`, while the
limit branch is an `Option` that is skipped entirely when `limit.cores` is unset, leaving whatever
the pod template specified.

internal/webhook/sparkconnect_validator.go
DUKartik: Done

internal/webhook/sparkconnect_validator.go
DUKartik: Fixed — the comment now matches the body:
    // It rejects a nil quantity as well as any quantity that is zero or negative. Callers must
    // therefore only invoke it for fields that are actually set.

I also noted that both callers nil-guard before calling, so the nil branch is defensive only.

internal/webhook/sparkconnect_validator.go
DUKartik: Added. `validateSparkConfCPUKeys` is called from `validateSpec` before the CPU cross-validation, so
the effective-value computation can't run on values that failed to parse. For each of

  spark.kubernetes.executor.request.cores
  spark.kubernetes.executor.limit.cores
  spark.executor.cores

internal/controller/sparkconnect/reconciler.go
DUKartik: Done

internal/controller/sparkconnect/reconciler.go
DUKartik: Done — all three assignments are now copies:

- pod.Labels = maps.Clone(template.Labels)
- pod.Annotations = maps.Clone(template.Annotations)
- template.Spec.DeepCopyInto(&pod.Spec)

examples/sparkconnect/spark-connect.yaml
DUKartik: Done. `spark-connect.yaml` is reverted to its master content — no `coreRequest`/`coreLimit`, no
added `cores` — and `examples/sparkconnect/spark-connect-custom-resource.yaml` is added alongside it,
built from the `spark-pi.yaml` / `spark-pi-custom-resource.yaml` pattern you pointed at.

test/e2e/sparkconnect_resources_test.go
DUKartik: Done — all eight `Spec.Containers[0]` sites in this file now go through a `containerByName` helper
keyed on `common.SparkDriverContainerName`, the same constant the reconciler passes when it builds
the server container. The bare `"spark-kubernetes-driver"` string in the pod template is the same
constant now too, so the template and the assertions can't drift apart.

test/e2e/sparkconnect_resources_test.go
DUKartik: dded. There is a new It in the "Precedence" Context that waits for the executor pod and asserts on
a real one: CPU request `500m` and limit `1500m`, i.e. the CRD fields, not the `1` / `2` the executor
pod template carries.

docs/website/user-guide/working-with-sparkconnect.md
DUKartik: Done, and it turned out to need more than the four rows, because the table is the only complete field
list for this CRD and two existing rows became ambiguous next to a physical CPU field.

DUKartik:
Hi @tariq-hasan, just following up on #3096. All outstanding review comments have now been addressed, including the latest reconciler, validator, e2e, and docs changes. The PR is ready for another look whenever you get a chance. Thanks again for the detailed reviews and guidance!

PR #3098, opened 2026-08-15
feat(ci): publish the operator binary extracted from the released image

vikas-saxena02:
Follow-up to #3078, as discussed there.

#3078 built the release binaries with a separate `make build-operator` on the
runner. Same commit and same make target as the images, but not the same
artifact — the image compiles inside `golang:1.25.11` (digest-pinned in the
`Dockerfile`) while the runner used `setup-go` with `go-version-file: go.mod`,
so the two could resolve to different Go patch releases, and `BUILD_DATE`
differed because it was wall-clock.

This extracts the binary from the image build instead of rebuilding it.

### Changes

- **`Dockerfile`**: adds a `scratch`-based `artifacts` stage that copies the
  binary out of the existing `builder` stage. It is deliberately not the final
  stage, so `docker build .`, `make docker-build` and `make docker-buildx` are
  unchanged and the stage is only built when requested via `--target=artifacts`.

- **`build_operator_images`**: runs a second buildx invocation with
  `target: artifacts`, mirroring the inputs of the image build so the `builder`
  stage is a cache hit. The exported binary is the object that shipped in the
  image, not an equivalent one. Archives are packaged per-arch in this job and
  uploaded as workflow artifacts.

- **`draft_release`**: downloads and publishes those archives rather than
  building. `Set up Go` and the ordering constraint around `helm package` are
  both gone, since nothing compiles in this job anymore. Checksum generation and
  verification are unchanged.

- **`Makefile`**: `BUILD_DATE` now derives from the commit timestamp via
  `SOURCE_DATE_EPOCH` rather than the wall clock, so metadata is identical
  regardless of which job produced the binary. This also fixes the existing
  divergence between the amd64 and arm64 images, which are built on separate
  runners and therefore carry different build dates today at the same commit.

### Alternative considered

Having the `Dockerfile` `COPY` a prebuilt binary would also give a single
compilation, but it breaks `docker build .` from a clean checkout and moves the
arch-matching contract outside the `Dockerfile` — a mismatch would then produce
an image that builds fine and fails at runtime. Exporting from the image build
keeps cross-compilation inside buildx via `TARGETARCH` and leaves the local
build path untouched.

### Testing notes

- Verified locally that the binary exported via `--target=artifacts` is
  byte-identical (sha256 `0db9dc25…`) to the one inside the image produced by a
  default `docker build .`, and that the image build reuses the cached `builder`
  layers rather than recompiling. Note this was an arm64-native build on macOS;
  the amd64 cross-compile path is only exercised in CI.
- `BUILD_DATE` confirmed to track the commit timestamp rather than the wall
  clock on both GNU and BSD `date`.
- As with #3078, the verification steps only execute during a real
  `draft_release` run, which cannot be triggered from a fork PR — a maintainer
  dry-run is the first genuine exercise of that path.

Note: BSD `date` does not support the `%:z` specifier, so local macOS builds
produce a slightly malformed timezone suffix in `buildDate`. This predates the
change (plain `date -u +"%:z"` behaves the same) and CI uses GNU `date`, but
flagging it so it is not mistaken for a regression here.

Part of #3073

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -105,6 +105,16 @@ jobs:
           VERSION=$(cat VERSION)
           echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
+      - name: Collect build metadata
+        run: |
+          set -euo pipefail
+          {
+            echo "BUILD_VERSION=${VERSION#v}"
+            echo "GIT_COMMIT=$(git rev-parse HEAD)"
+            echo "GIT_TREE_STATE=$([ -z "$(git status --porcelain)" ] && echo clean || echo dirty)"
+            echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)"
+          } >> "$GITHUB_ENV"
+
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -93,6 +93,7 @@ jobs:
         run: |
           platform=${{ matrix.platform }}
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
+          echo "ARCH=${platform#linux/}" >> $GITHUB_ENV
 
       - name: Checkout source code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -133,6 +134,31 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           outputs: type=image,name=${{ env.IMAGE_REGISTRY }}/${{ env.OPERATOR_IMAGE_REPOSITORY }},push-by-digest=true,name-canonical=true,push=true
 
+      - name: Export operator binary
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
+        with:
+          platforms: ${{ matrix.platform }}
+          target: artifacts
+          outputs: type=local,dest=${{ runner.temp }}/binary
+
```
tariq-hasan, line 163: The two Docker build steps should pass in git metadata such as `VERSION`, `GIT_COMMIT`, `GIT_TREE_STATE` and `SOURCE_DATE_EPOCH` using the build-arg flag.

This will ensure that the provenance metadata stays uniform in both binaries.

tariq-hasan, line 163: The checksums of the binaries also need to be matched to ensure that the second Docker build did not have a cache miss which could have led to a rebuild.

This can be done by pulling the image by its digest and matching against the exported binary.

Makefile
```diff
@@ -14,7 +14,9 @@ SHELL = /usr/bin/env bash -o pipefail
 
 # Version information.
 VERSION := $(shell cat VERSION | sed "s/^v//")
-BUILD_DATE := $(shell date -u +"%Y-%m-%dT%H:%M:%S%:z")
+SOURCE_DATE_EPOCH ?= $(shell git log -1 --format=%ct 2>/dev/null || date -u +%s)
+BUILD_DATE := $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" +"%Y-%m-%dT%H:%M:%S%:z" 2>/dev/null \
+                   || date -u -r $(SOURCE_DATE_EPOCH) +"%Y-%m-%dT%H:%M:%S%:z")
```
tariq-hasan: This makes the build date equal to the date of last commit - which may not be the actual date when the build occurred. Can we retain the default behavior in the build target and let the CI override with SOURCE_DATE_EPOCH explicitly?

tariq-hasan (commented):

Makefile
vikas-saxena02: code change done

.github/workflows/release.yaml
vikas-saxena02: code change done

vikas-saxena02: code change done

nabuskey (approved):
/approve

Thank you very much!

PR #3100, opened 2026-08-16
docs: document webhook failure when a service mesh sidecar is injected

### LINKED ISSUE #3090: Webhook fails with TLS handshake EOF when the Istio sidecar is injected, blocking all pod creation
### What happened?

In a mesh-enabled cluster where the webhook pod receives an Istio sidecar, every pod the operator tries to create is rejected because the mutating webhook is unreachable. The operator retries indefinitely and no driver or executor pod is ever created.

The failure is difficult to diagnose because every obvious check passes:

- the webhook pod is `2/2 Running` with no restarts
- its certificate is valid and the CA bundle was patched into the `MutatingWebhookConfiguration` successfully at startup
- there is nothing in the operator logs beyond a generic `EOF`
- the `SparkConnect` resource is accepted and shows **no events at all**, so `kubectl describe` gives no indication of a problem

The only real clue is in the webhook's own logs, where the source address of the failing handshakes is `127.0.0.6` — the Istio sidecar's loopback address, not a genuine client. The connection is being intercepted by the pod's own `istio-proxy`, which attempts mTLS, while the Kubernetes API server is speaking plain webhook TLS. Neither side can complete the other's handshake and both observe `EOF`.

This affects any pod the operator creates, so `SparkApplication` and `SparkConnect` are both blocked.

### Reproduction Code

On a cluster with Istio installed and automatic sidecar injection enabled for the namespace the operator runs in (this is the default in a Kubeflow installation):

1. Install Spark Operator into that namespace with the webhook enabled.
2. Confirm the webhook pod has two containers:

```console
$ kubectl get pods -n kubeflow | grep spark-operator-webhook
spark-operator-webhook-8465b98f5c-nhhkg   2/2     Running   0   9d
```

3. Create any `SparkConnect` or `SparkApplication` resource.
4. Observe that the resource is accepted but no pod is created:

```console
$ kubectl get sparkconnect -n kubeflow-user-example-com
NAME                     AGE   STATUS   PODNAME
spark-connect-e28720a8   69s

$ kubectl get pods -n kubeflow-user-example-com
NAME      READY   STATUS    RESTARTS   AGE
vikas-0   2/2     Running   0          69m
```

### Expected behavior

The operator creates the server and executor pods, and the resource reaches a ready state.

Failing that, the operator surfaces an error that points at the cause. A mesh-intercepted webhook currently produces no events on the resource and a bare `EOF` in the controller log, neither of which suggests a service mesh is involved.

### Actual behavior

The controller cannot create the server pod. It retries with backoff indefinitely:

```
ERROR controller/controller.go:347 Reconciler error
  {"controller": "sparkconnect",
   "SparkConnect": {"name":"spark-connect-e28720a8","namespace":"kubeflow-user-example-com"},
   "error": "failed to create or update server pod: Internal error occurred:
             failed calling webhook \"mutate--v1-pod.sparkoperator.k8s.io\":
             failed to call webhook:
             Post \"https://spark-operator-webhook-svc.kubeflow.svc:9443/mutate--v1-pod?timeout=10s\": EOF"}
```

The webhook's own logs show the handshakes failing, with `127.0.0.6` as the source — the Istio sidecar loopback address:

```
2026/08/13 11:03:24 http: TLS handshake error from 127.0.0.6:52335: EOF
2026/08/13 11:03:24 http: TLS handshake error from 127.0.0.6:60629: EOF
2026/08/13 11:03:25 http: TLS handshake error from 127.0.0.6:44915: EOF
2026/08/13 11:03:29 http: TLS handshake error from 127.0.0.6:50097: EOF
2026/08/13 11:03:44 http: TLS handshake error from 127.0.0.6:41303: EOF
2026/08/13 11:04:46 http: TLS handshake error from 127.0.0.6:45689: EOF
```

Startup is clean, so the certificate and CA bundle are not the problem:

```
INFO mutatingwebhookconfiguration/controller.go:72 Updating CA bundle of MutatingWebhookConfiguration {"name": "spark-operator-webhook"}
INFO mutatingwebhookconfiguration/event_handler.go:68 MutatingWebhookConfiguration updated {"name": "spark-operator-webhook", "namespace": ""}
```

### Environment & Versions

- Kubernetes Version: v1.35.0 (kind)
- Spark Operator Version: 2.3.0 (Helm chart 2.3.0)
- Apache Spark Version: 4.0.4
- Istio: 1.26.1, as vendored by Kubeflow manifests v1.10.0
- Kubeflow: v1.10.0

### Additional context

**Workaround**

The chart already supports pod annotations on the webhook, so no code change is needed to work around this. Excluding the webhook port from mesh interception is the narrower of the two options:

```yaml
webhook:
  annotations:
    traffic.sidecar.istio.io/excludeInboundPorts: "9443"
```

Removing the webhook from the mesh entirely also works, but is a broader exemption:

```yaml
webhook:
  annotations:
    sidecar.istio.io/inject: "false"
```

For clusters where mesh coverage matters, a port-level `PeerAuthentication` is a narrower exemption than excluding the port. The sidecar still handles the traffic — so telemetry and `AuthorizationPolicy` continue to apply on 9443 — it simply accepts plain TLS in addition to mTLS:

```yaml
apiVersion: security.istio.io/v1
kind: PeerAuthentication
metadata:
  name: spark-operator-webhook
  namespace: kubeflow
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: spark-operator
      app.kubernetes.io/component: webhook
  mtls:
    mode: STRICT
  portLevelMtls:
    9443:
      mode: PERMISSIVE
```

This is a separate CR rather than a chart value, so it is not a candidate for a chart default — but it is worth documenting alongside the annotation for installations running STRICT mesh-wide.

**Why this is worth addressing rather than leaving to users**

The Kubernetes API server cannot participate in the mesh, so a mesh-injected admission webhook can never work without an exemption of some kind. That makes this deterministic rather than environment-specific: any mesh-enabled installation with injection enabled in the operator's namespace hits it.

It is also relevant to anyone running Spark Operator inside Kubeflow, where a service mesh is part of the default installation.

**Suggestions, in increasing order of intervention**

1. Document it. A note in the Helm chart README or the installation docs covering the symptom, the `127.0.0.6` clue, and the annotation would save considerable time — the observable behaviour points nowhere near a service mesh.
2. Ship the port exclusion as a default in `values.yaml`. It is inert on clusters without Istio, since the annotation is simply ignored.
3. Surface a clearer error. The operator could detect repeated webhook handshake failures and record an event on the resource, rather than leaving it with an empty status.

Happy to raise a PR for (1), and for (2) if maintainers think a default exclusion is appropriate.

**Possibly related**

#1148 covers Istio interaction but predates the current webhook architecture and is closed.

### Impacted by this bug?

vikas-saxena02:
**What this PR does / why we need it**:

Documents the webhook becoming unreachable when its pod receives a service mesh sidecar, and how to work around it.

Adds a `Mutating Admission Webhook in a service mesh` subsection to the getting-started guide, alongside the existing `Mutating Admission Webhooks on a private GKE or EKS cluster` subsection — both describe the API server being unable to reach the webhook in particular cluster configurations.

The subsection covers:

- the symptom: no driver or executor pods created, a generic `EOF` in the operator log, and no events on the `SparkApplication` or `SparkConnect` resource
- why it is easy to misdiagnose: the webhook pod is `Running`, its certificate is valid, and the CA bundle is patched successfully at startup
- the `127.0.0.6` source address in the webhook logs, which is the sidecar loopback rather than a real client, and the only real clue
- why it happens: the API server does not participate in the mesh, so any mesh-injected admission webhook needs its port exempted
- two remedies: excluding the webhook port from sidecar interception via `webhook.annotations`, and a port-level `PeerAuthentication` for clusters running STRICT mTLS mesh-wide

Both examples were rendered locally against master to confirm they work — the `--set-string` annotation escaping produces the expected pod annotation, and the `PeerAuthentication` selector matches the labels the chart generates.

As discussed in the issue, this is not specific to Spark Operator, so documenting the symptom and the remedies seemed the right level of intervention rather than changing chart defaults.

**Which issue(s) this PR fixes**:

Fixes #3090

**Checklist**:

- [ ] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/docs/website/getting-started/index.md
+++ b/docs/website/getting-started/index.md
@@ -332,3 +332,74 @@ helm install my-release spark-operator/spark-operator \
    --set webhook.enable=true \
    --set webhook.port=443
 ```
+
+### Mutating Admission Webhook in a service mesh
+
+If the webhook pod receives a service mesh sidecar (for example Istio with automatic
+injection enabled in the operator's namespace, which is the default in a Kubeflow
+installation), the Kubernetes API server cannot reach the webhook and the operator will be
+unable to create any driver or executor pods.
+
+The failure is easy to misdiagnose, because everything else looks healthy: the webhook pod
+is `Running`, its certificate is valid, and the CA bundle is patched into the
+`MutatingWebhookConfiguration` successfully at startup. The `SparkApplication` or
+`SparkConnect` resource is accepted but never progresses, and has no events.
+
+The operator logs a generic `EOF`:
+
+```
+ERROR Reconciler error {"error": "failed to create or update server pod: Internal error occurred:
+  failed calling webhook \"mutate--v1-pod.sparkoperator.k8s.io\": failed to call webhook:
+  Post \"https://spark-operator-webhook-svc.spark-operator.svc:9443/mutate--v1-pod?timeout=10s\": EOF"}
+```
+
+The webhook's own logs show the handshakes failing, with `127.0.0.6` as the source address:
+
+```
+http: TLS handshake error from 127.0.0.6:52335: EOF
+http: TLS handshake error from 127.0.0.6:60629: EOF
+```
+
+`127.0.0.6` is the sidecar's loopback address, not a real client. The connection is being
+intercepted by the pod's own sidecar, which attempts mTLS, while the API server is speaking
+plain webhook TLS. Neither side can complete the other's handshake.
+
+This is not specific to this operator. The Kubernetes API server does not participate in the
+mesh, so any mesh-injected admission webhook needs the webhook port exempted from mesh mTLS.
+
+The simplest fix is to exclude the webhook port from sidecar interception:
+
+```shell
+helm install my-release spark-operator/spark-operator \
+   --namespace spark-operator \
+   --create-namespace \
+   --set webhook.enable=true \
+   --set-string webhook.annotations."traffic\.sidecar\.istio\.io/excludeInboundPorts"=9443
+```
+
+For clusters running STRICT mTLS mesh-wide, a port-level `PeerAuthentication` is a narrower
+exemption. The sidecar continues to handle the traffic, so telemetry and any
+`AuthorizationPolicy` still apply on the webhook port; it simply accepts plain TLS in
+addition to mTLS:
+
+```yaml
+apiVersion: security.istio.io/v1
+kind: PeerAuthentication
+metadata:
+  name: spark-operator-webhook
+  namespace: spark-operator
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: spark-operator
+      app.kubernetes.io/component: webhook
+  mtls:
+    mode: STRICT
+  portLevelMtls:
+    9443:
+      mode: PERMISSIVE
+```
+
+Removing the webhook from the mesh entirely
+(`sidecar.istio.io/inject: "false"`) also works, but is a broader exemption than either of
+the above.
--- a/docs/website/getting-started/index.md
+++ b/docs/website/getting-started/index.md
@@ -400,6 +400,10 @@ spec:
       mode: PERMISSIVE
 ```
 
+If more than one release of the operator is installed in the namespace, add
+`app.kubernetes.io/instance: <release-name>` to the selector so the policy applies only to
+the intended webhook.
+
 Removing the webhook from the mesh entirely
 (`sidecar.istio.io/inject: "false"`) also works, but is a broader exemption than either of
 the above.


nabuskey (approved):
We really need to clean up docs in general. We need to move troubleshooting sections and topics to another doc. 

I am ok with this doc update as of now. Will open an issue with proposed documentation layout.

/approve

docs/website/getting-started/index.md
```diff
@@ -332,3 +332,78 @@ helm install my-release spark-operator/spark-operator \
    --set webhook.enable=true \
    --set webhook.port=443
 ```
+
+### Mutating Admission Webhook in a service mesh
+
+If the webhook pod receives a service mesh sidecar (for example Istio with automatic
+injection enabled in the operator's namespace, which is the default in a Kubeflow
+installation), the Kubernetes API server cannot reach the webhook and the operator will be
+unable to create any driver or executor pods.
```
juliusvonkohout, line 341: Are you sure that this is correct when you install the community distribution? this might be wrong information.

juliusvonkohout: We have tests for spark-operator in the community distribution

vikas-saxena02: @juliusvonkohout but this is for those cases where spark-operator is installed separately

PR #2814, opened 2026-01-19
Fix SparkConnect nil executor template panic

sameerdattav:
Supersedes #2810
Fixes https://github.com/kubeflow/spark-operator/issues/2807

@ChenYi015 I rebased the branch onto upstream master and reapplied only the intended SparkConnect fix and its test.
This removes the unintended changes that were present in the previous PR.

Thanks!

PR #3103, opened 2026-08-19
fix(sparkconnect): resolve images from selected template containers

immanuwell:
## Purpose of this PR

SparkConnect validation accepted an image from any pod template container. 
Reconciliation only used the selected Spark container, so some accepted resources failed forever with `image is not specified`

This restores Spark's first-container fallback for executor templates and validates the container runtime will actually select

Related: #2807, #2814

## Change Category

- [ ] Documentation update

### Reproduction

Apply this with the default webhook enabled:

```yaml
apiVersion: sparkoperator.k8s.io/v1alpha1
kind: SparkConnect
metadata:
  name: image-repro
spec:
  sparkVersion: 4.0.4
  server:
    template:
      spec:
        containers:
        - name: spark-connect
          image: apache/spark:4.0.4
  executor:
    template:
      spec:
        containers:
        - name: executor
          image: apache/spark:4.0.4
```

Before this change the CR is accepted, then reconciliation fails with `image is not specified`. After it, the first executor container supplies the image as Spark documents.

## Checklist

- [x] Documentation is not needed for this behavior fix.

Tests: `make unit-test`, `make go-fmt`, `make go-vet`, `make go-lint`

### DIFF
--- a/internal/controller/sparkconnect/options.go
+++ b/internal/controller/sparkconnect/options.go
@@ -84,7 +84,7 @@ func imageOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 
 	template := conn.Spec.Executor.Template
 	if template != nil && len(template.Spec.Containers) != 0 {
-		index := -1
+		index := 0
 		for i, container := range template.Spec.Containers {
 			if container.Name == common.Spark3DefaultExecutorContainerName {
 				index = i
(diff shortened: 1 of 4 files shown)


tariq-hasan (commented):
Thanks!
/lgtm

A shared utility in `pkg/util` for the Spark container-selection rules would be nice as they appear in multiple places - the [server pod mutation](https://github.com/immanuwell/spark-operator/blob/eb7d3cfdec4cbe46e841af565e2680f3067f6de5/internal/controller/sparkconnect/reconciler.go#L361-L367) logic, the image option selection logic and the webhook validation. I have raised a separate issue for it: https://github.com/kubeflow/spark-operator/issues/3105

nabuskey (approved):
/approve

PR #3109, opened 2026-08-22
feat(sparkconnect): add service account fallback for Spark Connect server pod

### LINKED ISSUE #3049: Add fallback mechanism for driver pod service account

vikas-saxena02:
## Purpose of this PR

Part of #3049. Adds a service account fallback for the Spark Connect server pod,
completing the issue alongside #3092 .

Two changes:

1. A new optional `.spec.server.serviceAccount` field on the `SparkConnect` CRD,
   for parity with `.spec.driver.serviceAccount` on `SparkApplication`.
2. The server pod now falls back to the `--default-service-account` controller
   flag introduced in #3092 when neither the `SparkConnect` nor its
   server pod template specifies one.

## Precedence

Resolved in `resolveServerServiceAccount`, highest first:

1. `.spec.server.serviceAccount`
2. `.spec.server.template.spec.serviceAccountName`
3. `--default-service-account`

If none is set the field is left empty and the namespace default applies, so
behaviour is unchanged for existing users.

I proposed this ordering on the issue and nobody objected, but it wasn't
explicitly confirmed — happy to change it if the pod template should win over
the dedicated field.

## Scope

Server pod only. Per @tariq-hasan's guidance on #3049, the fallback exists
because the driver needs permission to create executor pods; executors don't
have the same requirement. That's also why the field sits on `ServerSpec`
rather than the embedded `SparkPodSpec`, which `ExecutorSpec` shares.

## Implementation note

The resolution happens in `mutateServerPod` *after* the pod template is copied,
since `pod.Spec = template.Spec` is a wholesale assignment that would otherwise
discard it. It's also inside the `pod.CreationTimestamp.IsZero()` guard, as
`serviceAccountName` is immutable on an existing pod.

## Stacked on #3092

This branches off the `--default-service-account` PR and won't build without it,
so the diff includes that commit. I'll rebase onto master once it merges.

## Testing

- `resolveServerServiceAccount` unit tests covering all four precedence cases
  plus an empty-string server value.
- `mutateServerPod` specs asserting the resolved value survives the pod template
  copy.
- `make generate`, `make update-crd`, `make detect-crds-drift` — CRD, deepcopy,
  openapi, swagger and Python model regenerated and in sync.
- `make go-vet`, `make build-operator`, `make unit-test`, `make helm-unittest`
  all pass.

Closes #3049


api/v1alpha1/sparkconnect_types.go
```diff
@@ -88,6 +88,13 @@ type SparkConnectSpec struct {
 type ServerSpec struct {
 	SparkPodSpec `json:",inline"`
 
+	// ServiceAccount is the name of the Kubernetes service account used by the server pod.
+	// If not specified, the service account of the server pod template is used, and if that
+	// is also not specified, the operator falls back to the service account configured via
+	// the --default-service-account flag.
+	// +optional
+	ServiceAccount *string `json:"serviceAccount,omitempty"`
+
```
tariq-hasan: Actually thinking about it having a CRD field here is not justifiable given that pod template already gives the provision to add a service account.

Historically speaking SparkApplication introduced service account on the CRD when Spark began to support it: https://github.com/kubeflow/spark-operator/pull/924. Pod template support was introduced at a later date: https://github.com/kubeflow/spark-operator/pull/2141. The parity argument I proposed may be weak in light of this.

I think we should remove this field for now unless there is further justification for adding more API surface and introducing additional complexity.

cmd/operator/controller/start.go
```diff
@@ -597,8 +597,9 @@ func newScheduledSparkApplicationReconcilerOptions() scheduledsparkapplication.O
 
 func newSparkConnectReconcilerOptions() sparkconnect.Options {
 	options := sparkconnect.Options{
-		Namespaces:        namespaces,
-		NamespaceSelector: namespaceSelector,
+		Namespaces:            namespaces,
+		NamespaceSelector:     namespaceSelector,
+		DefaultServiceAccount: defaultServiceAccount,
```
tariq-hasan: Since default service account flag now threads through to spark connect controller as well we should update the flag documentation to reflect the scope - the documentation now only scopes to SparkApplication.

https://github.com/vikas-saxena02/spark-operator/blob/b6c78895f5bd6d0e781b06553403ebe852091aec/charts/spark-operator-chart/values.yaml#L134-L138

cmd/operator/controller/start.go
tariq-hasan: Since default service account flag now threads through to spark connect controller as well we may want to update the wording here to reflect the scope - the wording now only scopes to SparkApplication.

Same in the below places: [values.yaml](https://github.com/vikas-saxena02/spark-operator/blob/b6c78895f5bd6d0e781b06553403ebe852091aec/charts/spark-operator-chart/values.yaml#L134-L138), SparkConnect [docs](https://github.com/vikas-saxena02/spark-operator/blob/b6c78895f5bd6d0e781b06553403ebe852091aec/docs/website/user-guide/spark-connect.md) and SparkApplication [docs](https://github.com/vikas-saxena02/spark-operator/blob/b6c78895f5bd6d0e781b06553403ebe852091aec/docs/website/user-guide/writing-sparkapplication.md).

tariq-hasan (commented):

api/v1alpha1/sparkconnect_types.go
vikas-saxena02: Agreed — removed the field. The pod template already covers this, and the parity argument doesn't hold given the historical ordering you pointed out. Precedence is now just pod template → controller flag.

cmd/operator/controller/start.go
vikas-saxena02: Updated in all four places: the flag help text, `values.yaml` (and the regenerated chart README), the SparkApplication docs, and the SparkConnect docs.

docs/website/user-guide/writing-sparkapplication.md
```diff
@@ -93,7 +93,7 @@ spec:
...
... be used. The driver pod by default uses the `default` service account in the namespace it is running in to talk to the Kubernetes API server. The `default` service account, however, may or may not have sufficient permissions to create executor pods and the headless service used by the executors to connect to the driver. If it does not and a custom service account that has the right permissions should be used instead, the optional field `.spec.driver.serviceAccount` can be used to specify the name of the custom service account. Cluster administrators can also configure a cluster-wide default with the `--default-service-account` controller flag (`controller.defaultServiceAccount` in the Helm chart), which applies when neither `.spec.driver.serviceAccount` nor a service account in `.spec.driver.template` is set. The same flag also applies to the Spark Connect server pod. The same flag also applies to the Spark Connect server pod. When a custom container image is needed for the driver, the field `.spec.driver.image` can be used to specify it. This overrides the image specified in `.spec.image` if it is also set. It is invalid if both `.spec.image` and `.spec.driver.image` are not set.
```
tariq-hasan: We should remove the duplicate sentence.

`The same flag also applies to the Spark Connect server pod. The same flag also applies to the Spark Connect server pod.`

internal/controller/sparkconnect/reconciler.go
```diff
@@ -351,6 +356,12 @@ func (r *Reconciler) mutateServerPod(_ context.Context, conn *v1alpha1.SparkConn
 			pod.Spec = template.Spec
 		}
 
+		// Resolve the service account for the server pod. This must happen after the pod
+		// template has been copied above, which would otherwise overwrite it.
+		if serviceAccount := resolveServerServiceAccount(conn, r.options.DefaultServiceAccount); serviceAccount != "" {
+			pod.Spec.ServiceAccountName = serviceAccount
+		}
+
```
tariq-hasan: We can remove the helper altogether since precedence is now simpler. And we can add a log line as well.

```suggestion
		// Fall back to the operator-level default service account. This must happen after the
		// pod template has been copied above, which would otherwise overwrite it.
		if pod.Spec.ServiceAccountName == "" && r.options.DefaultServiceAccount != "" {
			pod.Spec.ServiceAccountName = r.options.DefaultServiceAccount
			ctrl.LoggerFrom(ctx).Info("Applied default service account to Spark Connect server pod",
				"serviceAccount", r.options.DefaultServiceAccount)
		}
```

internal/controller/sparkconnect/reconciler_test.go
```diff
@@ -252,5 +252,41 @@ var _ = Describe("mutateServerPod", func() {
...
+				},
+			}
+			pod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: conn.Namespace,
+				},
+			}
+			Expect(reconciler.mutateServerPod(context.TODO(), conn, pod)).To(Succeed())
+			Expect(pod.Spec.ServiceAccountName).To(Equal("template-sa"))
+		})
```
tariq-hasan, line 290: We can have a case for when the template is present with an empty `serviceAccountName` that should still get the fallback.

cmd/operator/controller/start.go
tariq-hasan, line 611: Given that the CRD field is gone and we are just wiring in the default service account or using pod template we should update https://github.com/vikas-saxena02/spark-operator/blob/7a3bb192b6fa231199c828a22a52f7ac605ca6b4/internal/webhook/sparkconf_validator.go#L35-L36 to say

```
	common.SparkKubernetesAuthenticateDriverServiceAccountName:   "configure the service account via the CRD spec or pod template instead",
	common.SparkKubernetesAuthenticateExecutorServiceAccountName: "configure the service account via the CRD spec or pod template instead",
```

tariq-hasan (commented):
Just have a few minor comments. Otherwise lgtm. Thanks.

docs/website/user-guide/writing-sparkapplication.md
vikas-saxena02: done

internal/controller/sparkconnect/reconciler_test.go
vikas-saxena02: done

cmd/operator/controller/start.go
vikas-saxena02: done

internal/controller/sparkconnect/reconciler.go
vikas-saxena02: done

andreyvelich (commented):
Sure, thanks @vikas-saxena02!
/lgtm
/approve
/cc @RobuRishabh @ChenYi015

PR #3099, opened 2026-08-15
fix(webhook): generate valid volume names for ConfigMaps

immanuwell:
## Purpose of this PR

ConfigMap names may contain dots, but Pod volume names cannot. With the webhook enabled, mounting `spark.application.conf` through `spec.driver.configMaps` creates `spark.application.conf-vol`, so Kubernetes rejects the Pod with `must not contain dots`

Repro: create that ConfigMap, add it to `spec.driver.configMaps`, then submit the SparkApplication

**Proposed changes:**

- Generate valid, unique volume names for dotted or long ConfigMap names
- Keep ordinary names unchanged and add regression tests

## Change Category

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that could affect existing functionality)
- [ ] Documentation update

### Rationale

This makes the documented ConfigMap mount flow work with valid Kubernetes ConfigMap names

## Checklist

- [x] I have conducted a self-review of my own code.
- [ ] I have updated documentation accordingly - N/A here
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.

Documentation is unchanged because this only corrects generated volume names.

### Additional Notes

Vet, lint, unit tests, and the operator build pass locally.

PR #2072, opened 2024-06-25
Use controller-runtime to reconsturct spark operator

ChenYi015:
## Purpose of this PR

Use [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) to reconsturct spark operator.

Close #547
Close #507 
Close #2052 

### Spark Operator Executable

- Split the operator into two separate components i.e. controller and webhook.

- Use [cobra](https://github.com/spf13/cobra) to implement `spark-operator` command, and it has two sub commands `controller` and `webhook`, one can start controller/webhook by `spark-operator controller start` and `spark-operator webhook start` repectively. 

    ```shell
    $ make build-operator
    $ bin/spark-operator --help          
    Spark operator
    
    Usage:
      spark-operator [flags]
      spark-operator [command]
    
    Available Commands:
      completion  Generate the autocompletion script for the specified shell
      controller  Spark operator controller
      help        Help about any command
      version     Print version information
      webhook     Spark operator webhook
    
    Flags:
      -h, --help   help for spark-operator
    
    Use "spark-operator [command] --help" for more information about a command.
    ```

- SparkApplication/ScheduledSparkController controller only watches resources in the spark job namespaces.
- Add finalizer for SparkApplication to ensure all sub resources e.g. pods, services, ingress related to that app will be deleted when deleting the SparkApplication.

### Helm Chart

- Change `imagePullSecrets` to `image.pullSecrets`.
- All controller configurations are prefixed with `controller` e.g. `controller.replicaCount` and `controller.resyncInterval`
- All webhook configurations are prefixed with `webhook` e.g. `webhook.replicaCount` and `webhook.failurePolicy.
- All monitoring configurations are prefixed with `promethues` e.g. `promethues.metrics` and `promethues.podMonitor`.
- Enable leader election for controller/webhook by default to ensure only an controller instance is working when upgrading the chart.
- The update strategy of controller/webhook deployment will be the rolling update, not recreate.
- Deploy webhook server with a separated deployment and enable webhook server by default.
- The webhook secret will not be created and updated by operator, and the certificates will not change during chart upgrading/rollback process.
- Change the default of webhook `timeoutSeconds` from `30` to `10`.
- Change the default of webhook `failurePolicy` from `Ignore` to `Fail`.
- Change the default spark job namespace from `[]` to `["default]`, thus the SparkApplication under `examples` directory can be running directly without creating rbac resources manually.
[...]

PR #3111, opened 2026-08-23
fix(webhook): generate valid volume name for Prometheus ConfigMap

### LINKED ISSUE #3102: Fix Prometheus ConfigMap volume name
### What happened?

In https://github.com/kubeflow/spark-operator/pull/3099, ConfigMap volume name is now sanitized to handle cases where the volume name end up being invalid. This also needs to be extended to `addPrometheusConfig`

rjgoyln:
## Summary

`addPrometheusConfig` still derives `<configmap name>-vol`, the shape #3099 replaced for `spec.driver.configMaps`. A SparkApplication name only has to be a DNS-1035 label, so once it reaches 50 characters the derived volume name passes 63 and Kubernetes rejects every driver and executor pod of that application. The Prometheus ConfigMap itself stays valid, so nothing catches it earlier.

`TestGetConfigMapVolumeName` gains the two cases either side of that limit, since today its `<=` can be flipped to `<` unnoticed.

Closes #3102

## Change Category

## Checklist

### Additional Notes

The `TestPatchSparkPod_PrometheusConfigMaps` block commented out in #2072 sits below the new test and still asserts the old naming. Happy to drop it here.

### DIFF
--- a/internal/webhook/sparkpod_defaulter.go
+++ b/internal/webhook/sparkpod_defaulter.go
@@ -383,7 +383,7 @@ func addPrometheusConfig(pod *corev1.Pod, app *v1beta2.SparkApplication) error {
 	}
 
 	name := util.GetPrometheusConfigMapName(app)
-	volumeName := name + "-vol"
+	volumeName := getConfigMapVolumeName(name)
 	mountPath := common.PrometheusConfigMapMountPath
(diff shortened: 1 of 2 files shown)


nabuskey (approved):
LGTM thank you very much!

PR #3120, opened 2026-08-26
feat(sparkconnect): add a `Dependencies` field to `SparkConnectSpec`

### LINKED ISSUE #2959: Add a `Dependencies` field to `SparkConnectSpec` for structured JAR/package management

Add a `Dependencies` field to `SparkConnectSpec`, mirroring the existing `Dependencies` struct already present on `SparkApplicationSpec`.

### Why is this needed?

The Kubeflow SDK's `SparkClient` ([kubeflow/sdk#470](https://github.com/kubeflow/sdk/issues/470)) wants to offer first-class dependency management so that users can attach table-format and storage connectors (Iceberg, Delta Lake, Hudi, S3A) without hand-writing Spark config strings. Today, there is no structured way to declare dependencies in `SparkConnectSpec`.

`SparkApplication` already solves exactly this problem: its `Dependencies` struct is consumed by the submission path, which emits real `--jars`, `--packages`, `--exclude-packages`, and `--repositories` flags. `SparkConnect` is the only workload type in the operator without this capability, which is an inconsistency in the API surface and forces every higher-level tool to reinvent dependency handling on top of raw `sparkConf`.

_No response_

_No response_

### Additional context

Tracking issue on the SDK side: [kubeflow/sdk#470](https://github.com/kubeflow/sdk/issues/470).

### Love this feature?

Mr-Neutr0n:
Fixes #2959

Added a `Dependencies` struct and `spec.dependencies` field to `SparkConnectSpec`, and wired the Spark Connect controller to pass dependency flags through to `start-connect-server.sh` the same way `SparkApplication` does for `spark-submit`.

Local tests pass.

---
This change was prepared with AI assistance under human direction and review.


api/v1alpha1/common_types.go
```diff
@@ -49,6 +49,36 @@ const (
...
+	// +optional
+	ExcludePackages []string `json:"excludePackages,omitempty"`
+	// Repositories is a list of additional remote repositories to search for the maven coordinate
+	// given with the "packages" option.
+	// +optional
+	Repositories []string `json:"repositories,omitempty"`
+	// Archives is a list of archives to be extracted into the working directory of each executor.
+	// +optional
+	Archives []string `json:"archives,omitempty"`
+}
```
tariq-hasan, line 80: I think we should keep to only `jars`, `packages`, `excludePackages` and 
`repositories`. They should address the Maven dependencies for now.

`pyFiles`, `files` and `archives` can be revisited later as more use cases occur.

internal/controller/sparkconnect/options.go
```diff
@@ -132,6 +168,10 @@ func shellQuoteSparkConfig(key, value string) string {
 	return fmt.Sprintf("'%s'", strings.ReplaceAll(config, "'", `'"'"'`))
 }
 
+func shellQuoteSparkArg(value string) string {
+	return fmt.Sprintf("'%s'", strings.ReplaceAll(value, "'", "'\"'\"'"))
+}
+
 func hadoopConfOption(conn *v1alpha1.SparkConnect) ([]string, error) {
 	if conn.Spec.HadoopConf == nil {
 		return nil, nil
```
tariq-hasan: We also need to specify precedence rules in the reconciler for when both one of these CRD fields and the equivalent spark config is provided. The CRD field should win in this case as per the general pattern that's followed in the operator.

api/v1alpha1/common_types.go
tariq-hasan, line 80: The comments also need to change to refer to sparkconnect and not sparkapplication.

tariq-hasan (commented):
Hi @Mr-Neutr0n! I have added some more comments.

Please also update `docs/website/user-guide/spark-connect.md`, add a new `examples/sparkconnect/spark-connect-dependencies.yaml` and add/update e2e tests.

Mr-Neutr0n:
Addressed the review in `244c6dd`:

- `Dependencies` now carries only `jars`, `packages`, `excludePackages`, and `repositories`. `files`, `pyFiles`, and `archives` are gone from the API type, the controller, and the regenerated artifacts (deepcopy, OpenAPI, Swagger, CRDs, Python API). The type and field comments now describe the Spark Connect server instead of a Spark application.
- `.spec.deps` is documented in `docs/website/user-guide/spark-connect.md`.
- Added `examples/sparkconnect/spark-connect-dependencies.yaml`.
- Added `test/e2e/sparkconnect_dependencies_test.go`, which starts the server from the new example and asserts that the dependency flags reach `spark-submit`.

Verification: `make unit-test` (with envtest), `make go-vet`, `make go-lint` (0 issues) and a build of the e2e module all pass. I could not execute the e2e suite locally (no container runtime for kind here), so that test compiles but was not run in this environment.

Mr-Neutr0n:
Two CI checks needed generated updates rather than code changes: the Python model was missing the generator's trailing blank lines, and `docs/api-docs.md` had to be regenerated for the new v1alpha1 type. I rebased onto current `master`, regenerated all artifacts on that base, and re-ran the checks locally (`make unit-test`, `go vet`, `golangci-lint` 0 issues, root and e2e builds). The e2e matrix had been cancelled by fail-fast on the previous run, so it should run fresh on this head.

Mr-Neutr0n:
The e2e run was right: the dependencies test timed out waiting for the server pod to become ready, which needs the kind cluster to resolve the Maven artifacts (the rest of the suite only uses artifacts bundled in the Spark image). I narrowed the test to what this change owns: the SparkConnect is reconciled into a server pod whose `spark-submit` command carries the `--jars`, `--packages`, `--exclude-packages`, and `--repositories` flags. Happy to extend it to a full startup check if the e2e cluster has Maven egress.

PR #2827, opened 2026-01-31
feat: add configurable timestamp precision for ScheduledSparkApplication names

puwun:
## Purpose of this PR
Added a global controller option to configure the timestamp precision used when generating SparkApplication names from ScheduledSparkApplications. This addresses the issue where nanosecond timestamps (19 digits) combined with long application names can exceed Kubernetes' 63-character name limit.

Fixes #2602 

**Proposed changes:**

- Add `--scheduled-spark-application-timestamp-precision` flag to controller with valid values: `nanos` (default), `micros`, `millis`, `seconds`, `minutes`
- Add formatTimestamp() helper function and update createSparkApplication() to use configurable precision
- Add `controller.scheduledSparkApplicationTimestampPrecision` Helm chart value and pass it to controller deployment
- Add unit tests for all precision levels

## Change Category

- [x] Feature (non-breaking change which adds functionality)

### Rationale

When using ScheduledSparkApplication with long names, the generated SparkApplication names can exceed Kubernetes' 63-character limit due to the 19-digit nanosecond timestamp suffix. This feature allows operators to configure shorter timestamp precisions (e.g., `seconds` = 10 digits, `minutes` = 8 digits) to stay within the limit while maintaining uniqueness for their scheduling frequency.
The default remains `nanos` for full backward compatibility.

## Checklist

- [x] I have conducted a self-review of my own code.
- [x] I have updated documentation accordingly.
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] Existing unit tests pass locally with my changes.

PR #3121, opened 2026-08-28
fix(controller): keep scheduled run names within DNS label limit

immanuwell:
## Purpose of this PR

A 55 character `ScheduledSparkApplication` name is accepted, but even the shortest `minutes` suffix creates a 64 character run name. 
The child webhook rejects it, so the schedule never starts. 
so this is still reachable after #2827

Related to #2602.

Proposed changes:

- Reuse the existing hash and truncation helper for scheduled run names
- Keep names unchanged when they already fit
- Test the real parent admission and child creation path, plus collision safety

## Reproduction

1. Create a schedule with `metadata.name` set to 55 lowercase letters
2. Set timestamp precision to `minutes`
3. Wait for a run. The parent exists, but child creation fails with `must be no more than 63 characters`

## Change Category

- [ ] Documentation update

### Rationale

Only names that already fail are changed. Working names stay byte for byte the same, so yeah this is pretty contained.

## Checklist

- [x] Documentation is not needed for this behavior fix.

### Additional Notes

`go test ./...`, `go vet ./...`, and `make go-lint` pass locally.

### DIFF
--- a/internal/controller/scheduledsparkapplication/controller.go
+++ b/internal/controller/scheduledsparkapplication/controller.go
@@ -275,7 +275,7 @@ func (r *Reconciler) createSparkApplication(
 	}
 	app := &v1beta2.SparkApplication{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-%s", scheduledApp.Name, formatTimestamp(t, r.options.TimestampPrecision)),
+			Name:      util.GetScheduledSparkApplicationRunName(scheduledApp, formatTimestamp(t, r.options.TimestampPrecision)),
 			Namespace: scheduledApp.Namespace,
 			Labels:    labels,
 			OwnerReferences: []metav1.OwnerReference{{
--- a/pkg/util/sparkapplication.go
+++ b/pkg/util/sparkapplication.go
@@ -201,6 +201,12 @@ func generateName(name, suffix string) string {
 	return fmt.Sprintf("%s-%s-%s", name[:maxNameLength], hash[:8], suffix)
 }
 
+// GetScheduledSparkApplicationRunName returns the name for a SparkApplication
+// created by a ScheduledSparkApplication.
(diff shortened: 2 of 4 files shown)


pkg/util/sparkapplication.go
```diff
@@ -201,6 +201,12 @@ func generateName(name, suffix string) string {
 	return fmt.Sprintf("%s-%s-%s", name[:maxNameLength], hash[:8], suffix)
 }
 
+// GetScheduledSparkApplicationRunName returns the name for a SparkApplication
+// created by a ScheduledSparkApplication.
+func GetScheduledSparkApplicationRunName(app *v1beta2.ScheduledSparkApplication, timestamp string) string {
+	return generateName(app.Name, timestamp)
+}
+
```
tariq-hasan, line 209: With this I think we'd want to hash only the app name and not the app name and timestamp combination. That would lead to a more predictable and ordered list of scheduled spark application names.

Add to that the ordering is going to become a problem if we want to sort each app by its timestamp as well since the hash is in the middle.

internal/controller/scheduledsparkapplication/controller.go
```diff
@@ -275,7 +275,7 @@ func (r *Reconciler) createSparkApplication(
 	}
 	app := &v1beta2.SparkApplication{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-%s", scheduledApp.Name, formatTimestamp(t, r.options.TimestampPrecision)),
+			Name:      util.GetScheduledSparkApplicationRunName(scheduledApp, formatTimestamp(t, r.options.TimestampPrecision)),
 			Namespace: scheduledApp.Namespace,
```
tariq-hasan, line 279: Logging if the name changes would help with better observability.

tariq-hasan (commented):
Hi @immanuwell! I have added a few comments.

PR #3129, opened 2026-08-31
feat(helm): support restricted pod security for CRD upgrade hook Job

### LINKED ISSUE #3070: Helm hook Job cannot run under restricted Pod Security admission
### What happened

The Helm hook Job created by `hook.upgradeCrd` cannot be scheduled in a namespace that enforces the `restricted` Pod Security Standard, so the CustomResourceDefinition upgrade never runs there. There is no chart value that can fix it.

### Details

`charts/spark-operator-chart/templates/hook/job.yaml` sets a container security context:

```yaml
securityContext:
  readOnlyRootFilesystem: true
  privileged: false
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
```

The `restricted` profile additionally requires `runAsNonRoot: true` and a `seccompProfile`. Neither is set, there is no pod-level `securityContext`, and the `hook:` values block exposes only `upgradeCrd`, `image`, `nodeSelector`, `affinity` and `tolerations` — so an operator cannot supply them either.

The controller and webhook are unaffected: `controller.podSecurityContext`, `controller.securityContext`, `webhook.podSecurityContext` and `webhook.securityContext` all exist. The hook Job appears to be the one component that was missed.

### Reproduction

```bash
kubectl create namespace spark-operator
kubectl label namespace spark-operator \
  pod-security.kubernetes.io/enforce=restricted

helm upgrade --install spark-operator spark-operator/spark-operator \
  --namespace spark-operator \
  --set hook.upgradeCrd=true
```

The hook pod is rejected by admission, so `helm upgrade` does not complete.

### A complication worth flagging

`docker/Dockerfile.kubectl` builds from `alpine` and sets no `USER`, so the image runs as UID 0. Adding `runAsNonRoot: true` alone would therefore make the pod fail to start with *"container has runAsNonRoot and image will run as root"*. A fix needs a `runAsUser` as well, or a `USER` line in the Dockerfile.

Setting `runAsUser` in the chart looks viable — `kubectl` does not need root, `readOnlyRootFilesystem` is already enabled, and the service account token is world-readable — but it should be verified on a cluster rather than assumed.

### Proposed change

Add `hook.podSecurityContext` and `hook.securityContext` values, wired into `templates/hook/job.yaml` with the same `with` pattern used by the controller and webhook templates.

One decision worth your input before I send a pull request:

1. **Restricted-compliant defaults** — the hook works out of the box on hardened clusters, at the cost of changing current rendering.
2. **Empty `{}` defaults** — no behaviour change, and each operator opts in.

I lean towards the first, since the container already drops all capabilities and disallows privilege escalation, but it is your call.

### Context

The Kubeflow Community Distribution is adding a Helm chart that wraps this one ([kubeflow/community-distribution#3575](https://github.com/kubeflow/community-distribution/pull/3575)). It ships with `hook.upgradeCrd: false` and documents that the definitions are not upgraded by `helm upgrade`, because the hook cannot run in the `kubeflow` namespace.

Happy to send the pull request, including helm-unittest coverage in `tests/hook/job_test.yaml` and a regenerated `README.md`, once you have indicated which default you prefer.

KHARSHAVARDHAN-eng:
## Description

Fixes #3070.

This PR updates the CRD upgrade hook Job so it can run in Kubernetes namespaces enforcing the `restricted` Pod Security Standard.

## Changes

- Run the kubectl hook image as a non-root user by default.
- Add configurable `hook.podSecurityContext` and `hook.securityContext` values.
- Add restricted Pod Security compliant defaults, including:
  - `runAsNonRoot: true`
  - `seccompProfile.type: RuntimeDefault`
  - `readOnlyRootFilesystem: true`
  - `allowPrivilegeEscalation: false`
  - `privileged: false`
  - `capabilities.drop: [ALL]`
- Add Helm unit tests for default and custom security contexts.
- Regenerate the Helm chart README documentation.

## Testing

- `./bin/helm unittest charts/spark-operator-chart --strict --file "tests/**/*_test.yaml"`
  - 235/235 tests passed
- Verified the rendered Helm template with default and custom security context values.
- Ran `git diff --check` successfully.

### DIFF
--- a/charts/spark-operator-chart/tests/hook/job_test.yaml
+++ b/charts/spark-operator-chart/tests/hook/job_test.yaml
@@ -228,6 +228,7 @@ tests:
           privileged: false
           allowPrivilegeEscalation: false
           runAsNonRoot: true
+          runAsUser: 65534
           capabilities:
             drop:
               - ALL
--- a/charts/spark-operator-chart/tests/hook/job_test.yaml
+++ b/charts/spark-operator-chart/tests/hook/job_test.yaml
@@ -215,3 +215,69 @@ tests:
     - equal:
         path: spec.template.metadata.labels.key2
         value: value2
+
+- it: Should use the default container securityContext
+  set:
+    hook:
+      upgradeCrd: true
+  asserts:
+    - equal:
+        path: spec.template.spec.containers[0].securityContext
+        value:
+          readOnlyRootFilesystem: true
+          privileged: false
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
+          capabilities:
+            drop:
+              - ALL
+          seccompProfile:
+            type: RuntimeDefault
+
+- it: Should add container securityContext if `hook.securityContext` is set
+  set:
+    hook:
+      upgradeCrd: true
+      securityContext:
+        readOnlyRootFilesystem: true
+        runAsUser: 1000
+        runAsGroup: 2000
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+        runAsNonRoot: true
+        privileged: false
+        seccompProfile:
+          type: RuntimeDefault
+  asserts:
+    - equal:
+        path: spec.template.spec.containers[0].securityContext
+        value:
+          readOnlyRootFilesystem: true
+          runAsUser: 1000
+          runAsGroup: 2000
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          runAsNonRoot: true
+          privileged: false
+          seccompProfile:
+            type: RuntimeDefault
+
+- it: Should add pod securityContext if `hook.podSecurityContext` is set
+  set:
+    hook:
+      upgradeCrd: true
+      podSecurityContext:
+        runAsUser: 1000
+        runAsGroup: 2000
+        fsGroup: 3000
+  asserts:
+    - equal:
+        path: spec.template.spec.securityContext
+        value:
+          runAsUser: 1000
+          runAsGroup: 2000
+          fsGroup: 3000
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -83,6 +83,21 @@ hook:
     # key1: value1
     # key2: value2
 
+  # -- Security context for the Helm hook Job pod.
+  podSecurityContext: {}
+
+  # -- Security context for the Helm hook Job container.
+  securityContext:
+    readOnlyRootFilesystem: true
+    privileged: false
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
+    capabilities:
+      drop:
+      - ALL
+    seccompProfile:
+      type: RuntimeDefault
+
 controller:
   # -- Number of replicas of controller.
   replicas: 1
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -92,6 +92,7 @@ hook:
     privileged: false
     allowPrivilegeEscalation: false
     runAsNonRoot: true
+    runAsUser: 65534
     capabilities:
       drop:
       - ALL
--- a/docker/Dockerfile.kubectl
+++ b/docker/Dockerfile.kubectl
@@ -36,4 +36,6 @@ COPY charts/spark-operator-chart/crds /etc/spark-operator/crds
 
 COPY --from=builder /workspace/kubectl /usr/bin/kubectl
 
+USER 65534:65534
+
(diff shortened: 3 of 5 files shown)


sxivansx (commented):
Pulled the branch. Renders restricted-compliant, unittest 235/235, README regen looks right. `USER 65534:65534` is the correct fix for the Dockerfile.

One problem though. `runAsNonRoot: true` with no `runAsUser` fails against any image built before this PR. I pulled the config blobs from ghcr:

```
kubectl:2.5.2       User=''
kubectl:2.5.0-rc.0  User=''
```

Both UID 0, and no released tag has a USER line. Kubelet rejects that with `container has runAsNonRoot and image will run as root`, so the Job never starts and helm upgrade fails.

Hits anyone pinning `hook.image.tag`, pulling through a mirror, or using their own kubectl image. Chart appVersion is still 2.5.0-rc.0, so the default points at a root image too until a new one ships.

Neither gate catches this. helm unittest only checks rendered YAML, and integration never exercises `hook.upgradeCrd`.

`runAsUser: 65534` in the default securityContext fixes it and still passes restricted.

@danish9039 suggested runAsNonRoot alone above. I think that's backwards: runAsNonRoot alone is what breaks other images, runAsUser is what makes it portable. Unless there's a reason not to pin the UID?

Minor: `--set hook.securityContext=null` now drops all the hardening that used to be hardcoded. Matches how controller.securityContext works so probably intentional, just flagging it.

charts/spark-operator-chart/values.yaml
```diff
@@ -83,6 +83,22 @@ hook:
...
+    privileged: false
+    allowPrivilegeEscalation: false
+    runAsNonRoot: true
+    runAsUser: 65534
+    capabilities:
+      drop:
+      - ALL
+    seccompProfile:
+      type: RuntimeDefault
+
```
tariq-hasan, line 101: I can understand `runAsUser` was set to help enable backward-compatibility but it would end up doing more harm than good - since we are forcing all new installations to run as this user when that may not be the intention to begin with.

To align with controller and webhook let's drop `runAsUser`. For platform admins that do need compatibility with older images they are free to add this option as needed - we can add a note above the key for this?

```suggestion
  # -- Security context for the Helm hook Job pod.
  podSecurityContext: {}

  # -- Security context for the Helm hook Job container.
  securityContext:
    readOnlyRootFilesystem: true
    privileged: false
    allowPrivilegeEscalation: false
    runAsNonRoot: true
    capabilities:
      drop:
      - ALL
    seccompProfile:
      type: RuntimeDefault

```

The general principle here is regarding portability. We'd want to ship a chart that expresses security properties but without exact implementation details. In this case, imposing a concrete user identity would conflict with platform-level UID allocation such as in vendor-based platforms such as OpenShift. From that point of view it makes perfect sense to add `runAsNonRoot: true` and retain a default identity in the image but enable platform admins to override with appropriate UIDs or security policies as necessary.

This would then enable a clean abstraction boundary and separation of responsibilities  between what the chart, image and platform individually accomplish - chart expresses security properties, image supplies the identity, platform overrides.

docker/Dockerfile.kubectl
```diff
@@ -36,4 +36,6 @@ COPY charts/spark-operator-chart/crds /etc/spark-operator/crds
 
 COPY --from=builder /workspace/kubectl /usr/bin/kubectl
 
+USER 65534:65534
```
tariq-hasan, line 39: For consistency should we use 185 as UID and GID? Granted 185 exists in the controller image since that's native to the Spark image itself but I am wondering if we can retain the same identity here for consistency here as well.

https://github.com/kubeflow/spark-operator/blob/761686ee016287e0c2afc856ad15d601de32cdad/Dockerfile#L58-L72

charts/spark-operator-chart/tests/hook/job_test.yaml
```diff
@@ -215,3 +215,70 @@ tests:
...
+        runAsUser: 1000
+        runAsGroup: 2000
+        fsGroup: 3000
+  asserts:
+    - equal:
+        path: spec.template.spec.securityContext
+        value:
+          runAsUser: 1000
+          runAsGroup: 2000
+          fsGroup: 3000
```
tariq-hasan, line 284: Should we also add a negative test for `hook.securityContext: null`?

tariq-hasan (commented):
Hi @KHARSHAVARDHAN-eng thanks so much for working on this and @sxivansx for guiding the review! I have added a few comments.

Also @KHARSHAVARDHAN-eng please follow these instructions to sign your commits: https://github.com/kubeflow/spark-operator/pull/3129/checks?check_run_id=99506904919.

Issue #3130, opened 2026-08-31
Migrate off `gen-crd-api-reference-docs`


tariq-hasan:
### What feature you would like to be added?

`gen-crd-api-reference-docs` [README](https://github.com/ahmetb/gen-crd-api-reference-docs) says the project is "not super actively maintained" and points at [crd-ref-docs](https://github.com/elastic/crd-ref-docs).

Kueue uses [genref](https://github.com/kubernetes-sigs/reference-docs) with a [config file listing each API package](https://github.com/kubernetes-sigs/kueue/blob/main/hack/genref/config.yaml), which handles the multi-version case natively and emits one file per version.

Worth evaluating both.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

PR #3138, opened 2026-09-02
docs: add SparkConnect to the website API reference

### LINKED ISSUE #3137: Fix website navigation/reference
Update `docs/website/reference/api-docs.md` and `docs/website/reference/index.md` to include SparkConnect.

sxivansx:
Fixes #3137

The reference section documented only the two `v1beta2` kinds, so `SparkConnect` (`sparkoperator.k8s.io/v1alpha1`) was missing from it entirely.

- `docs/website/reference/api-docs.md` — retitled to "API Reference" and now lists all three kinds with their API versions, plus a `SparkConnect (v1alpha1)` section giving the group/version, kind, resource name and scope. Carries the same alpha-API warning the user guide uses.
- `docs/website/reference/index.md` — updated the landing-page card title and description to cover all three.

One thing worth flagging: `make build-api-docs` passes `-api-dir` for `api/v1beta2` only, so the generated `docs/api-docs.md` contains no SparkConnect types and there is nothing generated to link to. Rather than change the generator here, the page states that limitation and points at `api/v1alpha1/sparkconnect_types.go` and the Spark Connect user guide. Generating for both API packages looks like #3130 — the genref config that issue mentions is what would list each package.

Verified locally with the same command CI runs, `sphinx-build -b html -W --keep-going`: build succeeds with no warnings, and the relative link resolves to `../user-guide/spark-connect.html`.


docs/website/reference/api-docs.md
```diff
@@ -18,8 +20,32 @@ repository:
 
 - [**`SparkApplication` / `ScheduledSparkApplication` API definition**](https://github.com/kubeflow/spark-operator/blob/master/docs/api-docs.md)
 
+The generated reference covers the `v1beta2` types only. For `SparkConnect`, see the Go
+types in
+[`api/v1alpha1/sparkconnect_types.go`](https://github.com/kubeflow/spark-operator/blob/master/api/v1alpha1/sparkconnect_types.go).
+
```
tariq-hasan: The wording here needs to change as https://github.com/kubeflow/spark-operator/pull/3126 is merge-ready.

docs/website/reference/api-docs.md
```diff
@@ -18,8 +20,32 @@ repository:
...
+| Resource | `sparkconnects` (short name `sparkconn`) |
+| Scope | Namespaced |
+
+:::{warning}
+`SparkConnect` uses the `sparkoperator.k8s.io/v1alpha1` API. An alpha API can change in
+a future release.
+:::
+
+See the [Spark Connect user guide](../user-guide/spark-connect.md) for a worked example.
+
```
tariq-hasan: We should either add for all three CRDs or for none - to avoid inconsistency.

tariq-hasan: We should either add for all three CRDs or for none - to avoid inconsistency.

tariq-hasan (commented):
Hi @sxivansx! I've added a few comments.

Please also update the following.

https://github.com/kubeflow/spark-operator/blob/ff244c12b36cacfb2d6702de26ab765178773461/docs/website/index.rst#L124

docs/website/reference/api-docs.md
sxivansx: Dropped the version claim instead of switching it to v1alpha1. #3126 is not merged yet, so naming either version would be wrong in one of the two merge orders. The link is now just "Spark Operator API definition" and the regeneration note says `api/` rather than `api/v1beta2/`, which holds whichever lands first.

docs/website/reference/api-docs.md
sxivansx: Went with none and removed the table. Once #3126 is in, `docs/api-docs.md` generates group, kind, resource and scope for all three, so repeating it here by hand would just drift.

I kept the one-line alpha warning, since that is a stability caveat rather than reference data and only SparkConnect is alpha. Happy to add tables for all three instead if you would rather have them on this page.

vara-bonthu (approved):
/approve

PR #3140, opened 2026-09-03
fix(webhook): validate ConfigMap mount-path uniqueness

### LINKED ISSUE #3134: Validate mount-path uniqueness for ConfigMaps
Duplicate path values within `spec.{driver,executor}.configMaps` produce a pod the API server rejects. In addition build a single volume per distinct ConfigMap and emit a single volume mount per entry.

pujitha24:
## Purpose of this PR

`spec.{driver,executor}.configMaps` entries are validated for a repeated `name`, but not for a repeated `path`. Two entries with different names and the same mount path pass admission today, and the mutating webhook then builds two `volumeMounts` with an identical `mountPath`, which the API server rejects at pod creation with a generic error that gives no hint the `SparkApplication` spec is the cause. Conversely, mounting the *same* ConfigMap at two different paths — a legitimate configuration — was rejected outright by the duplicate-name check.

**Proposed changes:**

- Replace the duplicate-`name` check in `validateConfigMapList` with a duplicate-`path` check, so the webhook rejects the actual conflict (two mounts at the same path) instead of a configuration that is actually safe (the same ConfigMap mounted twice at different paths).
- Update `addGeneralConfigMaps` in the mutating webhook to build one pod volume per distinct ConfigMap and one `volumeMount` per entry, instead of one volume per entry, so mounting the same ConfigMap at multiple paths no longer produces two pod volumes sharing a name.

## Change Category

- [ ] Documentation update

### Rationale

This closes out a known follow-up: the previously merged ConfigMap-reference validator (#3117) left an explicit comment noting that once the defaulter builds one volume per distinct ConfigMap, mount paths become the thing that needs to stay unique instead of names, and pointed at this issue. This PR makes that change.

## Checklist

### Additional Notes

Validation performed locally:
- `go test ./internal/webhook/...`, including the ginkgo envtest suite (`TestWebhooks`), run with `KUBEBUILDER_ASSETS` pointed at the `make setup-envtest` binaries — passes, including new/updated cases for the duplicate-path rejection, the same-ConfigMap-at-two-paths acceptance, and one-volume/two-volumeMounts pod shape.
- `go test $(go list ./... | grep -v -e /drift -e /test/e2e)` (full repo unit tests) — passes.
- `make go-vet`, `make go-fmt` (no diff produced), `make go-lint` (golangci-lint, 0 issues) — all pass clean.
- `go mod tidy` — no changes.
- Branch is cut from `upstream/master` tip; upstream's own Integration Test workflow is green on recent `master` runs.

No documentation changes were needed beyond the code comments and tests; the CRD schema and field docs for `configMaps[].path` already describe it as the mount path and are unaffected by this change.

Fixes #3134

### DIFF
--- a/internal/webhook/configmap_validator.go
+++ b/internal/webhook/configmap_validator.go
@@ -27,8 +27,8 @@ import (
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
 )
 
-// validateConfigMaps rejects ConfigMap references no ConfigMap could satisfy, and names
-// repeated within one list, which would mount two pod volumes under the same name. Neither
+// validateConfigMaps rejects ConfigMap references no ConfigMap could satisfy, and mount
+// paths repeated within one list, which would mount two volumes at the same path. Neither
 // surfaces before the API server rejects the pod the mutating webhook has already built.
 func validateConfigMaps(spec *v1beta2.SparkApplicationSpec, root *field.Path) error {
 	var errs []error
@@ -52,18 +52,18 @@ func validateConfigMaps(spec *v1beta2.SparkApplicationSpec, root *field.Path) er
 
 func validateConfigMapList(path *field.Path, configMaps []v1beta2.NamePath) []error {
 	var errs []error
-	// A repeated name only collides because the mutating webhook builds one volume per entry.
-	// Once it builds one volume per distinct ConfigMap, mount paths become the thing to keep
-	// unique instead; tracked at https://github.com/kubeflow/spark-operator/issues/3134
+	// The same ConfigMap may be mounted more than once, at different paths, but two entries
+	// mounted at the same path collide: the mutating webhook would emit two volumeMounts with
+	// identical mountPaths, which the API server rejects.
 	seen := make(map[string]bool, len(configMaps))
 	for i, configMap := range configMaps {
 		if err := validateConfigMapName(path.Index(i).Child("name"), configMap.Name); err != nil {
 			errs = append(errs, err)
 		}
-		if seen[configMap.Name] {
-			errs = append(errs, fmt.Errorf("%s has duplicate ConfigMap name %q", path.Index(i).Child("name"), configMap.Name))
+		if seen[configMap.Path] {
+			errs = append(errs, fmt.Errorf("%s has duplicate mount path %q", path.Index(i).Child("path"), configMap.Path))
 		}
-		seen[configMap.Name] = true
+		seen[configMap.Path] = true
 	}
 	return errs
 }
--- a/internal/webhook/configmap_validator.go
+++ b/internal/webhook/configmap_validator.go
@@ -25,11 +25,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/kubeflow/spark-operator/v2/api/v1beta2"
+	"github.com/kubeflow/spark-operator/v2/pkg/common"
+	"github.com/kubeflow/spark-operator/v2/pkg/util"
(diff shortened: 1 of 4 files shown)


internal/webhook/sparkpod_defaulter.go
```diff
@@ -335,10 +335,14 @@ func addGeneralConfigMaps(pod *corev1.Pod, app *v1beta2.SparkApplication) error
...
+	addedVolumes := make(map[string]bool, len(configMaps))
 	for _, namePath := range configMaps {
 		volumeName := getConfigMapVolumeName(namePath.Name)
-		if err := addConfigMapVolume(pod, namePath.Name, volumeName); err != nil {
-			return err
+		if !addedVolumes[volumeName] {
+			if err := addConfigMapVolume(pod, namePath.Name, volumeName); err != nil {
+				return err
+			}
+			addedVolumes[volumeName] = true
```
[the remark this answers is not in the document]
RobuRishabh: Small suggestion on the volume dedup: `addedVolumes` is keyed on `volumeName` (from `getConfigMapVolumeName`), but dedup should be per ConfigMap *source* name, not per generated volume name.

If two different ConfigMap names ever collide to the same volume name, we'd skip creating the second volume but still add its volumeMount, silently mounting the first ConfigMap's data at the second path. 
Could we key on `namePath.Name` instead?

    if !addedVolumes[namePath.Name] {
        ...
        addedVolumes[namePath.Name] = true
    }

That keeps the "same ConfigMap at multiple paths → one volume, N mounts" behavior without the cross-ConfigMap collision risk.

RobuRishabh (commented):
Hi @pujitha24 Thanks for the contribution. Left some minor comment.

RobuRishabh (commented):
Thanks for the quick fix, keying on namePath.Name looks good to me.

For hash-collision test, I don't think we need one for this PR. Forcing a collision would require either mocking `getConfigMapVolumeName` or finding two real names that share the same 8-hex suffix, and the failure mode is already the safe outcome.

For now it looks good to me. 
@nabuskey could you please trigger the CI.

internal/webhook/configmap_validator.go
```diff
@@ -52,18 +52,18 @@ func validateConfigMaps(spec *v1beta2.SparkApplicationSpec, root *field.Path) er
...
 		if err := validateConfigMapName(path.Index(i).Child("name"), configMap.Name); err != nil {
 			errs = append(errs, err)
 		}
-		if seen[configMap.Name] {
-			errs = append(errs, fmt.Errorf("%s has duplicate ConfigMap name %q", path.Index(i).Child("name"), configMap.Name))
+		if seen[configMap.Path] {
+			errs = append(errs, fmt.Errorf("%s has duplicate mount path %q", path.Index(i).Child("path"), configMap.Path))
 		}
-		seen[configMap.Name] = true
+		seen[configMap.Path] = true
```
RobuRishabh, line 109: Empty `path` passes admission but generates an invalid `VolumeMount`. Require a non-empty mount path.

internal/webhook/configmap_validator.go
RobuRishabh: Duplicate checks ignore Spark, Hadoop, and Prometheus reserved mount paths, allowing conflicting mounts that Kubernetes rejects.

RobuRishabh (commented):
Hi @pujitha24, I added few more comments, these are actually pre-existing gaps, but neither is a regression from your PR. It would be nice, if we fix them here, or you can open a follow-up.

vara-bonthu (approved):
/approve

PR #3141, opened 2026-09-04
docs: replace Enterprise Gateway notebooks guide with Kubeflow SDK

vikas-saxena02:
## What this does

Replaces the Jupyter Enterprise Gateway integration guide with the Kubeflow SDK approach.

This follows the discussion on kubeflow/website#4437, where @juliusvonkohout and @thesuperzapper asked for the Enterprise Gateway page to be replaced with the SDK-based approach rather than updated in place. The page has since moved to this repository as part of the docs migration in kubeflow/website#4445, so the change lands here.

The SDK's `SparkClient` creates a `SparkConnect` resource directly, so the gateway deployment, the kernelspec customization, and the extra kernel images are no longer needed.

## Changes

- `user-guide/notebooks-spark-operator.md` — rewritten around `SparkClient().connect()`. Same filename, so the published URL and toctree entry are unchanged.
- `user-guide/notebooks-spark-connect.svg` — new architecture diagram, replacing `notebooks-spark.png`. Keeps the multi-tenant framing of the original (two users, isolated namespaces) and separates the control plane from the data path, since the operator is not involved once the session is established.
- `user-guide/notebooks-spark.png` — removed, it depicted the gateway architecture.
- `user-guide/index.md` — card description updated.

Kept short and high level per the feedback on kubeflow/website#4437.

## Notes for reviewers

**Kubeflow Platform specifics are stated inline rather than linked.** The RBAC, driver ServiceAccount, Istio sidecar exclusion, and executor container name requirements will be documented properly in the Kubeflow Workspaces guides tracked by kubeflow/notebooks#1362. Those pages don't exist yet, so this PR states the requirements directly and I'll replace them with links in a follow-up once they land.

**Question — committed Python bytecode.** While building the docs locally I noticed `docs/website/_ext/kubeflow_topnav/__pycache__/__init__.cpython-312.pyc` is tracked in the repository, and the root `.gitignore` has no `__pycache__/` or `*.py[cod]` entry. It shows up as modified for anyone who runs a docs build, which makes it easy to stage by accident.

The fix is small — `git rm --cached` on the file plus two lines in `.gitignore`. @nabuskey @vjanelle @tariq-hasan  Would you prefer I add it as an extra commit here, or raise it as a separate PR? Leaving it out for now.

## Verification

`make html` and `make linkcheck` both clean. Diagram renders correctly in the built output.

The SDK flow documented here is exercised end to end by the Workspaces Spark Connect test in kubeflow/community-distribution#3590.

/cc @juliusvonkohout @thesuperzapper @christian-heusel


nabuskey (approved):
/approve

PR #3150, opened 2026-09-07
feat: add operator flags to inject default pod labels and annotations

### LINKED ISSUE #3148: Support operator flag to inject labels/annotations on Spark pods

The goal of this issue is to add an operator flag that enables Spark pods to be injected with platform admin-defined labels or annotations.

### Why is this needed?

_No response_

_No response_

_No response_

### Additional context

_No response_

### Love this feature?

Fumer057:
Fixes #3148

### Description
This PR adds support for platform administrators to inject operator-level default labels and annotations onto every Spark driver and executor pod.

**Changes:**
- Adds two new CLI flags to the controller: --default-pod-labels and --default-pod-annotations (both accept JSON strings)
- Adds corresponding Helm chart values under \.Values.spark.defaultPodLabels\ and \.Values.spark.defaultPodAnnotations\
- Merges the global defaults with per-app labels/annotations before submitting Spark applications, ensuring that user-provided labels on the \SparkApplication\ spec correctly override the global defaults.
- Applies the global defaults directly to Spark Connect server pods upon creation.

### Testing Done
- Added comprehensive unit tests in \pkg/util/sparkapplication_test.go\ verifying default label application and user override behaviors.

### DIFF
--- a/charts/spark-operator-chart/templates/controller/deployment.yaml
+++ b/charts/spark-operator-chart/templates/controller/deployment.yaml
@@ -138,6 +138,12 @@ spec:
         {{- with .Values.controller.defaultServiceAccount }}
         - --default-service-account={{ . }}
         {{- end }}
+        {{- with .Values.controller.defaultPodLabels }}
+        - --default-pod-labels={{ . | toJson | quote }}
+        {{- end }}
+        {{- with .Values.controller.defaultPodAnnotations }}
+        - --default-pod-annotations={{ . | toJson | quote }}
+        {{- end }}
         {{- if .Values.controller.driverPodDisruptionBudget.enable }}
         - --enable-driver-pdb=true
         {{- end }}
--- a/charts/spark-operator-chart/values.yaml
+++ b/charts/spark-operator-chart/values.yaml
@@ -139,6 +139,16 @@ controller:
   # in `spark.jobNamespaces`.
   defaultServiceAccount: ""
 
+  # -- Default labels to inject globally into every Spark driver and executor pod.
+  # These act as fallback defaults; labels defined directly in a SparkApplication
+  # spec will override these defaults if there are key conflicts.
+  defaultPodLabels: {}
+
+  # -- Default annotations to inject globally into every Spark driver and executor pod.
+  # These act as fallback defaults; annotations defined directly in a SparkApplication
+  # spec will override these defaults if there are key conflicts.
+  defaultPodAnnotations: {}
+
   # -- Driver PDB feature gate. When true, the controller creates a PodDisruptionBudget for each SparkApplication that sets spec.driverPodDisruptionBudget=true. Default false.
   driverPodDisruptionBudget:
     enable: false
--- a/cmd/operator/controller/start.go
+++ b/cmd/operator/controller/start.go
@@ -117,6 +117,9 @@ var (
 	ingressTLS         []networkingv1.IngressTLS
 	ingressAnnotations map[string]string
 
+	defaultPodLabels      map[string]string
+	defaultPodAnnotations map[string]string
+
 	// Leader election
 	enableLeaderElection        bool
 	leaderElectionLockName      string
@@ -164,6 +167,8 @@ var (
 func NewStartCommand() *cobra.Command {
 	var ingressTLSstring string
 	var ingressAnnotationsString string
+	var defaultPodLabelsString string
+	var defaultPodAnnotationsString string
 	var command = &cobra.Command{
 		Use:   "start",
 		Short: "Start controller and webhook",
@@ -178,6 +183,16 @@ func NewStartCommand() *cobra.Command {
 			if ingressAnnotationsString != "" {
 				if err := json.Unmarshal([]byte(ingressAnnotationsString), &ingressAnnotations); err != nil {
 					return fmt.Errorf("failed parsing ingress-annotations JSON string from CLI: %v", err)
+				}
+			}
+			if defaultPodLabelsString != "" {
+				if err := json.Unmarshal([]byte(defaultPodLabelsString), &defaultPodLabels); err != nil {
+					return fmt.Errorf("failed parsing default-pod-labels JSON string from CLI: %v", err)
+				}
+			}
+			if defaultPodAnnotationsString != "" {
+				if err := json.Unmarshal([]byte(defaultPodAnnotationsString), &defaultPodAnnotations); err != nil {
+					return fmt.Errorf("failed parsing default-pod-annotations JSON string from CLI: %v", err)
 				}
 			}
 
@@ -251,6 +266,9 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().StringVar(&ingressURLFormat, "ingress-url-format", "", "Ingress URL format.")
 	command.Flags().StringVar(&ingressTLSstring, "ingress-tls", "", "JSON format string for the default TLS config on the Spark UI ingresses. e.g. '[{\"hosts\":[\"*.example.com\"],\"secretName\":\"example-secret\"}]'. `ingressTLS` in the SparkApplication spec will override this value.")
 	command.Flags().StringVar(&ingressAnnotationsString, "ingress-annotations", "", "JSON format string for the default ingress annotations for the Spark UI ingresses. e.g. '[{\"cert-manager.io/cluster-issuer\": \"letsencrypt\"}]'. `ingressAnnotations` in the SparkApplication spec will override this value.")
+
+	command.Flags().StringVar(&defaultPodLabelsString, "default-pod-labels", "", "JSON format string for the default labels to be injected into Spark driver and executor pods. e.g. '{\"billing-team\":\"xyz\",\"env\":\"prod\"}'.")
+	command.Flags().StringVar(&defaultPodAnnotationsString, "default-pod-annotations", "", "JSON format string for the default annotations to be injected into Spark driver and executor pods. e.g. '{\"security.example.com/scan\":\"true\"}'.")
 
 	command.Flags().BoolVar(&enableLeaderElection, "leader-election", false, "Enable leader election for controller manager. "+
 		"Enabling this will ensure there is only one active controller manager.")
@@ -582,6 +600,8 @@ func newSparkApplicationReconcilerOptions() sparkapplication.Options {
 		EnableDriverPDB:              enableDriverPDB,
 		DefaultTimeToLiveSeconds:     defaultTimeToLiveSeconds,
 		DefaultServiceAccount:        defaultServiceAccount,
+		DefaultPodLabels:             defaultPodLabels,
+		DefaultPodAnnotations:        defaultPodAnnotations,
 	}
 	if enableBatchScheduler {
 		options.KubeSchedulerNames = kubeSchedulerNames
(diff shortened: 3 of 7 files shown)


charts/spark-operator-chart/templates/controller/deployment.yaml
```diff
@@ -138,6 +138,12 @@ spec:
         {{- with .Values.controller.defaultServiceAccount }}
         - --default-service-account={{ . }}
         {{- end }}
+        {{- with .Values.controller.defaultPodLabels }}
+        - --default-pod-labels={{ . | toJson | quote }}
+        {{- end }}
+        {{- with .Values.controller.defaultPodAnnotations }}
+        - --default-pod-annotations={{ . | toJson | quote }}
+        {{- end }}
```
tariq-hasan, line 146: Please remove the quotes.

```suggestion
        {{- with .Values.controller.defaultPodLabels }}
        - --default-pod-labels={{ . | toJson }}
        {{- end }}
        {{- with .Values.controller.defaultPodAnnotations }}
        - --default-pod-annotations={{ . | toJson }}
        {{- end }}
```

tariq-hasan, line 146: Please also add Helm unit tests for these new flags.

internal/controller/sparkconnect/reconciler.go
```diff
@@ -364,6 +370,29 @@ func (r *Reconciler) mutateServerPod(ctx context.Context, conn *v1alpha1.SparkCo
...
+			if pod.Annotations == nil {
+				pod.Annotations = make(map[string]string)
+			}
+			for k, v := range r.options.DefaultPodAnnotations {
+				if _, exists := pod.Annotations[k]; !exists {
+					pod.Annotations[k] = v
+				}
+			}
+		}
+
```
tariq-hasan, line 395: In addition to these server pod changes please also wire up the SparkConnect executor pods to receive those same labels and annotations.

charts/spark-operator-chart/values.yaml
```diff
@@ -139,6 +139,16 @@ controller:
...
+  # -- Default labels to inject globally into every Spark driver and executor pod.
+  # These act as fallback defaults; labels defined directly in a SparkApplication
+  # spec will override these defaults if there are key conflicts.
+  defaultPodLabels: {}
+
+  # -- Default annotations to inject globally into every Spark driver and executor pod.
+  # These act as fallback defaults; annotations defined directly in a SparkApplication
+  # spec will override these defaults if there are key conflicts.
+  defaultPodAnnotations: {}
+
```
tariq-hasan, line 151: Given that you've updated `values.yaml` please run `make helm-docs` to update `charts/spark-operator-chart/README.md`.

cmd/operator/controller/start.go
```diff
@@ -180,6 +185,16 @@ func NewStartCommand() *cobra.Command {
...
+			if defaultPodLabelsString != "" {
+				if err := json.Unmarshal([]byte(defaultPodLabelsString), &defaultPodLabels); err != nil {
+					return fmt.Errorf("failed parsing default-pod-labels JSON string from CLI: %v", err)
+				}
+			}
+			if defaultPodAnnotationsString != "" {
+				if err := json.Unmarshal([]byte(defaultPodAnnotationsString), &defaultPodAnnotations); err != nil {
+					return fmt.Errorf("failed parsing default-pod-annotations JSON string from CLI: %v", err)
+				}
+			}
```
tariq-hasan: Other than these please also validate that

- the JSON is syntactically valid e.g. `{foo:bar}` is not valid
- the values are semantically valid e.g. `{"foo":false}` is not valid

internal/controller/sparkapplication/controller.go
```diff
@@ -1016,6 +1022,12 @@ func (r *Reconciler) submitSparkApplication(ctx context.Context, app *v1beta2.Sp
 		logger.Info("Applied default driver service account", "serviceAccount", r.options.DefaultServiceAccount)
 	}
 
+	// Apply default pod labels and annotations globally
+	if len(r.options.DefaultPodLabels) > 0 || len(r.options.DefaultPodAnnotations) > 0 {
+		submitApp = util.ApplyDefaultPodLabelsAndAnnotations(submitApp, r.options.DefaultPodLabels, r.options.DefaultPodAnnotations)
+		logger.Info("Applied default pod labels and annotations")
+	}
+
```
tariq-hasan: Please add precedence logic to ensure that CR field wins over operator flag.

Please also note that operator-reserved labels/annotations should not be overwritten by admin-controlled operator flags. Please add this constraint to the implementation as well.

internal/controller/sparkconnect/reconciler.go
tariq-hasan: The SparkApplication comment regarding precedence logic and operator-reserved labels/annotations applies here as well.

Please add precedence logic to ensure that CR field wins over operator flag. Also please note that operator-reserved labels/annotations should not be overwritten by admin-controlled operator flags. Please add this constraint to the implementation as well.

cmd/operator/controller/start.go
tariq-hasan, line 216: Please also add validation for K8s labels and annotations for allowed characters, maximum lengths, qualified names, prefixes, DNS-style syntax, etc.

You can use validation helpers from k8s libraries to complete this gap.

tariq-hasan (changes requested):
Hi @Fumer057! Thanks for raising the PR. I have added suggestions for changes.

Also please note that operator-reserved labels/annotations should not be overwritten by admin-controlled operator flags. Please add this constraint to the implementation as well.

And also please add user documentation for the feature in `docs/website/user-guide/writing-sparkapplication.md` and `docs/website/user-guide/spark-connect.md`.

Fumer057:
��@tariq-hasan I have implemented this. SparkConnect executor pods are now properly wired up to receive the default labels and annotations.

charts/spark-operator-chart/values.yaml
tariq-hasan: The defaults apply for all CRDs.

```suggestion
  # -- Default labels to inject globally into every Spark driver and executor pod.
  # These act as fallback defaults; labels defined directly in a CRD
  # spec will override these defaults if there are key conflicts.
  defaultPodLabels: {}

  # -- Default annotations to inject globally into every Spark driver and executor pod.
  # These act as fallback defaults; annotations defined directly in a CRD
  # spec will override these defaults if there are key conflicts.
  defaultPodAnnotations: {}

```

cmd/operator/controller/start.go
```diff
@@ -117,6 +117,9 @@ var (
 	ingressTLS         []networkingv1.IngressTLS
 	ingressAnnotations map[string]string
 
+	defaultPodLabels      map[string]string
+	defaultPodAnnotations map[string]string
+
```
tariq-hasan: Please keep the ordering intact.

```suggestion
	// Fallback service account for Spark driver pods when the custom resource does
	// not specify one. Empty by default, which preserves existing behavior.
	defaultServiceAccount string

	// Default pod labels and annotations
	defaultPodLabels      map[string]string
	defaultPodAnnotations map[string]string

	// Spark web UI service and ingress
	enableUIService    bool
	ingressClassName   string
	ingressURLFormat   string
	ingressTLS         []networkingv1.IngressTLS
	ingressAnnotations map[string]string

```

cmd/operator/controller/start.go
```diff
@@ -252,6 +286,9 @@ func NewStartCommand() *cobra.Command {
 	command.Flags().StringVar(&ingressTLSstring, "ingress-tls", "", "JSON format string for the default TLS config on the Spark UI ingresses. e.g. '[{\"hosts\":[\"*.example.com\"],\"secretName\":\"example-secret\"}]'. `ingressTLS` in the SparkApplication spec will override this value.")
 	command.Flags().StringVar(&ingressAnnotationsString, "ingress-annotations", "", "JSON format string for the default ingress annotations for the Spark UI ingresses. e.g. '[{\"cert-manager.io/cluster-issuer\": \"letsencrypt\"}]'. `ingressAnnotations` in the SparkApplication spec will override this value.")
 
+	command.Flags().StringVar(&defaultPodLabelsString, "default-pod-labels", "", "JSON format string for the default labels to be injected into Spark driver and executor pods. e.g. '{\"billing-team\":\"xyz\",\"env\":\"prod\"}'.")
+	command.Flags().StringVar(&defaultPodAnnotationsString, "default-pod-annotations", "", "JSON format string for the default annotations to be injected into Spark driver and executor pods. e.g. '{\"security.example.com/scan\":\"true\"}'.")
+
```
tariq-hasan: Please move these flags under the one for default service account to keep the ordering intact.

internal/controller/sparkconnect/reconciler.go
```diff
@@ -390,6 +419,30 @@ func (r *Reconciler) mutateServerPod(ctx context.Context, conn *v1alpha1.SparkCo
...
+		if len(defaultArgs) > 0 {
+			// Insert defaults right after the initial start-connect-server.sh command (args[0])
+			// so that resource-specific conf options appear later and override.
+			combined := make([]string, 0, len(args)+len(defaultArgs))
+			combined = append(combined, args[0])
+			combined = append(combined, defaultArgs...)
+			combined = append(combined, args[1:]...)
+			args = combined
+		}
+
```
tariq-hasan: Please remove the sparkconf settings for driver since this is client mode and sparkconf settings for the driver are silent no-ops on the Spark runtime.

For the executor side please also remove the sparkconf settings and add the default labels and annotations directly on the executor pod template (making sure not to override labels/annotations on the pod template). This is because the executor pod template is the ultimate source of truth on the Spark runtime.

internal/controller/sparkapplication/controller.go
```diff
@@ -1016,6 +1022,12 @@ func (r *Reconciler) submitSparkApplication(ctx context.Context, app *v1beta2.Sp
 		logger.Info("Applied default driver service account", "serviceAccount", r.options.DefaultServiceAccount)
 	}
 
+	// Apply default pod labels and annotations globally
+	if len(r.options.DefaultPodLabels) > 0 || len(r.options.DefaultPodAnnotations) > 0 {
+		submitApp = util.ApplyDefaultPodLabelsAndAnnotations(submitApp, r.options.DefaultPodLabels, r.options.DefaultPodAnnotations)
+		logger.Info("Applied default pod labels and annotations")
```
tariq-hasan: Please also log the labels and annotations within `logger.Info("Applied default pod labels and annotations")` similar to how logging for service account is done just above it.

cmd/operator/controller/start.go
```diff
@@ -180,6 +185,35 @@ func NewStartCommand() *cobra.Command {
...
 				}
 			}
+			if defaultPodLabelsString != "" {
+				if err := json.Unmarshal([]byte(defaultPodLabelsString), &defaultPodLabels); err != nil {
+					return fmt.Errorf("failed parsing default-pod-labels JSON string from CLI: %v", err)
+				}
+				for key, value := range defaultPodLabels {
+					if strings.HasPrefix(key, common.LabelAnnotationPrefix) {
+						return fmt.Errorf("invalid key %q in --default-pod-labels: keys with prefix %q are reserved by the operator", key, common.LabelAnnotationPrefix)
+					}
```
tariq-hasan: Please also ensure that the following are not being overriden - spark-role, spark-app-selector, spark-app-name, spark-version, spark-exec-id, and spark-exec-resourceprofile-id. These are labels set by Spark on the pods and they cannot be overridden so it is better to error here than error during submission.

Most of these are defined in `pkg/common/spark.go`. For the ones that do not exist in `pkg/common/spark.go` please define them there.

Issue #2961, opened 2026-05-28
 Executor pods are not cascade-deleted when driver pod is deleted


hahahaheihei:
### What happened?

When a SparkApplication is suspended (spec.suspend=true) or otherwise has its driver pod deleted, the executor pods are not cleaned up by Kubernetes garbage collection — even though the driver pod carries controller: true in the executor's ownerReferences. This is because Spark Operator additionally attaches the SparkApplication CR as a second owner of each executor pod, which prevents K8s GC from cascading the deletion when only the driver disappears.

### Reproduction Code

Apply a SparkApplication and let it reach the running state:
```
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: spark-test
spec:
  type: Python
  mode: cluster
  image: docker.io/apache/spark:3.5.3-python3
  mainApplicationFile: local:///opt/spark/examples/src/main/python/pi.py
  sparkVersion: "3.5.3"
  driver:
    cores: 1
    coreRequest: "1"
    memory: 512m
  executor:
    cores: 1
    coreRequest: "1"
    memory: 512m
    instances: 2
  restartPolicy:
    type: Never
```
Inspect an executor pod's owner references:
```console
$ kubectl get pod spark-test-<hash>-exec-1 \
    -o jsonpath='{range .metadata.ownerReferences[*]}{.kind}/{.name}  controller={.controller}{"\n"}{end}'
SparkApplication/spark-test    controller=
Pod/spark-test-driver          controller=true
```
The executor has **two** owners: the SparkApplication CR (controller=false) and the driver pod (controller=true).

Now delete the driver pod and observe that executors are NOT garbage collected:

```console
$ kubectl delete pod spark-test-driver
pod "spark-test-driver" deleted

$ sleep 40 && kubectl get pod
NAME                            READY   STATUS      RESTARTS   AGE
spark-test-<hash>-exec-1        0/1     Completed   0          7m
spark-test

### Expected behavior

When the driver pod is deleted, the executor pod should be deleted in cascade.

### Actual behavior

executor pod still exist

### Environment & Versions

- Kubernetes: v1.33.x
- Spark Operator: v2.5.0
- Apache Spark: 3.5.3
- SparkApplication suspended via Kueue (or by manually setting spec.suspend=true / deleting driver)

### Additional context

_No response_

### Impacted by this bug?

Give it a 👍 We prioritize the issues with most 👍

hahahaheihei:
[...]

PR #3152, opened 2026-09-08
fix: stop adding SparkApplication owner reference to executor pods

KR-Ravindra:
## Purpose of this PR

**Proposed changes:**

- Stop adding the SparkApplication as a second (non-controller) ownerReference on executor pod templates at submission time, so that executor pods are garbage-collected as soon as their controller owner (the driver pod) is deleted.
- Update the expectations in `TestExecutorPodTemplateContents` and `TestSubmitPodTemplates` so executor templates carry only user-supplied owner references; driver templates keep the controller reference to the SparkApplication (`TestDriverPodTemplateContents` is unchanged).

## Problem

Since #2670 every executor pod has two owners: the driver pod (`controller: true`, set by Spark itself) and the SparkApplication (`controller` unset, added by the operator). Kubernetes garbage collection only deletes a dependent once *all* of its owners are gone, so deleting the driver pod (manually, or via `spec.suspend=true`, e.g. when Kueue preempts the workload) leaves the executor pods behind until the SparkApplication itself is deleted. Across Kueue requeue cycles this accumulates orphaned executor pods.

## Root cause

`internal/controller/sparkapplication/submission.go:1199-1209` (`buildExecutorPodTemplate`) appends `util.GetOwnerReference(app)` with `Controller`/`BlockOwnerDeletion` cleared to the executor pod template. That extra owner is what blocks the cascade deletion from the driver pod. The webhook path (`internal/webhook/sparkpod_defaulter.go` `addOwnerReference`) already returns early for non-driver pods, so submission is the only place that adds it.

## Fix

Remove the executor ownerReference block in `buildExecutorPodTemplate`. Executor pods keep their controller reference to the driver pod, and the driver pod keeps its controller reference to the SparkApplication, so deleting the SparkApplication still cascades to the driver and then to the executors, while deleting only the driver now cascades to the executors as well. This follows the approach suggested in the issue discussion (drop the executor's SparkApplication reference and rely on GC cascade deletion instead of adding operator-managed deletion of executor pods).

The original motivation for the extra reference was to let other controllers (e.g. Kueue) recognize executor pods as children of the SparkApplication. Kueue's ownership walk (`FindAncestorJobManagedByKueue`) and its `IsOwnerManagedByKueueForObject` check both use `metav1.GetControllerOf`, i.e. only controller references, and its SparkApplication integration identifies executor pods by the `spark-role` label and reads readiness from `status.executorState`, so the non-controller reference is not consumed there.

## How tested

Unit tests (no cluster required) plus an e2e assertion in the existing suspend/resume test.

Before the fix, with the updated expectations, three subtests fail. Two are shown below; the third, `TestSubmitPodTemplates/sends_default_templates_with_owner_ref_when_not_specified` (`internal/controller/sparkapplication/rest_submission_test.go:203`), fails the same way because the submitted executor template still carries the SparkApplication owner reference (`Empty` assertion on `OwnerReferences`) and is only absent from the output because of the `-run` filter. The "pod template already has owner references" case is unchanged and passes both before and after, because user-supplied owner references are preserved as-is.

```
$ go test ./internal/controller/sparkapplication/ -run 'TestExecutorPodTemplateContents|TestDriverPodTemplateContents' -count=1
--- FAIL: TestExecutorPodTemplateContents (0.00s)
    --- FAIL: TestExecutorPodTemplateContents/with_pod_template (0.00s)
        submission_test.go:645:
            	Diff:
            	--- Expected
            	+++ Actual
            	@@ -6,2 +6,6 @@
            	     uid: ""
            	+  - apiVersion: sparkoperator.k8s.io/v1beta2
            	+    kind: SparkApplication
            	+    name: test
            	+    uid: 2d2d2c08-fcc9-4e41-9d93-13540605f24e
            	 spec:
    --- FAIL: TestExecutorPodTemplateContents/no_pod_template (0.00s)
        submission_test.go:645:
            	Diff:
            	--- Expected
            	+++ Actual
            	@@ -1,2 +1,7 @@
            	-metadata: {}
            	+metadata:
            	+  ownerReferences:
            	+  - apiVersion: sparkoperator.k8s.io/v1beta2
            	+    kind: SparkApplication
            	+    name: test
            	+    uid: 2d2d2c08-fcc9-4e41-9d93-13540605f24e
            	 spec:
FAIL
FAIL	github.com/kubeflow/spark-operator/v2/internal/controller/sparkapplication	0.031s
```

After the fix:

```
$ go test ./internal/controller/sparkapplication/ -run 'TestExecutorPodTemplateContents|TestDriverPodTemplateContents' -count=1 -v
--- PASS: TestDriverPodTemplateContents (0.00s)
--- PASS: TestExecutorPodTemplateContents (0.00s)
ok  	github.com/kubeflow/spark-operator/v2/internal/controller/sparkapplication	0.030s

$ KUBEBUILDER_ASSETS=... go test ./internal/controller/sparkapplication/... ./internal/webhook/... -count=1
ok  	github.com/kubeflow/spark-operator/v2/internal/controller/sparkapplication
ok  	github.com/kubeflow/spark-operator/v2/internal/webhook
```

`gofmt -l`, `go vet`, `go build ./...` and `golangci-lint run ./internal/controller/sparkapplication/...` are clean.

**e2e** (added on request of the reviewers): the `spark-pi: Suspend and Resume` case in `test/e2e/sparkapplication_test.go` now waits for an executor pod to exist before suspending and, once the application reaches `Suspended`, asserts that no pod labelled `sparkoperator.k8s.io/app-name=spark-pi` remains (driver or executor). Without this fix the executor pod survives the driver deletion because the SparkApplication is still an owner, so the assertion fails; with the fix the executors are cascade-deleted with the driver. `gofmt -l` and `go vet ./...` in `test/e2e` are clean; the CI e2e matrix runs it.

## Links

- Issue: https://github.com/kubeflow/spark-operator/issues/2961
- Change that introduced the executor owner reference: https://github.com/kubeflow/spark-operator/pull/2670
- Related Kueue report: https://github.com/kubernetes-sigs/kueue/issues/11675

## Change Category

- [ ] Documentation update

### Rationale

Covered above. Executor pods created by an operator built before this change keep their existing owner references; only newly submitted applications are affected.

## Checklist

### Additional Notes

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


RobuRishabh (commented):
Hi @KR-Ravindra thanks for the contribution, the PR looks good to me
@tariq-hasan need a suggestion here unit tests are enough for merge but how about add a e2e test here that submits an app, deletes the driver, and asserts executors disappear would lock in the fix or do you think its over-engineering.

and a quick question if anyone has answer to it, Is there any other tool (not Kueue, not spark-operator) that depends on executors having a direct SparkApplication owner ref?

internal/controller/sparkapplication/submission.go
```diff
@@ -1196,17 +1196,11 @@ func buildExecutorPodTemplate(app *v1beta2.SparkApplication) *corev1.PodTemplate
...
-	// are the children of the SparkApplication
-	ownerReference := util.GetOwnerReference(app)
-	ownerReference.Controller = nil
-	ownerReference.BlockOwnerDeletion = nil
-	if !slices.ContainsFunc(template.OwnerReferences, func(r metav1.OwnerReference) bool {
-		return reflect.DeepEqual(r, ownerReference)
-	}) {
-		template.OwnerReferences = append(template.OwnerReferences, ownerReference)
-	}
+	// Executor pods are controlled by the driver pod (Spark sets the controller
```
vjanelle: I feel like this was done for a reason, potentially because of executor tracking?

vjanelle:
> 

Reasonably sure that the e2e tests could add support for the situation.

vara-bonthu (approved):
/approve

PR #3154, opened 2026-09-08
docs(kep): propose workload-aware scheduling for SparkApplication

RobuRishabh:
Part of [#2962](https://github.com/kubeflow/spark-operator/issues/2962).

Design-only PR for Kubernetes-native Workload-Aware Scheduling (WAS) in Spark Operator. It adds [KEP-2962](proposals/2962-workload-aware-scheduling/README.md) proposing a top-level `SparkApplication.spec.scheduling` API aligned with [KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis), [kubernetes/enhancements#6342](https://github.com/kubernetes/enhancements/pull/6342), and the Kubeflow [Trainer KEP](https://github.com/kubeflow/trainer/pull/3219).

## Summary

- **SparkApplication is the root controller** and sole compiler of native `Workload` objects via upstream `workloadbuilder`.
- **Opt-in API:** omitted `.spec.scheduling` preserves current behavior; `SparkApplicationWorkloadAwareScheduling` defaults to disabled.
- **Driver schedules independently;** executor cohort is one leaf `PodGroup`, avoiding the cluster-mode bootstrap deadlock.
- **Gang `minCount` is controller-computed** from Spark's resolved initial executor count; users express scale through `executor.instances` / `sparkConf`.
- **One stable `Workload` per SparkApplication;** each `status.submissionID` gets an isolated attempt-scoped `PodGroup`.
- **Native WAS is mutually exclusive** with Volcano, YuniKorn, and scheduler-plugins per application.
- **Alpha scope is static executors only:** dynamic allocation with native WAS is rejected in this KEP and deferred to a follow-up KEP (topology, DRA, client modes, and Kueue are also out of scope).
- **Phase 0 prerequisite:** pin Kubernetes v1.37+ and the newest KEP-6089 served API stack at implementation time (prefer `scheduling.k8s.io/v1` building blocks from #6342 over v1.36 prototype wire types).

**Alpha vs upstream beta:** "Alpha" here is the Spark Operator feature stage (`kep.yaml` `stage: alpha`, gate off by default, narrow MVP). It does not mean KEP-6089 stays alpha — reusable building blocks are graduating to beta via #6342 for Kubernetes v1.38.

The KEP is restructured to match the Trainer layout and includes API examples, validation tables, lifecycle semantics, RBAC, graduation criteria, alternatives, and seven Mermaid diagrams (architecture, compile path, sequence, ownership, bootstrap deadlock, and lifecycle).

Thanks to @Rajadi16 for prototype work in #3093, #3113, and #3116. Those PRs are credited as implementation evidence; this PR asks the community to agree on the durable API and lifecycle before implementation continues.

## Review focus

Highest-value feedback:

1. Open decisions in the KEP — especially empty `scheduling: {}` defaulting (Basic vs Gang), CRD surface area, and behavior when a deployment-level default batch scheduler is also configured.
2. Alpha scope boundary — static gang only in MVP; dynamic allocation explicitly deferred.
3. Kubernetes version and Phase 0 API pinning (v1.37+ provisional minimum).
4. Workload / PodGroup lifecycle — retry isolation, suspend/resume, and immutable `.spec.scheduling` in alpha.

## Test plan

- [x] Design-only change; no runtime code.
- [ ] Reviewers: read `proposals/2962-workload-aware-scheduling/README.md` and `kep.yaml`.
- [ ] Confirm alignment with KEP-6089 root-controller model and upstream `workloadbuilder`.
- [ ] Confirm alpha scope and deferred topics match expectations for [#2962](https://github.com/kubeflow/spark-operator/issues/2962).

## References

- Kubernetes KEP-6089: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis
- KEP-6089 beta promotion (#6342): https://github.com/kubernetes/enhancements/pull/6342
- Kubeflow Trainer KEP: https://github.com/kubeflow/trainer/pull/3219


sxivansx:
One thing worth checking before Phase 0. I looked at the runtime shape in `k8s.io/api v0.35.4`, which this repo already pins in `go.mod`. In `scheduling.k8s.io/v1alpha1`, `PodGroup` isn't a standalone object — it's an inline struct in `WorkloadSpec.PodGroups`, and that field is immutable. Only `Workload`/`WorkloadList` are registered, and `Workload` has no `Status`. Pods join via `pod.spec.workloadRef{name, podGroup, podGroupReplicaKey}`, not `spec.schedulingGroup`.

If that shape carries into the version you pin, the per-`submissionID` PodGroup object has nothing to create or GC, and updating `minCount` in place isn't possible either.

`WorkloadReference.PodGroupReplicaKey` looks like it covers the first case — its doc says policy is applied separately per replica. Setting it to `status.submissionID` would give retry isolation with one long-lived Workload, which is the invariant you already want.

Smaller: the RBAC block grants `podgroups` and `podgroups/status`, neither of which exists in that shape.

RobuRishabh:
Hi @sxivansx thanks for the review. Agreed on Phase 0 must pin the exact served API before implementation.
The shape you inspected (`scheduling.k8s.io/v1alpha1` in `k8s.io/api` v0.35.x) is the superseded prototype: inline `Workload.spec.podGroups`, Pods via `spec.workloadRef` + `podGroupReplicaKey`, no standalone `PodGroup` CRD.

This KEP targets Kubernetes v1.37+ and `scheduling.k8s.io/v1beta1`, where:
- Workload / PodGroup are standalone resources (PodGroup has status)
- Pods join via spec.schedulingGroup, not workloadRef
- Gang minCount is mutable on the runtime PodGroup
- RBAC for podgroups / podgroups/status applies

Phase 0 will bump k8s.io/api / component-helpers to the v1.37+ stack and compile workloadbuilder against those types, IMO we should not implement against the v1alpha1 wire shape in today’s go.mod.
The per-submissionID PodGroup + schedulingGroup model follows KEP-6089’s centralized management pattern (also used by Job and Trainer).

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+Spark Operator already knows information users should not repeat: driver versus executor role, resolved initial executor count, submission attempt identity, executor template content, and suspend/resume/retry lifecycle. [KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis) calls the highest-level controller with this complete view the **root controller** and expects it to compile native `Workload` objects. For `SparkApplication`, that is the SparkApplication controller.
+
+Earlier prototype PRs ([#3093](https://github.com/kubeflow/spark-operator/pull/3093), [#3113](https://github.com/kubeflow/spark-operator/pull/3113), [#3116](https://github.com/kubeflow/spark-operator/pull/3116)) proved the operator can create scheduling resources and inject executor membership, but they also exposed prototype API choices such as `minMember` through legacy batch scheduler options and may have targeted superseded WAS wire shapes. Agreeing on the public API and lifecycle before implementation avoids encoding those prototype shapes in the Spark CRD.
```
tariq-hasan: I suppose native WAS would be unavailable to most Spark Operator users for the foreseeable future - since `GenericWorkload` is beta-but-disabled-by-default in v1.37 and managed Kubernetes providers generally do not expose non-default gates.

So I guess the motivation is really just to build now for the platform teams who run their own control planes and to keep Volcano/YuniKorn first-class indefinitely.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+1. Add an optional `SparkApplicationSpec.Scheduling`, serialized as `.spec.scheduling`.
+2. Compose [KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis) leaf-level building blocks in a Spark-owned wrapper.
+3. Make `SparkApplication` the root workload and the SparkApplication controller the sole compiler
+  and owner of its native `Workload`.
+4. Use upstream `workloadbuilder` from
+  `k8s.io/component-helpers/scheduling/schedulingv1/workloadbuilder` for translation, defaulting,
+   and supported-policy validation.
+5. Support cluster-mode executor Pods in one Basic or Gang `PodGroup` with the driver scheduling
+  independently.
+6. Compute Gang `minCount` from the same resolved initial-executor value used for Spark submission.
```
tariq-hasan, line 88: I suppose this may be the correct choice since Spark has multiple existing scale configuration surfaces - coming from both the CRD as well as spark config. So `minCount` would add a third source of truth and complicate things.

At the same time though this is stricter than other adopting controllers. k8s Job, for example, allows an explicit `minCount`. JobSet also allows it - freezing it as a floor that can never be lowered.

So perhaps we could have this restriction more so as a validation in the alpha stage  but not as a permanent API promise. Perhaps a future beta could relax it to a floor semantics so that explicit user `minCount` becomes a minimum floor and alpha’s prohibition does not become permanent.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+      minCount: 4
+```
+
+The ephemeral executor template sent to Spark contains:
+
+```yaml
+spec:
+  schedulingGroup:
+    podGroupName: spark-pi-executors-<attempt-hash>
+```
```
tariq-hasan, line 226: Other than the `Static executor gang for Spark Pi` story we might also want to have user stories that cover failure modes.

One of them would be the case when the gang never admits. The idea behind this would be that the driver is already in a `Running` state but the executor gang remains `Unschedulable` because of capacity constraints resulting from the `minCount` restriction. This would in turn lead to a consideration of how we handle status, conditions, events, metrics and possibly timeout for this kind of failure mode.

A second case may be for retry identity. If one submission fails and a second submission is running with pending executor pods the question then may be as to which PodGroup belongs to this submission attempt. That would naturally lead to a consideration of `submissionID` metadata, status identity and a separation of old vs new group.

A third case may be for suspension and resumption semantics - for example, when Kueue admits the quota but WAS is not able to do placement due to topology or capacity constraints. This would force a consideration of coexistence and timeout semantics and an understanding of what the user may see and who releases the quota.

I am thinking these and other stories may stress the KEP a bit more so we are able to expose missing design considerations.

tariq-hasan: I am thinking the ephemeral executor template may need to be fleshed out in more detail - in the Design Details section at least.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+```mermaid
+flowchart LR
+    User[User] --> App[SparkApplication]
+    App --> Controller[SparkApplication controller]
+    Controller -->|spark-submit| Driver[Driver Pod]
+    Controller -->|executor Pod template| Driver
+    Driver -->|creates| Executors[Executor Pods]
+    Scheduler[kube-scheduler] --> Driver
+    Scheduler --> Executors
+```
```
tariq-hasan, line 259: The fact that the operator submits a driver and passes it an executor template and it's the driver pod that then creates executor pods may very well be one of the major factors that distinguish the spark operator from the other adopting controllers - just by virtue of the fact that the spark operator, unlike a Job-like controller, does not create or own the worker (executor) pods directly.

This fact has several implications for the spark operator
- the operator cannot simply set `schedulingGroup` at pod creation time
- the membership must travel indirectly through template/REST/webhook
- the operator must verify a field it did not itself send on the final pod create request

Therefore the injection mechanism as it relates to the executor template needs to be fleshed out in more depth in the KEP - possibly under a separate heading in the Design Details section.

tariq-hasan, line 259: To speak to the details of this transport chain from executor template mutation to an actual executor pod there are several steps involved
- operator writes into the executor template file
- upstream Spark materializes that executor template for the driver (after the operator or the submission service invokes `spark-submit`)
- the Spark driver uses a Fabric8 K8s client to create the executor pods

The third step is what carries a real integration risk - if the Spark version uses a Fabric8 K8s client whose model predates `schedulingGroup`. In that case we'd probably need to test if the Spark template load/build/serialize pipeline still preserves this unknown field. So essentially we'd need to resolve a possible correctness question in terms of whether Fabric8 preserves unknown Pod fields.

tariq-hasan, line 259: To add to all of this we'd want to ensure strong defense-in-depth by layering in the template injection mechanism with two additional steps.

The first is to use the webhook as a fallback injection mechanism if the template injection fails.

The second is to have the controller run a post-hoc membership check by inspecting actual executor Pods to verify that `pod.spec.schedulingGroup.podGroupName` is the current attempt `PodGroup` name and emit a condition along the lines of `WorkloadSchedulingReady=False
Reason=ExecutorMembershipMissing` otherwise.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+not apply to the superseded v1alpha1 inline model.
+
+### Open Questions
+
+1. Should `.spec.scheduling: {}` default to `Basic` or `Gang`?
+2. Should the first CRD expose only `schedulingPolicy`, or the full wrapper with allow-list rejection?
+3. Is v1.37+ an acceptable minimum cluster version for Phase 0 pinning? (Provisional yes; confirm
+   after Phase 0 module bump and API discovery.)
+4. Should native WAS be rejected when a deployment-level default batch scheduler would also apply?
+5. Are Events and errors enough for alpha, or should a `WorkloadSchedulingReady` condition be added?
```
tariq-hasan: The issue here is that gang scheduling fundamentally changes the shape of how failure is reported.

With a normal scheduler a Spark user can discover capacity problems from Pod events such as `Pending` or `FailedScheduling`. But with gang scheduling the group members would be held earlier in scheduler queueing before normal per-Pod scheduling or binding so a user may see a running driver and a running SparkApplication and would wonder why the executors are not making any progress.

I am thinking therefore we need to expose the `PodGroup.status.conditions` scheduler status through the operator with something along the lines of `status.workloadScheduling` that may possibly have `workloadName`, `podGroupName`, `submissionID`, resolved policy, `minCount`, mirrored `PodGroup` scheduling state among other things - essentially exposing workload scheduling readiness in addition to what we currently have for application lifecycle. This would enable the users to get observability on the scheduling and also give the controller a direct mapping from the current submission ID to the current pod group name without having to resort to recompute hashes.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+| Empty `scheduling: {}`                       | Accept; default Basic proposed         | Safe explicit opt-in.           |
+| `schedulingPolicy.gang` without `minCount`   | Accept; controller computes `minCount` | Spark scale is source of truth. |
+| User sets `gang.minCount`                    | Reject                                 | Duplicate scale configuration.  |
+| Both Basic and Gang set                      | Reject                                 | Upstream policy is a union.     |
+| Native WAS + dynamic allocation              | Reject                                 | Requires separate KEP.          |
+| `scheduling` + `batchScheduler`              | Reject                                 | No native translation defined.  |
+| User sets `spec.schedulingGroup`             | Reject                                 | Operator-managed per attempt.   |
+| Unsupported topology, disruption, or claims  | Reject by allow-list                   | Fail closed.                    |
+| Feature gate disabled + `scheduling` set     | Reject admission                       | Avoid unusable contract.        |
```
tariq-hasan: Is the intention that the webhook be used here to reject admission?

I say this because the webhook runs in another process from the controller and the feature gate state is not propagated from the controller to the webhook.

I am thinking in this case we can either plumb feature-gate state into webhook or do the rejection during reconciliation.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+| Required Kubernetes APIs not served          | Reject at reconcile preflight          | Cached API discovery.           |
+| Client or in-cluster-client mode             | Reject until supported                 | Unverified injection path.      |
+| `ScheduledSparkApplication` template invalid | Reject the schedule                    | Fail before child creation.     |
```
tariq-hasan, line 412: Should we also have validation for non-default scheduler names such as `spark.kubernetes.scheduler.name`, `spark.kubernetes.driver.scheduler.name` and `spark.kubernetes.executor.scheduler.name` passed in as Spark config that might conflict with native WAS?

Same for `spark.kubernetes.allocation.maxPendingPods < minCount` to avoid a liveness deadlock.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+For alpha:
+
+1. Use effective executor instances after typed-field, `sparkConf`, and API-default precedence.
+2. Require at least one executor for Gang.
+3. Reject dynamic allocation enabled through typed fields or `sparkConf`.
+
+### Compatibility with Existing Integrations
+
+- **Volcano, YuniKorn, scheduler-plugins:** unchanged; mutually exclusive with native `.spec.scheduling`.
```
tariq-hasan, line 676: I am thinking that while we keep `spec.scheduling` on the public API we might still want to register the native Kubernetes WAS as a scheduler backend just as we do with Volcano, Yunikorn, etc.

The reason I say this is because the operator already has an internal scheduler registry/interface that manages the full lifecycle of a scheduler backend and native WAS may fit in somewhat within this hierarchy. As a sidenote this would also ensure that mutual exclusion is not something that's enforced only through validation but also as part of the backend registry.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+rules apply when the cluster serves standalone `PodGroup` resources (Kubernetes 1.37+); they do
+not apply to the superseded v1alpha1 inline model.
+
+### Open Questions
+
+1. Should `.spec.scheduling: {}` default to `Basic` or `Gang`?
+2. Should the first CRD expose only `schedulingPolicy`, or the full wrapper with allow-list rejection?
+3. Is v1.37+ an acceptable minimum cluster version for Phase 0 pinning? (Provisional yes; confirm
+   after Phase 0 module bump and API discovery.)
+4. Should native WAS be rejected when a deployment-level default batch scheduler would also apply?
```
tariq-hasan: I'm thinking the following may be a good hierarchy but open to other thoughts.

- explicit `spec.scheduling` beats `--default-batch-scheduler`
- explicit `spec.scheduling` + `explicit spec.batchScheduler` is an admission error
- neither means unchanged behavior

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+- one long-lived `Workload` blueprint per `SparkApplication`; and
+- one attempt-scoped `PodGroup` per submission ID.
+
+The cluster-mode driver is not a member of the executor gang. The driver must run before it can create executor Pods, so requiring driver and executors to schedule together would deadlock.
+
+This KEP is **provisional**. Open decisions are listed in [Open Questions](#open-questions).
+
+## Motivation
+
+Spark Operator users who need gang scheduling today choose an external integration such as Volcano, YuniKorn, or scheduler-plugins. Those integrations remain useful, but they have provider-specific APIs and installation requirements.
```
tariq-hasan: Add to this Volcano, YuniKorn, and scheduler-plugins do not really do gang-scheduling in the truest sense of the word.

These schedulers enable the driver to start and leave enough headroom in the cluster for executors to just take up capacity as they come in.

One would think true gang scheduling is about admission control not about capacity reservation and just hoping worker processes come in and take up the capacity.

This does tie in with the original motivation for the issue https://github.com/kubeflow/spark-operator/issues/2962 - which is that Spark's dynamic allocation poses elasticity challenges that may require consideration on the upstream WAS stream of work.

The existing "gang" scheduling support in the spark operator bypasses admission control altogether so this would be a much stronger motivation for integrating WAS into spark operator.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+For alpha:
+
+1. Use effective executor instances after typed-field, `sparkConf`, and API-default precedence.
+2. Require at least one executor for Gang.
+3. Reject dynamic allocation enabled through typed fields or `sparkConf`.
+
+### Compatibility with Existing Integrations
+
+- **Volcano, YuniKorn, scheduler-plugins:** unchanged; mutually exclusive with native `.spec.scheduling`.
+- **Kueue:** `kueue.x-k8s.io` Workload and `scheduling.k8s.io` Workload are different objects; queue admission is out of scope for this KEP.
```
tariq-hasan: It is true that the Kueue and the WAS workload objects are different but regardless Kueue has interactions with spark operator that would need to be addressed.

One point would be that Kueue quota allocation would need to agree with the WAS placement count to ensure that there is no deadlock and over-reservation.

A second point would be that mutations from Kueue would need to be preserved as the WAS integration mutates as well with the scheduling policy.

Yet another point would be for the case when Kueue admits the quota but WAS cannot place the gang. In this case an owner would need to be decided to time out on the workload. Otherwise there would be a race between operator retry and Kueue eviction and requeue.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+1. Detect `.spec.scheduling`.
+2. Verify the Spark feature gate and served Workload/PodGroup APIs.
+3. Reject legacy scheduler or user-provided membership conflicts.
+4. Resolve initial executor count with the same resolver used for `spark-submit`.
+5. Persist `status.submissionID` before creating attempt resources.
+6. Build or reconcile the long-lived `Workload`.
+7. Materialize or discover the current attempt `PodGroup`.
+8. Deep-copy the executor template and inject `schedulingGroup.podGroupName`.
+9. Submit the driver through the existing cluster-mode path.
+10. Observe objects and emit events; delete stale attempt PodGroups only after member Pods disappear.
```
tariq-hasan: I'm thinking probably ideal to delete the PodGroup itself as that can can hold final removal while members still exist - just because API-enforced deletion protection would be safer than unbounded controller-side waiting.

https://github.com/kubernetes/enhancements/blob/0469acb1d9444635135faf5a6c6711486b1a9040/keps/sig-scheduling/4671-gang-scheduling/README.md#L835-L854

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+### Non-Goals
+
+1. Replace, translate, or deprecate existing batch scheduler integrations.
+2. Put the cluster-mode driver in the executor gang.
+3. Expose a user-managed `schedulingGroup` field in the Spark CRD.
+4. Let users provide Gang `minCount`; scale remains expressed by Spark executor settings.
+5. Support dynamic allocation in this KEP; alpha rejects opted-in applications with dynamic
+  allocation enabled.
+6. Add a `CompositePodGroup` hierarchy in alpha.
```
tariq-hasan: I suppose the only way a `CompositePodGroup` hierarchy would ever apply is if a driver count of 1 could be made part of a composite pod group in addition to an executor count of N.

That is only feasible if upstream Spark ever enables a bootstrap mechanism whereby executors can be represented or reserved before the driver actually runs.

Since this requires changes in upstream Spark a `CompositePodGroup` hierarchy looks to be completely out of scope for the foreseeable future.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+
+Discovery rules:
+
+1. No matching owned object: create it.
+2. Exactly one matching owned compatible object: reuse it.
+3. More than one match, foreign owner, or incompatible immutable policy: stop and emit a conflict;
+  do not adopt or arbitrarily choose.
+
+Every step is idempotent. The controller persists the submission ID first and treats compatible already-existing objects as success after restart.
```
tariq-hasan, line 616: Should the PodGroup be labeled or annotated with the submission ID as `sparkoperator.k8s.io/submission-id=<id>` so as to help with the discovery?

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+```text
+driver:    ordinary independent scheduling
+executors: one Basic or Gang PodGroup
+```
+
+### Naming Conventions
+
+- Workload: `<truncated-app-name>-<stable-hash>`
+- PodGroupTemplate: `executors`
+- PodGroup: `<truncated-workload-name>-executors-<attempt-hash>`
```
tariq-hasan, line 581: Is the intention that the attempt hash is a hash of the submission ID?

I am thinking this would help restart discovery be reproducible and also create a deterministic mapping.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+    Running --> Retrying: application failure with retry
+    Retrying --> Running: new submissionID and PodGroup
+    Running --> Suspending: suspend requested
+    Suspending --> Suspended: members cleaned up
+    Suspended --> Running: resume with new submissionID
+    Running --> Completed: terminal success
+    Running --> Failed: terminal failure without retry
+    Completed --> [*]
+    Failed --> [*]
+```
```
tariq-hasan, line 649: I suppose it may help to clarify that retry does not involve **driver restart** (only a fresh submission with a new submission ID).

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+### Workload Lifecycle
+
+
+| Event | Workload | PodGroup | Notes |
+| --- | --- | --- | --- |
+| New opted-in app | Create or discover | Create after submission ID | Inject membership, then submit. |
+| Retry | Preserve blueprint | New group per submission ID | Never attach Pods to old group. |
+| Static executor-count change | Update blueprint if needed | New group with resolved `minCount` | Uses invalidation flow. |
+| Suspend | Preserve blueprint | Delete after members gone | Existing stop/delete flow. |
+| Resume | Reuse blueprint | New group per submission ID | Fresh driver submission. |
```
tariq-hasan: What would be the reason for deleting pod group on suspension? Comparing with other adopter controllers the behavior differs hence asking.

proposals/2962-workload-aware-scheduling/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+```go
+type ResolvedExecutorAllocation struct {
+    InitialExecutors int32
+}
+```
+
+For alpha:
+
+1. Use effective executor instances after typed-field, `sparkConf`, and API-default precedence.
```
tariq-hasan: We need to clarify the precedence rule among typed-field, `sparkConf`, and API-default.

tariq-hasan (commented):
Hi @RobuRishabh! I have added comments inline on the KEP.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Make sense, I will add a sentence to motivation, something like 
```
GenericWorkload is beta and off by default in v1.37, and most managed providers will not expose it. I added that to Motivation: native WAS is for clusters that can enable the gates, and Volcano, YuniKorn, and scheduler-plugins stay first-class. The feature stays opt-in with the gate disabled by default.
```

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: ok, will update it so that this should not be a permanent promise. Alpha still rejects user `gang.minCount` and in Graduation Criteria beta may allow it only as a floor that stays consistent with Spark's resolved executor count.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: ok, Added two stories: gang never admits (driver can be Running, no per-Pod fallback, alpha uses Events) and retry identity (new submissionID and PodGroup). The Kueue-admits-but-WAS-cannot-place case is real, but it needs quota ownership, so I listed it under Future Plans rather than designing it in this KEP.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: make sense. Will add Executor Template Injection under Design Details for that path.

RobuRishabh: Agreed, this is a correctness risk, and I will treat it as a Phase 0 gate rather than an assumption. Before implementation I will:
- Take the Spark versions this operator supports.
- Check the Fabric8 client each of those releases uses, and whether its Pod model includes schedulingGroup or preserves unknown Pod fields.
- Round-trip an executor template that sets spec.schedulingGroup through that Spark version's template load and serialize path.
- Allow native WAS only for versions where the field is still present on the created Pod.
- Reject opted-in applications on Spark versions that drop the field. I will not submit them and fall back to ordinary per-Pod scheduling.

RobuRishabh: I agree this is the right defense-in-depth, but not for alpha. Alpha proves the primary template path. A webhook fallback injector and a post-create check such as `WorkloadSchedulingReady=False` / `ExecutorMembershipMissing` are beta, after we know the primary path works. Alpha still rejects a user-set `schedulingGroup` on the stored template.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: For alpha, the KEP uses Events and admission or reconcile errors, and records the attempt PodGroup name on the object. A mirrored condition or status.workloadScheduling is now a beta graduation requirement, not an alpha API

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: make sense. Added both to Validation. A non-default spark.kubernetes.scheduler.name, driver scheduler name, or executor scheduler name is rejected with native WAS. spark.kubernetes.allocation.maxPendingPods below the resolved Gang minCount is rejected so the gang cannot deadlock on liveness.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: So, I am not sure here, @nabuskey @vara-bonthu can advice on this. I would rather not make this a public backend. `.spec.scheduling` is the opt-in, and admission enforces mutual exclusion with `batchScheduler`. Registering `was` next to Volcano would make the native API look like another third-party scheduler. Internal reuse of registry helpers can still happen later without a user-facing name. I recorded that under Alternatives.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Adopted this order: explicit `.spec.scheduling` wins over the deployment default; explicit `.spec.scheduling` plus an explicit batch scheduler is an admission error; if neither is set, behavior is unchanged.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: yes these interactions exist: quota versus minCount, preserving Kueue mutations, and who times out when Kueue has admitted quota but WAS cannot place the gang. They need an owner for quota and eviction, so this KEP only lists them under Future Plans. I do not want to specify that protocol inside the alpha scheduling API.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Updated the lifecycle to match KEP-4671.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Yes. The attempt PodGroup gets `sparkoperator.k8s.io/submission-id`. Discovery still also uses owner UID, controllerRef, and template name.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Yes. <attempt-hash> is a hash of status.submissionID, so a restart can rebuild the same name from the persisted ID.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: ok, clarified in the retry story and in Workload Lifecycle.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Spark suspend ends the attempt, and resume is a new submission. Keeping the old PodGroup would let a later attempt attach to stale scheduler state. The Workload blueprint stays. Only the attempt PodGroup is deleted. That differs from controllers that keep one long-lived group for a stable replica set.

proposals/2962-workload-aware-scheduling/README.md
RobuRishabh: Documented the same order `spark-submit` already uses. `spec.executor.instances`, when set, is passed after `sparkConf` and wins. Otherwise `spark.executor.instances` is used. Otherwise the API default is 1 when dynamic allocation is off. `minCount` uses that resolved value.

Issue #3119, opened 2026-08-25
Support Gateway API HTTPRoute for the Spark UI, next to the existing Ingress


xman1980:
## What

Add an option to expose the Spark UI through a Gateway API `HTTPRoute` instead of an `Ingress`, keeping the current behaviour as the default.

## Why

ingress-nginx is deprecated and its retirement is announced for March 2026. The operator is the only reason ingress-nginx cannot be removed from our staging cluster: everything else there has already moved to Envoy Gateway.

The docs say it plainly in getting-started: "The generated ingress resources are intended for the Ingress NGINX Controller". `SparkUIConfiguration` and `DriverIngressConfiguration` only expose `ingressURLFormat`, `ingressAnnotations` and `ingressTLS`, and the chart only has `controller.uiIngress`. There is no way to ask for anything else.

A data point on scale. In one namespace of our staging cluster the operator currently owns alot of Ingress objects, one per SparkApplication. Every one of them carries basic auth, a rewrite-target and a proxy-redirect annotation, all specific to ingress-nginx. Converting them by hand is not an option, and neither is a static route: the backend service name is derived from the application name, which is dynamic.

Two earlier requests, #2781 and #2214, were closed by the stale bot rather than by a decision, so this is not a new ask.

## Proposal

Mirror the existing ingress support rather than replace it:

- `controller.uiHTTPRoute.enable`, off by default, next to `controller.uiIngress.enable`
- values for `parentRefs` and the hostname or URL format, the same way `ingressURLFormat` works today
- per-application override in `SparkUIConfiguration`, matching `ingressAnnotations` and `ingressTLS`
- the created object is owned by the SparkApplication, same as the Ingress, so it is cleaned up with the job

Path handling stays as it is. `spark.ui.proxyBase` is already set by the operator, so a route with a path prefix and a rewrite filter reproduces what the nginx annotations do now.

Behaviour with `uiIngress` enabled must not change, so nobody is forced to migrate.

I am happy to work on the implementation if the direction is acceptable.

sxivansx:
/assign

I'd like to work on this. My plan is to keep the current `Ingress` behaviour as the default and add Gateway API `HTTPRoute` as an opt-in alternative, so existing deployments are unaffected.

Starting by mapping the existing Ingress code paths (`SparkUIConfiguration`, `DriverIngressConfiguration`, and `controller.uiIngress` in the chart) before proposing the API shape here for review.

[...]

Issue #2214, opened 2024-10-02
[QUESTION] httpRoute suported


stephbat:
Hello

I would like to know if httpRoute is supported for accessing to spark UI.
In fact, I would like that the spark operator creates a httpRoute (new way to do exposition in kubernetes), instead of an ingress for the spark UI

Thanks in advance

jacobsalway:
Hey, currently there is no support for the Gateway API in the Spark operator. The current Spark UI functionality creates Ingress resources that work best with the [NGINX ingress controller](https://github.com/kubernetes/ingress-nginx) because of automatically added re-write annotations. Obviously this doesn't preclude us from supporting it in the future as a feature request.

PR #3160, opened 2026-09-12
docs(proposals): add KEP-3119 for Gateway API HTTPRoute support

sxivansx:
Design document for #3119, requested by @tariq-hasan before implementation continues.

`proposals/` has held only a README so far, so this follows the template in [kubeflow/community](https://github.com/kubeflow/community/tree/master/proposals/NNNN-template) and takes the issue number as the KEP number.

**The two decisions that need maintainer direction** are in [Open Questions](https://github.com/kubeflow/spark-operator/blob/kep/gateway-api-httproute/proposals/3119-gateway-api-httproute/README.md#open-questions). Both are laid out with options and trade-offs rather than a single proposal, since each determines API surface that cannot be walked back:

1. **Where does the switch live** — chart and operator flag only, a per-application field on `SparkUIConfiguration` and `DriverIngressConfiguration`, or both with the field overriding. My recommendation is the flag alone for a first iteration, on the grounds that adding a `v1beta2` field is a one-way door and the flag can grow into the override later.
2. **Require the Gateway API CRDs, or detect and no-op** — detection is already the house pattern via `getPreferredAvailableAPIs`, but silence when someone explicitly enables the feature is a configuration that looks like it works and does not. My recommendation is to detect, and fail fast only when it was explicitly requested.

Happy to be overruled on either. The tables are meant to make the trade-offs decidable rather than to argue for one answer.

A few things the document records that may be useful independently:

- #2781 and #2214 both asked for this and were closed `not_planned` by the stale bot rather than on the merits.
- GSoC 2025 already delivered the first phase of this migration across other Kubeflow repositories, and kubeflow/notebooks#1301 is doing the equivalent for workspaces with a `ROUTING_PROVIDER` selector. Aligning naming with that is worth considering.
- The `httproutes` RBAC rule in #3125 is granted unconditionally, matching the existing `ingresses` rule. Gating it on the feature flag would be tighter but splits the role template — flagged for your preference rather than decided.

#3125 remains open and green. It deliberately adds no CRD surface, so it is unaffected by Question 1 whichever way it goes, but I have asked on #3119 whether you would rather hold it until this is agreed.


proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Summary
+
+The operator can expose the Spark web UI through a Kubernetes `Ingress`. This KEP adds
+Gateway API `HTTPRoute` as an opt-in alternative, selected by configuration, with the
+Ingress path unchanged and still the default.
+
+The operator creates one routing object per `SparkApplication`, named and owned by that
+application. Nothing about that lifecycle changes. Only the kind of object created
+changes, and only when a cluster administrator or user asks for it.
```
tariq-hasan: "Nothing about that lifecycle change" - I don't think this statement is true given that Ingress carriess semantics that does not exist in HTTPRoute.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+Two design questions are deliberately left open for maintainers to settle in review, since
+they determine the API surface and cannot be walked back once released. They are listed
+under [Open Questions](#open-questions).
```
tariq-hasan: The actual list of open questions is more extensive and covers the boundary of migration semantics.

Below is a sample representative. The KEP would need to address these questions.
- Where does `urlFormat` belong?
- What happens to `ingressTLS`?
- What happens to auth annotations?
- Can Ingress and HTTPRoute both be enabled?
- Must HTTPRoute have a hostname?
- Is route readiness exposed?
- How does `driverIngressOptions` migrate?
- Should this use existing feature gates?

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+name is derived from the application name, so a single static route cannot replace them.
+
+### Goals
+
+- Expose the Spark web UI through a Gateway API `HTTPRoute` as an opt-in alternative to
+  `Ingress`.
+- Leave existing behaviour unchanged when the feature is not enabled.
+- Reuse the existing per-application object lifecycle, including ownership and garbage
+  collection.
+- Cover both `SparkUIConfiguration` and `DriverIngressConfiguration`, which today follow
```
tariq-hasan: The KEP does not explain the design considerations for `DriverIngressConfiguration`.

It's better to take a phased approach on this KEP and explicitly state that Phase 1 will only cover `SparkUIConfiguration` and that `DriverIngressConfiguration` is being relegated to a subsequent phase pending further discussion.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+  the same Ingress pattern.
+- Degrade safely on clusters that do not have the Gateway API CRDs installed.
```
tariq-hasan: This might need clarification in the KEP in terms of whether the degradation is a silent no-op (operator continues with no external route), fallback (operator creates Ingress objects) or fail fast (operator refuses to start).

At the same time though the KEP later expands on capability discovery to determine the APIs exist in this cluster and feature enablement in terms of what behavior the platform administrator configures. This is more in line with explicit configuration validation as opposed to degradation.

So I believe the goal is rather to have a defined, documented behavior instead of safe degradation itself.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+The scale makes manual conversion impractical. The operator creates one `Ingress` per
+`SparkApplication`, each carrying controller-specific annotations, and the backend service
+name is derived from the application name, so a single static route cannot replace them.
+
+### Goals
+
+- Expose the Spark web UI through a Gateway API `HTTPRoute` as an opt-in alternative to
+  `Ingress`.
+- Leave existing behaviour unchanged when the feature is not enabled.
```
tariq-hasan: Since the KEP has a compatibility goal for users who leave HTTPRoute disabled we should also have a corresponding goal for those who enable it. This is also where the consideration of migration safety would come into play.

So the goal would be, for every user-visible behavior provided by current Ingress exposure, to define whether HTTPRoute preserves it, relocates it, rejects it, or intentionally drops it.

And that may then be accompanied by a migration matrix that covers the following in the design section.

Existing behavior | HTTPRoute treatment
-- | --
hostname | translate
path | translate
backend Service | translate
rewrite | typed filter, where supported
TLS | Gateway listener / reject old config
basic auth annotation | external policy / reject
ingress class | replaced conceptually by Gateway attachment
arbitrary annotation | translate, reject, or warn
status | route-parent conditions

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+- Reuse the existing per-application object lifecycle, including ownership and garbage
+  collection.
+- Cover both `SparkUIConfiguration` and `DriverIngressConfiguration`, which today follow
+  the same Ingress pattern.
+- Degrade safely on clusters that do not have the Gateway API CRDs installed.
+
```
tariq-hasan: We should also have a goal around observability to report whether the generated route was actually programmed by the Gateway - given that creating an HTTPRoute object is not equivalent to establishing working exposure. This could happen because parent may not exist, listener may not match, namespace may not be allowed, hostname may not intersect listener, filter may be unsupported, reference may be invalid, etc.

Given that the Gateway API gives substantially richer attachment status than merely “object exists” we'd need to tell the user whether the generated route was accepted/programmed to the extent Gateway API status permits.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+### Non-Goals
+
+- Removing, deprecating, or changing the `Ingress` code path. It stays the default.
+- Migrating existing `Ingress` objects to `HTTPRoute` automatically.
+- Supporting Gateway API kinds beyond `HTTPRoute` (`GRPCRoute`, `TLSRoute`, `TCPRoute`).
+- Managing `Gateway` or `GatewayClass` objects. Those are cluster infrastructure and
+  remain the administrator's responsibility.
+- Istio `VirtualService` support, or any other mesh-specific routing object.
```
tariq-hasan: Another non-goal would be for the operator to create authentication or traffic-policy objects (SecurityPolicy, AuthorizationPolicy, or equivalents) on behalf of the user.

If a user currently express some security behavior through Ingress annotations such as `nginx.ingress.kubernetes.io/auth-type: basic` then switching to HTTPRoute should not automatically preserve the authentication. The operator should only be creating the HTTPRoute. Authentication policy would need to be owned or configured elsewhere.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+  [@madmecodes](https://github.com/madmecodes), mentors Julius von Kohout and Kimonas
+  Sotirchos) delivered 25+ PRs described as "pioneering the migration to Gateway API
+  (HTTPRoute)", including HTTPRoute-based path routing for KServe.
+- **GSoC 2026 Project 5** "Platform Scalability and Security" (mentor
+  [@juliusvonkohout](https://github.com/juliusvonkohout)) lists "migration from Istio
+  Gateway to Kubernetes Gateway API" among its deliverables.
+- [kubeflow/notebooks#1301](https://github.com/kubeflow/notebooks/pull/1301)
+  ([@aojea](https://github.com/aojea)) implements a Gateway API routing provider for
+  workspaces, generating `HTTPRoute` instead of Istio `VirtualService`, selected by a
+  `ROUTING_PROVIDER` setting.
```
tariq-hasan: This has been closed unmerged so we should remove this reference.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+This KEP covers the Spark Operator's part of that migration. It does not attempt to
+coordinate the other repositories.
+
+## Proposal
+
+Mirror the existing Ingress support rather than replace it.
```
tariq-hasan: Mirroring the behavior across the two APIs may not be possible given that TLS, authentication annotations and ingress class name do not translate over from Ingress to Gateway.

I think the idea would be to add HTTPRoute as an alternative exposure mechanism alongside Ingress, with a defined contract for configuration that cannot be carried across.

proposals/3119-gateway-api-httproute/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+When the feature is enabled and the cluster supports it, the operator creates an
+`HTTPRoute` in place of the `Ingress` it would otherwise have created, attached to
+administrator-supplied `parentRefs`, targeting the same Spark UI `Service`, with the same
+name, namespace, labels and owner reference.
```
tariq-hasan: The issue with having the same name is that the Ingress name is of the form `<app>-ui-ingress`.

I think, in addition to `driverInfo.webUIIngressName` on the CRD status we may want to have `driverInfo.webUIExposureKind` so we can separate Ingress from HTTPRoute resources.

This would also enable cleanup to happen conveniently during mode switching.

tariq-hasan (commented):
Thanks for raising the PR.

I think we need to be more explicit about the semantic contract when switching from Ingress to HTTPRoute. So we need to decompose the Ingress semantics (routing, TLS, authentication, rewrite behavior, controller selection, annotations/policies, status/attachment, etc.) and remap individually into the Gateway API architecture. We may find that some translate cleanly, some move elsewhere, some disappear while some require an explicit rejection. Either way that contract needs to be made explicit within the KEP.

Another point would be to consider how driverIngressOptions is handled - whether we take a phased approach and in Phase 1 only consider Spark UI and relegate the driverIngressOptions to phase 2.

PR #3164, opened 2026-09-13
test: add unit tests for the scheduler Registry

magic-peach:
## Purpose of this PR

internal/scheduler's Registry backs the actual scheduler wiring in cmd/operator/controller/start.go (Register is called there for each of the volcano, yunikorn, and kubescheduler factories) and had no test coverage, including the duplicate-registration error path.

**Proposed changes:**

- Add unit tests for Register, covering a normal registration and the duplicate-name error case
- Add unit tests for GetScheduler, covering a registered name and a not-found error
- Add unit tests for GetRegisteredSchedulerNames
- Add a unit test for GetRegistry confirming it returns the same singleton instance across calls

## Change Category

- [ ] Documentation update

### Rationale

Test-only change, no behavior modified.

## Checklist

### Additional Notes

No existing issue tracks this; found while reading the scheduler registry.


tariq-hasan (changes requested):
Please rewrite the tests using Ginkgo/Gomega as that would be the default going forward.

RobuRishabh (commented):
Hi @magic-peach thanks for the contribution, Please add a case where the factory returns an error. `GetScheduler` should return that error to the caller. Right now every fake factory always succeeds, so that path is untested and Optionally you can also pass a non-nil config into `GetScheduler` and check the factory actually receives it. Today every call uses `nil`.

